PR #2, opened 2026-07-30
fix: resolve the adversarial-review findings on the 0.1.653 batch

vitali87:
Fixes the eight findings an adversarial review raised against the 0.1.653 batch, each verified against source (the tcsh one reproduced live) and fixed RED-test-first.

- Grid overlay desync: in a 3+ editor grid, slot 0's image gate, layout key and baked pixels now all resolve to the active group ("group_on_side" applies the same clamp as "focused_image_side" and the bake reads through it), so a focused PDF actually previews, pages turn, and links click.
- csh-family PATH repair: the login-shell probe marks itself a login shell via a leading-dash argv[0] (tcsh refuses "-l" next to "-c"), so "~/.login" is finally sourced; verified against real tcsh with an injected HOME.
- Collab snapshot resend: the exponential backoff caps at 1s so the schedule (0.5s / 1.5s / 2.5s) fits inside the 3s bootstrap window; unbounded doubling never re-asked an owner that registered in the back half of the window.
- PDF link slicing: a bare "&" no longer swallows the rest of a run (bounded entity lookahead + a clamp keeping every slice inside its run rect), and zero-extent runs clamp to one page unit so they stay clickable.
- Test hardening: the dock-launch stub is a strict login-shell emulator pinning the probe's exact argv and nested quoting; poppler and tcsh join the nix dev shell so the capability-gated tests run for real on Linux instead of skipping.
- Docs, release notes, and the 0.1.654 bump.

Full suite: 2751 + 5 CLI tests, 0 failed, clippy zero warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PDF previews for focused editor groups and improved PDF link handling.
  * Improved csh/tcsh PATH detection when launching from the Dock.
  * Reduced duplicate multiplayer snapshot requests with capped retry backoff.
  * Corrected image handling in complex editor layouts.
* **Documentation**
  * Updated architecture and multiplayer documentation.
* **Tests**
  * Added coverage for PDF links, shell login behavior, layout overlays, and snapshot retries.
* **Release**
  * Updated the package version to 0.1.654.


vitali87:
@greptile review

PR #3, opened 2026-07-31
fix: bind session and collab sockets 0600 without touching the process umask

vitali87:
Both the session mux and the collab relay bound their unix sockets under a save/restore umask dance. umask is process-global: two concurrent binds interleaving their restores could leave one socket created under the caller's loose mask (the relay had no chmod fallback, so its socket could stay 0644 for its whole life) and could corrupt the process umask permanently. Observed as a real interleaving in the parallel test suite; latent for any future in-process threading.

The fix extracts a single shared binder, "session::bind_socket_0600", which touches no process-global state: it binds inside a fresh 0700 staging dir next to the target (same filesystem, short name for the AF_UNIX path budget), sets the socket itself to 0600 while nobody can traverse to it, then renames it into place atomically. RED-tested: a two-thread bind stress demonstrated the umask corruption against the old code (process umask left at 0o077) and is deterministic-green against the new binder.

Full suite: 2754 + 5 CLI tests, 0 failed, clippy zero warnings. Ships as 0.1.655.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security for session and collaboration sockets by restricting access to the owner.
  * Prevented race conditions during socket creation and improved handling of stale or active sockets.
  * Socket setup no longer changes the application’s process-wide permissions behavior.

* **Documentation**
  * Updated multiplayer documentation to describe secure, serialized socket creation.

* **Release**
  * Updated the package version to 0.1.655.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft"
-version = "0.1.654"
+version = "0.1.655"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft"
-version = "0.1.654"
+version = "0.1.655"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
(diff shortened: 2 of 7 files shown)


vitali87:
@greptile review

PR #4, opened 2026-07-31
fix: resolve the adversarial-review findings on the Finder/explorer span

vitali87:
An adversarial review of the twelve-commit Finder/explorer span (99eadc7..b66a51f, judged against current HEAD) produced eight findings; six are fixed here RED-test-first and two were refuted against the source.

Fixed:
- Croft.app launcher: every path component now crosses both shell layers ("do shell script", then Ghostty's /bin/sh -c) through its own "quoted form of", paths are escaped into the AppleScript literals, and "on open" loops over every selected/dropped file instead of silently keeping only the first.
- PDF wheel: same-direction momentum bursts coalesce (each page step is a synchronous rasteriser run on the UI thread; a flick used to queue 10-30 of them), and the wheel clamps at the document ends instead of wrapping (arrow-key wrap stays).
- A PDF rebuilt on disk re-bakes its preview: the overlay re-emit key gains a content generation, so an unchanged path/rect/page no longer pins the pre-rebuild pixels.
- Opening a file while the terminal is maximized restores the editor (VS Code's convention) instead of moving focus into a zero-height pane where typing dirtied an invisible buffer.
- On iTerm2/Sixel (no image z-layer) an open context menu overlapping the editor preview suppresses the image blit that overpainted it; Kitty keeps its deep-negative z path.
- Minors: End on a PDF with an unknown page count reports honestly instead of rendering page 4294967295, and the render scratch dir moves out of shared /tmp into croft's own cache dir.

Refuted (no change):
- "Caret paints a hint-width left of a click on the anchor char": the caret-before-hint painting is deliberate VS Code affinity, documented and pinned by cursor_screen_pos_shifts_past_inlay_hints.
- "Inlay click inversion wrong on wrapped rows": impossible scenario; row_inlay_spans returns empty whenever wrap is enabled, so hints are never rendered in wrap mode and the inversion degrades to the identity.

Full suite: 2765 + 5 CLI tests, 0 failed, clippy zero warnings. Ships as 0.1.656.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF wheel navigation with smoother, throttled paging that stops at document boundaries.
  * Fixed PDF and image previews refreshing reliably after content or page changes.
  * Prevented image overlays from appearing beneath context menus.
  * Restored maximized terminals when returning to the editor.
  * Improved handling of unknown PDF page counts.
  * Fixed launching multiple files from Finder, including paths with spaces and special characters.
  * Stored temporary PDF files in the application cache.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft"
-version = "0.1.655"
+version = "0.1.656"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft"
-version = "0.1.655"
+version = "0.1.656"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2113,10 +2113,12 @@ pub struct App {
     /// Required to bake OSC-1337 images at exact viewport pixel size so
     /// iTerm draws them with no stretching or letterboxing.
     cell_pixel: Option<(u32, u32)>,
-    /// Last wheel step applied to a PDF preview (when, and in which
-    /// direction), so a momentum flick's same-direction burst coalesces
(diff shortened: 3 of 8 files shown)


vitali87:
@greptile review

PR #5, opened 2026-07-31
fix: resolve the adversarial-review findings on the pair/navigator span

vitali87:
An adversarial review of the four-commit AI-pair/navigator span (ad4e1f7..eef9d3f, judged against current HEAD) produced ten findings; all ten are addressed here, fixed RED-test-first.

- Scrollbar with comment boxes: the drag now maps through the same content length the bar was drawn with (lines + box rows). A short file with a tall navigator comment had a completely dead scrollbar; a long file's thumb ran away from the pointer.
- F4 review key: next-note walking is ordered by (row, index), so both notes on a shared row are reachable. Replying anchors the navigator's answer on the answered note's own row, which made the answer permanently unreachable by F4 before.
- Tall comment boxes in non-wrap mode: the viewport top is now a content row (line plus rows into its box), mirroring the wrap path, so a box taller than the window scrolls into view and its Reply/Ignore footer is reachable; the wheel, the scrollbar drag, and the thumb position all go through the same model.
- Proactive looks: tree-sitter-md wraps list bullets and quote lines in paragraph nodes, so a TODO list fired one unsolicited turn per bullet, and a paragraph split in two fired anchored at the wrong half; both are now excluded (only purely-added top-level prose counts). A single struct field or enum variant no longer counts as a completed construct.
- Reply UX: the footer field windows horizontally around the caret (typing past the field width used to go blind with no caret drawn), and the "you: ..." line is appended only after a successful send, so a failed send no longer duplicates the reply on retry.
- Workspace re-root drops the old workspace's collab carets (they are keyed by root-relative paths and would dress same-named files in the new root), and turn commentary anchors at the active cursor row only when the noted file IS the active tab.
- Test hardening: the Cmd+K Y OUTPUT assertion only accepts lines past its own baseline on the process-global channel, so sibling-test residue cannot satisfy it.

Full suite: 2775 + 5 CLI tests, 0 failed, clippy zero warnings. Ships as 0.1.657.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft"
-version = "0.1.656"
+version = "0.1.657"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft"
-version = "0.1.656"
+version = "0.1.657"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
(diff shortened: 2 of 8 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #7, opened 2026-07-31
fix(pair): navigator lifecycle hardening (note anchors, host lock liveness, draft loss, child reaping)

vitali87:
Six confirmed bugs from an adversarial review of the 0.1.635 resident-navigator span, each fixed RED first.

## Fixes

1. **Aborted edits drifted comment anchors.** `open_region` and `EditBody` shift note offsets for the streamed edit's delete and inserts, but `revert_region` restored the text without the inverse shift, so every cancel/abort left the turn's notes anchored below where the model put them. The revert now applies the inverse span. The REPL note path also pushed its note after a lock gap (`say` re-locks internally), letting a concurrent remote edit strand the new note's anchor; the note now lands under the same lock as its offset computation.

2. **A failed spawner deadlocked the workspace's navigator.** The single-host flock was claimed before the spawn and never released on failure, while the death latch blocked retries: one window with a broken claude blocked every other window from hosting, and the "hosted by another croft window" status was a lie. A failed spawner now releases the lock and undoes its same-tick owner self-appointment (a host whose pilot died later keeps both, since it is still the live collab owner and re-seats on re-activation).

3. **`croft pair` after a failure did nothing.** The death latch only cleared on an enabled-to-disabled transition, so re-running `croft pair` (record already enabled) never retried, despite the CLI's printed promise. Any record rewrite (mtime change) now re-arms seating.

4. **The refusal status clobbered the losing window forever.** The non-hosting window re-announced the lock refusal every second, permanently overwriting its status line and dirtying a frame at 1 Hz. It now announces once and polls silently for takeover.

5. **A proactive look could eat a typed ask.** `tick_proactive_navigator` did not check for an open input prompt, so it could steal the seat while the user typed in the ask box; their Enter then failed against a busy host and the submit path had already closed the prompt, destroying the draft. Proactive looks are now suppressed while a prompt is open (without consuming the scan latch), and a failed ask keeps the box open with the draft intact.

6. **A quick quit after unseat orphaned the claude child.** `Drop` detaches the 2s grace-kill onto a thread; if the process exited within that window the thread died mid-kill and the child (plus its MCP subprocesses) survived. Detached teardowns are now registered and joined on the exit path.

Plus: turn note counters reset on unseat/death (a re-seated navigator's first TurnDone no longer inherits a dead seat's count), and the TurnDone hint only names a file when every note landed in it ("3 comments across 2 files" otherwise).

## Testing

Every fix landed RED first: 9 new/extended tests (note un-shift, lock release + second-window hosting, record-rewrite re-arm, refusal announced once, prompt suppression, draft preservation, teardown reaping via a live sleeper child, counter reset, summary wording). Full suite 2797 + 6 CLI green, clippy zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigator recovery after pairing records are rewritten or reactivated.
  * Prevented workspace locks and ownership from remaining held after failed setup.
  * Preserved ask-box drafts and prompts after unsuccessful submissions.
  * Improved note placement, multi-file summaries, and edit reversion behavior.
  * Ensured background shutdown tasks and child processes finish cleanly.
* **Release**
  * Updated the package version to 0.1.659.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft"
-version = "0.1.658"
+version = "0.1.659"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft"
-version = "0.1.658"
+version = "0.1.659"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/docs/MULTIPLAYER.md
+++ b/docs/MULTIPLAYER.md
@@ -561,11 +561,18 @@ interaction into turn-based driver/navigator pairing:
   `◆ <name> seated` badge; the orange streaming badge takes over whenever
   it types. If the claude child dies (or fails to seat), the host surfaces
   it and unseats, and does NOT respawn it every second — the latch clears
-  only when the record is deactivated, so a `croft pair --off` / `croft
-  pair` (or palette off/on) re-activates. Teardown on unseat runs on a
+  when the record is deactivated OR rewritten, so `croft pair --off` /
+  `croft pair`, a palette off/on, or simply re-running `croft pair` after
+  fixing the backend all re-activate. A window whose spawn failed also
+  releases the single-host lock (and its same-tick owner self-appointment),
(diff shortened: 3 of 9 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #8, opened 2026-07-31
fix(pair): pair-host hardening (diagnostics routing, UI deadlock cycle, send-failure atomicity, busy-gate ordering)

vitali87:
Six confirmed bugs from an adversarial review of the 0.1.634 croft-pair feature span, each fixed RED first.

## Fixes

1. **Diagnostics were anchored as the navigator's own comments.** The claude child's stderr tee and three host notices (edit suppressed on a comment-only turn, no live session serves the file, target file not live) rode `PairEvent::Commentary`, which the App accumulates and lands as an anchored comment box authored by the navigator — inflating the turn's comment count with text the model never said. All four now go to the OUTPUT channel when seated (matching the documented contract and the existing inverted-fence precedent); the REPL driver keeps stdout.

2. **Same-process deadlock cycle between the UI thread and the pilot.** `is_busy`, `caret_sites`, `notes_snapshot`, `last_seen`, the send paths, and the note actions all locked the pilot's state mutex from the render/tick path, while pilot threads hold that mutex across `write_frame_blocking` relay writes that spin on WouldBlock with no deadline — and only the App's own collab drain (same UI thread) completes them. Render reads now use an instant try_lock with a last-answer cache; everything that decides or acts (busy checks, sends, note edits, note landing) uses a bounded 100ms lock, so microsecond pump contention cannot spuriously reject a keypress or misreport a quiet seat as busy, while a genuinely wedged pilot can no longer hang the editor.

3. **A failed send wedged the seat.** `write_user_turn` latched `turn_active` and consumed the pending cancel note before the send; a dead stdin left the seat permanently "mid-turn" and silently dropped the cancel notice owed to the model. Both are restored on error.

4. **`is_busy` went false before the App consumed the turn's end.** The pilot's reader clears `turn_active` before queueing `TurnEnd`, so a tick in that window could start a new turn that stole the finished turn's commentary anchor and note counters. A `turn_pending` gate now holds the seat busy from a successful send until `poll` delivers `TurnDone`.

5. **`begin_turn` committed state before the send could fail.** A failed yield destroyed the yield-diff baseline (a retry diffed the file against itself) and left `comment_only` latched to edit-suppress the next successful turn. The look is staged and undone on send failure.

6. **An unterminated fence header leaked into commentary.** A turn ending exactly at `<<<NOTE f.rs:12>>>` (no trailing newline) flushed the raw marker into a comment box. A tail that parses as a valid fence header is dropped like an unterminated body; malformed tails keep the malformed-header-as-commentary rule.

## Testing

Every fix landed RED first: six new tests plus a rewrite of the yield-suppression test (diagnostics asserted on OUTPUT and absent from Commentary, with the fake claude now emitting stderr noise so every host e2e exercises the tee). The bounded-lock design was itself driven by a reproducible interference failure between the ask-box and proactive tests under parallel load. Full suite 2805 + 6 CLI green, clippy zero.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented diagnostic messages from appearing as model commentary.
  * Improved responsiveness during busy or stalled operations.
  * Failed sends now restore the correct turn state and pending notes.
  * Prevented incomplete protocol markers from appearing as commentary.
  * Improved turn-completion handling to avoid overlapping turns.
  * Improved app startup by preventing shell probes from taking over the terminal.
* **Documentation**
  * Clarified diagnostic messages shown in Navigator output.
* **Chores**
  * Updated the package version to 0.1.660.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #9, opened 2026-08-01
fix(collab): harden the Phase D engine against traversal, dead links, wedged peers, stale panes, and bootstrap lockout

vitali87:
Five defects in the 0.1.632-0.1.634 collab engine, each verified against HEAD, fixed RED-GREEN:

1. **Wire file keys escaped the workspace root.** `owner_buffer_text` joined the `SnapshotRequest` key straight onto the root, so `../../../.ssh/id_rsa` from any guest (the MCP collab agent forwards caller input verbatim) made the owner read, broadcast, and even edit files outside the workspace. `collab::contained_path` now rejects any key with a non-plain component, at every wire-key resolution site.

2. **A guest with no owner was locked out forever.** The 3s bootstrap timeout removed the doc, and the app tick immediately re-requested any open file that was neither live nor bootstrapping, re-arming the input gate 16ms after every give-up; the guest could never type, and the save gate blocked the only author from persisting anything. The give-up is now a `DocState::LocalOnly` latch: plain re-requests respect it, local-only files are exempt from the guest save gate, and the collab agent's `collab_open` explicitly rejoins.

3. **A dead relay was indistinguishable from an idle one.** `drain` treated EOF like `WouldBlock` and every `send` result was discarded, so a killed relay left `is_live` true while every op vanished into RAM. EOF and failed sends now latch the channel dead; the app drops the session and reconnects through the existing backoff, and files re-bootstrap. The owner's site-id counter is also pid-seeded so a restarted owner can never re-issue an id a surviving guest still holds.

4. **One wedged peer deadlocked the whole session.** The relay blocking-wrote to every peer while holding the global client mutex, and `write_frame_blocking` spun forever, so one SIGSTOPped participant wedged the relay and froze every other participant's UI thread mid-send. Relay writes now carry `SO_SNDTIMEO` (2s; a failed peer is shut down and dropped), and `write_frame_blocking` gives up after 5s. Setting the timeout is tolerated to fail on macOS for a peer that already hung up (the liveness probe), which otherwise killed the accept loop.

5. **Splitting or reopening a live file reverted the session.** The duplicate pane loaded stale disk text, incoming spans landed on it at nonsense offsets ("first found" is unstable under focus swaps), and the first keystroke broadcast a diff of the live doc against the old disk copy, reverting every peer. Buffers now carry the `CollabDoc::text_gen` they last synced at: never-attached panes never extract and are seeded from the replica, behind panes mirror, and remote spans land on every attached pane.

Nine new tests (session-, relay-, frame-, and two-app-level); full suite 2819 + 6 green; clippy clean. docs/MULTIPLAYER.md and docs/ARCHITECTURE.md updated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shared-file synchronization across split panes and open buffers.
  * Added reliable relay disconnect detection, reconnection, and offline edit recovery.
  * Prevented blocked connections from hanging indefinitely.
  * Added local editing and saving when collaboration setup times out.
  * Blocked unsafe workspace paths, including traversal, absolute paths, and symlinks.
  * Improved recovery from stale or corrupted shared-file snapshots.
  * Improved PDF scrolling responsiveness.

* **Documentation**
  * Updated collaboration and architecture documentation.

* **Chores**
  * Updated the package version to 0.1.661.


vitali87:
@greptile review

vitali87:
@greptile review

PR #10, opened 2026-08-01
fix: adversarial-review fixes for the themes/dtach/search-seed span

vitali87:
Fixes the six confirmed findings from the adversarial review of span 580624f..d9070bd (M7 test-runner registry, ten themes, ANSI palettes, frame prefill, dtach persistence, grep/rg Search seeding), judged against current HEAD.

## Fixes

- **Search seed panic**: a text selection left in the files-to-include field survived the seed as a stale byte range into the shorter seeded text; the next keystroke panicked out of bounds (`replace_range`). `SearchPanel::seed` now replaces query + both filter lists wholesale, resets both selections, and refocuses the query field.
- **Stale filters falsified the seed**: a previously-set include/exclude glob silently narrowed the seeded search while the status line claimed a faithful seed (feeding Replace All a wrong subset). The seed now describes the terminal command exactly.
- **`rg -s` dropped**: the explicit case-sensitive flag fell through the boolean catch-all, inverting case semantics (the long form was handled). Now honored.
- **Negated/multiple globs**: `rg -g '!node_modules'` was copied into the include filter, which has no negation support, matching nothing; only the first `-g` was kept. Negated globs (and grep's `--exclude`/`--exclude-dir`) now land in files-to-exclude, and globs accumulate comma-separated.
- **`croft ls` phantom rows**: the workspace's collab relay socket (`<hash>.collab.sock`, same dir + hash keying) was listed as a live `(unknown)` session whose 8-char id collided with the real session's, and probing it cost the relay a connect-and-drop. `ls` now skips relay sockets entirely (pruning belongs to the relay's own bind path).
- **Tab strip ignored the theme**: the ten editor-inspired themes recolored the editor body but the tab strip stayed the built-ins' hardcoded navy (the chrome model was binary: brand teal vs navy). Tab chrome (strip, tab bodies, hover lift, close pill) is now per-theme data: the two built-ins pin their historical constants byte-for-byte (guarded by the manifest==const equality test), and every other theme derives the colors from its palette (strip from `search`, active tab from `selection`, hover/pill from `accent`), with optional `tab_*` manifest overrides.

## Verification

Each fix landed strict RED→GREEN (10 new tests: 4 parser, 2 seed, 1 session ls with a live relay listener, 1 tab-strip render, 2 theme pins). Full suite green: 2837 bin tests + 6 CLI, 0 skipped/deleted; clippy zero warnings; fmt clean. Bumped to 0.1.662 with release notes; ARCHITECTURE.md updated.

## Summary by CodeRabbit

* **New Features**
  * Search commands now support multiple include patterns and exclusion filters.
  * Terminal searches correctly populate queries, filters, and search options.
  * Editor tabs now use theme-specific colors, including hover and close controls.

* **Bug Fixes**
  * Preserved both collaborators’ offline edits after reconnecting.
  * Prevented stale search filters and selections from carrying over.
  * Excluded collaboration relay sockets from session listings and cleanup.
  * Updated built-in themes with improved tab styling.

* **Documentation**
  * Expanded documentation for search, sessions, collaboration recovery, and themed tabs.


vitali87:
@greptile review

vitali87:
@coderabbitai review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #11, opened 2026-08-01
fix: adversarial-review fixes for the testing/LSP span (M5/M6, call hierarchy, logpoints, occurrences)

vitali87:
Fixes the nine confirmed findings (plus one minor) from the adversarial review of span f8f3ff5~1..580624f (vitest/jest runners M5, debug-a-test M6, call hierarchy, logpoints, gutter play glyph, occurrences, SCM graph scrolling, SGR-2 DIM, testing re-root), judged against current HEAD.

## Fixes

- **Debug-a-test froze the UI**: the `cargo test --no-run` build ran synchronously on the main loop (minutes on a cold crate, no keys/redraw/cancel, the "Building…" status unreachable). It now runs on a background thread with the progress status visible; the tick drain launches lldb-dap when the binary arrives, with breakpoints read at launch time.
- **Wrong harness for lib+bin crates**: the binary picker was last-wins over cargo's artifacts, so a `src/lib.rs` test launched the bin harness, ran zero tests, and exited before a breakpoint could bind (empirically reproduced by the reviewer). Candidates now rank lib > bin > integration and, with several harnesses, each is probed with `--list` to find the one that actually contains the test.
- **`#[cfg(test)]` play beads**: any attribute containing "test" marked a fn as a test — croft's own source has 39 `#[cfg(test)] fn` sites that wore beads running nothing (and wiping the Testing tree). Only a `test`-final attribute path counts now; `def test_*` beads also require a pytest-collectable file, and unsaved buffers get none.
- **Clicking a breakpoint dot started a test run**: the sign-cell click hit-test ignored the render's precedence (stop arrow / breakpoint / AI-stream square above the bead). It now honors it — clicking a red ● on a test line no longer silently runs the test.
- **Resurrected logpoints**: removing a breakpoint left its logpoint message (and condition) behind, so F9-remove + F9-add produced a "breakpoint" that printed and never paused. Removal now discards both.
- **vitest `-t` unescaped**: vitest's `-t` is a regex like jest's; a title like `adds (1 + 1)` matched nothing and an unbalanced bracket errored. All vitest paths now escape, matching the jest paths.
- **Occurrence tint over find highlights**: the caret parked on a find match makes documentHighlight return the very cells the find layer painted black-on-gold; the occurrence bg painted over them left unreadable black-on-grey. Occurrences now paint beneath the find layer.
- **Stale menu hint for Keep Open**: the tab menu advertised `⌘K ⏎`, which now runs the test at the caret. Keep Open lives on `⌘K ⇧P` (documented), and the menu says so.
- **Dead splitter band**: the COMMITS graph's scrollbar rect refreshes only while Source Control renders; after switching views the stale rect kept exempting a band of the sidebar seam from dragging. Both scrollbar exemptions (graph and outline) are now gated on their view.
- **Late call-hierarchy replies** (minor): a slow rust-analyzer answer replaced whatever menu the user had opened while waiting (and yanked focus); it is now dropped with a status line when another menu is open.

## Verification

Every fix landed strict RED→GREEN (12 new/extended tests, including a fake-harness probe test and an off-thread build test that needs no lldb). Full suite green: 2855 bin tests + 6 CLI, 0 skipped/deleted; clippy zero warnings; fmt clean. Bumped to 0.1.663 with release notes; ARCHITECTURE.md and KEYBINDINGS.md updated.

## Summary by CodeRabbit

* **New Features**
  * Keep preview tabs open with `Cmd+K`, then `Shift+P`.
  * Improved Cargo test debugging by selecting the correct test binary and handling builds asynchronously.
* **Bug Fixes**
  * Fixed test detection for Rust and Python files, including Vitest title filtering.
  * Improved breakpoint cleanup, syntax highlighting visibility, and test indicator behavior.
  * Fixed stale scrollbar interactions and delayed call-hierarchy menu updates.
  * Corrected complete breakpoint metadata removal.
* **Documentation**
  * Updated architecture and keyboard shortcut documentation.
* **Release**
  * Updated the application version to 0.1.663.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #12, opened 2026-08-01
fix: span 20272ce..c3aebda review — inlay caret sides, Testing border clicks, queued graph refetch, alt-selection neighbour tie-breaks

vitali87:
Adversarial review of span 20272ce..c3aebda (inlay hints, Testing panel, commit graph, alt-screen selection) judged against current HEAD. Four confirmed bugs fixed RED to GREEN, plus one test-infrastructure race exposed by adopting cargo-nextest.

## Fixes

1. **Secondary multi-cursor blocks and ghost carets painted on the wrong side of an inlay hint.** A caret exactly at a hint's anchor (where Change All Occurrences puts every caret on a `let` binding with a `: i32` hint) used `inlay_cells_before` (`<=`), shifting its block a hint-width right of where the primary caret sits. Carets now use a strictly-before rule (`inlay_cells_strictly_before`), matching `cursor_screen_pos` and VS Code: caret at the anchor sits left of the hint, a character at the anchor paints right of it. (`src/widgets/editor.rs`)

2. **Clicking the Testing panel's border ran a whole suite.** `hit_at` bounded y but not x against the bordered rect, so the left border column beside a suite header satisfied the play-glyph threshold. Border columns are now excluded. (`src/widgets/testing.rs`)

3. **Make Root during an in-flight COMMITS fetch stranded the graph on Loading.** A refresh arriving while `graph_fetch_inflight` was set got silently dropped; the stale reply was then discarded by the root-tag check, so nothing repainted the panel until HEAD next moved. The swallowed refresh is now latched (`graph_refetch_queued`) and re-fires the moment the stale reply drains. (`src/app/mod.rs`)

4. **A one-row alt-screen selection on a repeated row (divider rules, continuation markers) re-anchored to the nearest identical row, not its own copy.** The anchor now also captures the row above and below at selection time; re-anchoring prefers candidates whose neighbours match too, then smallest movement. Highlight and copied text stay on the same row again. (`src/widgets/terminal.rs`)

5. **Clipboard test guard: cross-process exclusion.** The in-process `CLIP_LOCK` mutex cannot serialize tests across cargo-nextest's process-per-test model, and the macOS pasteboard is one shared system resource; two test processes raced and one saw the other's sentinel. The guard now also takes an exclusive advisory `File::lock()` on a well-known temp file (kernel-released on process death), and snapshots the pasteboard only after both locks are held. (`src/clipboard.rs`)

Each fix landed strict RED to GREEN: the new/extended tests failed exactly as predicted before the fix.

## Housekeeping

- Version 0.1.664, release notes replaced for this bump, ARCHITECTURE.md updated for the three behavioural changes.
- `.gitignore` gains `/.config/nextest.toml` (per-machine thread cap, same rule as `/.cargo/config.toml`).

## Validation

- clippy: zero warnings
- `cargo nextest run`: 2872 run, 2872 passed (10 skipped)
- `cargo test`: 2866 + 6 passed, zero failures

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved editor caret placement around inlay hints, including secondary and collaborator carets.
  - Prevented Testing panel border clicks from launching tests or opening source files.
  - Queued commit-graph refreshes now complete after in-progress fetches finish.
  - Improved terminal selection anchoring when repeated rows or overlays are present.
- **Documentation**
  - Updated architecture documentation to reflect refresh queuing and improved terminal selection behavior.


vitali87:
@greptile review

vitali87:
@coderabbitai review

PR #13, opened 2026-08-02
fix: themes/testing span review — disabled-runner guard, palette-true uncaptured source, preview theme rebuild, non-iTerm2 canvas bg

vitali87:
Adversarial review of span 580624f..d9070bd (testing-M7 registry, ten themes, themed terminal ANSI, frame prefill, quickfix grep seeding) judged against current HEAD. Five reviewer findings adjudicated: four confirmed and fixed RED to GREEN, one refuted (the stale search include-glob on terminal seeding was already fixed at HEAD by 2220c2b, where `Search::seed` replaces include and exclude unconditionally).

## Fixes

1. **Disabling a test-runner extension did not stop croft running that runner.** The worker's four match arms used a `_` catch-all that collapsed `Some(Cargo)` and `None` into cargo, and the four run entry points (`run_all_tests`, `run_test`, `run_suite`, `run_named_test`) gated only on `is_busy()`. With the Rust runner toggled off, the gutter play glyph still shelled `cargo test`; with only a Python project and its runner off, cargo ran in a directory with no Cargo.toml. Now the entry points refuse with a status (`testing_runner_available`), and the worker takes `runner_for` through a let-else with exhaustive matches, so `None` refuses and a future `Runner` variant is a compile error instead of silently cargo. (`src/app/mod.rs`, `src/testing/worker.rs`)

2. **Uncaptured source bytes hardcoded the Base16 foreground.** `highlight_text`'s uncaptured branch painted literal `#c0c5ce`, so under Gruvbox or Dracula the bytes tree-sitter's queries don't capture rendered cold slate beside cream tokens. The palette is now snapshotted once per pass and both branches style through it (`highlight_text_with_palette`, hermetically testable). The markdown preview's fenced-code gap filler and unhighlighted-fence fallback had the same literal and now use the theme's code foreground. (`src/highlight.rs`, `src/markdown.rs`)

3. **The markdown preview kept the previous theme after a theme switch.** Its lines bake theme colors at build time and rebuilt only when `built_seq` fell behind the buffer's edit counter, which a theme switch never bumps. `rehighlight_for_theme` now rebuilds an open preview, and `apply_theme` assigns the new theme onto each editor before re-highlighting so the rebuild reads the right theme. (`src/widgets/editor.rs`, `src/app/mod.rs`)

4. **Image, PDF placeholder, CSV, and diff canvases were host-black islands on Ghostty/Kitty.** They filled with `bg(Color::Reset)` to inherit the iTerm2 `SetColors` session background, a mechanism those hosts ignore, and the fill overrides the non-iTerm2 frame prefill because ratatui applies any `Some` bg (`Some(Reset)` included). A new `canvas_bg` helper keeps `Reset` on iTerm2 and paints `theme.editor_bg()` everywhere else, threaded through all four canvas painters including the unified-deletion diff. (`src/widgets/editor.rs`)

Findings 1-3 landed strict RED to GREEN with the failure observed first. Finding 4's fix required the signature change itself, so its test pins the new `canvas_bg` contract plus the actual cell fill; the pre-fix behavior was verified against ratatui's `set_style` semantics.

One existing test (`clicking_the_gutter_play_glyph_runs_that_test`) used a manifest-less tempdir and relied on the old cargo fallthrough; it now writes a minimal Cargo.toml since it pins the click-to-run wiring, not runner detection.

## Housekeeping

- Version 0.1.665, release notes replaced for this bump, ARCHITECTURE.md updated for all four behaviour changes.

## Validation

- clippy: zero warnings
- `cargo nextest run`: 2877 run, 2877 passed (10 skipped; baseline 2872 + 5 new tests)
- `cargo test`: 2871 + 6 passed, zero failures

## Summary by CodeRabbit

* **Bug Fixes**
  * Test commands remain idle and show a clear status when no test runner is configured.
  * Syntax highlighting and Markdown code blocks now follow the active theme, including unsupported languages and uncaptured text.
  * Markdown previews recolor immediately after theme changes.
  * Image, spreadsheet, PDF, and diff views use appropriate theme backgrounds across supported hosts.
* **Documentation**
  * Updated architecture and release notes to reflect these improvements.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@coderabbitai review

PR #14, opened 2026-08-02
fix: terminal span review — clamp panic, clock-anchored timestamps and annotations, copy-mode drift sync, sticky-header sanitization

vitali87:
Adversarial review of span fb8ed58..20272ce (terminal annotations, scrollback-to-editor, sticky command header, arrival-time gutter, click-to-move prompt cursor, copy-mode/selection re-anchoring) judged against current HEAD. Eight reviewer findings personally verified; all eight fixed, each RED to GREEN with the failure observed first (two via a temporary old-semantics body, noted below).

## Fixes

1. **Click-to-move-cursor could panic and take the app down.** `prompt_click_arrows` computed `vc.clamp(b_col, end_col.max(cur))`; a prompt row rewritten shorter than the recorded 133;B column (a backgrounded `\r\x1b[K` progress writer, a pane shrunk under the prompt) inverted the bounds and `u16::clamp` asserts. The lower bound now caps at the upper, degrading the gesture instead. (`src/widgets/terminal.rs`)

2. **The timestamps gutter erased itself on pagers and froze in long sessions.** The reader thread keyed arrival stamps on `history_size + cursor line` and treated any drop as an ED 3 wipe: entering `git log`/vim/htop (alt screen has no history) cleared every stamp and exit re-stamped the whole scrollback with the exit time; any multi-line progress redraw wiped the gutter; and past scrollback saturation `history_size` freezes, so ids collapsed to screen positions and the stall marker compared unrelated rows. Stamps now key on the pane's scroll clock (`ScrollClock`, extracted and shared with the reader thread; both sides lock term then clock), the alternate screen stamps nothing, and a cursor-up redraw re-stamps only the re-arrived row. The gutter reader and `row_time` translate with the same clock. (`src/widgets/terminal.rs`)

3. **Annotations mis-pinned and froze.** The Cmd+K N prompt captured its grid line at OPEN but the commit sampled a fresh history baseline, so a note typed while output streamed landed rows below its text; and the translation used the saturating history delta, freezing the amber span in panes past 5000 lines while content slid underneath. `PaneAnnotation` now anchors on `(line_rec, clock_rec)` captured together when the prompt opens (`PromptKind::AnnotateTerminal` carries the clock), and every reader (`annotations_current`, `annotation_at`, render) translates by clock movement. (`src/widgets/terminal.rs`, `src/app/mod.rs`)

4. **Copy-mode keys teleported the selection while output streamed.** The App-owned `CopyModeState` held absolute grid lines and re-pushed them wholesale on every keypress, while the pane re-anchored the painted highlight to streaming content between keys. The state now carries its own clock stamp and every key first folds the clock movement into the stored coordinates, so motions continue from the highlight the user sees. (`src/app/mod.rs`)

5. **The sticky command header printed a raw newline into the frame.** A command spanning several grid rows (quoted/heredoc newline, or a soft wrap shown in a later-widened pane) is extracted with its rows joined by `\n`, and the header painted each char verbatim into a cell — the screen-corruption class the pdftoppm/DAP capture fixes exist to stop. The join is now a space. (`src/widgets/terminal.rs`)

6. **Clicking an annotated span left a stray one-cell highlight.** The annotation-popup branch returned before the shared mouse-up cleanup, so the zero-area selection the mouse-down planted stayed painted and the pane-drag latch stayed armed. The branch now runs the cleanup itself. (`src/app/mod.rs`)

RED evidence: findings 1, 4, 5, 6 failed live before their fixes (the clamp test panicked at the exact line; copy-mode selected `x-4` instead of `row-40`; a cell held `"\n"`; the click selection persisted). Findings 2 and 3 changed anchor plumbing, so their tests were demonstrated RED against a temporary restoration of the old semantics (history-keyed stamping with the clear; commit-time sampling and history-delta translation), then GREEN against the real fix.

## Housekeeping

- Version 0.1.666, release notes replaced for this bump, ARCHITECTURE.md terminal bullet updated (clock-keyed timestamps, open-time annotation anchors, copy-mode drift fold).
- New test hooks: `push_mark_for_test`, `stamp_chunk_for_test`, `line_time_entries_for_test`.

## Validation

- clippy: zero warnings
- `cargo nextest run`: 2887 run, 2887 passed (10 skipped; baseline 2879 + 8 new tests)
- `cargo test`: 2881 + 6 passed, zero failures

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal timestamp stability during extensive scrollback and alternate-screen transitions.
  * Kept annotations and copy-mode selections anchored correctly as new output streams.
  * Fixed prompt-click cursor handling on rewritten rows.
  * Clicking annotations now clears temporary selections before opening details.
  * Improved multi-line command headers and removed stale highlights.

* **Documentation**
  * Updated terminal architecture documentation to explain stable timestamps and annotation anchoring.

## Review responses (round 1)

- **CodeRabbit Critical (yank reads the selection before drift correction): refuted.** `selection_text()` applies the identical scroll-clock correction internally — it computes `delta = clock_now - sel_scrolled` under the term lock and shifts the stored coordinates before extraction (src/widgets/terminal.rs, `selection_text`), exactly as the `selection()` getter does. The yank therefore reads drift-corrected coordinates that equal what pushing the folded copy-mode state first would produce; there is no window between the fold at the top of the key handler and the read in which the clock can move, because no output is processed on this thread in between.
- **Nitpick (sanitize every control char in the sticky header): applied** in 6e07c18.
- **Nitpick (lock-order documentation and consistency): applied** in 6e07c18 — the order is documented as term → clock → line_times on `ScrollClock`, and the gutter now takes the clock reading before locking `line_times`.
- **Nitpick (clamp test asserts the degraded result): applied** in 6e07c18.

## Review responses (round 2)

- **Greptile P1 (annotation anchor capture races): fixed** in 948f9df. `begin_annotate_terminal` now captures the drift-corrected selection and its scroll-clock reading from one terminal snapshot (`selection_and_clock`), so PTY output can no longer land between the two reads and mis-pair them.
- **CodeRabbit Major/Minor (primary annotations painted and hit-tested over alt-screen content): fixed** in 948f9df — `render` builds no annotation spans and `annotation_at` resolves nothing while `TermMode::ALT_SCREEN` is active, with a regression test covering paint-suppress, no-hit, and reappearance after exit.
- **CodeRabbit Minor (copy-mode open pairing cursor/bounds with a separate clock read): fixed** in 948f9df via `cursor_bounds_and_clock`, the same one-snapshot capture.

## Review responses (round 3)

- **Greptile P1 (existing-annotation lookup mixes snapshots): fixed.** `annotations_at_clock` translates the annotations against the SAME clock reading captured with the selection, so output scrolling between the capture and the lookup can no longer shift them off the fixed selection line; the regression test reproduces T-Rex's exact interleaving (translated line -9 without the fix) and pins the corrected lookup.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@coderabbitai review

PR #15, opened 2026-08-02
fix: copy-mode span review — WRAPLINE-aware copy, capture jump in narrow panes, pane-bound copy mode, tiny-frame panics, host-scoped history

vitali87:
Adversarial review of span 769a8c8..fb8ed58 (copy mode, durable command history, capture triggers, undo close pane, session restore, OSC 9;4 gauge) judged against current HEAD. Six reviewer findings personally verified; all six fixed, plus one bonus crash the new tests flushed out. Every fix demonstrated RED first (a batched restoration of the old behaviors failed all five behavioral tests, each at its predicted assertion; the history-host fix is pinned by a new scope test).

## Fixes

1. **Soft-wrapped lines copied, stored, and pinned with a bogus newline.** `extract_selection_text` joined every grid row with `\n` regardless of alacritty's WRAPLINE flag, so a wrapped command entered the durable history with an embedded newline and typing it back from the popup ran only its first fragment on non-bracketed-paste shells. The join now consults the WRAPLINE flag on the wrapping row's last cell: soft wraps rejoin seamlessly (selection copy, history, sticky header alike), hard row breaks keep their newline. (`src/widgets/terminal.rs`)

2. **The CAPTURES jump could never match in a pane narrower than 60 columns.** The jump compared a fixed 60-char needle with `row.starts_with(needle)`, but a grid row's text is bounded by the pane width, so any captured line wider than the pane was reported "no longer in the scrollback" while visibly on screen. The matcher (extracted as `capture_row_matches`) now also accepts a row that is a prefix of the needle with a meaningful overlap. (`src/app/mod.rs`)

3. **Copy mode hijacked sibling panes.** `CopyModeState` had no pane identity: a click on another pane moved `active_terminal` while the mode stayed up, so the next keystroke folded the OTHER pane's independent scroll clock into the stored coordinates and painted a garbage selection there, leaving a stale green cursor on the original pane. The state now records its pane; a gesture that activates a sibling closes the mode (tmux-style), eagerly on the click and defensively in both the key handler and the apply path. (`src/app/mod.rs`)

4. **`e` was a dead key at a row's last word.** The end-of-word scan ran into the row's trailing padding and never advanced; `w` and `b` both wrap. `e` now hops to the next row's first word end like vim. (`src/app/mod.rs`)

5. **A tiny host window crashed croft twice over.** The command-history popup (and the two sibling popup geometries) clamped with inverted bounds below 40x10 (`Ord::clamp` asserts), and before even reaching that, a terminal pane squeezed to a sliver panicked inside alacritty's `Term::resize`. Both clamped: popup minimums cap at the frame, and `PtyTerminal::resize` floors at a 2x2 grid. (`src/app/mod.rs`, `src/widgets/terminal.rs`)

6. **Directory-scoped history conflated machines.** Entries stored only the OSC 7 path, so a command run at `/x` over an in-pane SSH session matched the Dir scope of a local pane at `/x`. `HistoryEntry` now carries the OSC 7 reporting host (serde-defaulted, so existing history files load unchanged), and the Dir scope filters on the (host, cwd) pair. (`src/command_history.rs`, `src/widgets/history_popup.rs`, `src/app/mod.rs`)

## Housekeeping

- Version 0.1.667, release notes replaced for this bump, ARCHITECTURE.md updated (WRAPLINE-aware extraction, pane-bound copy mode with wrapping word motions).

## Validation

- clippy: zero warnings
- `cargo nextest run`: 2895 run, 2894 passed + the one documented PTY load flake (`cmd_click_on_a_file_line_reference`, passes alone and under cargo test)
- `cargo test`: green (2888 bin + 6 cli on re-run)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved copy mode for wrapped lines, including navigation and copying across rows.
  * Copy mode now closes when switching terminal panes.
  * Fixed popup sizing in very small terminal windows.
  * Separated command history results by machine, including local-host aliases.
  * Improved terminal resizing and preserved soft-wrapped lines during text selection.

* **Documentation**
  * Updated architecture documentation for copy-mode behavior.

* **Release**
  * Updated the application version and current release highlights.

## Review responses (round 1)

- **Greptile P1 (history host read at drain, not completion): fixed** in 2a1fcfe — `FinishedCommand` now carries the OSC 7 host captured with the 133;D mark on the reader thread, alongside the cwd it already captured, so an in-pane ssh/exit between completion and drain can no longer misattribute the entry.
- **Greptile P1 + CodeRabbit Major (reverse-prefix capture match accepts lookalikes and continuation rows): fixed** in 2a1fcfe — the search moved into the pane as `find_captured_line`, which gates the reverse arm on the row actually soft-wrapping (WRAPLINE on its last cell) and skips continuation rows (previous row wraps) in both arms; the length heuristic is gone, and the regression test pins a non-wrapped lookalike rejecting and the wrapped first row winning over its continuation.
- **CodeRabbit Minor (local OSC 7 hosts hide legacy entries): fixed** in 2a1fcfe — `canon_host` collapses "", "localhost", and this machine's own hostname (short-name, case-insensitive) into one local value before the Dir-scope comparison, so shim-reported local entries and legacy host-less ones are one history while real remote hosts stay separate.
- **CodeRabbit Nitpick (persisted-schema coverage): applied** — the scope test now reloads the JSONL and asserts a host-bearing record round-trips and a pre-host record defaults to the local empty host.

## Review responses (round 2)

- **Greptile P1 (`e` stops before word ends): the motion is correct; the TEST's staging was environment-sensitive, now fixed** in 36e1231. The word-end scan resolves to each word's final cell (from column 0 on `alpha beta`, the extend loop stops at the `a` of `alpha`, column 4), and the test passes locally on every run. T-Rex's column-3 observation arises because the old staging jumped to the scrollback TOP (`g`) and assumed the fed text sat there — whatever the sandbox's shell printed first shifted the row under the cursor. The test now feeds onto a fresh row and navigates cursor-relative (`k`, `0`), so it is hermetic against prior shell output in any environment.

## Review responses (rounds 3–5)

- **CodeRabbit Nitpick (triplicated popup-clamp math): applied** in db930a1 — the frame-safe clamp lives once in `fallback_popup_rect(full, max_w_cap)`; the zoxide-jump fallback, the history popup, and the branch-picker fallback all call it.
- **Greptile P1 (zero-sized frame receives a 1×1 popup outside the frame): fixed** in bb051fa — a frame with no cells on either axis now gets an empty rect before any floor applies (the floors themselves predate the refactor; they existed to keep `Ord::clamp` from inverting). RED→GREEN: `a_zero_sized_frame_gets_no_fallback_popup` failed on the old floors and pins 0×0, 80×0, and the in-frame containment of a real tiny frame.
- **Greptile P1 (remote "localhost" histories collide with local): refuted.** The host field is the authority of an OSC 7 `file://` URI, and RFC 8089 §2 defines that authority: the string "localhost" "is interpreted as 'the machine from which the URI is being interpreted,' exactly as if no authority were present". `file://localhost/…` from any shell, local or over SSH, denotes "this machine" by specification; that is exactly why shell-integration shims fall back to it. Un-collapsing it would misread the URI and fragment the common local case (shim-reported "localhost" vs the machine's real hostname vs legacy empty hosts) to guard a machine misconfigured with the literal hostname `localhost`, which no hostname-string scheme can distinguish anyway. `canon_host` keeps the RFC semantics: "", "localhost", and this machine's own name are one local identity; every real remote hostname stays separate.
- **CodeRabbit Nitpick (`e` stops at column 0 on a run of blank rows): applied** in 0770a8a — the row hop became a scan that skips word-less rows down to the bottom and lands on the first word end found, with the next-row column 0 fallback kept for a grid with no further words. RED→GREEN: `copy_mode_e_skips_blank_rows_to_the_next_word_end`.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #16, opened 2026-08-02
fix(terminal): six inline-image and scrollback-key defects from the imgcat span review

vitali87:
Adversarial review of the imgcat/ANSI-palette/scrollback-keys span (769a8c8) judged against current main. Six confirmed defects, each fixed RED→GREEN (v0.1.668).

## Fixes

- **Kitty image id collision**: `KITTY_ID_TERMINAL` and `KITTY_ID_SCM_BADGE` both sat at `KITTY_ID_BASE + 19`, so on Ghostty/Kitty the SCM badge's 2s keepalive and the pane picture's per-frame re-emit replaced each other's placement in turn. The terminal slot moved to the free `+25`; `kitty_image_ids_are_pairwise_distinct` pins the whole id table.
- **Frozen anchor past scrollback saturation**: `pane_images()` derived the current line from the `history_size` delta, which saturates at the scrollback cap while content keeps rotating — in any long-lived pane the picture froze onto a viewport row over all later output. Images now record `clock_rec` from the shared ScrollClock and translate exactly like annotations. `an_inline_image_rides_scrollback_past_saturation` demonstrated the freeze (off by exactly the rotation) before the fix.
- **Clear left the picture floating**: neither the pane's Clear nor a program-emitted ED 3 (`clear` via modern terminfo) touched the captured images, so the picture floated over the fresh prompt. `clear_screen_and_scrollback` now drops them, and the reader thread detects a scrollback wipe (history shrinking across a chunk — only ED 3/RIS do that) for the typed-`clear` path, verified end-to-end through a real PTY.
- **Full re-decode per scrolled row**: the overlay's layout key contains the anchor row, so every scrolled line re-ran decode + Lanczos3 resize + PNG encode of the original payload on the render thread, stalling the UI for the pane height on any real photo. A pure anchor move (same seq/pane/cell size) now reuses the baked payload; `a_scrolled_inline_image_reuses_its_baked_payload` pins the bake count across a move (OLD-DEMO run confirmed the old path rebaked).
- **Picture overpainted the pane's right-click menu**: the terminal overlay baked at Kitty z=0 and its payload never yielded to `context_menu`, unlike the editor preview. It now bakes at `KITTY_Z_BELOW_TEXT_AND_BG` (pane default-bg cells are `Color::Reset`, so the picture still shows) and suppresses itself under an open menu on iTerm2/Sixel via the same `context_menu_covers_rect` rule.
- **Shift+End swallowed in the alternate screen**: the bottom-reset arm hardcoded `true` while `reset_scrollback` had no alt-screen guard, contradicting the block's own comment and all three sibling motions — vim/less/htop never received the chord. `reset_scrollback` now returns false in the alt screen and the chord encodes through to the program (pinned via a new test-only PTY write capture).

## Validation

Staged ladder: each fix RED→GREEN with a targeted test, clippy zero warnings, full suite 2904 tests: 2903 passed, 1 failed — `cmd_click_on_a_file_line_reference_in_the_terminal_opens_the_file_there`, the documented PTY load flake (passes in isolation).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inline-image positioning during extensive scrollback activity.
  * Cleared stale terminal images when scrollback, panes, or programs reset the screen.
  * Improved image rendering performance while scrolling.
  * Fixed image layering around context menus and overlays.
  * Resolved Kitty image-ID conflicts.
  * Fixed Shift+End handling for alternate-screen applications.

* **Documentation**
  * Updated terminal architecture documentation with scrolling and image-handling details.

* **Chores**

## Review responses

- **Greptile P1 (empty-history wipes preserve images): fixed** in 53ce0c8 — the history-shrink heuristic was replaced by a byte-level `WipeSniffer` (ED 2 / ED 3 / RIS with a bounded cross-read carry) that drops images by erased region, so a wipe with nothing yet scrolled into history still clears them. RED→GREEN through a real PTY (`a_zero_history_screen_wipe_drops_captured_images`).
- **Greptile P1 (wipe followed by alt entry in one chunk kept the images): fixed** in f3fefc2 — the sniffer became positional: it tracks DECSET/DECRST 47/1047/1049 through the stream (seeded from the terminal's mode at chunk start), so a wipe counts exactly where it landed and `clear && vim` drops the primary images even though the chunk ends inside the alt screen. RED→GREEN through a real PTY holding the alt screen across the chunk boundary.
- **CodeRabbit Minor (process-global bake counter races parallel tests): fixed** in f3fefc2 — `FIT_IMAGE_BAKES` is thread-local; the bake runs on the asserting test's own thread, so the count is isolated under both thread-parallel `cargo test` and process-parallel nextest.
- **CodeRabbit Minor (docs said "alt-screen chunks exempt"): fixed** in 375e87a — the ARCHITECTURE sentence now states the exemption positionally, matching the implementation.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #17, opened 2026-08-02
fix(terminal): four quick-select and trigger-scanner defects from the b1dbc63 span review

vitali87:
Adversarial review of the quick-select/triggers/broadcast span (b1dbc63) judged against current main. Four confirmed defects, each fixed RED→GREEN (v0.1.669).

## Fixes

- **Trigger scanner judged the whole chunk by its final mode**: the alt-screen gate read `t.mode()` after the entire 64KB chunk had advanced, so alt-screen content could fire false notifications when the chunk ended on the primary screen, a banner printed just before entering a full-screen app was silently dropped, and skipped chunks desynced the scanner's line buffer into splicing phantom lines (or wedging in a string state forever). `TriggerScanner` now tracks DECSET/DECRST 47/1047/1049 and RIS positionally through the byte stream — the same principle the image WipeSniffer established — and the reader feeds it every chunk unconditionally. An alt-screen boundary also resets the pending partial line on both edges. Unit RED covered all three failure modes plus multi-parameter DECSET lists; a real-PTY RED pinned the banner-before-vim gesture end to end.
- **Quick-select hijacked whichever pane became active**: `QuickSelectState` had no pane binding, so a pane switch left stale labels painted on the old pane while a typed prefix pushed its coordinates onto the new pane, and an UPPERCASE commit pasted the old pane's scraped text into the new pane's shell. The state now records its origin pane; a per-frame invariant in `sync_focus_flags` closes the mode whenever anything moves `active_terminal` (covering all seventeen assignment sites at once), with immediate closes at the key handler and the sibling-click site.
- **Labels squatted on fixed rows while content streamed**: hint lines were raw absolute grid lines captured once, never re-based, so under streaming output the gold label drifted off its match and the typed label copied text no longer on screen. Hints now carry the scroll-clock reading of their capture (one `visible_lines_and_clock` snapshot — coordinates and clock from a single term lock) and the render loop re-bases them every frame, the annotation model.
- **Label overflow dropped the bottom of the screen**: with more matches than labels (>676), `take(n)` kept the top-left matches and the reverse handout gave the cheapest label to the 676th-from-top — the exact inverse of the documented bottom-most-first priority, leaving everything near the prompt unlabelled. The pairing moved into `quick_select::label_matches`, which keeps the BOTTOM matches and gives the bottom-most the cheapest label (OLD-DEMO run confirmed the inverted behavior); the status line now counts any unlabelled overflow.

## Validation

Staged ladder: each fix RED→GREEN with a targeted test (unit scanner tests, two real-PTY trigger tests, three app-level quick-select tests, one pure-core overflow test), clippy zero warnings, full suite 2911 tests: 2909 passed, 2 failed — `cmd_click_on_a_file_line_reference` and `gui_path probing_a_real_shell`, both documented parallel-load flakes that pass in isolation.

## Review responses

- **Greptile P1 (CAN/SUB retained as CSI parameters): fixed** in ca01a42 — CAN/SUB now cancel an incomplete CSI and a nested ESC begins a new escape (VT100 semantics), in BOTH byte sniffers (`TriggerScanner` and the image `WipeSniffer`), so a cancelled sequence can no longer swallow the following alt-screen entry or wipe. RED→GREEN unit tests in each.
- **Greptile P1 (CAN/SUB don't cancel control strings): fixed** in 2c26617 — `InString` and `StringEsc` now abort to ground on CAN/SUB (ECMA-48), so an unterminated OSC/DCS/APC/PM/SOS cancelled mid-body releases the scanner instead of swallowing ordinary output and the alt boundary until an unrelated BEL/ST. The regression test sweeps all five string families times both cancel bytes; T-Rex had confirmed the defect existed before this PR as well, latent until alt tracking made the string states load-bearing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Quick-select labels now stay aligned with matches while terminal output scrolls.
  - Quick-select closes correctly when switching panes or changing focus.
  - Label overflow is handled more reliably, with clearer counts for unlabelled matches.
  - Trigger detection now works correctly around alternate-screen transitions.
  - Interrupted terminal control sequences no longer prevent later output and triggers from being processed.
- **Tests**
  - Added regression coverage for scrolling labels, pane switching, label overflow, and terminal-screen transitions.


vitali87:
@greptile review

vitali87:
@greptile review

PR #18, opened 2026-08-02
fix(terminal): four shell-integration defects from the 4a7b33d span review

vitali87:
Adversarial review of the shell-integration cluster (4a7b33d) judged against current main. Four confirmed findings, each fixed RED→GREEN (v0.1.670).

## Fixes

- **Mark GC was dead code and a clear inverted the drift math**: the retain predicate reduced algebraically to `line_rec + hist_rec >= 0` (always true), and `clear_screen_and_scrollback` zeroed `history_size` without touching marks, so stale PromptStart/CommandStart/CommandEnd trios later resurfaced as phantom decoration dots on unrelated rows — clicking one copied or re-ran the wrong text. OSC 133 marks now ride the shared scroll clock exactly like annotations and images (`(line_rec, clock_rec)`, stable past scrollback saturation, GC against the real floor); the pane's Clear drops them and the reader-thread WipeSniffer path retains marks by erased region alongside images. RED covered both the clear (decorations must vanish) and saturation (the mark must follow its prompt through ring rotation).
- **bash array `PROMPT_COMMAND` defeated the DEBUG-trap guard**: `${PROMPT_COMMAND[*]}` joins with the first char of IFS — a space — so in the array form (bash 5.1+, ble.sh and prompt frameworks) the guard's `;entry;` pattern never matched and every empty Enter emitted a spurious 133;C + 133;D, creating a phantom decoration that hijacked Cmd+K Shift+R/C/S. The trap now scopes `IFS=';'` locally before the join. RED reproduced end-to-end through a real bash 5.2 with `PROMPT_COMMAND=(mytheme)`.
- **Remote-reported cwds were trusted for local actions**: `next_terminal_cwd` (split-in-cwd) and `save_terminal_session` gated the OSC 7 path only on local `is_dir()`, so an in-pane ssh session reporting `file://prod-db-1/var/log` opened local splits (and restored sessions) in the local `/var/log`. Both now route through `local_shell_cwd()`, which requires the reporting host to be this machine via the command history's canonicalizer (RFC 8089 localhost semantics included); the history popup already paired cwd with host correctly.
- **zsh/bash shims emitted raw `$PWD` while the sniffer percent-decodes**: a real directory named `a%41b` was reported as `aAb`, corrupting the cwd for history scoping and splits. Both shims now escape `%` as `%25` (fish always escaped); the shim files self-update on content change. RED drove a real zsh through the shim inside a literal `a%41b` directory.

## Validation

Staged ladder: each fix RED→GREEN with a targeted test (two real-shell e2e tests, one real-PTY host-gate test, two mark-model tests), clippy zero warnings, full suite 2919 tests: 2918 passed, 1 failed — `cmd_click_on_a_file_line_reference`, the documented PTY load flake (passes in isolation).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal command markers, including placement, clearing, and persistence through scrollback changes.
  * Removed phantom commands from Bash array-based prompt configurations.
  * Prevented remote SSH working directories from appearing as local paths.
  * Fixed percent-sign handling in Zsh and Bash working-directory paths.
  * Cleared stale terminal marks and image references when content is erased.
  * Improved local host detection for terminal working directories.
* **Documentation**
  * Updated architecture documentation for terminal command decorations.
* **Release**
  * Updated the package version to 0.1.670.

## Review responses

- **Round 1 (Greptile P1): first-label hostname matching let `prod.evil.com` pass as local `prod`.** Fixed in b00635b: `is_local_hostname` now requires a full-hostname match after lowercasing and one mDNS `.local` strip; a bare dotless reported name may still match the local first label, but a dotted FQDN never matches by first label alone.
- **Round 2 (Greptile P1, security): any foreground process could impersonate a local cwd report.** Fixed in a3f3db6: `local_shell_cwd()` additionally required the shell to own the tty (`foreground_is_shell`) so a claim printed by a foreground job (an ssh client) was not trusted.
- **Round 3 (Greptile P1, security): a stale claim became trusted once the forging job died and the shell reclaimed the tty.** Fixed in 2f3ae29 by sampling tty ownership when the report was captured rather than when it was consulted.
- **Round 4 (Greptile P1, security, T-Rex verified): capture-time sampling still races buffered parsing — the reader can parse the claim's bytes after the job died, and the sample then sees the shell at its prompt.** Fixed in 5196ade by dropping timing-based provenance entirely: PTY bytes carry no author, so no tty-ownership sample at any moment is sound. A local claim is now trusted only when it names the directory the kernel says the shell is in (`cwd_of_pid`, canonicalized for macOS symlinks). A forged claim equal to the kernel's truth grants nothing, and one that differs is rejected. The regression test stalls parsing with the term lock (the production delay) across the forging job's death and asserts the buffered claim stays untrusted.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #19, opened 2026-08-02
fix: hyperlink-span review — OSC 8 scheme allowlist, Ctrl+F passthrough, pane-bound clock-anchored terminal find, encoding-reopen history

vitali87:
Adversarial review of the span 90b5e39 / 5b8ddd1 / 31b5071 (editor redo, find-in-terminal, OSC 8 hyperlink click / bell / wide-char copy) judged against current main. Five confirmed findings, each fixed RED first:

- **OSC 8 hyperlink clicks opened any URI scheme (security).** The cell carries whatever URI the printing program chose while showing unrelated text, and `open`/`xdg-open` would launch a `file://` or custom-scheme target off one disguised click; the plain-text fallback directly below it was already https-only, and on a drop-relay session the URI was forwarded to the local Mac. `open_detected_url` now refuses non-web links and surfaces the real destination in the status line.
- **Plain Ctrl+F was stolen from the terminal child.** `is_editor_find_key` accepts Ctrl or Cmd, so the pane's find bar swallowed the 0x06 control byte programs bind (vim page-forward, emacs forward-char) with no way to send it at all. A new `is_terminal_find_key` keeps Cmd+F everywhere and accepts Ctrl+F only off macOS, matching VS Code's terminal find defaults and iTerm2 (croft's iTerm2 setup already relocates iTerm's own Find off Cmd+F to forward it).
- **Find-in-terminal was not bound to its pane.** Every operation targeted whichever pane was active, so switching panes stranded the highlight on the old pane forever and applied the stored jump anchor to the wrong grid. Find now carries its pane like quick-select and copy mode: a per-frame invariant plus the sibling-click path close it, and closing clears every pane's highlight.
- **The active match was a raw grid line.** Streaming output shifted the stored line so the bright highlight detached from its text and Enter/Shift+Enter walked from a phantom row, re-finding the same occurrence. The match is now stamped with the scroll clock (one-lock `grid_lines_and_clock` snapshot) and both the render pass and navigation re-base it first — the annotation/hint/image model.
- **Reopen with Encoding kept both history stacks.** Every other whole-buffer swap clears them; here a redo popped after the swap reinstated the entire buffer as decoded under the old encoding. Both stacks are now cleared.

Version 0.1.671; release notes replaced; ARCHITECTURE updated for the scheme allowlist, the platform-split find chord, pane binding, and the clock-anchored match. Full suite green (2928 collected; the four failures under the full parallel run are the documented PTY-load flake family, each verified passing in isolation).

## Summary by CodeRabbit

* **New Features**
  * Added terminal scrollback search with platform-specific shortcuts: Cmd+F on macOS and Ctrl+F elsewhere.
  * Search highlights and navigation now remain synchronized as terminal output streams.
  * Find state is cleared when switching panes.

* **Bug Fixes**
  * Restricted terminal links to secure or standard HTTP(S) destinations.
  * Restored macOS Ctrl+F input forwarding to terminal applications.
  * Cleared undo and redo history when reopening files with a different encoding.

* **Documentation**
  * Updated terminal architecture and shortcut documentation.

## Review responses

Greptile's free-plan credits are exhausted this billing period, so per the review-bot pecking order CodeRabbit gates this PR.

- **CodeRabbit: clearing a pane left stale find state.** Fixed in e7b1bbb: `clear_terminal_at` drops the anchored match and active highlight and recounts over the emptied grid. The regression test for this exposed a pre-existing bug: `clear_screen_and_scrollback` ran ED 3 before ED 2, and alacritty's primary-screen ED 2 scrolls the viewport INTO history, so the "cleared" screen contents survived in scrollback. The order is now ED 2 then ED 3 (xterm.js / VS Code's clear order).
- **CodeRabbit: the encoding-reopen test used ASCII content that decodes identically under both encodings.** Fixed in e7b1bbb: the file now holds 0xC2 0xA2, which decodes to one char as UTF-8 and two as Windows-1252, and the test asserts both decodes around the history-clearing assertions.


vitali87:
@greptile review

vitali87:
@coderabbitai review

PR #28, opened 2026-08-03
fix(remote,outline): keep the background install off the attached session's terminal and stop the outline drain discarding the live reply

vitali87:
## Adversarial span review: background remote install and the Outline drain

Round 16 of the standing bug-hunt loop. An adversarial reviewer examined the span `f80d76a^..8867cbe` (inline Kitty images, outline seq drop, instant remote attach, dtach repaint) against current HEAD. It reported three findings. **Two are fixed here; the third is refuted below.** v0.1.678.

### 1. The background install reads the terminal the attached session is reading (`src/remote.rs`, `src/remote_bulk.rs`)

`ssh` forwards the caller's stdin to the remote command unless `-n` is given. On the launch-now path the caller's stdin is the tty that the attached remote croft has in raw mode, so two processes `read()` the same terminal and each keystroke reaches exactly one of them: typing into the remote session silently vanishes for as long as the background command runs.

Verified against a real host rather than reasoned about:

```
$ printf 'AAAABBBBCCCCDDDD' > probe
$ { ssh    host 'sleep 1'; echo -n "REMAINING:["; cat; echo "]"; } < probe
REMAINING:[]
$ { ssh -n host 'sleep 1'; echo -n "REMAINING:["; cat; echo "]"; } < probe
REMAINING:[AAAABBBBCCCCDDDD]
```

All 16 bytes are drained even though the remote command never reads stdin. The exposure is worst on the fallback path: the remote `cargo install` holds the tty for the whole multi-minute build.

Every ssh invocation on the background thread now goes through a builder that passes `-n` and closes all three streams (inherited stderr also wrote ssh diagnostics straight onto the alt screen), and `run_command_streaming` closes stdin for the non-ssh commands it runs — rsync and zigbuild — while keeping stdout/stderr piped so the install log still streams. rsync deliberately gets no `-n`: there it means dry-run.

The interactive `run_remote_croft` is the one caller that must keep the terminal, and it does.

RED tests: `background_ssh_never_reads_the_attached_sessions_stdin` and `every_lane_ssh_command_gives_up_the_callers_stdin`. The second exists because the first covers the wrong builder on its own — the multi-minute `cargo install` runs through `remote_bulk::ssh_command`, not `SshControl::command`, and was previously protected only by a single untested line. Verified RED by removing the flag.

### 2. The Outline drain discarded the reply it had just received (`src/app/mod.rs`)

`drain_lsp_document_symbols` collapsed a tick's replies to whichever **arrived last** and threw the rest away, despite a doc comment claiming the newest per file wins. This needs no out-of-order delivery to lose work: switch files while a slow reply for the previous file is in flight, both land in one drain window, and the stale reply for the file you left evicts the active file's. `apply_outline_symbols` then rejects the survivor on path, and `sync_outline` never re-requests because its `(path, seq)` key has not changed. The Outline and breadcrumb stay tree-sitter-only until the next edit.

The fix removes the pre-selection entirely rather than replacing it with a better rule. Every reply is offered to `apply_outline_symbols`, which is the only place that knows which request is outstanding.

**Note for reviewers, since the first attempt at this was wrong.** The obvious replacement — keep the highest seq per path — is also broken, and would have reproduced the exact bug being fixed. `seq` is `tab.edit_seq`, a per-`Editor` counter that starts at 0 (`src/widgets/editor.rs:1698`), and `split_editor_dir` builds a fresh `EditorTabs::default()`. After a Cmd+\ split or a close-and-reopen, the outstanding request carries a *lower* number than a reply still in flight from the old group, so "highest wins" keeps the stale one and the exact-match guard rejects it. Ties are the same story: duplicate requests at one seq are routine (tabbing A → B → A with no edit re-requests), and an empty reply arriving alongside a good one must not be allowed to shadow it. Offering everything to the guard makes arrival order, seq magnitude and ties all irrelevant.

`an_outline_reply_is_matched_by_request_not_by_the_larger_seq` pins the invariant that makes size-based selection wrong. Stated plainly: it does **not** fail on the old drain code, because the guard it pins was already correct — it is there to stop the size-based rule being reintroduced.

### 3. Refuted: the "Updating" indicator does not stick (`src/remote.rs`)

The reviewer reported that an interrupted install leaves `~/.cache/croft/updating` set, pinning the remote's indicator at "Updating" forever, because only a successful activation removes it.

The marker is genuinely left behind — but it does not persist, and the proposed reap makes things worse. If an install dies, the remote's install-stamp is still stale, so the *next* attach finds `remote_install_needed == true`, never reaches the up-to-date branch, re-marks, reinstalls, and `activate_command` clears the marker. The stale marker self-heals on the next attach, and the "Updating" shown in the meantime is honest, because an install really is running.

That leaves the up-to-date branch able to fire only when a marker exists while no install of our stamp is needed — which is precisely the case where **another checkout is mid-install against the same host** (`~/.cache/croft/updating` is host-global, and this workspace routinely runs two checkouts). Reaping there deletes a live install's marker, and every attached croft that had announced `InProgress` sees it vanish without the stamp moving, reporting "Background croft update failed" about an install that is still running and will succeed.

I had implemented the reap before working this through; it is reverted. No change ships for this finding.

### Reviewer areas probed clean

Reported as unbreakable and spot-checked here: the Kitty image-eviction path skipping `terminal.clear()` (the one explicit dependant, `flush_run_debug_icon_overlay`, still full-redraws via the main loop's OR chain); `consume_any_image_clear`'s eager array evaluation, which strictly fixes an older short-circuit that left later latches armed a frame too long; minimap `placement_moved`; and the dtach repaint latch, where crossterm 0.29 emits `Event::Resize` on every SIGWINCH without comparing sizes, so a same-size reattach still reaches `on_resize`. `SOURCE_STAMP_INPUTS` covers every `include_bytes!`/`include_str!` target.

### Verification

Both stdin tests observed failing before their fixes, then green. Clippy at zero. Full suite: cargo exit 0, **2994 passed, 0 failed, 10 ignored** — the previous baseline of 2991 plus exactly these three tests, with no new ignores or skips.

**Unrelated, flagged not fixed:** the repo is already fmt-dirty against its own pinned rustfmt (1.9.0, per `rust-toolchain.toml`) — 28 diffs at HEAD in files this branch never touches, such as `src/git.rs`. Confirmed by running `cargo fmt --check` on a clean worktree at HEAD. This branch adds zero new drift, so the count is still 28. Sweeping it belongs in its own PR rather than buried under two bug fixes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Background remote operations no longer consume terminal keystrokes or display unwanted diagnostics.
  - Outline views now retain language-server symbols for the active file when responses arrive out of order.
  - Improved reliability for remote and bulk SSH operations.

- **Documentation**
  - Updated architecture documentation to reflect remote session input handling and outline response processing.

- **Release**
  - Updated the package version and release highlights.

## CodeRabbit round 1 (4 items: 1 Major confirmed, 2 Minor confirmed, 1 nitpick answered)

**Major — `-n` broke the bulk lane's throughput probe. Confirmed, and introduced by this PR.** `measure_throughput` is the one lane command that *writes* to ssh: it pipes `THROUGHPUT_PROBE_BYTES` (768 KB) into a remote `cat` and times it. `-n` points ssh's stdin at `/dev/null` and drops the pipe's read end, so the probe would have measured nothing, `establish` would have fallen back to `BWLIMIT_FLOOR_KBPS`, and every background install would have crawled at the 256 KB/s floor.

The builders are now split: `ssh_command` keeps `-n` and closed stdin for background work, `probe_ssh_command` keeps the stdin it feeds, and both share a private base so the option list stays in one place. Terminal isolation is not weakened by the probe, because that caller replaces stdin with its own pipe either way. RED test: `the_throughput_probe_keeps_the_stdin_it_feeds`.

Worth recording how this got in: `remote_bulk.rs` was changed *after* this branch's pre-push adversarial review had already run, in response to that review's own finding that the `cargo install` path was unprotected. Its "`-n` breaks no caller" clearance had covered `SshControl::background_command`, a different builder from the one I then went on to change. A fix written after the review is a fix nobody reviewed.

**Minor — the `-n` claim in `docs/ARCHITECTURE.md` overstated. Confirmed.** `BulkLane::rsync_ssh_arg` builds rsync's remote shell without `-n`. Terminal isolation there comes from `run_command_streaming` spawning rsync with stdin closed, which its nested ssh inherits. The line now distinguishes the two mechanisms and names the probe as the exception.

**Minor — the outline architecture line described a rule that no longer exists. Confirmed.** It still said replies are "reduced to the newest per file", which was the pre-selection removed in this PR after the pre-push review showed it was wrong. Stale text I left behind when reverting. It now states that the drain forwards every reply to `apply_outline_symbols`, which filters on active path and exact edit-seq equality, and says why no pre-selection is correct.

**Nitpick — test `drain_lsp_document_symbols` itself rather than `apply_outline_symbols`. Correct in principle, not done, reason stated.** Driving the drain means injecting replies into `LspManager`, and `LspManager::new` spawns a tokio runtime, starts `worker_loop`, and loads the user's real prefs to build the server registry. Standing that up in a unit test trades a hermetic suite for coverage of a function whose body is now a bare `for` loop with no selection logic left in it. The behaviour that can go wrong lives in `apply_outline_symbols`, which is what the test exercises. A genuine end-to-end test here needs an LSP stub server, and that harness remains unbuilt.

Re-verified after these fixes: clippy at zero, `cargo fmt --check` still at the pre-existing 28-diff baseline with nothing added, full suite cargo exit 0, **2995 passed, 0 failed, 10 ignored** — the 2994 above plus exactly the one new probe test.


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

Issue #37, opened 2026-08-04
Tests mutating $HOME race with $HOME readers, and a failed external reload is silently dropped


vitali87:
Two separate defects, found while diagnosing a one-off suite failure of
`widgets::editor::tests::pdf_reload_keeps_the_current_page`.

## 1. Tests that repoint the process-global `$HOME` race with every test that reads it

`src/pdf.rs:404` builds the rasteriser's scratch directory from
`session_state::dirs_cache_croft()`, which resolves `$HOME` at call time.
Three tests mutate `$HOME` for the whole process:

- `src/widgets/file_finder.rs:1016`
- `src/app/tests.rs:12409`
- `src/widgets/zoxide_jump.rs` (fixed in v0.1.681: the test now exercises a
  pure `collapse_home(path, home)` and touches no env var)

`HOME_LOCK` serialises those mutators against each other, but nothing
serialises them against readers, so any test thread resolving `$HOME` during
the window sees the substituted value.

The zoxide instance pointed `$HOME` at `/Users/v`, which does not exist and
cannot be created without root, so a concurrent PDF render failed outright:

```
$ HOME=/Users/v ./target.noindex/debug/deps/croft-<hash> \
    widgets::editor::tests::pdf_reload_keeps_the_current_page --exact
PDF render failed: Permission denied (os error 13)
```

The two remaining mutators point `$HOME` at temp directories that do exist,
so the failure mode is narrower: a concurrent render writes into the fake
home's cache dir, and the read fails only if the temp tree is removed between
`pdftoppm` writing the PNG and `std::fs::read`. Still a real race.

The durable fix is to stop mutating `$HOME` in tests at all, the way
`collapse_home` now does: take the value as an argument and let the thin
env-reading wrapper stay untested.

## 2. A failed external reload is silently discarded

`EditorTabs::reload_externally_changed_tabs` (`src/widgets/editor.rs:8549`)
drops `ExternalChange::ReloadFailed` on the floor:

```rust
ExternalChange::Unchanged | ExternalChange::ReloadFailed => {}
```

`reload_or_flag_conflict` reaches that arm whenever `reload_from_disk`
returns `Err` (`src/widgets/editor.rs:4448`), which for a PDF means the
rasteriser could not run. The user is reading a PDF, the file is rebuilt on
disk, the re-render fails, and nothing is reported: the tab keeps showing the
stale page with no status line, no toast, and no indication the sync sweep
even tried. The same applies to a text file that becomes unreadable
[...]

Issue #32, opened 2026-08-04
main does not satisfy its own pinned rustfmt: 28 diffs across 8 files at HEAD


vitali87:
A clean checkout of `main` does not satisfy its own pinned rustfmt. Verified in a fresh `git worktree` at HEAD, so this is not local working-tree noise:

```
$ cargo fmt --check | grep -c '^Diff in'
28
```

Across 8 files:

- src/app/mod.rs
- src/app/tests.rs
- src/git.rs
- src/testing/locate.rs
- src/testing/worker.rs
- src/widgets/diff.rs
- src/widgets/editor.rs
- src/widgets/editor_find.rs

Toolchain: `rust-toolchain.toml` pins `channel = "1.97.1"`, whose rustfmt is `1.9.0-stable (8bab26f4f6 2026-07-14)`.

### Why it is worth fixing rather than living with

It makes `cargo fmt --check` useless as a gate. Every branch has to be judged against "28 pre-existing diffs" instead of zero, which means the check cannot be automated in CI and each PR relies on someone remembering the baseline number. The failure mode is quiet: run `cargo fmt` while fixing a bug and unrelated files get swept into the diff, which is how formatting churn ends up mixed into behavioural commits.

### Suggested fix

One commit that runs `cargo fmt` and touches nothing else, so it can be reviewed as pure formatting and skipped in `git blame` via `.git-blame-ignore-revs`. Then `cargo fmt --check` can join clippy as a hard gate.

Keeping it as a standalone PR matters: folding it into a feature branch is what makes formatting churn indistinguishable from real changes.

PR #38, opened 2026-08-04
fix(editor,lsp,app): format-on-save writes the file you asked for, folds and sticky scroll stop hiding the caret

vitali87:
Round 19 of the adversarial sweep over main, reviewing the span below `587e58b`
(folding, format-on-save, sticky scroll, breadcrumbs). Every finding was
confirmed against current HEAD before being fixed, and every fix has a test
that was seen failing first.

## Fixed

**Format on Save wrote the wrong file.** `save()` armed a bare flag and
returned; the write happened later, from the format reply, through
`write_current_to_disk`, which saves whatever tab is ACTIVE. Switching tabs
while the formatter was thinking left the file the user pressed Cmd+S on
unwritten and still dirty, and wrote the other tab instead, snapshot and
`didSave` included. Reproduced before the fix: `a.py` on disk still read
`a=1\n` while `b.py` had been written unasked. The latch now carries the path
(`save_after_format`), and `write_tab_to_disk` finds that tab in any editor
group.

**A second Cmd+S during a format discarded the first file.** The latch holds
one path and the manager tracks one `format_request_id`, so arming a second
file abandons the first reply; overwriting the latch lost that save outright,
leaving the file dirty with nothing reported. `arm_deferred_save` now flushes
the earlier path first. It goes to disk unformatted, which is the honest
outcome once its formatter reply can no longer arrive.

**A format request that could not reach a server stranded the save.** Three
bail paths in the worker's `request_formatting` returned without answering, so
the deferred write never fired and Cmd+S did nothing at all. All three now
answer through `answer_unsupported`; the app reports "No formatter available
for this file" and still performs the save.

**The caret could end up inside a collapsed fold.** Arrow keys walked into
hidden lines, so typing changed content the user could not see. Movement now
steps over a collapsed block, and `reveal_cursor_fold` runs both from
`pin_on_edit` and at the top of the render, where every direct `cursor_row`
setter converges (Go to Definition, search, Ctrl+G, a restored session). A
caret on a hidden line has no painted row at all, so before this nothing was
drawn anywhere on screen.

**Sticky scroll painted over the caret, then disappeared entirely.** The band
floated over the caret's own row, which cannot be seen underneath it. Stopping
the band above the caret was not enough on its own: `scroll_view_to` drags the
caret to `scroll`, so the guard trimmed the band to nothing for the whole of a
wheel scroll, deleting the feature for the gesture it exists to serve. The
caret is now carried to the first row the band does not cover
(`caret_floor_row`), and the guard measures the caret's PAINTED row from the
frame's own layout, so a collapsed fold or comment box between the top of the
viewport and the caret cannot make the band overshoot either.

**Folds did not survive a reload, and did survive a file switch.** `open`
cleared the fold set unconditionally, which is right for a preview tab reused
for the next single-click but wrong for the same-path reload behind every
FS-sync sweep: a `git checkout`, an external formatter, or another split saving
the file popped every block open. The clear is now gated on the incoming path
differing; the line-count case is already covered by the render-time
`fold_epoch_lines` guard.

**Breadcrumbs measured characters, not cells.** A CJK segment is twice its
`char` count in display cells, so the next crumb painted over the tail of the
previous one (`文档文档` rendered as `文档`) and clicks landed on the wrong
symbol. Both the paint and the recorded hit rect now advance in cells via
`set_stringn`.

**Fold All made every frame expensive.** `is_line_hidden` re-derived the fold
ranges per rendered row per frame. The merged spans are now cached in
`hidden_ranges`, rebuilt on each write to `folded` and binary-searched per
query; a test pins that the rebuild count tracks fold changes, not queries.

**A test broke unrelated tests through the process-global `$HOME`.**
`display_path_collapses_home_to_tilde` repointed `$HOME` at `/Users/v`, which
cannot be created without root, and `pdf.rs`'s `unique_temp_dir` resolves
`$HOME` to reach the cache dir. Any PDF render on another thread during that
window failed outright, which is what made
`pdf_reload_keeps_the_current_page` fail in a full-suite run:

```
$ HOME=/Users/v ./target.noindex/debug/deps/croft-<hash> \
    widgets::editor::tests::pdf_reload_keeps_the_current_page --exact
PDF render failed: Permission denied (os error 13)
```

The test now exercises a pure `collapse_home(path, home)` and touches no env
var. `HOME_LOCK` only ever serialised the mutators against each other, never
against readers; issue #37 tracks the two remaining mutators and the separate
problem that a failed external reload is silently discarded.

## Verification

Full suite green: 3011 passed, 0 failed, 10 ignored (3021 collected against the
3008 baseline, so 13 new tests and no new ignores), plus `tests/cli.rs` 6
passed. Clippy is at zero warnings. `cargo fmt --check` sits at exactly the 28
pre-existing hunks tracked by issue #32, with no new drift.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Format on Save when switching between files or tabs.
  * Fixed formatting requests for untracked files and unavailable formatting services.
  * Improved navigation through folded code, including automatic expansion when editing hidden content.
  * Preserved folds when reloading the same file and reset them when opening another.
  * Improved sticky scroll positioning, Fold All performance, and breadcrumb sizing for wide characters.

* **Documentation**
  * Updated architecture and keyboard shortcut documentation for folding and navigation behavior.


vitali87:
@coderabbitai review

Issue #30, opened 2026-08-04
collab: a_live_relay_is_not_stolen_from flakes under parallel load, mechanism still unknown


vitali87:
`collab::tests::a_live_relay_is_not_stolen_from` (src/collab.rs:1776) fails intermittently in the full parallel suite (`RUST_TEST_THREADS=8`, ~3000 tests) and passes in isolation. Most recent occurrence 2026-08-04; previously seen 2026-07-14.

The failing assertion is `assert!(!crate::session::is_alive(&socket))` at src/collab.rs:1786, after the test drops the never-accepting `UnixListener` it bound. For it to fail, `connect()` must have succeeded on a socket whose only listener was just closed.

### Mechanism: still unknown

Two theories have been empirically refuted and should not be re-proposed without new evidence:

1. **macOS connect-after-close race** — 0 reproductions in 2000 iterations under a loaded stress loop.
2. **A relay spawned by `ensure_relay` binding the socket.** Under `cargo test`, `std::env::current_exe()` is the libtest harness, not the croft binary. Running the harness with `collab-relay --socket <path>` exits 101 without binding (verified directly). `ensure_relay` cannot produce a listener in this environment.

Also already fixed and therefore not the cause: `session::bind_socket_0600` used to save/restore a process-global umask around `bind`, and creation was not serialized. Since 0.1.655 it binds in a 0700 staging dir and renames into place, with a `<socket>.bind.lock` flock held across probe, stale-removal, bind and publish, so racing creators can no longer strand a live listener.

### Next step

Instrument the test to dump `lsof` for the socket path on failure, so the next occurrence identifies what actually holds it. Guessing has cost two wrong theories already.

### Separate hardening candidate (not the cause of this failure)

`session::is_alive` (src/session.rs:244) reads any `connect` error as "nothing is listening":

```rust
pub(crate) fn is_alive(socket: &Path) -> bool {
    std::os::unix::net::UnixStream::connect(socket).is_ok()
}
```

`ECONNREFUSED` and `ENOENT` do mean dead; `EMFILE`, `ENFILE` and `EAGAIN` mean the probe could not be performed. In production — where `current_exe()` really is croft — a false "dead" sends `ensure_relay` (src/collab.rs:601-628) down the spawn path. The flock added in 0.1.655 prevents the worst outcome, but distinguishing "dead" from "could not ask" is still the correct behaviour. Worth doing on its own merits; it does not explain the test failure above.

PR #40, opened 2026-08-05
fix(editor,app): saving keeps a file's encoding, Reopen with Encoding stops eating unsaved work, split-pane git bars refresh

vitali87:
Round 21 of the adversarial sweep over main, reviewing `caa30f8~1..00c17e6` (git
gutter, Go to Symbol picker, LSP signature help, the status bar redesign, and
the three activity-bar count badges). Every finding was confirmed against
current HEAD before being fixed, and every fix has a test that was seen failing
first.

## Fixed

**"Reopen with Encoding" destroyed unsaved work.** No dirty check, no
confirmation, and it clears both undo stacks, so the edits were unrecoverable.
The encoding segment is a clickable status-bar target a few cells from the
Ln/Col readout, and picking the encoding already in use looks like a no-op, so
it is easy to hit by accident. It now refuses on a dirty buffer.

It deliberately points at **undo**, not at saving. The buffer may be showing
mojibake precisely because it was decoded wrongly (the repo's own test has a
Windows-1252 `café` whose `0xE9` survives only on disk), and saving writes those
replacement characters over the original bytes. Advising "save first" would gate
the recovery path behind the one action that makes recovery impossible.

**Saving discarded the file's encoding and byte-order mark.** `encoding_rs` is
decode-only for UTF-16 by the WHATWG spec, so `Encoding::encode` routes
UTF-16LE/BE through `output_encoding()` to UTF-8 and hands back the string's
UTF-8 bytes; the returned "encoding actually used" was discarded. A UTF-8 BOM
was stripped by `decode` and never re-emitted, so every save dropped it.

**UTF-16 now round-trips through croft at all.** Fixing only the encoder would
have been worse than the bug: the output is genuinely UTF-16, and `is_binary`
rejects NUL bytes *before* the BOM is sniffed, so croft could write a UTF-16
file and then never open it again. Both ends are fixed. UTF-16 output always
carries its BOM (VS Code writes it unconditionally for the same reason: UTF-16
is otherwise indistinguishable from binary), and `open` sniffs the BOM first,
treating it as a positive declaration of text that settles what the heuristic is
only guessing at. The test drives the whole cycle: open, edit, save, open again.

**`decode` overrides the encoding you pass it.** When the file carries a BOM,
`enc.decode(bytes)` ignores `enc` and returns the encoding it actually used.
That return was discarded, so reopening a UTF-8-with-BOM file as Windows-1252
left the buffer holding UTF-8 text while claiming windows-1252, and the BOM was
dropped on the next save. The encoding actually used is now recorded.

**Git change bars went stale in a split pane.** Both the baseline fetch and the
HEAD-move invalidation iterated only the active group, while inactive split
panes render their editors too. After a commit, the unfocused pane kept marking
lines that were already clean, and it never self-corrected: focusing it makes
`git_baseline_for == path`, so the fetch is skipped from then on. Both loops now
cover every group, with the per-sweep `git show` results shared across groups
(a split usually holds the same file in both panes, and a HEAD move invalidates
everything at once).

**The parameter-hints popup painted outside the editor.** The overflow clamp
pushed a wide signature left with no floor at the pane's own edge, and the
trailing width clamp cannot repair it because `x - viewport.x` saturates to 0.
On a 120-column terminal with the sidebar open, a 97-character signature put 17
columns over the Explorer. `hover_popup` already has the missing clamp; this
popup's `MAX_WIDTH` of 100 (against the completion popup's 60) is what makes it
wide enough to hit.

**Clicking "Ln 1, Col 1" opened PROBLEMS.** The diagnostics hit rect was clipped
to the frame edge but never to where the right status cluster starts painting
over the left paragraph, so on a narrow window the cells that visibly read the
position readout still answered the diagnostics click, with the counts they
claimed not on screen at all.

## Verification

Full suite: 3023 passed, 10 ignored, 3034 collected against the 3027 baseline,
so 7 new tests and no new ignores. The single failure,
`collab::a_live_relay_is_not_stolen_from`, is the documented suite-only flake
(same family as issue #30): it passes 3/3 serially in isolation, and this branch
touches no collab or relay file. Clippy is at zero warnings. `cargo fmt --check`
sits at exactly the 28 pre-existing hunks tracked by issue #32, with no new
drift.

## Bot review

**CodeRabbit, `editor.rs` encoding reset on a same-path reload: confirmed and
fixed.** `reload_from_disk` re-enters `open`, so the FS-sync sweep and every
revert re-ran auto-detection and dropped an encoding picked through "Reopen
with Encoding"; the next save then re-encoded as UTF-8 over the original bytes.

The suggested patch was not taken as written. It guards the `self.encoding` /
`self.bom` assignment but leaves `enc.decode(&bytes)` running on the freshly
sniffed UTF-8, so the buffer would hold UTF-8-decoded mojibake while claiming
windows-1252, and the save would re-encode that as windows-1252: two
corruptions in place of one. The encoding used for the decode is sticky too,
which is what the new test's second assertion pins.

Suite after the fix: 3025 passed, 0 failed, 10 ignored, 3035 collected against
the 3027 baseline, so 8 new tests and no new ignores. The collab flake passed
this run. Clippy at zero; `cargo fmt --check` still at the 28 pre-existing
hunks of issue #32.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reopening files with a different encoding is blocked when unsaved changes exist.
  * UTF-16 files and UTF-8 BOMs are detected and preserved more reliably.
  * Git change indicators now refresh across split editor panes after commits.
  * Signature-help popups remain within the visible window.
  * Status-bar controls have improved click targeting in narrow layouts.
* **Documentation**
  * Updated architecture documentation to reflect encoding preservation and editor refresh behavior.


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

PR #48, opened 2026-08-06
fix(remote,ci): ship only build inputs to a remote, resolve rustup targets against the pinned toolchain, and add CI covering linux musl and android builds

### LINKED ISSUE #46: The cross-build preflight check and setup-cross probe a different toolchain than the build uses, so a missing target is never reported
The remote fast path silently degraded to a full `cargo install` on the VPS for
four days (2026-08-02 to 2026-08-06) and the guard built to announce exactly
that could not fire.

## The two toolchains

`rust_target_installed` (`src/remote.rs:1785`) builds its command with
`cross_tool_command("rustup")`, which sets `PATH` and nothing else - **no
`current_dir`**. It therefore inherits croft's own working directory, which is
the user's project/home, where `rust-toolchain.toml` does not apply. rustup
resolves the **default** toolchain.

`cargo_zigbuild_command` (`src/remote.rs:1685`) ends with `.current_dir(source)`
where `source = CARGO_MANIFEST_DIR`, i.e. the croft checkout, where
`rust-toolchain.toml` **pins** the channel.

When those two differ, the check passes and the build dies:

```
-- stable-aarch64-apple-darwin: aarch64-apple-darwin aarch64-unknown-linux-musl x86_64-unknown-linux-musl
-- 1.97.1-aarch64-apple-darwin: aarch64-apple-darwin
```

`~/.cache/croft/install.log`:

```
Cross-compiling croft locally for x86_64-unknown-linux-musl
error[E0463]: can't find crate for `core`
  = note: the `x86_64-unknown-linux-musl` target may not be installed
Local cross-build skipped (cargo zigbuild exited with exit status: 101);
falling back to remote cargo install
```

Note what is missing: the message at `src/remote.rs:481`, `Local cross-build
skipped: rustup target X missing (run rustup target add X ...)`. It never
printed, because as far as the check was concerned the target was installed.

## setup-cross cannot repair it either

`install_rust_target_if_missing` (`src/cli.rs:917`) runs bare
`std::process::Command::new("rustup")` for both the list and the add, with no
`--toolchain` and no `current_dir`. Run `croft setup-cross` from anywhere
outside the checkout and it prints `rustup target <triple> is already
installed.` for both targets and exits successfully, having changed nothing.
The documented remedy for this failure is a no-op.

## Fix

Pin the toolchain everywhere the cross path talks to rustup, so the probe, the
install and the build cannot disagree:

- give `cross_tool_command` the same `.current_dir(source)` the build uses, or
  pass an explicit `+<channel>` / `--toolchain <channel>` read from
  `rust-toolchain.toml`;
- do the same in `install_rust_target_if_missing`, and make `setup-cross`
  report which toolchain it acted on;
- consider having the check verify the target is usable rather than merely
  listed - the whole failure mode here is a listing that belongs to another
  toolchain.

Worth a test: a missing target must produce the `rustup target add` line, not a
raw exit-101 fallback.

Root cause of the outage itself is not a code defect - `e89015e` bumped the pin
from 1.95.0 to 1.97.1 without re-adding `CROSS_TARGETS`, and rustup targets are
per-toolchain. But that mistake should have been a one-line log message on the
next connect instead of four days of slow updates.

vitali87:
Three defects made every remote update crawl, and nothing in the repo could have caught any of them. This fixes the defects and adds the CI that would have.

## What was broken

**The source sync shipped the entire build directory.** rsync excluded `target`; this repo's build directory is `target.noindex`, so nothing was ever excluded. Three servers were 23 to 27 GB deep in a 176 GB transfer, throttled to a couple of MB/s, delivering a source tree whose only purpose was to then be compiled on the box. Measured mid-transfer they were gaining ~600 MB/minute with roughly five hours left each, all saturating the same uplink.

The same rule appeared in three places. `sync_local_source_to_remote_streaming` is the update path; `sync_via_rsync` and `sync_via_tar` are the FIRST-install path, and the tar one packed `.` into a gzip pipe with no throttle at all. All three now build their file set from `SOURCE_STAMP_INPUTS`, the same list the source stamp hashes, as an allow-list. A deny-list has to be updated every time the tree grows a directory; an allow-list derived from the definition of "the source" cannot drift. `SOURCE_SKIP_NAMES` is shared by the stamp walk and the sync filters, so nested `assets/.DS_Store` or a vendored `target/` is skipped identically by both.

**A toolchain bump orphaned the cross targets.** rustup targets belong to one toolchain. `e89015e` moved the pin from 1.95.0 to 1.97.1 without re-adding them, so `cargo zigbuild` died with `E0463` in about a second and every install fell back to `cargo install` on the remote. Four days, every box.

**The guard against exactly that could not fire.** `rust_target_installed` ran `rustup` with no working directory, resolving the DEFAULT toolchain, while `cargo zigbuild` runs with `current_dir(source)` where `rust-toolchain.toml` pins the channel. The target was installed on stable and absent from the pin, so the check reported success immediately before the build failed. `croft setup-cross` had the identical defect, which meant the documented remedy printed "already installed" and changed nothing.

Only the toolchain-sensitive query moved into the checkout. The presence probes deliberately did not: their answer is the same in any directory, and running `cargo --version` under `rust-toolchain.toml` would make an availability check trigger a rustup auto-install. Both are guarded by `is_dir()`, so a croft with no checkout at runtime (an installed release, a Nix build) fails the fast path with its normal reason rather than `ENOENT`.

## What now catches it

`.github/workflows/ci.yml`, the repo's first CI, on every push to main and every pull request:

- **clippy + tests** on Linux. Neither has ever run there. Two Linux-only errors were found and fixed while writing this: a deprecated `libc::time_t` under musl, and `WATCH_WALK_DIR_BUDGET` being dead code outside macOS.
- **ship-path build** for `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl`: byte-for-byte the `cargo zigbuild --profile remote-fast` that `croft <host>` runs, asserting the artifact exists and is a Linux ELF.
- **Android**, `aarch64-linux-android` through the runner's NDK. Termux builds natively on the phone so nothing ships this artifact, but compiling it is what catches code that cannot build there at all. Previously the only way to find out was on the device, by hand.

These run the real build rather than `cargo check`: croft's tree-sitter grammars are C, and their build scripts need a working cross C compiler, so a check fails before it can prove anything.

`remote::tests::the_pinned_toolchain_has_every_cross_target` is the local half, failing the suite on the machine that actually ships remote installs. It is gated on the cross toolchain being installed, so a fresh clone, a CI lint runner, Termux and a Nix shell skip it with a printed reason instead of failing over a target they will never use.

## Cleanup

An EXCLUDED path is protected from `--delete`, so the artifacts already on those hosts would have outlived the fix. `REMOTE_SOURCE_DIR_PREP` evicts `~/.cache/croft/source/target.noindex` on every sync. It cannot be anything but junk croft shipped: `target.noindex` is a macOS-local name a Linux box never creates. The remote's own `target/` is untouched, since that is the incremental build cache every install depends on. The four affected hosts were also cleaned by hand: 73 GB, plus 2.4 GB on a fourth found afterwards.

## Verification

Every fix carries a test, and the two behavioural ones run the real tools rather than asserting on argument lists:

- `the_source_sync_ships_only_build_inputs` builds a checkout containing `target.noindex`, `target`, `.git`, `assets/.DS_Store` and a nested `src/vendor/target`, runs rsync for real, and asserts what lands.
- `the_tar_fallback_packs_only_build_inputs` pins the tar member list, including that a non-existent input is never named (tar fails if it is).
- `source_sync_honors_the_bulk_lane_and_allow_lists_the_build_inputs` replaces a test that asserted `--exclude=target` under the comment "shipping target/ would be gigabytes through the lane". It passed on every run while 176 GB crossed the wire, because it checked the rule and never the outcome.

```
suite    3030 passed, 0 failed, 10 ignored   (3040 collected vs 3035 baseline, exactly the 5 new tests)
clippy   0 warnings on macOS and on x86_64-unknown-linux-musl
fmt      28 hunks, the pre-existing baseline tracked in #32, unchanged
cross    x86_64 and aarch64 musl both verified locally as static Linux ELF binaries
```

Fixes #46.

Stated plainly: the CI `test` job will be the first time croft's suite has run on Linux. If it goes red it will be for real Linux reasons rather than workflow syntax, and that is the point of adding it.

## Summary by CodeRabbit

* **New Features**
  * Added prebuilt remote-update binaries for Linux x86-64, Linux ARM64, and Android.
  * Improved remote source synchronization by excluding unnecessary files and removing stale build artifacts.
  * Added pinned Rust toolchain validation for remote builds.

* **Bug Fixes**
  * Improved remote transfer efficiency and reliability.
  * Resolved platform-specific build warnings.

* **Documentation**
  * Added guidance for updating the Rust toolchain and documented remote cross-build behavior.

* **Chores**
  * Updated the application version to 0.1.685.
  * Added automated CI coverage for tests and cross-platform builds.


vitali87:
@coderabbitai review

PR #59, opened 2026-08-07
fix(editor): keep the image/PDF canvas at the default background on Kitty hosts

### LINKED ISSUE #58: Image/PDF preview invisible on Kitty-graphics hosts since 0.1.665: themed canvas cells occlude the deep-z Kitty image
Opening a PNG (or PDF) on any Kitty-graphics host — Ghostty locally, or a remote hop (croft exports `TERM_PROGRAM=ghostty` across SSH) — shows the metadata header pill and an empty black canvas instead of the picture. iTerm2 is unaffected.

## Mechanism

Two commits that are each correct in isolation compose into an occlusion:

- `99eadc7` (v0.1.641) placed the editor image preview at `KITTY_Z_BELOW_TEXT_AND_BG` (z = −2 000 000 000) so right-click menus paint over it. That placement relies on the canvas cells keeping the **default** background: Kitty draws any cell with a *non-default* background color **over** images below z −1 073 741 824.
- `47338bc` (v0.1.665) made the image/CSV/diff canvases paint `theme.editor_bg()` **explicitly on non-iTerm2 hosts** (fixing the host-black-island look of CSV/diff tabs). For the image canvas this breaks the invariant the deep z depends on: every cell now wears an explicit RGB background, and Kitty paints those cells over the picture.

So since v0.1.665 the image is transmitted every frame and then hidden behind its own canvas.

## Proven with bytes

Headless capture at v0.1.685 (`script` + `croft <file.png>` with `TERM_PROGRAM=ghostty`, session env scrubbed):

- the preview APC is emitted each frame: `\x1b_Gf=100,a=T,c=76,r=21,…,z=-2000000000` — the image reaches the terminal;
- the same canvas region is painted with 490 cells of explicit `48;2;0;0;0` (the theme's editor bg as a non-default RGB color) — which Kitty renders on top of the deep-z image.

The other deep-z consumers (terminal imgcat images, sidebar hero/SSH illustrations) sit over cells that keep `Color::Reset`, so only the editor image/PDF canvas is affected.

## Direction

The image/PDF canvas must keep the default background whenever the active inline-image protocol layers images below cells (Kitty), while CSV/diff canvases keep the v0.1.665 themed fill. The baked image already letterboxes with the theme bg pixel, so the themed look is preserved by the picture itself.

vitali87:
Fixes #58.

Opening a PNG or PDF on a Kitty-graphics host — Ghostty locally, or any remote hop (croft exports `TERM_PROGRAM=ghostty` across SSH) — showed the metadata header and an empty canvas. iTerm2 was unaffected.

## Root cause

Two individually-correct changes composed into an occlusion:

- `99eadc7` (v0.1.641) placed the editor preview at `KITTY_Z_BELOW_TEXT_AND_BG` (z = −2 000 000 000) so right-click menus paint over it. Kitty draws any cell with a **non-default** background color over images below z −1 073 741 824, so this placement relies on the canvas cells keeping the default background.
- `47338bc` (v0.1.665) made the image/CSV/diff canvases paint `theme.editor_bg()` explicitly on non-iTerm2 hosts (the host-black-island fix for CSV/diff tabs). For the image canvas this broke the invariant: every cell now wore an explicit RGB background, and Kitty painted those cells over the picture.

Proven with a headless byte capture (`script` + `croft <file.png>` under `TERM_PROGRAM=ghostty`), replayed through a VT emulator:

| build | preview APC emitted | canvas cells (76×21) |
|---|---|---|
| v0.1.685 | `a=T,c=76,r=21,…,z=-2000000000` every frame | 1573/1596 explicit `48;2;0;0;0` → image occluded |
| this fix | same APC | 1596/1596 default bg → image visible |

## Fix

`image_canvas_bg` (src/widgets/editor.rs) is a separate canvas choice for the image/PDF tab: `Reset` on iTerm2 (the existing `SetColors` reason) **and on Kitty**, where the deep-z image must show through; the themed fill stays on sixel (which blits into the cell buffer over the canvas) and on no-graphics hosts (where the placeholder is all the user sees). CSV/diff canvases keep the v0.1.665 behavior unchanged.

No untinted island appears while a picture is up: the bake letterboxes with the theme bg pixel, covering every canvas cell.

Test: `the_kitty_image_canvas_keeps_the_default_bg_so_the_deep_z_preview_shows` (red before the fix) pins the per-protocol choice and that the Kitty fill actually lands `Reset` on every canvas cell.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed image and PDF previews disappearing or rendering incorrectly in Ghostty, Kitty, and WezTerm.
  * Improved preview backgrounds across supported terminal graphics protocols while preserving themed letterboxing.
  * Added regression coverage for protocol-specific preview rendering.

* **Documentation**
  * Clarified preview canvas layering and background behavior in the architecture documentation.

* **Chores**
  * Incremented the package version to 0.1.688.


vitali87:
Applied the full-canvas assertion in 453b0f9, scoped to the picture band: the test now asserts every cell below the metadata pill row is `Color::Reset` (and that the pill row keeps its own bg — it sits above the image band by design; `update_editor_image_overlay` carves the header row off before anchoring the picture, so the suggested all-rows diff fails on the pill's intentional explicit background).

Not driving the test through `Widget::render`: the production `ibg` comes from `detect_inline_image_protocol()`, which reads `TERM_PROGRAM`/`TERM` from the process environment with no injection seam — mutating the environment in-process races sibling tests (#37 class), so the hermetic pin stays on the pure `image_canvas_bg` + `render_image_placeholder` seam. The real render path is covered end-to-end by the headless byte capture in the PR description (image APC + 1596/1596 canvas cells default-bg under `TERM_PROGRAM=ghostty`).

PR #113, opened 2026-08-13
feat(dap): debugger WATCH section

### LINKED ISSUE #112: Debugger WATCH section: expressions re-evaluated at every stop
## Summary

The Run and Debug sidebar shows CALL STACK and VARIABLES but has no WATCH section — pinned expressions re-evaluated at every stop-event, the standard debugger-frontend feature (VS Code, gdb display, IntelliJ, nvim-dap-ui). The DAP plumbing already exists: `src/dap/session.rs` has the `evaluate` request builder with a `watch` context.

## Scope

- WATCH section under VARIABLES in the Run and Debug panel: one row per expression, value aligned right of it, a per-row remove affordance, and an add affordance in the section header.
- Palette: "Debug: Add Watch Expression" (input modal) and "Debug: Remove All Watch Expressions".
- On every stopped event, each expression fires one `evaluate` (context `watch`) against the top frame of the stopped thread; responses update the rows. A value that changed since the previous stop highlights; an expression the adapter rejects renders as `<not available>`, never an error popup.
- While the debuggee is running, values dim rather than clear.
- Session-scoped for now (like the debug session itself); persistence per workspace can follow separately.

Cost: one DAP request per expression per stop, only while a session is paused; zero outside debugging.

vitali87:
Fixes #112.

The Run and Debug panel gains a WATCH section — expressions re-evaluated at every stop, the standard debugger-frontend feature — built on the `evaluate` plumbing the DAP client already had (its doc comment listed the `watch` context before anything used it).

## What ships

- **WATCH section** under VARIABLES in the paused-state tree: one row per expression, a right-edge `✕` per row (hit rects recorded per frame), and a trailing "+ Add Expression" row.
- **Add / remove**: the panel's add row and the palette's "Debug: Add Watch Expression" open the standard input popup (croft's inputs-in-popups rule); duplicates are refused, whitespace trimmed; "Debug: Remove All Watch Expressions" clears the list. Session-scoped.
- **Evaluation**: every stop re-evaluates each expression against the selected frame (`evaluate`, context `watch`). A value that changed since the previous stop paints highlighted; an expression the adapter rejects renders `<not available>` — the raw adapter error never leaks into the row (`DapEvent::Evaluated` gained a `success` flag).
- **Baseline correctness**: `InspectionUpdated` fires several times per stop (scopes, variables, frame switches, expansions). The changed-value baseline rotates exactly once per stop — armed by `Stopped`, consumed by the first `InspectionUpdated` — later ones re-evaluate without rotating; rotating every time clobbered the baseline with the stop's own values and the highlight never fired (caught live, fixed, re-verified).
- Row text clips three cells before the right edge so long values never run under the `✕`.

## Tests

- Panel render: WATCH header, value row, `<not available>` degradation (asserting the raw adapter error does NOT leak), pending placeholder, add row, and `watch_remove_at` mapping by render order.
- App: palette prompt round-trip (purpose, typed value, submit), duplicate/trim rules, removal by display index keeping later rows' values, clear-all.

## Verified live against a real debugpy session

Breakpoint in a Python loop → F5 (croft provisioned the debugpy venv itself) → palette-added `total + n * 100` showed `300` against live state → F10 re-evaluated to `303` painted in the changed highlight → `no_such_var` rendered `<not available>` → its `✕` removed it ("Stopped watching no_such_var") → the "+ Add Expression" row opened the prompt.

v0.1.708; release note replaced; KEYBINDINGS.md and ARCHITECTURE.md document the section and the baseline invariant.

## Summary by CodeRabbit

* **New Features**
  * Added a WATCH section to the debugger for pinning and evaluating expressions.
  * Watch values refresh at each debugging stop, with changed values highlighted.
  * Unavailable or pending expressions now display clear status indicators.
  * Added controls to add, remove, and clear watch expressions through the UI and command palette.
* **Documentation**
  * Updated architecture, keybinding, and release documentation with watch-expression guidance.
* **Chores**


vitali87:
All six review findings addressed in the follow-up commit:

1. **ARCHITECTURE contract vs selected_frame** — the doc now states the selected-frame behavior explicitly: the stop selects the top frame, a call-stack click re-evaluates every watch against the clicked frame.
2. **Session scoping documented** — KEYBINDINGS now says the expression list lives for the croft run while values and the changed-baseline reset when a session ends.
3. **Stale state across sessions** — new `reset_watch_runtime` clears `watch_vals` / `watch_prev` / `watch_baseline_pending` in both `debug_stop` and the `Terminated` arm; expressions themselves survive.
4. **Silent add rejection** — empty input reports "Enter an expression to watch", duplicates report "Already watching …".
5. **Phase gating** — both evaluate call sites now filter on `SessionPhase::Stopped`.
6. **Scroll-corrupted remove index (Major)** — `DebugRowKind::Watch` now carries its expression index; the render-order counter is gone entirely, and the widget test gained a scrolled variant asserting the first visible row keeps its true index.

Full suite green (3075 + 6) after the round.

PR #116, opened 2026-08-13
feat(lsp): Peek Definition (Alt+F12)

### LINKED ISSUE #115: Peek Definition: Alt+F12 excerpt popup with Enter-to-jump
## Summary

croft jumps to definitions (F12 family, Cmd/Ctrl+click) but has no way to READ a definition without leaving the current position — VS Code's Peek Definition (Alt+F12), a daily-driver navigation feature.

## Scope

- `Alt+F12` and a "Peek Definition" palette entry: same LSP definition request, but the response opens a popup at the caret instead of jumping — a header (`workspace-relative-path:line`) and a numbered excerpt of the target region with the definition line marked.
- `Enter` converts the peek into the real jump (nav history recorded as usual); `Esc` closes; any other key closes the popup and keeps its normal meaning.
- The excerpt comes from the open buffer when the target file is already open (unsaved edits visible), from disk otherwise.
- Bare `F12`'s tolerance of a stray ALT bit (an iTerm2 Cmd-fold quirk) moves to the peek: on those rare folded events the user gets a popup instead of a jump — strictly less disruptive than the old behavior.

Documented in the F12 navigation family in KEYBINDINGS.md.

vitali87:
Fixes #115.

Peek Definition — VS Code's Alt+F12 — in croft's idiom: the same LSP definition request, tagged separately (`peek_definition_request_id`), whose response opens a caret-anchored excerpt popup instead of moving the caret.

## What ships

- **Popup**: `workspace-relative-path:line` header, a clamped excerpt window (`peek_excerpt`: 4 lines above, 12 total, pulled up at file edges), the definition row marked `▶`, and a one-line key legend. The open buffer supplies the text when the target file is already open (unsaved edits visible); disk otherwise.
- **Keys**: `Enter` converts the peek into the real jump (ordinary nav history); `Esc` closes; **any other key closes the popup and keeps its normal meaning** — a glance, never a mode.
- **Chord**: `Alt+F12`, claiming the ALT bit bare-F12 used to tolerate as iTerm2 Cmd-fold noise; on those rare folded events the user now gets a popup instead of a mis-jump — strictly less disruptive. The F12-family predicate tests pin the new routing.
- Palette "Peek Definition"; F1 overlay entry (the shortcuts-sync guard test caught its absence — good invariant).

## Tests

`peek_excerpt` window clamping (center, top, bottom, short file, empty); the popup contract driven through the drain's consumer (header, marked line, legend, Esc-no-move, Enter-jumps-with-caret, plain-key dismissal); the chord family routing.

## Verified live

Against the `ty` language server on a Python workspace: caret on a cross-file call, `Alt+F12` opened `util.py:1` with `▶1 def helper(x):` and the docstring visible, `Enter` landed on util.py line 1 ("Jumped to util.py line 1").

v0.1.709; release note replaced; KEYBINDINGS.md F12 family and ARCHITECTURE.md updated.

## Summary by CodeRabbit

- **New Features**
  - Added Peek Definition with `Alt+F12`/`Opt+F12`.
  - View the target source in an excerpt popup, including unsaved changes.
  - Press Enter to navigate to the definition or Esc to dismiss.
  - Added Peek Definition to the Command Palette.

- **Documentation**
  - Updated architecture, shortcut, and release documentation.

- **Chores**
  - Incremented the package version.


vitali87:
All five findings addressed in the follow-up commit:

1. **Empty reply hung the status** — a peek reply with no target now clears the request id and reports "No definition found".
2. **Unsaved edits in background tabs/splits** — the excerpt lookup now searches every open buffer: the active group's tabs plus every inactive split leaf's (new `EditorLayout::inactive_leaf_tabs`), falling back to disk only when the file is open nowhere.
3. **Anchor fallback** — `(0,0)` replaced with the editor pane's own corner, matching the sibling popups.
4. **Press/Repeat gate** — the peek block now filters `key.kind` like every other modal, so a key Release can neither dismiss nor jump.
5. **Forwarded-key test** — the dismissal test now uses a two-line buffer and asserts `Down` still moved the caret after closing the popup.

Full suite green (3078 + 6) after the round.

PR #122, opened 2026-08-14
feat(editor): auto-closing pairs

### LINKED ISSUE #121: Auto-closing pairs: brackets and quotes close, type over, surround, and delete together
## Summary

croft has no auto-closing pairs — typing `(` inserts a lone parenthesis, a selection is replaced (not wrapped) by a typed quote, backspace inside `()` deletes one side, and Enter between braces gives no block shape. VS Code's autoClosingPairs family is the most-used typing behavior in any modern editor.

## Scope (VS Code's default semantics)

- **Auto-close**: typing `(` `[` `{` or a quote with no selection inserts the pair with the caret between — for openers only when the next char is whitespace/EOL/a closer (never producing `()foo`); quotes additionally require the previous char not be a word char or the same quote (an apostrophe inside a word stays single). One undo step.
- **Type-over**: typing a closer or quote when that exact char sits at the caret advances over it instead of inserting (stateless, no provenance tracking — VS Code's observable default). No undo entry (nothing changed).
- **Selection surround**: typing an opener or quote with a selection wraps it, keeping the selection on the inner text.
- **Pair backspace**: backspace between an empty matched pair deletes both.
- **Enter between braces**: `{|}` + Enter opens the block — indented middle line with the caret, closer on its own line at the original indent.
- A settings toggle ("Auto Closing Pairs", default on) beside Format on Save / Auto Save, persisted in config.json.

Verified in code as absent: nothing in the insert/backspace/newline paths knows about pairs (`bracket_match` is display-only highlighting).

vitali87:
Fixes #121.

Auto-closing pairs — the most-used typing behavior in any modern editor, previously absent from croft entirely (`bracket_match` was display-only).

## What ships (VS Code's default semantics)

- **Auto-close**: `(` `[` `{` and quotes insert their pair with the caret between — openers guarded so a closer is never jammed into a following word; quotes additionally refuse after a word character or the same quote (the apostrophe-in-`don't` case). The pair is one undo step.
- **Type-over**: typing a closer or quote when that exact character sits at the caret advances over it — stateless, no provenance tracking, no undo entry (nothing changed).
- **Selection surround**: an opener or quote typed over a selection wraps it (multi-row included), the selection surviving on the inner text; one undo step.
- **Pair backspace**: backspace between an empty matched pair deletes both, so `(` then backspace round-trips to nothing.
- **Enter between braces** was already shipped (`is_bracket_pair_split`) — verified, not duplicated.
- **Toggle**: "Auto Closing Pairs" (default on) leads the settings gear's list, persisted as `disable_auto_close_pairs` (default-false keeps old configs valid), synced onto the editor beside the blame flag.

Scope note: the multi-caret typing path (`multi_insert_char`) keeps plain inserts in v1 — per-caret pair logic with column shifting is its own change.

## Tests (red first)

Seven behaviors pinned: auto-close with caret placement and single-step undo, the word-guard, type-over, the apostrophe case, surround with selection coordinates and undo, pair backspace, and the toggle disabling everything. One pre-existing guard test (settings first-row identity) updated for the new row order.

## Verified live

Typed `foo(` `"hi` `"` `)` in a real session → `foo("hi")` with both closers stepped over; backspaces collapsed it cleanly.

v0.1.712; release note replaced; KEYBINDINGS.md documents the behavior set.

## Summary by CodeRabbit

* **New Features**
  * Added automatic closing for brackets, braces, parentheses, and quotes.
  * Typing over existing closers, wrapping selections, and deleting empty pairs are supported.
  * Added an “Auto Closing Pairs” setting, enabled by default and saved across sessions.
* **Documentation**
  * Documented auto-closing pair behavior and configuration options.
* **Release**
  * Updated release highlights and package version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.711"
+version = "0.1.712"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.711"
+version = "0.1.712"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -9,6 +9,7 @@ On macOS, the `Cmd` chords below only reach croft after you run a one-time setup
 | Keys | Action |
 |------|--------|
 | `Ctrl+s` / `Cmd+s` | Save the open file |
+| Typing `(` `[` `{` or a quote | Auto-closing pairs: the pair inserts with the caret between (openers never before a word, quotes never after one); typing the closer steps over; a selection is surrounded; backspace inside an empty pair deletes both. Settings gear → "Auto Closing Pairs" toggles |
 | `Ctrl+q` | Quit |
 | `F1` | Open the shortcuts modal |
 | `F6` | Cycle focus across panes (tree → editor → terminal → tree) |
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2040,6 +2040,9 @@ pub struct App {
     blame_fetched: Option<PathBuf>,
     /// User pref: show the current-line inline blame annotation (default on).
     inline_blame_enabled: bool,
+    /// Auto-closing pairs (#121), persisted; synced onto the active editor
+    /// beside the blame flag.
+    auto_close_pairs: bool,
     /// LSP inlay hints (inline type / parameter annotations), on by default
     /// like VS Code's `editor.inlayHints.enabled`; toggled from the palette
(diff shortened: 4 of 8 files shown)


vitali87:
All three findings addressed in the follow-up commit:

1. **Empty-buffer panic (Critical)** — `insert_char_with_pairs` seeds an empty `lines` before touching it; regression test types `(` into a fresh `Editor::new()`.
2. **Pair backspace on pre-existing pairs (Major)** — provenance, the cheap way: the auto-close records `(row, col, edit_seq)`; pair-backspace fires only while the caret still sits exactly there with no edits since. A `()` already in the file keeps its closer (regression test), while type-`(`-then-backspace still round-trips to nothing. This is the common-case slice of VS Code's tracked-pair model without carrying positions through arbitrary edits.
3. **`cfg!(test)` save guard (Major)** — the toggle now matches every sibling; the suite had in fact rewritten the real config during the settings-view test (verified: the written value equaled the default, so no user-visible harm; the guard ends the class).

Full suite green (3098 + 6) after the round.

Issue #109, opened 2026-08-13
flaky under load: a_percent_directory_round_trips_through_the_zsh_shim times out waiting for the OSC 7 report


vitali87:
## Summary

`widgets::terminal::tests::a_percent_directory_round_trips_through_the_zsh_shim` failed once during a full parallel suite run (`zsh never reported a cwd`, src/widgets/terminal.rs:5830) and passes 3/3 in isolation immediately afterwards. Linux, `cargo test` with 8 test threads, tree at v0.1.707-dev (the merge-conflict feature branch; the failure is in shell-integration plumbing untouched by that change).

## Reading

The test spawns a real interactive zsh through the shell-integration shim and waits a bounded interval for the OSC 7 cwd report. Under full-suite load, shell startup (rc files plus the shim) can exceed the wait before the first prompt fires the report — the same startup-window family as #94: the bounded wait measures machine load, not the shim.

## Direction

Extend the wait deadline and/or gate it on the first prompt mark rather than wall-clock alone, mirroring how the #94 and suppress-cd tests were hardened (wait for an unambiguous shell state before asserting).

PR #124, opened 2026-08-14
feat(search): workspace Replace All polish

### LINKED ISSUE #123: Workspace Replace All polish: confirmation, inline preview, focus handoff, dirty-buffer safety
## Summary

Workspace search-and-replace works (toggle chevron, Tab to the replace input, Enter runs it) but the UX around it is rough, found by driving it live:

1. **No confirmation**: Enter in the replace input immediately rewrites every hit file on disk — a workspace-wide destructive action with no modal, against croft's own confirm-destructive rule (Replace All replaced 4 occurrences the instant Enter landed).
2. **No preview**: result rows show the plain match while a replacement is typed; VS Code shows the match struck through with the replacement beside it, which is how users catch a bad pattern before committing.
3. **No focus handoff**: revealing the replace row leaves focus in the search input — typing continues into the query (observed: "betadelta", 0 matches). Tab works but nothing says so.
4. **Dirty-buffer blindness**: `replace_all_on_disk` writes every hit file including ones open with unsaved edits, forcing those tabs into disk-conflict divergence.

## Scope

- A confirm modal — "Replace N occurrence(s) across M file(s)?" — before any disk write; Enter/Y confirms, Esc cancels.
- Inline row preview while the replacement is non-empty: each match renders as the old text (dim, crossed) followed by the expanded replacement (green), regex capture references honoured per match.
- Toggling the replace row open moves focus to it; its ghost text teaches the Tab switch.
- Replace All skips files open with unsaved changes and names them in the result status; clean/closed files replace on disk as today.

### LINKED ISSUE #125: flaky on CI: a_dead_relay_marks_the_session_disconnected misses EOF on a single poll
## Summary

`collab::tests::a_dead_relay_marks_the_session_disconnected` failed once on CI (`EOF from the relay must mark the session disconnected`, src/collab.rs:2187) during PR #124's run — a PR touching only search/editor_find — and passes 5/5 locally in isolation. Sibling of the closed #30 (different test, same suite).

## Reading

The test drops the accepted peer socket and immediately calls `session.poll` ONCE, asserting the disconnect latch. On a loaded CI box the hangup may not be visible to a nonblocking read yet — the read returns WouldBlock instead of `Ok(0)`, the single poll observes nothing, and the assert fires. Locally the close-to-poll window is effectively instant.

## Direction

Bounded retry around the second poll (poll until `disconnected()` or a deadline), the same state-gated-wait shape #109/#110 got — the latch is edge-triggered by the kernel's delivery, not by the test's schedule.

vitali87:
Fixes #123. Fixes #125.

Workspace Replace All polish — four gaps found by driving the existing flow live, each fixed with its own test.

## What ships

- **Confirmation before the rewrite**: Enter (or the replace-all icon) now counts first — `count_replacements` reads every unique hit file without writing — and raises a modal: "Replace N occurrence(s) across M file(s) with …?". `Enter`/`Y` commits, `Esc` cancels. Previously Enter rewrote every hit file on disk instantly, against croft's own confirm-destructive rule (observed live: 4 occurrences gone the moment Enter landed).
- **Inline preview**: while a replacement is typed, every result row renders the match crossed out and dim with the expanded replacement beside it in green — `expand_replacement` honours `$1`/`${name}` capture references per match in regex mode — the bad-pattern catch before anything commits.
- **Focus handoff**: expanding the Replace row focuses it (typing used to keep extending the query — observed: "betadelta", 0 matches); collapsing returns focus; the ghost text teaches the Tab switch.
- **Dirty-buffer safety**: the confirmed rewrite skips files open with unsaved edits anywhere (active group or split leaves, via `inactive_leaf_tabs`) and names them in the status — a disk write under a dirty buffer forces that tab into conflict divergence. Clean and closed files rewrite as before.

## Tests

App-level: the confirm flow (modal raised with counts, nothing written; Esc leaves disk alone; Enter commits) and the dirty-skip (clean file rewritten, dirty-open file untouched and named). Widget-level: toggle-focuses-replace and `expand_replacement` capture semantics.

## Verified live

Toggle → typing landed in the replace input directly; rows previewed `beta`→`delta` inline; Enter raised the modal with "4 occurrence(s) across 1 file(s)"; Esc left the file intact ("Replace All cancelled"); Enter+Enter rewrote it ("replaced 4 occurrence(s)").

v0.1.713; release note replaced; KEYBINDINGS.md rows updated.

## Summary by CodeRabbit

* **New Features**
  * Added confirmation before applying Replace All, with match and file counts.
  * Supports keyboard confirmation or cancellation.
  * Shows inline replacement previews, including regex capture substitutions.
  * Skips files with unsaved changes and reports which files were skipped.
  * Automatically re-runs the search after replacements.
* **Documentation**
  * Updated search sidebar guidance for Replace All behavior.
* **Chores**
  * Incremented the package version to 0.1.713.

Rider: CI failed the same run twice on `a_dead_relay_marks_the_session_disconnected` — a single nonblocking poll racing the kernel's EOF delivery (green 5/5 locally; this branch's added tests shifted CI's parallel scheduling into the window). Filed as #125 and hardened here with the #109/#110 state-gated wait: poll until the disconnect latch flips or a 5s deadline.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.712"
+version = "0.1.713"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.712"
+version = "0.1.713"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/docs/KEYBINDINGS.md
+++ b/docs/KEYBINDINGS.md
@@ -157,8 +157,8 @@ All three live under `~/.config/croft/` (XDG-resolved, so the same paths on macO
 | Click `Aa` / `ab` / `.*` (inset at the right of the search field) | Toggle case-sensitive / whole-word / regex; active toggles show an accent chip |
 | Click the header refresh icon | Re-run the current query |
 | Click the header clear icon | Clear the search query and results |
-| Click the left chevron (`▸`/`▾`) | Expand / collapse the Replace row |
-| Type in Replace, then `Enter` or click the replace-all icon | Replace every match across all result files with the Replace text (regex mode honours `$1` capture references); the search re-runs afterward |
+| Click the left chevron (`▸`/`▾`) | Expand / collapse the Replace row (expanding focuses it); while a replacement is typed, every result row previews the match crossed out with the expanded replacement beside it |
+| Type in Replace, then `Enter` or click the replace-all icon | Raise the Replace All confirmation — "Replace N occurrence(s) across M file(s)?" — then `Enter`/`Y` rewrites on disk, `Esc` cancels (regex mode honours `$1` capture references); files open with unsaved changes are skipped and named, and the search re-runs afterward |
 | Click the `...` icon | Expand / collapse the "files to include" and "files to exclude" glob inputs |
 | Type globs into include / exclude | Restrict the search to / from matching files (comma-separated, VS Code style; a bare `*.rs` matches at any depth). Editing re-runs the search live |
 | `Tab` | Cycle focus through the visible inputs (search → replace → include → exclude) |
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2929,9 +2929,10 @@ pub struct App {
     build_diag_files_by_pane: std::collections::HashMap<u64, Vec<PathBuf>>,
     /// True while the "Discard All Changes" confirmation modal is up.
     pub pending_discard_all: bool,
-    /// The Replace All confirmation (#123): `(occurrences, files)` counted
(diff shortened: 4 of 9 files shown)


vitali87:
All three findings addressed in the follow-up commit:

1. **Streaming-hits race (Major)** — the search panel now tracks `complete` (set by the worker's `Done` for the current query, reset on submit); Replace All refuses while hits are still streaming, with a status explaining the retry.
2. **Count/skip mismatch + zero-width preview (Major)** — `count_replacements` excludes the dirty-open set, the modal carries and shows the skipped-file count, and `split_for_highlight_regex` now reports zero-width match positions AND advances past them (the "foofoo" row-doubling is pinned by a new unit test). The find bar's documented zero-width semantics are preserved at its own layer: `line_matches` and the replace loop skip empty match chunks, so its "replace exactly what the highlighter shows" invariant test passes unchanged while the workspace preview mirrors what `replace_in_text` will actually write.
3. **Modal height (Minor)** — derived from the wrapped prompt rows via `wrap_cells_variable_width`, plus a row when skips are announced; the Enter/Esc hint can no longer fall below the border.

Full suite green (3103 + 6) after the round.

PR #128, opened 2026-08-15
fix(editor): show the caret on blank lines and at end-of-line in soft-wrap mode (v0.1.715)

vitali87:
## Problem

In soft-wrap mode (on by default for Markdown) the blinking caret disappears whenever it sits on a blank line or one past the last character of a line — exactly where you are about to type. The editor blinks the hardware caret by calling `frame.set_cursor_position` only when `cursor_screen_pos()` returns a cell; the wrap branch's guard

```rust
if end == start || visible_col >= end - start {
    return None;
}
```

rejects the two positions the row matcher just above deliberately accepts: an empty visual row (`start == end`, a blank line) and the end-of-line caret (`cursor_col == end == line_char_len`). Non-wrap files clamp against the text width instead, which is why only Markdown showed the bug.

## Fix

Replace the guard with the check it was standing in for: return `None` only when the caret cell would fall outside the text column — past the pane edge or onto the vertical scrollbar. Mid-segment cells are unaffected; the end-of-segment caret now lands on the cell after the last character, matching the non-wrap branch and VS Code. This also fixes a latent overflow where a zero-width wrap column could compute a caret cell far past the pane (the old guard let `visible_col` run up to the full segment length).

## Tests

Red/green: two regression tests (`cursor_screen_pos_wrap_shows_caret_on_blank_line`, `cursor_screen_pos_wrap_shows_caret_at_end_of_line`) failed with `None` before the fix and pass after; a third (`cursor_screen_pos_wrap_boundary_still_maps_to_next_row_start`) pins that a wrap-boundary caret still maps to the next row's first cell rather than one past the previous row. Full suite: 3113 passed, clippy clean, fmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved caret positioning in wrapped Markdown text, including empty lines, line endings, and wrap boundaries.
  * Fixed caret visibility and placement when editing soft-wrapped content.
  * Added safeguards to prevent caret placement outside the available text area.

* **Release**
  * Updated the application version to 0.1.715.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,7 +471,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.714"
+version = "0.1.715"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.714"
+version = "0.1.715"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/release_notes.rs
+++ b/src/release_notes.rs
@@ -60,6 +60,6 @@ pub struct ReleaseNote {
 
 /// What shipped in the current release. Replace on every version bump.
 pub const RELEASE_NOTES: &[ReleaseNote] = &[ReleaseNote {
-    kind: NoteKind::Feature,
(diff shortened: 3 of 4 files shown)


vitali87:
b0cd641 adds `cursor_screen_pos_wrap_zero_width_column_rejects_out_of_pane_cells`, pinning the rejection side of the clamp: a zero-width wrap column degenerates to one whole-line segment and a caret deep in that line is rejected rather than painted past the pane edge.

A separate caret-on-the-scrollbar test is not constructible: with any nonzero column width, wrap segments are capped at that width, so every caret the row matcher accepts (including end-of-line) lands strictly left of the scrollbar column. The zero-width degenerate layout is the only path that reaches the clamp, and that is exactly the case the new test covers.

Issue #65, opened 2026-08-07
finished_commands_land_in_durable_history flakes under load: 50ms tolerance on an arrival-stamped C-to-D duration


vitali87:
`app::tests::finished_commands_land_in_durable_history_and_the_popup_types_them` (src/app/tests.rs:20182) flakes under full-suite parallel load:

```
duration must cover the C→D gap, got 195
```

## Mechanism

The test's shell script emits OSC 133 `C`, sleeps 0.25s, then emits `D;0`, and the assertion requires the recorded `dur_ms >= 200` — a 50ms tolerance. But the C and D marks are timestamped when the reader thread *parses* them, not when the child *wrote* them. Under scheduler load the reader can pick up the pre-sleep chunk (prompt + command + `C`) tens of milliseconds after the child wrote it, while `D` (written after the sleep, into a quiet pipe) is picked up promptly — so the measured C→D gap undershoots the real 250ms. The observed 195ms is exactly a ~55ms late C stamp.

Same family as #57 (`clicking_source_control`, 5ms budget vs preemption), #44, #47, #50: a wall-clock margin assumption racing scheduler preemption; passes in a quiet run, fails under `cargo test` parallel load.

## Direction

Arrival time is the only clock croft has for marks, so the measurement itself can't be sharpened — the test's margin has to absorb scheduling, not fight it. Either lengthen the simulated runtime so the tolerance is proportionally trivial (e.g. `sleep 1` with a `>= 500` floor), or assert the semantic fact only (a nonzero duration was recorded alongside exit 0) and leave precise duration accounting to a test that owns its clock (`stamp_chunk_for_test`-style injection instead of a live child).

PR #160, opened 2026-08-16
test(history): retry the C→D duration measurement's coalesce collapse

### LINKED ISSUE #154: Flaky under load: finished-command duration records 0 across the C→D gap
`app::tests::finished_commands_land_in_durable_history_and_the_popup_types_them` fails intermittently under full-suite parallel load with:

```
duration must cover the C→D gap, got 0
```

(src/app/tests.rs, the assertion on the recorded command duration). It passes standalone and on most full-suite runs; observed failing roughly once per several full runs on a loaded box (three sightings across 2026-08-15/16, two of them before the failure was captured by name).

The duration is measured in the PTY reader thread between the OSC 133;C and 133;D marks. Under heavy load the test's sleep between the marks can be swallowed — if both marks arrive in one read chunk (or the clock the reader stamps with is coarser than the actual gap under scheduler pressure), the measured duration is 0 despite the test having slept. Either the test needs a longer/checked gap that cannot collapse under load, or the C→D stamping needs to take its timestamps at mark-parse time per mark rather than per chunk.

Per the project rule, a load-only flaky test is a real timing bug, not a test to retry.

vitali87:
Fixes #154.

The finished-command duration is stamped when the reader thread PARSES the OSC 133 marks — C sets an `Instant`, D takes `elapsed()`. When the reader is descheduled across the whole between-marks gap (a loaded parallel suite), both marks arrive coalesced in one read and the arrival-time measurement honestly collapses to ~0: the true gap is unrecoverable from coalesced bytes, so no stamping change can fix this — it is an inherent lower-bound property of measuring at the observation point. (The test already carried the #65 scar for the milder late-pickup form: a 500ms floor with "10x the largest undershoot ever observed" margin, which this session's build-heavy load broke outright.)

The fix is in the test's design: the timed command runs in a bounded retry loop (max 3), asserting the 500ms floor per attempt and retrying only the collapse class (`dur_ms < 500`). A real wrong-span bug — measuring across the wrong marks — fails every attempt deterministically; three consecutive full-second reader deschedules do not happen. The popup half of the test continues against the last spawned pane unchanged.

Test-only: no binary change, so no version bump (the docs-only precedent). Suite 3175 + 6 green under load, clippy zero, fmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of terminal command-history timing checks by retrying measurements when results are shorter than expected.
  * Terminal tests now reset and rerun automatically before each retry, reducing intermittent failures.

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -20652,35 +20652,49 @@ fn finished_commands_land_in_durable_history_and_the_popup_types_them() {
     // 1s of "runtime" → D;0), then a read so the popup's Enter-typed
     // bytes are observable through the shell.
     let script = "printf '\\033]133;A\\007$ \\033]133;B\\007kubectl get pods\\n\\033]133;C\\007'; sleep 1; printf 'out\\n\\033]133;D;0\\007'; read x; echo typed-$x; sleep 30";
-    app.terminals[0] = crate::widgets::terminal::PtyTerminal::new_running(
-        "/bin/sh",
-        &[String::from("-c"), String::from(script)],
-        tmp.path(),
-    )
-    .unwrap();
-    let mut waited = 0u32;
-    while app.command_history.is_empty() {
-        app.drain_terminal_bells();
-        assert!(waited < 8000, "finished command never reached the store");
-        std::thread::sleep(std::time::Duration::from_millis(40));
-        waited += 40;
-    }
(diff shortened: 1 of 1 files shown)


vitali87:
@coderabbitai review

PR #219, opened 2026-08-19
feat(nav): go forward and go to last edit location (v0.1.752)

### LINKED ISSUE #210: Go Forward in navigation history and Go to Last Edit Location
croft has Go Back (Ctrl+-, navigate back through jump history) but no Go Forward (VS Code Ctrl+Shift+- / workbench.action.navigateForward) and no Go to Last Edit Location (VS Code Cmd+K Cmd+Q). After navigating back there is no way to return, and there is no jump to where the last edit happened. Parity ask: forward navigation over the same history the back command walks, and a last-edit-location jump, both with palette entries and keybindings.

vitali87:
Closes #210.

## What

Two-way navigation history plus the missing edit-location jump:

- **Go Forward** (`Ctrl+Shift+-`, palette "Go Forward", editor context menu): Go Back now threads the position it leaves onto a forward stack, and forward walks it back out, restoring each position exactly. A fresh jump forks history (the forward trail dies), matching VS Code. On hosts without the extended keyboard protocol, `Ctrl+-` and `Ctrl+Shift+-` are the same 0x1F byte; that spelling deliberately stays Go Back (the safer meaning) and the palette covers forward there. The F1 overlay documents the chord (its sync-guard test enforced this).
- **Go to Last Edit Location** (`Cmd+K Cmd+Q`, palette entry): jumps to the most recent buffer edit across all open editors, including background splits. The position is captured in `mark_buffer_changed` beside the existing auto-save timestamp, so every real edit path feeds it. The chord's second key carries Cmd, keeping it distinct from the navigator's plain `Cmd+K Q`. The jump records the current position on the back stack so `Ctrl+-` returns.

## Tests

- nav history unit tests: back/forward round-trip through the current position, forks clear the forward trail, plus the existing order/cap/dedupe suite updated for the two-stack model.
- App-level: go-to-definition then `Ctrl+-` then `Ctrl+Shift+-` restores the jump target and row; bare `Ctrl+_` (legacy `Ctrl+-`) still goes back; an edit in one file followed by navigation and `Cmd+K Cmd+Q` returns to the edited row.
- Full suite green (3294 lib + 6 CLI), clippy zero.
- Live drive in tmux: "Last edit: alpha.rs line 1", "Back to beta.rs", "Forward to alpha.rs line 1" statuses observed through the real chords (CSI-u encodings).


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

PR #233, opened 2026-08-21
feat(remote): confirm before compiling croft on the remote box

vitali87:
When the fast install path is unavailable, the installer silently fell back to rsyncing the source and running `cargo install` **on the remote box**. That build takes minutes and loads the machine — and because it was silent, a missing `croft setup-cross` never announced itself. It surfaced only indirectly, as a remote that stayed mysteriously out of date.

The fallback now requires an explicit yes, and the reason the fast path was unavailable travels with the question.

## The signature change

`try_local_cross_install` and `try_local_cross_install_streaming` returned `bool`, which could say *whether* the fast path ran but not *why* it didn't. They now return `Option<String>`:

- `None` — the prebuilt binary shipped, nothing to ask
- `Some(reason)` — the fast path is unavailable, and `reason` is what the user gets shown (`cargo-zigbuild not found`, `rustup target aarch64-unknown-linux-musl missing`, a build failure…)

`install_remote_croft_streaming` takes a `confirm_fallback: &mut dyn FnMut(&str) -> bool` gate, so each entry point supplies its own way of asking.

## Three ways of asking, one default

| Context | Behavior |
|---|---|
| TUI connect | `InstallEvent::FallbackPrompt` → the connect dialog's new `AwaitingFallbackConfirmation` phase → `respond_fallback` |
| CLI connect | `confirm_remote_compile_on_tty`; non-interactive stdin declines |
| Background update (already attached) | Declines automatically, logs the `setup-cross` hint |
| Dialog torn down mid-prompt | `respond_fallback(false)` — no deadlock |
| Answer channel dropped | `recv().unwrap_or(false)` — reads as a decline |

Every path defaults to **no**. The slow compile is always a choice, never a slide.

The background-update case is deliberate rather than incidental: when a croft is already running on the remote the user is attached to it, so there is nobody to ask and an unrequested compile would load the box under a live session. It declines and logs how to get the fast path instead.

## Tests

Three cases pin `answer_install_fallback`, the decision point behind both the Enter and the click handler:

- `install_fallback_proceeds_only_on_an_explicit_yes` — `y`/`Y`/`yes`/`YES`/`  yes  ` reach `DialogPhase::Installing`
- `install_fallback_declines_and_tears_down_on_anything_else` — `""`, `"   "`, `n`, `no`, `yep`, `later` close the dialog, clear the pending launch, and set a status naming `setup-cross`
- `install_fallback_without_a_dialog_does_not_panic` — both arms tolerate a torn-down dialog

The empty-string case is the one that matters: the footer offers bare Enter as Cancel, so it must never read as consent. Verified by mutation — forcing `let proceed = true` fails that test specifically.

Also `set_fallback_prompt_awaits_an_explicit_yes` on the dialog side.

## Also here

One drive-by fix, kept as its own commit: the remote install script's `$CROFT_NICE $CROFT_IONICE cargo install ...` line. zsh does not word-split unquoted parameters, so under a zsh login shell that tried to run a command literally named `nice -n 19`. Now `eval`-ed, which splits correctly under sh, bash and zsh alike.

## Test suite

3334 passed. Four failures (`accept_all_incoming_then_complete_merge_stages_the_resolved_file`, `repositories_overview_lists_every_folder_and_a_pin_overrides_the_follow`, `gui_path::probing_a_real_shell_returns_a_usable_path`, `workspace_file_round_trips_with_relative_paths_and_tolerates_comments`) reproduce identically on clean `origin/main` — shell-spawning and load-sensitive, unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added an interactive confirmation prompt when remote installation must fall back to slower compilation.
  * Displays the reason for the fallback and provides clear guidance for proceeding or canceling.
  * Accepts only “y” or “yes” to continue; other responses cancel the connection safely.
  * Supports confirmation through both keyboard and mouse submission.

* **Bug Fixes**
  * Prevents unattended or background sessions from waiting indefinitely for fallback confirmation.
  * Ensures canceled connections clean up authentication and update status guidance appropriately.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -20072,6 +20072,16 @@ impl App {
                     }
                 }
                 crate::install_session::InstallEvent::CanLaunch => can_launch = true,
+                crate::install_session::InstallEvent::FallbackPrompt { reason } => {
+                    if let Some(dialog) = self.connect_dialog.as_mut() {
+                        dialog.set_fallback_prompt(&reason);
+                    } else if let Some(session) = self.install_session.as_ref() {
+                        // No dialog to ask through: decline so the installer
+                        // errors out with the `croft setup-cross` guidance
+                        // instead of blocking forever on an answer.
+                        session.respond_fallback(false);
+                    }
+                }
                 crate::install_session::InstallEvent::Done(Ok(())) => done = true,
                 crate::install_session::InstallEvent::Done(Err(detail)) => {
                     failure = Some(detail);
@@ -20341,6 +20351,14 @@ impl App {
                 if !dialog.phase.awaits_input() {
                     return;
                 }
+                if matches!(
+                    dialog.phase,
+                    crate::widgets::connect_dialog::DialogPhase::AwaitingFallbackConfirmation
+                ) {
+                    let payload = dialog.input_for_submit();
+                    self.answer_install_fallback(&payload);
+                    return;
+                }
                 let payload = dialog.input_for_submit();
                 dialog.submitted = true;
                 dialog.status_line = format!("Verifying credentials with {}", dialog.host);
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile review

vitali87:
@coderabbitai review

Issue #100, opened 2026-08-11
"Opened" status transients show absolute paths; deep workspace roots truncate the filename out of view


vitali87:
## Summary

The "Opened …" status-bar transients print the file's absolute path. With a deeply nested workspace root, the status bar clips the message tail, so everything after the root prefix — including the file name itself — is invisible. Opening `src/main.rs` in a workspace rooted at a long path shows, at 140 columns:

```
Opened /tmp/<deep>/<workspace-root-prefix-only, cut mid-directory>
```

The one piece of information the message exists to convey (which file opened) is the piece that gets cut.

## Where

`format!("Opened {}", path.display())` and variants:

- `src/widgets/editor.rs:2784` and `:2888` (open + binary-view open)
- `src/app/mod.rs:18805` (Quick Open), `:18885` ("Opened … at line …", search-result jump), `:22402`

## Fix direction

Display the workspace-root-relative path in these transients — the tab titles and breadcrumbs already speak root-relative — falling back to the absolute path only for files outside the root. Middle-eliding the absolute path (keeping the tail) would also work but relative is shorter and matches the rest of the UI.

vitali87:
Same defect class in one more place, found while exercising the app: the F2 rename popup titles itself `Rename <absolute path>` (`src/app/mod.rs` ~30384), and under a deep workspace root the popup truncates the title before the filename is visible — the one thing the rename dialog needs to show. The `Renamed to <absolute path>` status transient (~30481) has the same shape. Both belong in this fix.

PR #268, opened 2026-08-23
feat(statusline): segment the bar with hairline dividers and hover fills

vitali87:
## What

Redesigns the bottom status bar so it reads as segments instead of one run of text. Design comparison (now vs new, with cell-level spacing analysis): https://claude.ai/code/artifact/693ad9a1-cf6d-4465-a522-0ddff95372fa

Three problems, three fixes:

1. **Scattered diagnostics** — each count carried its own side padding plus a joining space, putting the ⊗ and ⚠ icons three cells apart. They tighten to one cell, matching the single hit rect they already share.
2. **Unanchored transient** — the message ("Copied 133 chars…") was appended after two raw spaces in the same style as everything else. It is now fenced behind a hairline `│` (`#3a4150`) and dropped to the muted gray (`#9aa4b8`), which it keeps through middle-elision, so commentary reads apart from controls.
3. **Invisible buttons** — Ln/Col, Spaces, encoding, EOL and language were identical dim text with two-space seams. The right cluster gains `│` seams between segments, steps up a shade (`#c9d2e3`), and the four clickable segments take a hover fill (`#2a3140`) driven by `pointer_cell` against the frame's hit rects — the same affordance the activity bar uses. Ln/Col is a readout, not a control, so it never hovers.

A `│` also separates the brand/git zone from the diagnostics cluster.

## Implementation notes

- Right-cluster spans are now built after `rx` is known so hover is computed against this frame's rects; hit-rect recording moved into the same pass.
- The middle-elision logic for over-long transients (#100) and the diag-rect clipping under an overlapping cluster are untouched in behavior.
- The cluster is 4 cells wider (47 → 51 with dividers), so the two tests that pinned exact widths move to 56 and 153 cols to keep testing their intent: the diag hit-rect clip, and verbatim transients whenever they truly fit.

## Testing

- `cargo test --bin croft`: both adjusted status-bar tests pass; remaining failures in the full run are the known load-sensitive/environmental flakes (reproduced identically on `main`).
- `cargo fmt --check` clean.

## Summary by CodeRabbit

* **UI Improvements**
  * Added subtle dividers between status bar sections for clearer separation.
  * Reduced diagnostic spacing and dimmed transient status messages.
  * Improved text clipping while preserving message styling.
  * Enhanced right-side status segments with distinct styling and hover highlights for interactive controls.
  * Improved status bar layout across different terminal widths.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -12713,8 +12713,20 @@ impl App {
             ));
             spans.push(Span::raw(" "));
         }
-        spans.extend(git_status_spans(&scm_status));
-        spans.push(Span::raw("  "));
+        // Hairline zone divider (│): fences the brand/git zone, the
+        // diagnostics cluster, and the transient message off from each other
+        // so the bar reads as segments instead of one run of text. Quiet on
+        // both the navy and near-black bars.
+        let zone_div = Style::default().fg(self.theme.ui(Color::Rgb(0x3a, 0x41, 0x50)));
+        let scm_spans = git_status_spans(&scm_status);
+        // The wordmark already ends in a space; the git spans don't, so buy
+        // the divider one cell of air only when the branch segment painted.
+        let scm_present = !scm_spans.is_empty();
+        spans.extend(scm_spans);
+        spans.push(Span::styled(
+            if scm_present { " \u{2502}" } else { "\u{2502}" },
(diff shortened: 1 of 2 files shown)


vitali87:
@coderabbitai review

Issue #254, opened 2026-08-23
LSP completeness batch: folding ranges, expand/shrink selection, linked editing, on-type/range formatting, prepare-rename, colors, document links


vitali87:
## Motivation

Croft's LSP client (`src/lsp/client.rs`, `src/lsp/manager.rs`) covers the big requests: completion, hover, definition/references/implementations/type-definition/declaration, rename, document formatting, code actions with resolve, diagnostics, inlay hints, document highlight, workspace and document symbols, call hierarchy, semantic tokens. What is missing is a band of smaller protocol features that individually seem minor but together are the polish difference people feel moving from VS Code. None of these exist today (zero hits in `src/` for each method name).

This is an umbrella issue; each checkbox is independently shippable and reviewable. Split into per-item PRs.

### 1. `textDocument/foldingRange`
Folding today is indentation-based (`fold_all`/`toggle_fold` on indented blocks). Indentation folding breaks on brace-on-own-line styles, multiline strings, and comments, and cannot fold `#region` markers. Request folding ranges from the server on open/change (debounced), fall back to the current indentation scan when the server lacks the capability or is not warm yet (same two-phase posture as the tree-sitter-first outline). Honor `kind` (comment/imports/region) for "Fold All Comments" / "Fold All Regions" palette commands. Also add region-marker parsing (`#region`/`#endregion`, `// region`) to the fallback scanner.

### 2. `textDocument/selectionRange` (Expand/Shrink Selection)
VS Code's `Shift+Alt+→/←` semantic selection growth. `selectionRange` appears only as a capability string in `src/lsp/manager.rs`; there is no command. Implement expand/shrink over the server's range stack, with a tree-sitter node-ancestry fallback (croft already has parse trees for 13 grammars in `src/highlight.rs`), so the feature works even with no LSP. Maintain a per-cursor stack so shrink retraces exactly. Works with multi-cursor.

### 3. `textDocument/linkedEditingRange` (auto-rename paired tags)
Editing one HTML/JSX tag updates its pair live. Request on cursor move (debounced), mirror keystrokes across the returned ranges while the cursor stays inside one. vtsls and the html-language-server both support it.

### 4. `textDocument/onTypeFormatting`
Fires the server's on-type formatter on trigger characters (`}`, `;`, newline per server capability). Gate behind a setting (`editor.format_on_type`, default off, matching VS Code).

### 5. `textDocument/rangeFormatting` (Format Selection)
[...]

PR #274, opened 2026-08-24
feat(lsp): Format Selection (#254 item 5)

vitali87:
Item 5 of the #254 LSP completeness batch: `textDocument/rangeFormatting` behind a **Format Selection** command and VS Code's `Cmd+K Cmd+F` chord.

## What's here

Deliberately thin, as the issue predicted ("Straightforward: same apply-edits path as document formatting, scoped to the selection"):

- The request reuses whole-document formatting's `FormatResult` channel, one-undo-step apply path (`apply_rename_to_open_tab`), and the **always-answer** contract — every unroutable early-out reports `unsupported`, so the status line never strands. The drain distinguishes the two kinds by the in-flight id and words statuses accordingly ("Formatted selection" / "Selection already formatted" / "No range formatter available for this file").
- Selection endpoints are converted char-column → UTF-16 via a new `Editor::pos_to_utf16` before hitting the wire.
- Capability gated on `documentRangeFormattingProvider` (`one_of_supported`, bare-`false` treated as unsupported); nothing is sent unadvertised.
- **Chord**: `Cmd+K Cmd+F` — the SUPER-held second key distinguishes it from croft's existing plain `Cmd+K F` format-on-save toggle, so both stay reachable. No selection → refuses with "Select something to format".

## Acceptance criteria (per item)

- [x] Format Selection palette command + Cmd+K Cmd+F
- [x] Same apply-edits path as document formatting, scoped to the selection
- [x] Capability negotiation: silent degradation, nothing sent unadvertised
- [x] LAYOUT.md and KEYBINDINGS.md updated

## Tests

An unroutable-request-still-answers worker test (the format family's contract), and an app-level test covering the no-selection refusal plus the end-to-end unsupported verdict arriving through the shared drain with selection-specific wording. Full suite: 3407 passed, only the known pre-existing environmental failures. Clippy `--all-targets -D warnings` clean.

## Summary by CodeRabbit

* **New Features**
  * Added **Format Selection** using language-server support.
  * Use `Cmd+K`, then `Cmd+F` to format the selected text.
  * Formatting requires an active selection and applies as a single undo step.
  * Added status feedback when selection formatting succeeds, makes no changes, or is unavailable.
  * Retained `Cmd+K`, then `F` for the Format-on-Save toggle.
* **Documentation**
  * Updated keyboard shortcut and editor capability documentation.

PR #275, opened 2026-08-24
feat(lsp): on-type formatting behind editor.format_on_type (#254 item 4)

vitali87:
Item 4 of the #254 LSP completeness batch: `textDocument/onTypeFormatting`, gated behind a setting exactly as the issue asks (`editor.format_on_type`, **default off**, matching VS Code).

## What's here

**Trigger detection**: the editor records the last typed character with the `edit_seq` it produced (`insert_char_raw` and `insert_newline` — newline is a real trigger for several servers). The per-frame tick fires at most one request per typed trigger (a seq latch), and only when the character is in the trigger set some spawned server actually advertised (`firstTriggerCharacter` + `moreTriggerCharacter`, published per language at spawn). The worker then re-picks the serving client by the same capability — so nothing is ever sent unadvertised, at either layer, and a language with no provider stays completely inert.

**Stale-safety**: the reply applies quietly through the shared formatting edit path, but only while `edit_seq` still equals the value the request was computed against — keep typing and the edits are dropped rather than misapplied. On-type replies ride their own channel so the explicit Format Document/Selection statuses never see them.

**Setting**: `format_on_type` in config.json (serde default = off; an older config parses straight to disabled, with a round-trip test), toggled from the palette ("Editor: Toggle Format on Type") or the Settings list with provenance display.

- [x] Fires the server's on-type formatter on its advertised trigger characters (incl. newline)
- [x] Gated behind a setting, default off, matching VS Code
- [x] Capability negotiation: silent degradation, nothing sent unadvertised

## Tests

Prefs round-trip (default-off, old-config compatibility) and an app-level test pinning both gates: disabled → nothing fires even on a trigger character; enabled but no advertised trigger set → the request is never sent. Full suite: 3403+ passed; the failures are the known pre-existing environmental trio plus parallel-load flakes that pass alone. Clippy `--all-targets -D warnings` clean.


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:

CodeRabbit rate-limited on this PR (3 bounced re-triggers), so this review was run locally with the fallback reviewer.

---

## 1. Walkthrough

Adds VS Code's `editor.formatOnType` to croft: a new `format_on_type` pref (serde `default` = off), a per-frame tick that fires `textDocument/onTypeFormatting` when the last edit was a character in the trigger set a spawned server advertised, and a dedicated result channel whose reply is applied only while `edit_seq` still matches the value the request was computed against. Trigger detection is bolted onto the editor via a new `Editor::last_typed: Option<(char, u64)>` field, written from `insert_char_raw` and `insert_newline`. The toggle is reachable from the Command Palette (`Editor: Toggle Format on Type`) and the Settings list.

| File | Summary |
| --- | --- |
| `docs/KEYBINDINGS.md` | One row documenting the palette toggle and its semantics. |
| `docs/LAYOUT.md` | Adds "format on type" to the Editor feature paragraph. |
| `src/app/mod.rs` | `format_on_type`/`on_type_request`/`on_type_fired_seq` state; `tick_on_type_formatting`, `drain_lsp_on_type_format`, `toggle_format_on_type`; palette + settings-row wiring; two new calls in `main_loop`. |
| `src/app/tests.rs` | One test asserting inertness when disabled and when no trigger set is advertised. |
| `src/lsp/client.rs` | `LspClient::on_type_formatting` wrapping `DocumentOnTypeFormattingParams`. |
| `src/lsp/manager.rs` | `Cmd::RequestOnTypeFormatting`, `on_type_rx` channel, `on_type_triggers` per-client + per-language capability publishing, `language_on_type_triggers`, worker handler that re-picks a client by trigger set. |
| `src/prefs.rs` | `format_on_type` field, `save_format_on_type`, a round-trip test. |
| `src/widgets/command_palette.rs` | `Command::ToggleFormatOnType` (label, id, empty keybinding). |
| `src/widgets/editor.rs` | `last_typed` field + writes in `insert_char_raw` / `insert_newline`; new `pos_to_utf16`. |

## 2. Findings

- **`src/widgets/editor.rs:4895`** **[blocker/correctness]** — **The feature never fires for any character trigger; only newline works.** `last_typed` is written in `insert_char_raw` (line 4895), but the actual typing path is `Editor::insert_char` (line 4621), which inserts into `self.lines[row]` directly and never calls `insert_char_raw`. The auto-pairs branch `insert_char_with_pairs` (line 4655) likewise inserts directly. Verified empirically with a probe test:

  ```
  PROBE typed ';' -> last_typed=None edit_seq=2
  PROBE Enter    -> last_typed=Some(('\n', 3)) edit_seq=3
  PROBE '}'      -> last_typed=Some((';', 13)) edit_seq=13   // stale, from a paste
  ```

  So typing `;` or `}` — the two triggers the PR body and `docs/KEYBINDINGS.md:230` name explicitly — leaves `last_typed` untouched, and `tick_on_type_formatting` (`src/app/mod.rs:8220`) bails at the `seq != self.editor.edit_seq` guard. Only `insert_newline` works. Fix: set it in the real typing entry points (`insert_char`, both the pairs-handled and plain branches).

- **`src/widgets/editor.rs:5400`** **[blocker/correctness]** — **`pos_to_utf16` panics on an empty buffer.** `let row = row.min(self.lines.len().saturating_sub(1));` yields `0` when `self.lines` is empty, then `&self.lines[row]` indexes out of bounds. `Editor::new` initialises `lines: Vec::new()`, and the surrounding code guards `self.lines.is_empty()` in ~20 places precisely because this state is reachable. Verified by probe (panic: index out of bounds). `tick_on_type_formatting` calls this every frame while enabled, so an empty scratch buffer plus a stale `last_typed` crashes the editor. Fix: early-return `(0, 0)` on an empty buffer.

- **`src/widgets/editor.rs:4877`** **[major/correctness]** — **Paste (and snippet expansion) is treated as typing.** `insert_str` loops over `insert_char_raw`, so pasting any text whose last character is a trigger sets `last_typed` and — once the first blocker is fixed — fires an on-type format at the caret. `insert_str` is the clipboard-paste path and the snippet body path. VS Code fires `formatOnType` only on a user keystroke, never on paste. Fix: move the write out of `insert_char_raw` into `insert_char` — that resolves both defects at once.

- **`src/app/mod.rs:8220`** **[major/correctness]** — **`edit_seq` is per-tab, so the staleness latch does not survive a tab switch.** `self.editor` derefs to the active tab, and each `Editor` carries its own `edit_seq` counter. `tick_on_type_formatting` reads `last_typed`/`edit_seq` from whichever tab is active, and `drain_lsp_on_type_format` compares `self.editor.edit_seq != seq` against the tab active at reply time — switch tabs between request and reply and the comparison is between two unrelated counters. The `path` check covers the common case, but not all orderings. Fix: record the tab identity (path) alongside the seq in `on_type_request` and require both to match.

- **`src/config_layers.rs:92`** **[major/correctness]** — **The new key is invisible to the config-layer system, so the Settings row lies about provenance.** The Settings row renders `prov("format_on_type")`, but `format_on_type` is absent from `WORKSPACE_ALLOWED_KEYS` (its sibling `format_on_save` sits at line 94). Consequences: a workspace `.croft/config.json` cannot set it (refused as a non-allowed key); the provenance suffix always renders empty; and `load_vscode_subset` maps `editor.formatOnSave` but nothing maps `editor.formatOnType`, so a repo's `.vscode/settings.json` is ignored for this key while the PR body claims parity with VS Code's `editor.formatOnType`. Fix: add the key to `WORKSPACE_ALLOWED_KEYS` and map `editor.formatOnType` in `load_vscode_subset`.

- **`src/lsp/manager.rs:1985-1989`** **[minor/correctness]** — **The "not probed yet" sentinel is lost for on-type triggers.** `language_on_type_triggers` documents `None` as "until a server for the language has been probed, or when none carries the provider", but the publish always inserts, collecting into a `String` that is `""` when no client advertises the provider — so post-probe-with-no-provider is `Some("")`, not `None`. `.contains(ch)` on `""` is false so behavior is safe, but the two states the doc comment distinguishes are collapsed. Either insert only when non-empty, or fix the comments.

- **`src/prefs.rs:607`** **[minor/tests]** — `assert!(!Prefs::load_or_default().format_on_type || true)` is a tautology. `X || true` is unconditionally true; this line asserts nothing and additionally reads the *real* user config dir inside a test that otherwise uses a tempdir. Delete it.

- **`src/app/mod.rs:8207-8210`** **[minor/style]** — **The new doc comment is grafted onto the previous function's.** The doc block for `tick_occurrences` now runs straight into `/// On-type formatting (#254): …` with no blank line, so rustdoc attaches the entire occurrences explanation to `tick_on_type_formatting` and leaves `tick_occurrences` undocumented. Separate the two.

## 3. Test coverage

The single new app test only asserts two negatives — disabled, and no advertised triggers — both of which pass trivially; it would have passed identically if `tick_on_type_formatting` were an empty function, and in fact it masks the first blocker. Missing: a trigger-detector test (typed chars recorded, paste not), a seq-latch test, a stale-reply-drop test for `drain_lsp_on_type_format` (neither branch is exercised), a `pos_to_utf16` unit test (multi-byte/astral chars + the empty-buffer panic), and palette-dispatch coverage for `Cmd::ToggleFormatOnType`.

## 4. Nitpicks

- `src/widgets/editor.rs:4893` / `src/app/mod.rs:8210` — "THE last edit" in caps; surrounding comments use plain prose emphasis.
- `src/lsp/manager.rs:1985-1989` — concatenating trigger characters from multiple servers into one flat `String` means the app-side `contains(ch)` can pass on server A's trigger while the worker's re-pick finds only server B. Harmless today (the worker silently returns), but the two checks are doing different things under one name.
- `src/prefs.rs:604` — the test name says "defaulting off" but the meaningful default assertion is on `Prefs::default()`; consider `round_trips_format_on_type_and_old_configs_default_off` to match the sibling.
- `docs/KEYBINDINGS.md:230` — the row is 4 lines in a table whose other rows are one clause; it reads as release notes.

Not a finding: no `Cargo.toml` version bump / `src/release_notes.rs` replacement — checked the four most recent merged feature PRs; none touch either file, so the bump is applied at release time, consistent with repo practice.

## 5. Scores

- **CodeRabbit score (diff quality) — 1/5**: two blockers inside the new lines (the trigger is never recorded on the real typing path; `pos_to_utf16` panics on an empty buffer), plus a paste-fires-formatting defect and a tautological assertion.
- **Greptile score (codebase fit) — 1/5**: the change misreads its own host — `insert_char` vs `insert_char_raw`, per-tab `edit_seq` under a `Deref` to the active tab, and `WORKSPACE_ALLOWED_KEYS`/`load_vscode_subset` left unextended while the Settings row already calls `prov("format_on_type")`.

VERDICT: REQUEST_CHANGES

vitali87:

Re-review after the fixes in 05d7523; CodeRabbit is still rate-limited on this PR.

---

## Walkthrough

Commit `05d7523` addresses the prior review's seven findings. Verified fixed: `last_typed` now lives on `insert_char` (both the auto-pair and plain branches) so real keystrokes fire and `insert_str` paste does not (`editor.rs:4625`, `4654`, `5051`); `pos_to_utf16` guards the empty buffer (`editor.rs:5406`); `format_on_type` joined `WORKSPACE_ALLOWED_KEYS` and `load_vscode_subset` (`config_layers.rs:95`, `402`); the `Some("")` sentinel is gone (`manager.rs:1991`); the tautological prefs assert and the grafted doc comment are cleaned up. Four new tests pass and clippy `--all-targets` is clean.

The remaining problems are in the two mechanisms the fix commit introduced. The `take()`-based consumption does **not** deliver the "tab switches can never replay a stale trigger" guarantee the commit message claims, and the config-layer parity is only half wired — the reload path never applies the new key. I reproduced both with throwaway probes against the working tree (reverted; tree is clean).

## Findings

- **`src/app/mod.rs:8212`** **[major/correctness]** — `self.editor.last_typed.take()` only consumes the **active** tab's record, so a keystroke followed by a tab switch inside the same input burst leaves the trigger armed on the background tab indefinitely. `main_loop` drains every queued event in one `loop { event::read() }` before any tick runs, so `Char('}')` + `Ctrl+Tab` arriving together is enough. `last_typed` is written only at `editor.rs:4625/4654/5051` and cleared only by this `take()` — nothing clears it on tab switch, and switching tabs is a bare `self.active = idx` that never touches `edit_seq`. When the user returns to that tab minutes later, `seq == edit_seq` still holds and the request fires for a long-dead keystroke. Verified: a probe typing `;` into `a.rs`, opening `b.rs`, ticking, then reopening `a.rs` printed `tab A last_typed=Some((';', 2)) edit_seq=2`. The commit message's claim that "the tick consumes last_typed exactly once, so tab switches … can never replay a stale trigger" is therefore false. Fix: consume it at the source rather than at the reader — clear `last_typed` on every tab-activation path, or store the record on `App` rather than per-`Editor` so there is exactly one and `take()` genuinely means once.

- **`src/app/mod.rs:31399-31406`** **[major/correctness]** — `apply_merged_settings` applies `format_on_save`, `auto_save`, `copy_on_select`, `render_whitespace` and every other allowlisted key, but never `format_on_type`. This is the live-reload path: saving `.croft/config.json` or `.vscode/settings.json` inside croft reaches `reload_config_for_path` → `remerge_settings` → `apply_merged_settings`, and it is the only caller. So the PR's own new `WORKSPACE_ALLOWED_KEYS` entry and `editor.formatOnType` mapping have no effect until the next launch, while the Settings row already renders `prov("format_on_type")` — the UI shows a provenance layer whose value is not the one in force. Startup is fine (`mod.rs:4168` reads `loaded_prefs` from the merged chain); only the reload path is missing. Fix: add `self.format_on_type = p.format_on_type;` next to line 31406.

- **`src/widgets/editor.rs:4701-4705`** **[major/correctness]** — the auto-pair type-over branch returns `true` after only `self.cursor_col += 1`, with no `mark_buffer_changed()`, so `edit_seq` does not move; `insert_char` then records `last_typed = Some((c, self.edit_seq))` against the *unchanged* seq. That breaks the staleness latch in `drain_lsp_on_type_format`, whose sole "buffer unchanged" test is `self.editor.edit_seq != seq`. Concretely: type `;` at seq N (request fires with `seq=N`), then before the reply lands type `)` over an existing `)` — the caret has moved but `edit_seq` is still N, so the `;` reply passes the guard and applies as if nothing happened. Verified: a probe doing `insert_char('(')` then `insert_char(')')` with `auto_close_pairs = true` printed `after_pair=1 after_typeover=1 last_typed=Some((')', 1))`. Fix: record the caret position alongside the seq — make the record `(char, u64, usize, usize)` and have the drain also require `(cursor_row, cursor_col)` to be unchanged — since `edit_seq` alone provably cannot see a caret-only edit.

- **`docs/SETTINGS.md:84`** **[minor/docs]** — the workspace allowlist is enumerated verbatim (`theme`, `format_on_save`, `auto_save`, …) and the section closes with "Extending the allowlist is a deliberate review decision, not a default". This PR extended it at `config_layers.rs:95` and left the doc listing stale. No test enforces the sync. Fix: add `format_on_type` to the list.

- **`docs/SETTINGS.md:99-104`** **[minor/docs]** — the VS Code subset table states croft maps "exactly these keys (and nothing more — broad VS Code settings compatibility is a non-goal)" and lists `editor.formatOnSave` but not `editor.formatOnType`, which `load_vscode_subset` now maps (`config_layers.rs:402-404`). Fix: add a `editor.formatOnType` → `format_on_type` row.

- **`src/lsp/manager.rs:1991`** **[minor/correctness]** — every sibling capability writes unconditionally (`support.formatting.insert(lang, …)`, `support.code_action.insert(lang, …)`), but `on_type_triggers` inserts only when non-empty, so a re-probe of the same `lang` whose new server set advertises no provider leaves the previous entry in place. `capability_support` is keyed by `Language` alone while `clients` is keyed by `(Language, root)`, so a second root for the same language re-enters this block and every other flag is overwritten last-write-wins while this one is not. Blast radius is small — the worker independently re-picks by `on_type_triggers`, so nothing unadvertised is sent — but it makes the app fire requests the worker silently drops. Fix: `support.on_type_triggers.remove(&lang)` in the `else` branch, so the map is written on every probe like its neighbours.

## Test coverage

- **The tab-switch replay (finding 1) has no test.** The new `on_type_formatting_stays_inert_when_disabled_or_unadvertised` asserts `app.editor.last_typed.is_none()` twice, but only ever on the tab that was active at the tick — precisely the case that works. Write: open two files, type in tab A, switch to tab B **without** ticking, tick, reopen tab A, and assert `app.editor.last_typed.is_none()`.
- **The type-over seq hole (finding 3) has no test.** `typing_records_last_typed_but_paste_does_not` covers the auto-paired *opener* but stops before the type-over closer.
- **`apply_merged_settings` parity (finding 2) has no test.** The new `config_layers.rs` test asserts `m.prefs.format_on_type` and its provenance, but nothing drives `remerge_settings` and checks `app.format_on_type`.
- **`drain_lsp_on_type_format` is entirely unexercised** — no test constructs a `FormatResult` and drives the stale/fresh branches, so the id/seq/path triple guard, the empty-edit no-op case, and the `clamp_cursor` call are all untested.

## Nitpicks

- `src/lsp/manager.rs:3547` — `t.contains(ch)` treats the concatenated trigger string as a char set. LSP's `firstTriggerCharacter`/`moreTriggerCharacter` are `String`s; servers use single characters in practice, but a hypothetical multi-char entry would produce spurious membership hits. A `Vec<char>` (or `HashSet<char>`) would express the intent exactly.
- `src/app/mod.rs:8238-8240` — this is the repo's first correct UTF-16 position conversion for LSP; every other request site sends raw char columns (`trigger_completion`, `request_hover`, `request_definition`). Not this PR's bug, and the new code is the right side of the inconsistency, but the divergence is worth a follow-up issue.
- `src/lsp/manager.rs:3536-3554` — the sibling `request_formatting` carries an explicit "Every bail below MUST still answer" invariant; this function deliberately breaks it. The doc comment does explain why and nothing waits on the reply, so it is fine — but a one-line pointer at the divergence would stop the next reader from "fixing" it.

## Scores

- **CodeRabbit score (diff quality) — 2/5** — three majors: a staleness latch that does not hold across tab switches, a `edit_seq` proxy the type-over branch defeats, and a reload path that drops the new key.
- **Greptile score (codebase fit) — 2/5** — `apply_merged_settings` breaks the parity every other allowlisted preference follows, `on_type_triggers` breaks the unconditional-insert convention of its sibling capability maps, and `docs/SETTINGS.md` — the file that enumerates the allowlist this PR extended — was left stale.

VERDICT: REQUEST_CHANGES

vitali87:

Round 3, re-run after the round-2 fixes in `dfad904`; CodeRabbit is still rate-limited on this PR.

---

## 1. Walkthrough

PR #275 adds VS Code's `editor.formatOnType` (issue #254 item 4): a `format_on_type` pref (default off), a per-frame `tick_on_type_formatting` that fires `textDocument/onTypeFormatting` when the last edit was a real keystroke in a server-advertised trigger set, and a dedicated result channel whose reply is applied only while the buffer has not moved. This third-round diff (`dfad904`) hardens three things found in round 2: background tabs are disarmed so a keystroke + tab-switch in one input burst cannot replay later, `apply_merged_settings` now carries the key so live config reload applies it, and the empty trigger set is `remove`d rather than skipped so a re-probe is last-write-wins.

| File | Summary |
|---|---|
| `src/app/mod.rs` | New `format_on_type` / `on_type_request` fields; `tick_on_type_formatting` (called each main-loop iteration) now sweeps `last_typed = None` across every editor in every group before processing the taken record; `drain_lsp_on_type_format` applies a fresh reply; `apply_merged_settings` carries `format_on_type`; `toggle_format_on_type`; Settings row + palette dispatch. |
| `src/lsp/manager.rs` | `Cmd::RequestOnTypeFormatting`, `on_type_triggers: HashMap<Language, String>` capability map (now `remove`d on empty), `language_on_type_triggers`, dedicated `on_type_rx` channel, worker `request_on_type_formatting`. |
| `src/lsp/client.rs` | `on_type_formatting` wrapper over `DocumentOnTypeFormattingParams`. |
| `src/widgets/editor.rs` | `last_typed: Option<(char, u64)>` set on `insert_char` (auto-pair branch included) and `insert_newline`, never on `insert_str`; new type-over assertions in `typing_records_last_typed_but_paste_does_not`. |
| `src/prefs.rs` | `format_on_type` pref + `save_format_on_type`, symmetric with `format_on_save`. |
| `src/config_layers.rs` | `format_on_type` added to `WORKSPACE_ALLOWED_KEYS` and to the `.vscode/settings.json` subset (`editor.formatOnType`). |
| `src/widgets/command_palette.rs` | `Command::ToggleFormatOnType` across all five match arms. |
| `src/app/tests.rs` | Two new tests: tab-switch disarm, settings-reload parity. |
| `docs/SETTINGS.md`, `docs/KEYBINDINGS.md`, `docs/LAYOUT.md` | Allowlist, VS Code subset table, and feature-list entries. |

### Verification of the round-2 fixes (all hold)

1. **Sweep completeness** — verified exhaustively. `EditorTabs::editors` (`src/widgets/editor.rs:10738`) is the *only* owning `Vec<Editor>` in the codebase, and `App::editor` (`src/app/mod.rs:2210`) plus `LayoutNode::Leaf(Some(EditorTabs))` (`src/app/editor_layout.rs:58`, reached via `inactive_groups_mut`) are the only two `EditorTabs` storage sites. Markdown preview, diff, and preview/scratch state are *fields inside* `Editor` (`editor.rs:2130`, `:2146`, `:2003`), not separate instances. No orphan Editor exists. Clearing `last_typed` on non-active editors breaks nothing: `grep -rn last_typed src/` shows `tick_on_type_formatting` is its only reader.
2. **Tick frequency** — `main_loop` polls at 8 ms (`src/app/mod.rs:38963`) and calls the tick unconditionally at `src/app/mod.rs:38682`; there is no nested event loop that could skip it. The sweep cost is negligible and the `once(&mut self.editor).chain(inactive_groups_mut())` fan-out is already established per-frame practice at `src/app/mod.rs:6210` (`sync_git_gutters`).
3. **`edit_seq`-only staleness (round 2, item 3)** — **the author's reasoning is correct; I could not break it.** I probed `apply_span_edits` + `clamp_cursor` directly. Type-over (`editor.rs:4700-4705`) mutates only `cursor_col` and returns, leaving text byte-identical and `edit_seq` untouched, so an in-flight reply still applies to the text it was computed against. Server edits are absolute `(row, col)` spans (`text_edit_to_span`, `manager.rs:3729`) applied bottom-to-top — caret position is not an input to them. Adding a caret-unchanged check would discard valid replies after mere arrow motion without fixing anything. Separately, `sync_lsp()` (`:38655`) runs before the tick (`:38682`), so the server's document is current when the position is sent.
4. **`remove` on empty** — matches sibling behavior. Every sibling (`manager.rs:1978-1984`, `:1999-2000`) `insert`s a bool, so absence means "not probed"; `on_type_triggers` is a `String` map where absence means "no triggers", and the sole consumer `language_on_type_triggers` (`:1287`) plus its caller (`mod.rs:8249-8252`) treat `None` and `Some("")` identically. Nothing distinguishes them.
5. **The new tests** — both are revert-sensitive; I verified empirically. Removing the sweep fails `a_tab_switch_disarms_the_background_tabs_on_type_trigger` at `tests.rs:29041`; removing `self.format_on_type = p.format_on_type` fails `a_settings_reload_applies_format_on_type_live` at `tests.rs:29063`. `open_pinned` genuinely switches tabs rather than reopening (`editor.rs:11560-11564` returns early via `find_tab_matching` + `select`), so the tick does run with a different tab active.
6. **Repo conventions** — `cargo clippy --bin croft --all-features` is clean; `cargo test --bin croft on_type` passes 4/4. On the version-bump rule: `CONTRIBUTING.md` documents no such rule, and the four sibling #254 PRs (#271 folding ranges, #272 expand/shrink selection, #273 linked editing, #274 Format Selection) all merged touching neither `Cargo.toml` nor `src/release_notes.rs`. This PR matches that batching practice. **Not flagged.**
7. **`docs/SETTINGS.md`** — matches code exactly. The allowlist prose (`SETTINGS.md:84-87`) is character-for-character the same set as `WORKSPACE_ALLOWED_KEYS` (`config_layers.rs:92-107`), and the VS Code table row (`SETTINGS.md:102`) matches `load_vscode_subset` (`config_layers.rs:402-404`).

## 2. Findings

- `src/app/mod.rs:8273-8315` **[major/tests]** `drain_lsp_on_type_format` — the entire staleness-and-apply path — has **zero test coverage**. `grep -rn "drain_lsp_on_type_format" src/app/tests.rs` returns nothing; the only two on-type assertions (`tests.rs:28993`, `:29008`) check `on_type_request.is_none()` on the *request* side. This is the single most subtle piece of the PR and the exact logic round 2 disputed and round 3 now defends in a five-line comment (`:8291-8296`) — yet nothing verifies that a fresh reply applies, that a bumped `edit_seq` drops one, that a path mismatch drops one, or that a superseded `request_id` is ignored without clearing the newer request. A comment arguing for an invariant is not a test of it. Fix: push a synthetic `FormatResult` through the drain and assert each branch — see Test coverage below.

- `src/widgets/editor.rs:5313-5314` **[minor/correctness]** `multi_insert_char` does not set `last_typed`, so on-type formatting silently never fires during multi-cursor editing. I verified this is the *correct* behavior — `multi_apply` (`:5330-5348`) edits at N carets as one step, and a reply computed at a single position would be wrong for the others — but the choice is undocumented and untested, unlike the paste exclusion, which gets both a doc comment (`:5079-5081`) and an assertion (`:12077-12079`). A future refactor that routes multi-cursor typing through `insert_char` would silently arm a hazardous request. Fix: document the exclusion at the call site and assert it.

- `src/widgets/command_palette.rs:363` **[minor/style]** `Command::ToggleFormatOnType => "Editor: Toggle Format on Type"` uses the `Editor:` prefix while its immediate sibling on the next line uses `Command::ToggleFormatOnSave => "Preferences: Toggle Format on Save"`. VS Code names both under `Preferences:` (they are both `editor.*` settings toggles), and `docs/KEYBINDINGS.md:230` hardcodes the `Editor:` string, so the two must change together.

## 3. Test coverage

- **`drain_lsp_on_type_format` staleness branches (all four) — untested.** This is the gap behind the major finding. The test needs a `FormatResult` injected into the on-type channel; if the channel is not test-reachable, extract the decision into a pure helper and test that directly. Assert: (a) matching id + matching seq + matching path applies the edits; (b) `edit_seq` advanced by one keystroke → dropped; (c) `result.path != requested` → dropped; (d) a reply whose `request_id` does not match the armed one is skipped **without** clearing `on_type_request`, so the newer in-flight request still applies when its own reply lands. Branch (d) is the one most likely to regress silently, since it is the only path that must *not* run `self.on_type_request = None`.

- **Caret position after a line-count-changing reply — untested, and worth a characterization test.** I probed this empirically: with `lines = ["a","","","b","TARGET","x","y","z"]` and the caret on `"TARGET"` (row 4), an edit deleting rows 1–2 leaves the caret at row 4, which is now `"y"` — the caret silently lands on a different line, and `clamp_cursor` does nothing because row 4 is still in range. **Not filed as a finding**: `drain_lsp_format` (`src/app/mod.rs:9479-9481`) uses the identical `apply_rename_to_open_tab` + `clamp_cursor` pattern, so this is pre-existing repo behavior, not something #275 introduces. But on-type formatting makes it reachable *silently on ordinary keystrokes* rather than on an explicit Format Document, which raises how often a user would hit it.

- **Split-group disarm — untested.** `a_tab_switch_disarms_the_background_tabs_on_type_trigger` covers tabs within one group, but the sweep's second loop (`mod.rs:8222-8226`, over `inactive_groups_mut()`) has no test.

## 4. Nitpicks

- `src/app/mod.rs:8219-8226` — the sweep uses two separate `for` loops where the established repo idiom for exactly this fan-out is `std::iter::once(&mut self.editor).chain(self.editor_layout.inactive_groups_mut())` (`mod.rs:6210`).
- `src/app/mod.rs:8219` — the sweep re-clears `self.editor.editors[active]`, which `take()` at `:8212` already drained. Harmless and the comment at `:8217` calls it out, but the redundancy invites a future reader to wonder whether the `take` or the sweep is load-bearing.
- `docs/KEYBINDINGS.md:230` — the example trigger set `` `}`, `;`, newline `` is illustrative but does not match the most likely server: rust-analyzer advertises `=`, `.`, `>`, `{`. The clause "per its `documentOnTypeFormattingProvider`" hedges it correctly, so this is cosmetic.

## 5. Scores

- **CodeRabbit score (diff quality) — 3/5.** One major (`drain_lsp_on_type_format` entirely untested) plus two minors inside the changed lines. The new code itself is correct — I attacked the `edit_seq`-only staleness argument directly with probes and could not break it, both new tests are revert-sensitive, and clippy is clean — but the PR's most contested logic ships with a defending comment and no assertion.
- **Greptile score (codebase fit) — 5/5.** No finding's evidence lives outside the diff. The sweep provably reaches every `Editor` the process can hold; `remove`-on-empty matches sibling capability semantics and no consumer distinguishes absent from empty; `format_on_type` is symmetric with `format_on_save` across pref, toggle, persist, allowlist, VS Code subset, Settings row, and palette; `docs/SETTINGS.md` matches the code exactly; and the absent version bump matches the actual practice of the four sibling #254 PRs rather than violating it.

VERDICT: REQUEST_CHANGES

PR #276, opened 2026-08-24
feat(lsp): prepare-rename validation and prompt pre-fill (#254 item 6)

vitali87:
Item 6 of the #254 LSP completeness batch: `textDocument/prepareRename`. Rename no longer fires blind.

## What's here

**Two-phase F2**: Rename Symbol now sends `prepareRename` first and defers the prompt on the verdict. The worker **always answers** (the prompt is waiting):
- a validated range/placeholder → the prompt pre-fills the server's exact symbol text (placeholder wins over range extraction)
- `Ok(None)` → "Nothing renameable at the cursor" — fail fast, no prompt
- a server error → "Rename rejected: <the server's own message>" (root cause line, not a generic error)
- `unsupported` (no `prepareProvider`, unroutable doc) or the protocol's "default behavior" answer → the unchanged word-under-cursor prompt, so servers without the capability keep today's flow exactly

**Capability**: `prepareProvider: true` only exists on an options-shaped `renameProvider` — a bare `true` provider does NOT imply prepare support, and the shape test pins that. Nothing is sent unadvertised. Positions go over the wire as real UTF-16 (`pos_to_utf16`), unlike the pre-existing rename request's raw char columns.

Verdicts are gated on id + path + edit_seq, so a reply landing after an edit or a tab switch is dropped instead of opening a stale prompt.

- [x] Prepare-rename validates the position and pre-fills the rename prompt
- [x] Invalid positions fail fast with the server's message
- [x] Capability negotiation: silent degradation to the word prompt, nothing sent unadvertised

## Tests

Capability-shape test (options-with-prepareProvider only), unroutable-answers-unsupported worker test, and an app-level end-to-end test driving the real worker's unsupported verdict into the fallback prompt pre-filled with the word under the cursor. Full suite: 3408 passed; only the known pre-existing environmental failures. Clippy `--all-targets -D warnings` clean.

## Summary by CodeRabbit

* **New Features**
  * F2 Rename Symbol now validates rename requests with the language server when supported.
  * Rename prompts are pre-filled with the server-provided symbol name and range.
  * Unsupported or rejected renames fall back appropriately or show the server’s specific reason.

* **Documentation**
  * Updated keybinding and layout documentation to explain the improved rename behavior.

* **Chores**


vitali87:

---

## Walkthrough

F2 (Rename Symbol) becomes two-phase: `start_rename_symbol` now fires `textDocument/prepareRename` and defers the prompt until a worker verdict arrives, so the prompt pre-fills the server's validated range/placeholder and an invalid position fails fast with the server's own message. A new `prepare_rename_supported` gate requires an options-shaped `renameProvider` carrying `prepareProvider: true`; anything else answers `unsupported` and routes to the pre-existing word-under-cursor prompt. Verdicts are matched on request id, path, and `edit_seq` before opening the prompt.

| file | summary |
|---|---|
| `src/app/mod.rs` | New `prepare_rename_request` field; `start_rename_symbol` rewritten to send prepareRename; word-prompt logic extracted to `open_rename_prompt_fallback`; new `drain_lsp_prepare_rename` verdict handler wired into `main_loop`. |
| `src/lsp/manager.rs` | New `PrepareRenameResult`, `Cmd::RequestPrepareRename`, channel + `request_prepare_rename`/`drain_prepare_rename`, `supports_prepare_rename` on `ManagedClient`, worker `request_prepare_rename`, helpers `prepare_rename_supported`/`range_tuple`, two unit tests. |
| `src/lsp/client.rs` | Thin `prepare_rename` wrapper over the server's `prepare_rename`, contexted `"prepare_rename"`. |
| `src/widgets/editor.rs` | New `pos_to_utf16` and `utf16_col_to_char` position helpers. |
| `src/app/tests.rs` | One app-level test polling for the fallback prompt. |
| `docs/LAYOUT.md`, `docs/KEYBINDINGS.md` | Prose updated to mention prepare-rename validation and pre-fill. |

## Findings

- **`Cargo.toml:3` / `src/release_notes.rs:62` [blocker/docs]** — The shipped-change convention is violated: version is still `0.1.758` while `origin/main` is already at `0.1.759`, and `RELEASE_NOTES` still contains only the unrelated entry `"Croft Light: the selected activity-bar icon no longer disappears…"`. `src/release_notes.rs:8-16` states the rule in the module doc itself: *"Update `RELEASE_NOTES` on every version bump… REPLACE this list on every bump; do not append."* `docs/ARCHITECTURE.md:50` repeats it ("updated on every version bump"). This user-visible feature ships with neither. Fix: bump to `0.1.760` (past main) and replace the list with a single `NoteKind::Feature` entry describing prepare-rename.

- **`src/widgets/editor.rs:5390` [major/correctness]** — `pos_to_utf16` is a **duplicate definition** of a function that already exists on `origin/main` at `src/widgets/editor.rs:5827`, byte-for-byte identical, added by the linked-editing work merged after this branch forked. Git's textual auto-merge silently dedupes it only because the two blocks are identical; that is luck, not correctness — any non-identical resolution (or a rebase that reorders the hunk) yields `E0428 duplicate definitions with name pos_to_utf16`. Verified by diffing `origin/main:src/widgets/editor.rs` against the PR head and by inspecting the auto-merged tree `d24e2cf`, where the function appears once at line 5827 purely by textual coincidence. Fix: drop the new `pos_to_utf16` entirely and rebase onto main.

- **`src/widgets/editor.rs:5398` [major/correctness]** — `utf16_col_to_char` re-implements `utf16_col_to_char_pub`, which already exists on main at `src/widgets/editor.rs:4681`, **and does so less safely**. Main's version is `self.lines.get(row).map(|l| utf16_to_char_col(l, character)).unwrap_or(0)`; the new one is `utf16_to_char_col(&self.lines[row], character)` after `row.min(self.lines.len().saturating_sub(1))` — which yields index `0` on an empty `lines`, then **panics** on `self.lines[0]`. `self.lines.is_empty()` is guarded in roughly twenty places across `editor.rs` (2366, 2617, 3343, 4632, 4651, 4819, 5018, 5408, 5928, 6296, 6324, …), so the empty state is plainly reachable in this codebase. `pos_to_utf16:5392` has the identical unguarded `&self.lines[row]`. Confirmed the merged tree `d24e2cf:src/widgets/editor.rs` carries **both** `utf16_col_to_char_pub` (4681) and `utf16_col_to_char` (5836). Fix: delete both new helpers and call main's `cursor_position_utf16()` (main `editor.rs:5871`, which is exactly `pos_to_utf16(cursor_row, cursor_col)`) plus `utf16_col_to_char_pub`.

- **`src/lsp/manager.rs:4309-4369` [major/correctness]** — Capability negotiation is only half done: `text_document` never declares `rename`. The block runs completion → signature_help → semantic_tokens → document_symbol → inlay_hint → call_hierarchy → document_highlight → publish_diagnostics → code_action → `..Default::default()`, with no `rename` field. Per LSP 3.12, a server publishes `RenameOptions.prepareProvider` in response to the client's `rename.prepareSupport`; without it many servers advertise a bare `renameProvider: true`, which `prepare_rename_supported` (`manager.rs:3752`) correctly rejects — so the feature silently never engages. This file's own comments establish the exact opposite convention four times over, e.g. `manager.rs:4314-4316` *"Advertise inlay-hint support so servers that gate the provider on it (vtsls) publish `inlayHintProvider`"* and `4321-4323` for call-hierarchy/documentHighlight. Verified `RenameClientCapabilities { dynamic_registration, prepare_support, prepare_support_default_behavior, honors_change_annotations }` exists in the pinned lsp-types 0.95.1 (`src/rename.rs:33-62`). Fix, alongside the other declarations:
  ```rust
  rename: Some(lsp_types::RenameClientCapabilities {
      dynamic_registration: Some(false),
      prepare_support: Some(true),
      prepare_support_default_behavior: Some(
          lsp_types::PrepareSupportDefaultBehavior::IDENTIFIER,
      ),
      ..Default::default()
  }),
  ```

- **`src/lsp/manager.rs:3462` [major/correctness]** — The `PrepareRenameResponse::DefaultBehavior` arm is dead code, for the same root cause. lsp-types 0.95.1 `src/rename.rs:44-51` documents `prepare_support_default_behavior` as *"Client supports the default behavior result"*, and the spec only permits a server to return the `{ defaultBehavior: true }` shape when the client declared it. Since nothing declares it (previous finding), no conforming server can produce this arm. The PR body lists "the protocol's 'default behavior' answer" as a supported path; it is unreachable as written. Fixed by declaring `prepare_support_default_behavior` above.

- **`src/app/mod.rs:9279-9281` [major/correctness]** — A dropped verdict strands the UI. When the guard fails, the code does `self.prepare_rename_request = None; continue;` with `changed` still `false` — so no prompt opens, no status is written, no redraw is requested, and the screen keeps reading `"Preparing rename…"` (set at `mod.rs:9232`) indefinitely. This is reachable without any edit: `editor.rs:3464` bumps `edit_seq` on a **tab switch**, so F2 → switch tab → switch back leaves the status wedged with no way to tell rename died. The sibling handler `drain_lsp_rename` (`mod.rs:9349-9358`) always writes a terminal status, which is the established convention here. Fix:
  ```rust
  if self.editor.path.as_deref() != Some(path.as_path()) || self.editor.edit_seq != seq {
      self.status = String::from("Rename cancelled");
      changed = true;
      continue;
  }
  ```

- **`src/app/mod.rs:9297-9299` [minor/correctness]** — Row clamping is inconsistent inside the range-extraction closure. `utf16_col_to_char(sr, sc)` and `utf16_col_to_char(er, ec)` clamp `sr`/`er` internally to the last line, but the very next `self.editor.lines.get(sr)` uses the **unclamped** `sr`. For an out-of-range server range the columns are computed against the clamped last line while the lookup returns `None`, so the two disagree. It degrades to `unwrap_or_default()` → empty → fallback rather than crashing, but the mismatch is latent. Fix: bail out early when `sr >= self.editor.lines.len()` instead of relying on two different clamping behaviors.

- **`src/app/mod.rs:9226` vs `mod.rs:34974` [minor/correctness]** — The same caret is now sent to the server in two different column encodings within one F2 flow: prepareRename sends real UTF-16 via `pos_to_utf16`, while the follow-up `lsp.request_rename(path, row as u32, col as u32, new_name)` at `mod.rs:34974` still sends raw **char** columns. On any line containing astral-plane characters before the symbol, the server validates one position and renames at another. The PR body acknowledges the divergence ("unlike the pre-existing rename request's raw char columns") but leaves it. The pre-existing bug is not this PR's fault, but this PR is what makes the two positions disagree inside a single user action. Fix: `let (line, character) = self.editor.pos_to_utf16(row, col);` at the `PromptKind::RenameSymbol` commit site too.

- **`src/lsp/manager.rs:3749-3754` [minor/style]** — The new doc comment is appended to an already-broken comment stack rather than fixing it, leaving three unrelated doc blocks on `prepare_rename_supported` and `code_action_supported` (3761) undocumented. Lines 3742-3745 describe `one_of_supported` (defined far away at 3854) and 3746-3748 describe `code_action_supported`. The two-comment stack pre-dates this PR (confirmed at merge base `1a9da3a`, lines 3594-3601), but this diff inserts a third block and a whole function into the middle of it, moving the misattribution further from its owner. Fix: move 3742-3745 to `one_of_supported`, move 3746-3748 back down to `code_action_supported`, and leave only the prepare-rename doc on the new function.

## Test coverage

- **The verdict paths that carry the actual feature are entirely untested.** `an_unroutable_prepare_rename_answers_unsupported` (`manager.rs:5745`) only proves the early `self.docs.get(&path)` miss answers, and the app test only reaches the same `unsupported` branch (see below). Nothing exercises `RangeWithPlaceholder` → placeholder pre-fill, `Range` → range-text extraction, `Ok(None)` → `"Nothing renameable at the cursor"`, or `Err` → `"Rename rejected: <msg>"` — i.e. all four behaviors the PR body claims. Write a unit test constructing `PrepareRenameResult` values directly and driving `drain_lsp_prepare_rename`, asserting `app.prompt.buffer` for the placeholder case and `app.status` for the `None`/`Err` cases.
- **`rename_symbol_defers_on_prepare_and_falls_back_to_the_word_prompt` (`src/app/tests.rs:28978`) does not test what its name and comment claim.** The comment says "With no server, the worker's unsupported verdict" — but `docs` is only ever populated by `open_doc` (`manager.rs:2006`), and the test opens the file via `app.editor.open_pinned` directly, which never routes through the LSP open path. So it deterministically hits the `self.docs.get(&path)` miss at `manager.rs:3422` — the "document unknown to the LSP layer" branch, not capability negotiation. It therefore duplicates the manager-level test one layer up while burning a 10-second `Instant`-deadline poll loop with 20 ms `thread::sleep`s. Rewrite it to open the document through the app's real open path, or rename it to say it covers the unknown-document branch and drop the sleep loop.
- **The stranded-status path has no test.** Add: fire `start_rename_symbol`, bump `app.editor.edit_seq`, feed a matching verdict, and assert the status is no longer `"Preparing rename…"` — this test fails today.
- **Empty-buffer positions have no test.** Add an `editor.rs` unit test calling `pos_to_utf16(0, 0)` and `utf16_col_to_char(0, 0)` on an editor with `lines: vec![]`; both panic today.

## Nitpicks

- `src/app/mod.rs:9227-9229` — the `let Some(lsp) = self.lsp.as_mut() else { return; }` is dead: `mod.rs:9214` already returned when `self.lsp.is_none()` and nothing between them can clear it. It also returns silently without a status, unlike every other early exit in the function. Fold the two checks into one.
- `src/lsp/manager.rs:3413` — the `unsupported` closure takes `tx` as a parameter while already capturing `request_id` from the enclosing scope; capturing `tx` too would drop five repetitions of `(tx, path)` down to `(path)`.
- `src/lsp/manager.rs:3476` — `.unwrap_or("rename rejected")` is unreachable: `str::lines().next()` returns `None` only for an empty string, and `root_cause().to_string()` on an `anyhow` error is never empty in practice.

## Scores

- **CodeRabbit score (diff quality) — 2/5**: two majors inside the changed lines — the unguarded `self.lines[row]` panic in both new editor helpers, and the stranded `"Preparing rename…"` status on a dropped verdict — plus an untested feature surface where all four advertised verdict paths lack coverage and the one app test exercises a different branch than it claims.
- **Greptile score (codebase fit) — 1/5**: a blocker on the repo's documented version-bump + `release_notes.rs` replacement rule (self-documented at `release_notes.rs:8-16`, restated in `ARCHITECTURE.md:50`), compounded by three cross-file majors — `pos_to_utf16` and `utf16_col_to_char` duplicating helpers main already merged, and the missing `rename` client capability that contradicts the four adjacent "advertise so servers publish the provider" comments in the same block and renders the `DefaultBehavior` arm dead. The branch is also 38 commits behind main and conflicts in `docs/LAYOUT.md`.

VERDICT: REQUEST_CHANGES

vitali87:

Round 2, re-run after the round-1 fixes in `94edb21`. (CodeRabbit's rate limit has since reset and its own review is running on this PR; this one is kept as the recorded artifact for the round-1 loop.)

---

## Walkthrough

F2 (Rename Symbol) becomes two-phase: `start_rename_symbol` fires `textDocument/prepareRename` and parks with a "Preparing rename…" status instead of opening the prompt immediately. `drain_lsp_prepare_rename`, pumped from `main_loop`, resolves the server's verdict into one of five outcomes — server placeholder, validated-range text, refusal with the server's own message, "nothing renameable", or the word-under-cursor fallback for servers without `prepareProvider`. Commit `94edb21` addressed round one by declaring `rename.prepareSupport` in the client capabilities (without which the whole feature was inert), extracting the verdict handler so its shapes are unit-testable, and adding a "Rename cancelled" status for stale verdicts.

### Round-one items verified as fixed

All confirmed against the working tree, not just the diff:

- **Capability declaration** — `src/lsp/manager.rs:5346-5353`. The `rename:` field sits inside the `TextDocumentClientCapabilities` literal opened at line 5232, between `selection_range` and `publish_diagnostics`; the struct literal is well-formed. `PrepareSupportDefaultBehavior::IDENTIFIER` is `PrepareSupportDefaultBehavior(1)` in the pinned lsp-types 0.95.1, matching the spec's `Identifier = 1`. The claim that the feature was previously inert **holds**: `prepare_rename_supported` requires `OneOf::Right(opts)` with `prepare_provider == Some(true)`, and per LSP 3.12 a server only emits the options shape in answer to `prepareSupport`.
- **Merge-conflict resolution** — `git merge-base origin/main HEAD` == `origin/main` (4f533ef), and `git diff origin/main..HEAD` is exactly the 9 feature files with 512 insertions, containing only this feature's additions. `src/widgets/editor.rs` and the pre-existing content of `src/app/tests.rs` have a zero-byte delta against main. Nothing from main was dropped or duplicated; on-type formatting, range formatting, linked editing, documentColor/color presentation, folding ranges, and selection ranges all retain their code and tests. The `selection_range:` termination is intact.
- **Duplicate-helper removal** — `grep utf16_col_to_char` across `src/` returns only `utf16_col_to_char_pub`. No caller was missed. The row-clamp semantic difference is unreachable: the new call sites guard with `if sr == er` plus `.get(sr)`, so an out-of-range row yields `String::new()` either way. `pos_to_utf16`'s empty-buffer guard came from main and is covered by its own test.
- **Control flow of the extracted handler** — equivalent to the original loop. Both `return false` paths (no armed request, id mismatch) leave `prepare_rename_request` armed; every `return true` path clears it first. The drain loop collects queued results into a `Vec` before dispatching and accumulates with `changed |=`, correct for multiple queued verdicts.
- **Version bump + release notes** — `CONTRIBUTING.md:8-19` states the rule verbatim and CI enforces it. main is 0.1.762, this branch 0.1.763; `RELEASE_NOTES` is replaced, not appended.
- **Tests** — both new tests pass and test what their names claim. The `arm` closure re-reads `edit_seq` live per call, so the `edit_seq += 1` in the stale case does not leak into the superseded-reply case; the assertions are independent.

## Findings

- **`src/app/mod.rs:9780`** **[major/correctness]** `prepare_rename_request` is armed but never cleared on any cancellation path, so a cancelled rename leaves a live request that hijacks the next verdict. `grep -n "prepare_rename_request" src/app/mod.rs` returns exactly five sites: the field, the `None` initializer, the arm, the read, and the single clear inside `apply_prepare_rename_verdict`. There is no clear when the user presses Esc on the resulting prompt. Confirmed empirically with a throwaway test (since reverted): after arming and then dismissing the prompt, `prepare_rename_request.is_some()` is still true.

  Consequence: press F2, get the prompt, press Esc, then move the caret elsewhere in the same buffer *without editing* (pure cursor motion does not bump `edit_seq`). The stale request is still armed with the old `(row, col)`. If the user hits F2 again and the first server reply arrives after the second arm, `apply_prepare_rename_verdict` matches on id alone and the path/`edit_seq` gate passes, so a prompt reopens pre-filled for the abandoned position — and `PromptKind::RenameSymbol { row, col }` carries those stale coordinates straight into `lsp.request_rename(...)`. The rename lands on the wrong symbol.

  This also departs from the repo's own convention for deferred caret requests: `linked_request` — same `(id, path, seq)` tuple shape — is cleared at three sites whenever its preconditions lapse, and `hover_request_id` is cleared on every keypress. `prepare_rename_request` is the only one of the family with a single clear site.

- **`src/app/mod.rs:9835-9895`** **[minor/style]** `apply_prepare_rename_verdict`'s entire body is wrapped in a vestigial bare block. It is a leftover from the extraction out of the drain loop, where the block scoped the `self.lsp.as_ref()` borrow. In the extracted function there is no borrow to scope: the block indents 60 lines and buys nothing, and the final `true` reads as if something followed it.

- **`src/app/mod.rs:9762-9765`** **[minor/correctness]** The "No symbol under cursor" check moved behind the LSP round trip, changing which message a user sees and when. On main, `word_at` ran before the `self.lsp.is_none()` check, so F2 on whitespace reported "No symbol under cursor" instantly. Now it sets "Preparing rename…", makes a server round trip, and for a server that advertises `prepareProvider` yields "Nothing renameable at the cursor" instead. Arguably a better message, but an unremarked behavior change.

## Test coverage

- **The stale-request lifetime (the major above).** Nothing exercises what happens to `prepare_rename_request` after the prompt is dismissed or after a second F2.
- **The `main_loop` wiring.** `drain_lsp_prepare_rename` is called and OR'd into the repaint condition, but no test drives the drain function itself — both new app tests call `apply_prepare_rename_verdict` directly. The collect-then-dispatch structure is untested.
- **The multi-line validated range.** The handler returns `String::new()` when `sr != er`, routing to the fallback via the `word.is_empty()` guard. The verdict test only ever passes single-line ranges.

## Nitpicks

- `src/lsp/manager.rs:1504-1506` — `drain_prepare_rename` has no doc comment, though `drain_rename` below it is also bare, so this matches local practice.
- `src/app/mod.rs:9806-9808` — `drain_lsp_prepare_rename`'s doc comment describes what `apply_prepare_rename_verdict` does, not what the drain loop does.
- `src/app/mod.rs:9829` — "so the four verdict shapes are testable" undercounts: the handler distinguishes six, and the test covers all six plus the superseded reply.

## Scores

- **CodeRabbit score (diff quality) — 3/5.** One major (the never-cleared `prepare_rename_request`) plus two minors; the round-one fixes are all genuine and the capability declaration is correct against the pinned lsp-types.
- **Greptile score (codebase fit) — 3/5.** Same major, and it is specifically a departure from the codebase's own established pattern for deferred caret requests — `linked_request` and `hover_request_id` both clear their in-flight state on lapse, this one does not. Comment density, error-handling style, and test placement otherwise match the repo; the merge against main is verifiably clean.

VERDICT: REQUEST_CHANGES

---

**Addressed in `ca1c635`:** the major (Esc now disarms the in-flight request, with a revert-sensitive test), the vestigial block (removed), and both test-coverage gaps that were actionable (cancellation lifetime, cross-line range). The behavior-change minor is left as-is deliberately — routing whitespace-F2 through the server yields the server's own "nothing renameable" answer, which is the more accurate message.

PR #278, opened 2026-08-24
feat(lsp): document links via Ctrl+click (#254 item 8)

vitali87:
Item 8 — the last of the #254 LSP completeness batch: `textDocument/documentLink`, routed through the editor's existing Ctrl+click dispatch.

## What's here

**Cadence + storage**: link sets arrive on the open/change cadence beside inlay hints (fire-and-forget, seq-gated, fanned out to split panes), path-stamped so a set for another file never answers. Links the server left target-less are dropped at normalisation — resolving them needs a `documentLink/resolve` round trip croft deliberately skips for now.

**Click routing**: the Ctrl+click handler consults the link set before firing Go to Definition — a URL in a comment has no definition at all, and an import specifier's link and definition land in the same file anyway (VS Code follows the link too). Web targets go to the system opener, refused with guidance in a remote session (a headless host's opener would claim success invisibly — the same posture as the media opener); `file://` targets open in the editor. Unlinked ranges keep today's behavior exactly.

**Precedence note vs the issue**: the issue asks for lower precedence than croft's `path:line` file refs — those live in the *terminal* click path, which this PR doesn't touch; in the editor there is no competing file-ref handler, so the only precedence decision is link-vs-definition, resolved link-first as above.

**Capability**: `documentLinkProvider` presence, silent degradation, nothing sent unadvertised.

- [x] Ctrl+click on server-resolved URLs and import-path links
- [x] Routed through the existing click dispatch without disturbing file refs or definition
- [x] Capability negotiation: silent degradation, nothing sent unadvertised

## Tests

Editor: in-range lookup with exclusive end + wrong-file rejection. App: a `file://` link opens its target in the editor, and the dispatch's lookup answers for web links. Full suite: 3420 passed; the failures are the known environmental trio plus one PTY flake that passes alone. Clippy `--all-targets -D warnings` clean.


vitali87:

---

## Walkthrough

Adds `textDocument/documentLink` to croft's LSP client: the manager requests link sets on the existing open/change cadence (seq-gated, capability-gated on `documentLinkProvider`, target-less links dropped at normalisation), the app fans them out to every editor showing the file, and the editor stores them path-stamped with a `document_link_at(row, col)` lookup. The Ctrl+click dispatch consults that lookup before firing Go to Definition — `file://` targets open in the editor, everything else goes to the system opener (refused in a remote session). Docs in LAYOUT.md and KEYBINDINGS.md are updated to describe the new gesture.

| file | summary of its changes |
| --- | --- |
| `src/lsp/client.rs` | New `document_links(uri)` wrapper around `server.document_link`, mirroring the `inlay_hint` wrapper. |
| `src/lsp/manager.rs` | `DocumentLinkItem` / `DocumentLinksUpdate` types, a `RequestDocumentLinks` command, `links_rx` channel + `request_document_links`/`drain_document_links`, `supports_document_link` on `ManagedClient` from `caps.document_link_provider.is_some()`, and the spawned `request_document_links` worker that filters out target-less links. |
| `src/app/mod.rs` | Fires `request_document_links` on the doc-sync cadence; `drain_lsp_document_links` fan-out; `open_document_link` (file:// → editor, else system opener, refused when remote); link-first branch in `trigger_definition_at`; drain call in `main_loop`. |
| `src/widgets/editor.rs` | `doc_links` + `doc_links_path` fields, cleared in `open()`; `apply_document_links`; `document_link_at` with UTF-16→char conversion and exclusive end; one unit test. |
| `src/app/tests.rs` | One test opening a `file://` link and checking the editor-side lookup for a web link. |
| `docs/KEYBINDINGS.md`, `docs/LAYOUT.md` | Prose describing the new Ctrl+click behavior. |

## Findings

- **`Cargo.toml:3` / `src/release_notes.rs` (absent from diff) [blocker/docs]** — the PR changes shipped files (`src/app/mod.rs`, `src/lsp/*.rs`, `src/widgets/editor.rs`) but does not bump the version and does not touch `src/release_notes.rs`. Verified: the branch head is at `version = "0.1.758"` while `origin/main` is already at `"0.1.759"`, and `git diff origin/main...HEAD --stat` lists no `src/release_notes.rs`. `origin/main` commit `b6582c7` ("ci: gate PRs on a version bump and fresh release notes", closes #281) added a `release-hygiene` CI job whose script greps `^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)` minus `src/app/tests.rs` and `tests/`, then fails when `new = old`, when `old` sorts higher than `new`, or when `src/release_notes.rs` is untouched. This diff trips all three arms. The job did not appear in `gh pr checks 278` only because the PR head predates the gate; rebasing will surface it as a hard failure. Fix: rebase onto main, bump to `0.1.760`, and **replace** (not append) `RELEASE_NOTES` with a single entry describing document links, per the `REPLACE this list on every bump; do not append` rule in the module doc-comment.

- **`src/app/mod.rs:7051` [major/security]** — `open_document_link` hands any non-`file://` target straight to `open_url`, which spawns `open`/`xdg-open`/`cmd /C start` on it. The repo already ruled this exact pattern unsafe: `open_detected_url` at `src/app/mod.rs:24790-24798` lowercases the URL and refuses anything that is not `http://`/`https://`, with the comment *"a file:// or custom-scheme URI would hand `open`/`xdg-open` an arbitrary target (app launch, protocol handler) off one disguised click."* A document link is precisely a disguised click — the visible text and `l.target` are independent, and the target is attacker-influenced whenever the file is (a URL in a comment of a cloned repo, an import specifier in `node_modules`). A malicious `ms-msdt:`, `vscode://`, `smb://`, or custom-handler target fires with no confirmation. Fix, matching the existing helper:
  ```rust
  let lower = target.trim_start().to_ascii_lowercase();
  if !(lower.starts_with("http://") || lower.starts_with("https://")) {
      self.status = format!("Refused to open non-web link: {target}");
      return;
  }
  ```
  Note this also fixes the case-sensitivity hole: `target.starts_with("file://")` at line 7032 is case-sensitive, so a `FILE://…` target skips the file branch and falls through to `open_url` — `open_detected_url` lowercases first for this reason.

- **`src/widgets/editor.rs:2518-2534` (`mark_buffer_changed`, unchanged by this diff) [major/correctness]** — `doc_links` is never invalidated on edit, so stale ranges stay clickable and actionable between a keystroke and the next server round-trip. Verified: `grep -n "doc_links" src/widgets/editor.rs` shows the only clear site is inside `open()` (lines 3440-3441), i.e. file switch / external reload. `mark_buffer_changed` (line 2518) bumps `edit_seq` and already clears `self.occurrences` (line 2525) — the repo's own precedent for invalidating offset-sensitive server data on every edit — but leaves `doc_links` populated. Insert a line above the link, and the range that used to cover a URL now covers whatever text shifted into it; Ctrl+click there opens the old target. Unlike inlay hints (which tolerate staleness because a misplaced dim annotation is cosmetic), a document link is actionable, so this is a wrong-navigation / wrong-URL-opened bug. Fix:
  ```rust
  self.occurrences.clear();
  // Link ranges are actionable, not cosmetic: a stale one would fire the
  // wrong target on Ctrl+click. Cleared until the next server batch lands.
  self.doc_links.clear();
  self.doc_links_path = None;
  ```

- **`src/app/mod.rs:7036-7042` [major/correctness]** — the `file://` branch never records nav history, so Go Back cannot return from a followed link. Verified: `go_to_definition` at `src/app/mod.rs:7445-7452` opens with `self.nav.record(NavLoc { path: from, row: self.editor.cursor_row, col: self.editor.cursor_col })` before `open_at`, and `nav_back` (line 8915) pops that stack. `open_document_link` calls `self.editor.open_pinned(&p)` with no `nav.record`. The same gesture (Ctrl+click) that is documented in this very diff as having a "keyboard twin" `Ctrl+-` / `Shift+Ctrl+click` now silently loses the return trip whenever the click lands on a link instead of a symbol — and `KEYBINDINGS.md:265` still advertises `Shift`+`Ctrl`+click as navigating back for this row. Fix: record before opening, as `go_to_definition` does:
  ```rust
  if let Some(from) = self.editor.path.clone() {
      self.nav.record(NavLoc { path: from, row: self.editor.cursor_row, col: self.editor.cursor_col });
  }
  match self.editor.open_pinned(&p) { … }
  ```

- **`src/app/mod.rs:7012` [minor/correctness]** — the drain gates on `self.lsp_last_seen.get(&u.path) != Some(&u.seq)`, copied from `drain_lsp_inlay_hints` (line 6982). The stricter, offset-sensitive precedent is `drain_lsp_semantic_tokens` at lines 6916-6929, which additionally requires `self.editor.edit_seq == u.seq` (and `group.edit_seq == u.seq` per split) with the comment *"Drop a batch measured against text the buffer has moved past."* `lsp_last_seen` tracks what was last *sent*, not what the buffer currently holds, so a batch can be installed against a buffer that has already moved on. Given document links are actionable, match the semantic-tokens gate rather than the inlay one.

- **`src/app/mod.rs:7047-7050` [minor/correctness]** — the remote branch refuses outright, but croft has a purpose-built path for exactly this. `open_detected_url` (line 24799-24809) and `open_ssh_learn_more` (line 21289-21295) both check `self.drop_relay_active()` first and, when the relay is up, call `request_remote_url_open` / `pending_local_open` so the URL opens on the user's local machine. The PR body cites "the same posture as the media opener," but that precedent (`Cmd::MediaOpenExternal`, line 26267) is about a *local file* that cannot be relayed as a URL — a web link can be. As written, Ctrl+click on a URL in a remote session is strictly worse than clicking the same URL in croft's terminal pane. Fix: mirror `open_detected_url`'s relay branch before falling back to the refusal message.

- **`src/lsp/manager.rs:4172` (`build_client_capabilities`, unchanged by this diff) [minor/correctness]** — no `text_document.document_link` client capability is declared, yet the PR's acceptance criteria claim capability negotiation is handled. Verified: `grep -n "document_link" src/lsp/manager.rs` shows hits only on the *server* side (`caps.document_link_provider` at 1885); the function declares `inlay_hint` (4262), `call_hierarchy` (4267), `document_highlight` (4271), each with a comment saying servers *gate their provider on the client declaring it* (vtsls is named). `lsp_types` 0.95.1 has `ClientCapabilities::text_document.document_link: Option<DocumentLinkClientCapabilities>` (`lib.rs:1522`). Without it, vtsls-class servers may withhold `documentLinkProvider` and the feature silently never activates. (Noted as minor rather than major because the already-merged sibling `documentColor` PR has the same gap — `origin/main` declares no `color_provider` client capability either — so it is a batch-wide omission, not a regression introduced here.) Fix:
  ```rust
  document_link: Some(lsp_types::DocumentLinkClientCapabilities {
      dynamic_registration: Some(false),
      tooltip_support: Some(false),
  }),
  ```

- **branch state [minor/correctness]** — the head is 38 commits behind `origin/main` and conflicts in five files. Verified with `git merge-tree --write-tree origin/main pr-278-review`: `CONFLICT (content)` in `docs/LAYOUT.md`, `src/app/mod.rs`, `src/lsp/client.rs`, `src/lsp/manager.rs`, `src/widgets/editor.rs`. The conflicts are positional (main's `trigger_definition_at` is unchanged in substance, so no semantic drift with the merged folding / linked-editing / document-colors PRs), but they must be resolved before merge — and the rebase is required anyway for the version-bump finding above.

## Test coverage

- **Stale links after an edit.** Nothing exercises the interaction between `mark_buffer_changed` and `doc_links`. Test: `apply_document_links` a range on line 0, insert a newline above it, assert `document_link_at` no longer returns the old target at the old coordinates. This test fails on the current diff — it is the reproducer for the `mark_buffer_changed` finding.
- **`trigger_definition_at` precedence.** `ctrl_click_follows_a_document_link_before_go_to_definition` (`src/app/tests.rs:29054`) does not call `trigger_definition_at` at all — it calls `open_document_link` directly and separately asserts `document_link_at`. The test name and its comment ("the definition request is never fired for a linked range") assert something the body does not check. Test: set up an editor with a link range and a live `lsp`, call `trigger_definition_at` at a screen cell inside the range, and assert `app.definition_request_id.is_none()`; then call it outside the range and assert it is `Some(_)`.
- **`drain_lsp_document_links`.** Zero coverage of the seq gate and the split-pane fan-out — both headline claims in the PR body. Test: push a `DocumentLinksUpdate` with a mismatched `seq` and assert `document_link_at` stays `None`; push a matching one with the file open in an inactive split and assert that group's editor answers too. (`drain_lsp_inlay_hints` has ~9 editor-side tests for the same shape; this has one.)
- **Non-web scheme refusal.** No test that a `vscode://`/`ms-msdt:` target is refused rather than handed to `open_url`. Test: `app.open_document_link("ms-msdt:/id")`, assert `app.status` starts with `Refused to open non-web link` (paired with the security fix).
- **Nav history from a followed link.** Test: Ctrl+click a `file://` link, then `nav_back`, and assert the editor returns to the original file and cursor position.
- **Manager capability gating.** No test that `supports_document_link` is false for a server advertising no `documentLinkProvider`, nor that target-less links are dropped at normalisation. `color_provider_supported` in main has three assertions at `src/lsp/manager.rs:6640-6642`; the equivalent here has none.

## Nitpicks

- `src/lsp/manager.rs:1402` — `supports_document_link` is the only field in `ManagedClient` added without a doc comment, while its immediate neighbours `semantic_legend`, `semantic_supports_range`, and `supports_inlay_hints` (lines 1389-1401) each carry one naming real servers. Add e.g. `/// Whether the server advertises a documentLinkProvider (vtsls, gopls do; rust-analyzer does for paths in comments).`
- `src/widgets/editor.rs:4304-4313` — the `.unwrap_or(0)` fallbacks when `self.lines.get(sr)` / `.get(er)` miss mean a link whose `end_line` is past the buffer end collapses `ec` to 0, which can make the range test behave oddly rather than simply rejecting. `apply_occurrences` (line 4323) handles the same situation by `continue`-ing on a missing row. A `?`-style bail (return `false` from the closure) would be clearer than silently substituting 0.
- `src/app/mod.rs:7000-7002` — the doc comment says "Links change no pixels, so this never asks for a redraw," which is accurate but also flags that there is no Ctrl-hover underline or any other affordance. VS Code underlines a link on Ctrl-hover; without one, the user cannot tell which ranges will follow a link instead of a definition. Out of scope for this PR, but worth a follow-up issue.

## Scores

- **CodeRabbit score (diff quality) — 2/5**: two majors inside the changed lines (the unrestricted scheme hand-off to `open_url` at `mod.rs:7051`, and the missing `nav.record` at `mod.rs:7036`), plus a test whose name asserts precedence that its body never checks.
- **Greptile score (codebase fit) — 1/5**: a blocker against the repo's own documented, CI-enforced release-hygiene rule (no version bump, no `release_notes.rs` replacement), plus a major whose evidence is the untouched `mark_buffer_changed` clearing `occurrences` but not `doc_links`, and three minors where the diff diverges from established in-repo precedents (`open_detected_url`'s relay branch, `drain_lsp_semantic_tokens`'s stricter seq gate, `build_client_capabilities`'s per-feature declarations).

VERDICT: REQUEST_CHANGES

vitali87:

Round 2, re-run after the round-1 fixes in `45cc766`. CodeRabbit is rate-limited on this PR (its check passes as a no-op annotated "Review rate limited"), so this is the only review the change gets.

---

## Walkthrough

The change adds `textDocument/documentLink` support: `sync_lsp` requests link ranges on the same open/change cadence as inlay hints, a new worker path fetches and normalises them (target-less links dropped), and the editor stores a path-stamped `doc_links` set. The Ctrl+click dispatch consults `document_link_at` before firing Go to Definition — `file://` targets open in the editor via `open_pinned`, `http(s)` targets go to the system opener, and every other scheme is refused.

| File | Summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | 0.1.763 → 0.1.764 |
| `docs/KEYBINDINGS.md` | **Unresolved conflict markers committed** (lines 273/276/279) |
| `docs/LAYOUT.md` | **Unresolved conflict markers committed** (lines 24/26/28) |
| `src/app/mod.rs` | `request_document_links` on the sync cadence; `drain_lsp_document_links` fan-out; `open_document_link` with scheme guard + nav recording; link-before-definition dispatch |
| `src/app/tests.rs` | 3 new tests: link-beats-definition, scheme refusal + nav history, stale-link drop |
| `src/lsp/client.rs` | `document_links()` request wrapper |
| `src/lsp/manager.rs` | `DocumentLinkItem` / `DocumentLinksUpdate`, `Cmd::RequestDocumentLinks`, `links_rx` channel, `supports_document_link` capability, worker fetch + normalisation |
| `src/release_notes.rs` | Notes replaced (not appended) |
| `src/widgets/editor.rs` | `doc_links` / `doc_links_path` fields, `apply_document_links`, `document_link_at`, clearing in `mark_buffer_changed` and on tab reset |

### Round-1 fixes: verified holding

- **Scheme guard** (`src/app/mod.rs:7145-7152`) — empirically probed with a throwaway test (since reverted). `FILE:///etc/passwd`, `File:///etc/passwd`, `vscode://`, `javascript:`, `mailto:`, `ssh://` all refuse. `open_url` uses `Command::new("open").arg(url)` — argv, no shell — so an embedded newline passes one argv element and cannot inject a flag. `file://` targets go to `open_pinned`, an in-editor text open, so a `file://` pointing at an executable is *read*, never run.
- **`nav.record` placement** (`src/app/mod.rs:7125-7134`) — correct. `from` is captured *before* `open_pinned` mutates `self.editor`. Recording only on success is a deliberate improvement over the siblings, which record unconditionally; `NavHistory::record` dedups and clears the forward trail, so recording after a *failed* open would have polluted history.
- **`doc_links` clearing** (`src/widgets/editor.rs:2874-2878`) — the finding I most expected to be wrong, and it is fine. It fires on every keystroke, but `sync_lsp` re-fires `request_document_links` on every edit-seq change, so the feature is self-healing.
- **Version + release notes** — main is 0.1.763, HEAD 0.1.764; notes replaced rather than appended. CONTRIBUTING.md satisfied.
- **Merge integrity in `src/`** — `git diff origin/main..HEAD -- src/` contains exactly one deletion line (the intentional release-notes replacement); every other hunk is a pure addition. All seven recently-merged features retain implementation *and* tests. `CONFLICTED_BODY` is byte-identical to main; the marker fixtures elsewhere are untouched. No duplicated definitions.

## Findings

- `docs/KEYBINDINGS.md:273` **[blocker/correctness]** Unresolved git conflict markers (lines 273, 276, 279) were committed at `45cc766`. `git grep '<<<<<<<' origin/main -- docs/` is clean, so the branch introduced them. Worse than cosmetic: the `HEAD` side is the pre-merge text, so resolving toward HEAD silently deletes main's prepare-rename documentation. CI has no conflict-marker gate, so this would ship as-is. Fix: keep main's F2 row and this branch's F12 row.

- `docs/LAYOUT.md:24` **[blocker/correctness]** Same defect, same commit (markers at 24, 26, 28). The `HEAD` side is a stale Editor paragraph that omits expand/shrink selection, LSP folding ranges, format-on-type, prepare-rename, Format Selection, linked editing, and colour presentations — every one of which is documented on the `origin/main` side and has shipped code in the tree. Fix: keep main's paragraph and splice the document-links clause into it.

- `src/app/mod.rs:7153` **[minor/correctness]** `open_document_link` dead-ends a web link in a remote session with "Running remotely: open {target} on your machine", but the sibling `open_detected_url` first checks `drop_relay_active()` and, when a relay is up, forwards the URL to the user's local browser. A user on `croft remote` with the relay active can Ctrl+click an OSC 8 terminal link and have it open locally, but Ctrl+clicking the identical URL in a source comment just prints a message.

## Test coverage

- **The seq gate in `drain_lsp_document_links`** is untested. Every sibling drain has the same gate, but nothing asserts a stale-seq batch is dropped.
- **The split-group fan-out** is untested. `EditorTabs` derefs to the active editor, so `group.apply_document_links` only reaches each group's *active* tab.
- **Target-less link filtering** in the worker normalisation has no test.
- **Multi-line link ranges.** `document_link_at` handles spans crossing rows, but both tests use single-line ranges only.

On the tests that exist: all three test what their names claim. I reverted the `doc_links` clearing and `editing_drops_stale_document_links` failed as expected, then restored. The nav assertion is meaningful — it checks both `file_name()` and `(row, col) == (0, 4)` against a deliberately-set cursor, so it would catch a missing `record` *and* a record taken after `open_pinned` moved the cursor.

## Nitpicks

- `src/app/mod.rs:7145-7148` duplicates the `scheme_ok` block verbatim from the terminal path. Two copies of a security check drift apart.
- `src/app/mod.rs:7142` "Malformed file link" is the status for both an unparseable URI and a `file://` with a non-local host (`file://evilhost/share/x.rs` hits it).
- `src/lsp/manager.rs` — `DocumentLinkItem` derives `PartialEq, Eq` while the sibling `InlayHintItem` does not; the asymmetry is unexplained.

## Scores

- **CodeRabbit score (diff quality) — 1/5.** The Rust is clean and all three round-1 fixes verifiably hold, but two blockers: conflict markers committed into two docs files.
- **Greptile score (codebase fit) — 1/5.** `src/` matches repo conventions exactly and the merge preserved every main feature, but the same two blockers leave the docs tree contradicting itself and would delete main's feature documentation if resolved the wrong way.

VERDICT: REQUEST_CHANGES

---

**Addressed in `ff3b060`:** both blockers (markers removed; each file now carries main's text with the document-links clause grafted in — all of prepare-rename, format-on-type, Format Selection, linked editing, folding ranges and colour presentations verified still documented), and the remote-relay minor (a validated web link now goes through the relay so it reaches the user's own browser).

The nitpick about the duplicated `scheme_ok` block is **deliberately not** collapsed by delegating to `open_detected_url`, and the reason is worth recording: that path probes `crate::file_ref::editor_file_uri` *before* its scheme check, which resolves `vscode://file/<path>` (plus the Cursor/Windsurf/Zed schemes) into a file open. That is correct for a terminal link — text the user's own program printed — but a document link's target is chosen by the language server, so delegating would let a server open arbitrary paths from a click on unrelated-looking text. I tried the delegation and the refusal test caught it: `vscode://file/etc/passwd` opened `/etc/passwd`. The guard stays local, with a comment explaining why, and only already-validated http(s) targets are handed onward.

vitali87:

Round 3, re-run after the round-2 fixes in `ff3b060`. CodeRabbit is still rate-limited on this PR (its check passes as a no-op annotated "Review rate limited"), so this is the only review the change gets.

---

## Walkthrough

PR #278 adds `textDocument/documentLink` support: the LSP manager requests link ranges on the same open/change cadence as inlay hints, the editor stores them path-stamped and clears them on edit, and a Ctrl+click on a linked range calls `open_document_link` instead of Go to Definition. `file://` targets open in the editor (recording nav history); `http(s)` targets go to the system opener, or through the relay on a remote/relay session.

**Verified and closed from round 2** (not re-reported): no conflict markers remain anywhere in the repo — a whole-tree anchored grep returns nothing, and the fixture text inside string constants correctly does not match. Both docs files now differ from `origin/main` by exactly one line each, so nothing main documented (prepare-rename, format-on-type, Format Selection, linked editing, folding ranges, expand/shrink selection, colour presentations) was dropped. Version bump and release-notes *replacement* both satisfy CONTRIBUTING.md. `cargo fmt --check` and `cargo clippy --all-targets` are clean.

**The security reasoning holds, verified independently.** `src/file_ref.rs:105-117` does resolve `vscode`/`vscode-insiders`/`cursor`/`windsurf`/`zed` `://file/<path>` into a `FileRef`, and `open_detected_url` probes it *before* its scheme check — so the wholesale delegation suggested in round 2 would have been exploitable. Confirmed empirically by neutering the guard: the refusal test fails with `vscode://file/etc/passwd must be refused, got "Opened vscode://file/etc/passwd"`. The test earns its name. I also confirmed the *narrowed* delegation is safe: `editor_file_uri` and `diff_uri` both return `None` for `http(s)` inputs, so a pre-validated web target cannot re-enter the file-open probes. `drop_relay_active()` is `relay_dir().is_some()` — genuinely "relay-capable session" — so the added disjunct only widens the remote case and leaves a plain local user on `open_url` exactly as before.

I also swept every other opener caller (`open_url`, `open_detected_url`, `open_resolved_file_ref`). The only LSP-fed one is this new path; the PDF path has its own guard in `classify_href`, and the relay side re-validates via `url_is_safe_to_open`. No second unguarded LSP→opener path exists.

## Findings

- **`src/app/mod.rs:7154`** **[minor/security]** — the guard normalises with `trim_start()` but passes the *raw* `target` downstream, so `"\nhttps://evil.com\tx"` or `"  https://…"` passes `scheme_ok` and reaches `open_url`/`open_detected_url` unchanged. Not currently exploitable — `open_url` uses `Command::new("open").arg(url)` with no shell, and the relay's `url_is_safe_to_open` independently rejects `\0\n\r\t` — so the relay-injection path (a newline/tab-delimited log line) is closed by defence in depth on the far side. But this file relies on a validator in another module it never references. Fix: trim and reject control characters before the `starts_with` check, then forward the same value.

- **`docs/KEYBINDINGS.md:274`** **[minor/docs]** — the resolved line still says web targets are "refused with guidance in a remote session". `ff3b060` changed exactly that: a remote/relay web link now routes through `open_detected_url` and opens on the user's local machine. The commit updated `docs/LAYOUT.md` but left this clause describing pre-`ff3b060` behaviour.

- **`src/app/mod.rs:7121`** **[minor/correctness]** — the `file://` branch matches with case-sensitive `starts_with`, while the web guard below lowercases. A server emitting `FILE:///path` (legal per RFC 3986, and accepted by `Url::parse`) falls through to the web guard and is refused. Confirmed: `open_document_link("FILE:///etc/hosts")` → `Refused to open non-web link`. It fails closed, so correctness rather than security, but it is inconsistent with `file_ref`'s own `scheme.eq_ignore_ascii_case("file")`.

## Test coverage

- **The remote/relay web branch added by `ff3b060` has no test.** The `drop_relay_active() || is_remote_session()` disjunct is the one behaviour change in that commit and nothing exercises it.
- **The refusal test has no case distinguishing trim behaviour from a real bypass** — its `"  HTTP-ish-but-not://x"` case pairs leading whitespace with a bad scheme, so it passes for the wrong reason.
- **`drain_lsp_document_links`'s seq gate and split-group fan-out are untested.**

## Nitpicks

- `src/app/mod.rs:7106` — the doc comment still says remote sessions get a "claim success invisibly" refusal; the inline comment says the right thing but the doc comment above contradicts it.
- `src/app/tests.rs:30021` — the refusal loop calls `open_document_link` with live hostile URIs and depends on the guard to keep `open` from spawning. When I neutered the guard to verify, the test process really did shell out. Harmless today, but worth noting.
- `src/widgets/editor.rs:4692` — `document_link_at`'s two independent `unwrap_or(0)` fallbacks fail closed only because of the surrounding comparison; an explicit bail would read more clearly.
- `src/app/mod.rs:7145` — the eight-line "deliberately NOT delegated" comment is longer than the code it guards, and duplicates most of the commit message. It documents a real trap, so keep it.

## Scores

- **CodeRabbit score (diff quality) — 4/5** — three minors (raw-vs-trimmed target forwarded downstream, stale remote-session prose, case-sensitive `file://` match); no majors or blockers, fmt/clippy clean, and the security guard is correct with a test that provably fails without it.
- **Greptile score (codebase fit) — 4/5** — the drain/request/apply shape mirrors inlay hints and folding ranges exactly, capability probing and seq gating follow the established contract, and the version/release-notes convention is satisfied; the only fit gaps are one stale doc line and the untested new relay branch.

VERDICT: APPROVE

---

**Addressed in `b688915`:** all three minors. The target is now trimmed and control-character-refused once at the top, so the checked string is the forwarded string; `file://` matches case-insensitively like `file_ref` does; and the KEYBINDINGS clause describes the relay behaviour. Tests gained upper-case and Cursor-scheme refusals, control-character refusals, and coverage that `FILE://` and a whitespace-padded target both still open.

The relay-branch test gap is left open deliberately: exercising it needs `CROFT_RELAY_KEY` plus the relay cache-dir harness, and the branch is a two-line delegation to an already-tested function. Noted here rather than papered over.

PR #279, opened 2026-08-24
fix(spawn): keep child stderr off the tty during a live session

vitali87:
## Problem

Raw text painted over the TUI frame (torn macOS libmalloc diagnostics across the screen): a child process inherited croft's tty and wrote straight to it, bypassing the renderer.

## Audit

Swept every `Command` spawn site for tty inheritance:

- DAP adapters/servers: piped or nulled **plus** `setsid` — clean.
- LSP, git, pdf/sips, clipboard, voice helpers: explicitly piped/nulled — clean.
- `src/media.rs` ffmpeg poster grab: **inherited stdout and stderr**, and runs mid-session on first open of a video. Fixed: both nulled (`-loglevel quiet` already discards the useful output; failure is handled via exit status).
- `src/remote.rs` drop-relay ssh setup: **explicit `Stdio::inherit()` on stderr**. Fixed: captured and folded into the `bail!` message — better diagnostics and no tty writes.
- `sync_via_tar` (remote install): stderr inheritance kept **deliberately** — it runs in the pre-TUI CLI install flow where tar/ssh errors must reach the user's terminal.

## Notes

The message in the screenshot that prompted this ("can't turn on malloc stack logging…") most likely came from a memory tool/debugger attach outside croft; these fixes are defense-in-depth so no croft child can ever scribble on the frame.


vitali87:

---

## Walkthrough

Reviewed at head **`5c518c37f1bfb2c7196d4c0a5d07528b4ebccb8e`** (single commit, base `main`, same-repo branch).

The change closes two spawn sites that let a child process write onto croft's controlling tty and paint raw bytes over the ratatui frame. `src/media.rs`'s ffmpeg poster grab inherited both stdout and stderr; both are now `Stdio::null()`. `src/remote.rs`'s drop-relay ssh setup carried an explicit `.stderr(Stdio::inherit())`, now `Stdio::piped()` with the captured text folded into the `bail!` message. I verified the audit claim in the PR body by sweeping every `Command::new` site in `src/`: after this diff there are zero `Stdio::inherit` occurrences left, and every remaining implicit-inherit spawn is either `#[cfg(test)]`, a cooked-mode CLI flow, or already nulled/`setsid`'d.

| file | summary |
|---|---|
| `src/media.rs` | `poster_frame`'s ffmpeg child gets `.stdout(null())` and `.stderr(null())` plus a 2-line rationale comment. Behavior-preserving: `-loglevel quiet` already suppressed output, success is judged by exit status + `tmp.is_file()`. |
| `src/remote.rs` | `DropPump::start`'s setup ssh switches `stderr` from `inherit` to `piped`; the failure path now interpolates `String::from_utf8_lossy(&output.stderr).trim()` into the `bail!`. |

Verified in the worktree: `cargo fmt --check` clean, `cargo check --all-targets` clean.

## Findings

- **`Cargo.toml:3` [major/docs]** — the shipped change bumps no version and leaves stale release notes. The repo's documented rule lives in `src/release_notes.rs:8-16`: *"Update `RELEASE_NOTES` on every version bump"* and *"REPLACE this list on every bump; do not append… Each patch bump is one change."* `git log --oneline origin/main -- Cargo.toml` shows this is enforced without exception — every shipped commit carries a `(vX.Y.Z)` suffix or is a `chore: bump` (`69a76aa fix(theme): … (v0.1.758)`, `73617f7 feat(git): … (v0.1.757)`, and so on unbroken). This PR is a user-visible fix, not an exempt chore/docs change, so it needs both. Compounding it: the branch sits at `version = "0.1.758"` while `origin/main` is already at `0.1.759`, and the branch's `src/release_notes.rs:64` still describes the v0.1.758 activity-bar-icon fix, whereas main's describes the "Close All" fix. Fix: rebase onto `origin/main`, set `version = "0.1.760"`, and replace the single `RELEASE_NOTES` entry with a `NoteKind::Fix` describing this one (e.g. *"Video previews and remote drag-drop setup no longer let a child process print over the editor frame."*).

- **`src/remote.rs:879-895` [minor/correctness]** — the new comment's stated rationale does not match where this code runs, and the change loses output it claims to preserve. `DropPump::start` is called only from `run_croft_session` (`src/remote.rs:410`), which is the pre-TUI cooked-mode CLI flow — `src/remote.rs:446-448` documents it as *"Must be called only after the terminal has been returned to cooked mode and the alt-screen surrendered"*, and the immediate caller at `src/remote.rs:413` handles the error with a plain `eprintln!("Drag-drop relay disabled: {e}")`, alongside bare `println!`s at lines 426 and 686. So there is no TUI frame here to scribble on; this site is the same category as `sync_via_tar`, which the PR body exempts by exactly that reasoning. The change is still a net improvement (folding stderr into the error beats a bare exit code), but the comment justifies it with a hazard that does not apply, and piping now *discards* ssh stderr on the **success** path — previously an ssh warning (host-key notice, `Permission denied` on a non-fatal step) reached the user. Fix: retitle the comment to the real reason and drop the tty claim:
  ```diff
  -        // Capture stderr rather than inheriting the tty: the caller folds a
  -        // failure into its own message, and an inherited fd would let ssh
  -        // scribble raw bytes over whatever the terminal is showing.
  +        // Capture stderr so a setup failure carries ssh's own diagnosis in
  +        // the bail! message rather than a bare exit code. (This runs in the
  +        // pre-TUI cooked-mode flow, so inheriting would be safe — it just
  +        // produced worse errors.)
  ```

- **`src/remote.rs:890-895` [minor/correctness]** — an empty or whitespace-only stderr yields a message ending in a dangling `": "`. `output.status` on a signal-killed ssh renders as `signal: 9 (SIGKILL)` with no stderr at all, and the remote script is `set -e` with `mkdir -p`/`printf`, which can fail silently on a full disk. Guard the suffix:
  ```rust
  let err = String::from_utf8_lossy(&output.stderr);
  let err = err.trim();
  if err.is_empty() {
      anyhow::bail!("remote relay setup failed with {}", output.status);
  }
  anyhow::bail!("remote relay setup failed with {}: {err}", output.status);
  ```

## Test coverage

- **The ffmpeg child's fd wiring (`src/media.rs:392-397`)** is untested. `src/media.rs`'s test module (`427-555`) covers only the pure header parsers; `poster_frame` has no test at all, and it is a private fn with a hard `ffmpeg` dependency. A cheap regression guard that does not need ffmpeg: factor the `Command` construction into a `fn poster_cmd(path: &Path, tmp: &Path) -> Command` (mirroring the existing `probe_cmd`/`run_probe` split in `src/gui_path.rs:83-86`, which was carved out for exactly this reason) and assert in a test that the returned command's configured stdio are all null. Without that split, nothing stops a future edit from dropping the `.stderr(null())` again — which is precisely the bug this PR fixes.
- **`DropPump::start`'s new error-formatting path (`src/remote.rs:889-896`)** is untested, including the empty-stderr case in the third finding. The message assembly is the testable part: extract `fn relay_setup_error(status: ExitStatus, stderr: &[u8]) -> String` and unit-test it with (a) non-empty stderr producing `"…failed with exit status: 1: <msg>"` and (b) empty stderr producing no trailing `": "`.

## Nitpicks

- `src/media.rs:392-397` — the file uses fully-qualified `std::process::Stdio::null()` three times in one builder chain. The rest of the repo (`src/remote.rs`, `src/gui_path.rs:123-125`) imports `Stdio` and writes `Stdio::null()`. A `use std::process::Stdio;` at the top of `poster_frame` would match house style and shorten the chain.
- The commit message and PR body describe the remote.rs change as preventing tty scribbling, which the second finding shows is not the operative reason there. Worth correcting so the commit log stays accurate.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: no blockers or majors inside the changed lines; the two minors (a rationale comment that misstates why the code is correct, and the dangling-`": "` empty-stderr case) are real but non-breaking, and the diff compiles, formats clean, and preserves behavior.
- **Greptile score (codebase fit) — 3/5**: exactly one major — the violated version-bump + release-notes rule documented in `src/release_notes.rs:8-16` and enforced across every commit in `git log`, made worse by the branch trailing `origin/main` by 13 commits with a stale `0.1.758` version and last release's notes.

VERDICT: REQUEST_CHANGES

vitali87:

CodeRabbit's check on this PR reports `pass` but is annotated "Review rate limited", so no bot review ran. This is round 2, re-run after the round-1 findings were fixed in `b039ada`.

# Review: PR #279 — fix(spawn): keep child stderr off the tty during a live session

## Walkthrough

Two unrelated child-process stdio bugs, fixed together. The ffmpeg that extracts a video poster frame runs mid-session while the TUI owns the alt-screen (`open_media_preview` → `to_markdown` → `poster_frame`), and it inherited croft's stdout/stderr — so anything ffmpeg printed painted raw bytes over the frame; it is now `Stdio::null()` on both. Separately, the relay-setup ssh in `DropPump::start` moved from `Stdio::inherit()` to `Stdio::piped()` so its stderr is folded into the `bail!` message instead of being lost as a bare exit code. Round-2 fixes address the prior review: version bumped to 0.1.760 with matching release notes, main merged in, the misleading tty rationale in the remote.rs comment corrected, the dangling `": "` handled by an extracted `relay_setup_error` helper, the `std::process::` qualification collapsed to a local `use`, and a unit test added for the helper.

| file | summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | version 0.1.759 → 0.1.760 |
| `src/media.rs` | `use std::process::{Command, Stdio}` in `poster_frame`; ffmpeg's stdout and stderr now `Stdio::null()` with a rationale comment |
| `src/remote.rs` | new `relay_setup_error(ExitStatus, &[u8]) -> String` helper + doc comment; `DropPump::start` pipes stderr instead of inheriting and formats the failure through the helper; new `#[cfg(unix)]` unit test |
| `src/release_notes.rs` | `RELEASE_NOTES` replaced (not appended) with this release's single Fix entry |

Verified in the working tree: `cargo fmt --check` clean, `cargo clippy --bins --all-targets` produces no warnings or errors, and `cargo test --bins relay_setup_error` passes (1 passed).

## Findings

None. I attempted to break the change in five specific ways and each held up:

- **The remote.rs comment's corrected rationale is actually true.** `DropPump::start` is called only from `run_croft_session` (`src/remote.rs:410`), whose only two callers are `launch_only` (`src/remote.rs:449`) and `launch_croft_with` (`src/remote.rs:694`). Both are reached from `src/app/mod.rs:39117` after the TUI has been torn down — `restore_host_terminal_state()` is called on the *return* path at `src/app/mod.rs:39120`, and the surrounding code uses bare `println!`/`eprintln!` freely (`src/remote.rs:672`, `src/remote.rs:415`). The claim "this runs in the pre-TUI cooked-mode flow, so inheriting would be safe — it just produced worse errors" is accurate, including for the reconnect iteration of the loop, which is only reached after `run_remote_croft` has returned.
- **The dangling-`": "` fix is complete.** `relay_setup_error` (`src/remote.rs:865-874`) trims before the emptiness test, so whitespace-only stderr (`b" \n"`) takes the no-suffix branch, not just `b""`.
- **The extracted-helper test exists and is sound.** `src/remote.rs:2725-2740`. I independently confirmed `ExitStatus::from_raw(256)` yields `exit status: 1` / `success() == false` on this platform, so the fixture models a real ssh failure rather than a success. The three assertions cover the trim-and-append case, the whitespace-only case, and the empty case, which is exactly the surface the helper adds. `#[cfg(unix)]` matches the file's own convention (`src/remote.rs:1771`) and the repo's (`src/archive.rs:439`, `src/git.rs:140`).
- **Nulling ffmpeg's stderr drops no diagnostic that was ever surfaced.** `poster_frame` returns `Option`, and the sole call site (`src/media.rs:345-346`) discards `None` silently — no error path existed to degrade. `-loglevel quiet` was already in the argv, so the redirect is belt-and-braces for the cases ffmpeg ignores its own loglevel (usage errors, libav warnings).
- **Nulling ffmpeg's stdout is safe.** The output target is a path argument (`.arg(&tmp)`, `src/media.rs:392`), not `-` — nothing is read from the pipe.
- **CONTRIBUTING gates satisfied.** `src/` and `Cargo.toml`/`Cargo.lock` are shipped paths, and both required actions are present: 0.1.760 > main's 0.1.759 (verified via `git show origin/main:Cargo.toml`), and `src/release_notes.rs` is *replaced*, not appended, matching the module's "REPLACE this list on every bump; do not append" rule at `src/release_notes.rs:13-16`. The bumping commit `b039ada` carries the `(v0.1.760)` subject suffix. The CI job at `.github/workflows/ci.yml:34-72` will pass on both checks.
- **No sibling of this bug was missed in a mid-session path.** `Stdio::inherit` no longer appears anywhere under `src/`. I swept every `Command::new` site: the `git.rs` `.status()` calls that leave stderr inherited are all inside `mod tests` (which begins at `src/git.rs:1782`), and `src/launcher.rs:131-223` is reached only from the `croft --install-launcher` CLI handler (`src/cli.rs:821`, `src/cli.rs:845`), never with the TUI up.

## Test coverage

- **The ffmpeg stdio handles themselves** are untested, and — as previously established and not re-raised here — `std::process::Command` exposes no getters for its configured stdio, so this is not implementable in std. Nothing to write.
- **`DropPump::start`'s use of the helper** is untested: the test covers `relay_setup_error` in isolation, but not that `start` actually reaches it with `output.stderr` (a future refactor could pass `output.stdout` and the test would still pass). This needs a live ssh host, so it is not reasonably unit-testable, and the helper test is the right level to stop at.
- **`poster_frame` failure behavior** has no test, but it also has none on main and the diff does not change it (still `None` on any failure). No new gap.

The one gap the diff opens and closes itself. I would not block on additional coverage.

## Nitpicks

- `src/remote.rs:864` — the helper's doc comment is four lines for a six-line function. It is justified here (it records *why* the empty-stderr branch exists, which is the thing a future edit would delete), and it matches this file's unusually high comment density (`src/remote.rs:376-407` runs 30 lines of rationale for one `let`). Noting it only so it is not mistaken for an oversight.
- `src/media.rs:395-396` — the repo's own established sibling pattern (`src/pdf.rs:69`, `src/pdf.rs:369`) *captures* stderr via `.output()` and surfaces it in the error string rather than discarding it. Discarding is correct here because `poster_frame` has no error channel to surface it on, but if that ever changes, prefer the pdf.rs shape over `Stdio::null()`.

### Round-1 finding not carried forward

The round-1 review's test-coverage suggestion to assert the ffmpeg poster command's stdio handles are null was **dismissed, not fixed**: `std::process::Command` exposes no public getters for its configured stdio, so there is nothing a std-only unit test can read back. The stdio change is covered by review, not by assertion.

CodeRabbit score (diff quality): 5/5
Greptile score (codebase fit): 5/5

VERDICT: APPROVE

PR #283, opened 2026-08-24
fix(editor): make Close All close every split group

vitali87:
## Problem
With files open side by side (e.g. from `cgr duplicates`), right-clicking a tab and picking **Close All** only closed the tabs in the clicked group. The now-blank group then collapsed away and promoted the surviving group, so it looked like a single tab closed. The `⌘K W` chord had the same bug.

## Fix
Extract a shared `App::close_all_tabs` used by both the tab context menu and the `⌘K W` chord. It closes the tabs in **every** editor group, counts them all for the status line, and collapses any split back to a single blank pane (clearing seams, evicting the right image slot, and re-syncing focus flags — same cleanup as `collapse_split_if_empty`).

## Test
`close_all_closes_every_split_group`: opens a file, splits, opens a second file in the right group, dispatches `MenuAction::CloseAllTabs`, and asserts the split collapses, a single blank pane remains, and the status counts tabs from both groups.

Note: `app::tests::repositories_overview_lists_every_folder_and_a_pin_overrides_the_follow` fails on macOS on clean `main` too (/var vs /private/var canonicalization) — unrelated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.758"
+version = "0.1.759"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.758"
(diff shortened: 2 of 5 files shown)


vitali87:
@coderabbitai full review

PR #285, opened 2026-08-24
ci: gate PRs on a version bump and fresh release notes

vitali87:
Closes #281.

## Problem
Since v0.1.758, PRs #271, #273, and #274 merged shipped-code changes without bumping the version or touching `src/release_notes.rs`. So `main` builds share a version and differ only in commit hash, and the welcome panel's **IN THIS RELEASE** card still describes the previous release.

## Gate
New first CI job (`version bump + release notes`, PRs only). When the diff against the base touches anything compiled into the binary (`src/`, `assets/`, `build.rs`, `Cargo.toml`, `Cargo.lock` — excluding `src/app/tests.rs` and `tests/`), it fails unless:

1. `version` in `Cargo.toml` strictly increments vs the base, and
2. `src/release_notes.rs` changed.

Docs / CI / test-only PRs are exempt (the job prints why and passes).

Also documented the rule in CONTRIBUTING.md.

## Verified
Dry-ran the script against real history:
- `v0.1.758..main` (the violation): fails both checks ✔
- `v0.1.757..v0.1.758` (a proper release commit): passes ✔
- this PR itself touches no shipped files: exempt ✔

PR #287, opened 2026-08-24
ci: add SonarQube Cloud analysis (free plan)

### LINKED ISSUE #284: Add SonarqubeCloud CI free plan

vitali87:
Closes #284.

## What

- **`.github/workflows/sonarcloud.yml`** — runs on every PR and push to `main`: installs the pinned toolchain, generates a Clippy JSON report (`cargo clippy --all-targets --message-format=json`), and uploads it via `SonarSource/sonarqube-scan-action@v8`.
- **`sonar-project.properties`** — project binding (`vitali87` / `vitali87_croft`) plus `sonar.rust.clippy.enabled=false` so the scanner imports our pinned-toolchain report instead of running its own unpinned Clippy (which would be slower and duplicate every issue).

No `-D warnings` in the Sonar clippy run — the existing `clippy + tests` job stays the gate; this run only feeds the report.

Every step is gated on `SONAR_TOKEN`, so the job is a **green no-op** on fork PRs (secrets are withheld) and until the one-time setup below is done — CI never breaks on a missing token.

## One-time setup (needs the repo owner)

1. Sign in at [sonarcloud.io](https://sonarcloud.io) with GitHub (free plan covers public repos) and import `vitali87/croft`. If the generated keys differ from `vitali87` / `vitali87_croft`, fix `sonar-project.properties` to match.
2. In the SonarCloud project: **Administration → Analysis Method → turn Automatic Analysis OFF** (CI-based analysis conflicts with it).
3. Create a token (My Account → Security) and add it as the `SONAR_TOKEN` repository secret in GitHub.

Until step 3, the job logs a notice and passes.

## Scope

CI-only (`.github/` + `sonar-project.properties`): nothing ships, so no version bump / release-notes change per the #285 gate rules. Coverage import (`cargo llvm-cov` → lcov) is possible later, but croft's suite spawns real PTYs/zsh/poppler, so that's a deliberate follow-up rather than a freebie here.

## Summary by CodeRabbit

* **Chores**
  * Added automated SonarQube Cloud code-quality scanning for changes to the main branch and pull requests.
  * Configured Rust analysis, test paths, fixture exclusions, and Clippy report integration.
  * Added safeguards to avoid scan failures when credentials are unavailable.
  * Excluded generated analysis artifacts from version control.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -43,13 +43,15 @@ jobs:
         with:
           key: sonarcloud
       - name: Generate the Clippy report
+        if: env.SONAR_TOKEN != ''
         # No `-D warnings` here: the `clippy + tests` job in ci.yml is the
         # gate. This run only produces the JSON report SonarQube imports, so
         # warnings must land in the report, not kill the job.
-        run: cargo clippy --all-targets --message-format=json > clippy-report.json
-        if: env.SONAR_TOKEN != ''
+        #
+        # `|| true`: a compile error is already reported by ci.yml's gate job;
+        # failing here too would only add a second red X and skip the import
+        # of whatever diagnostics Clippy did emit before dying.
+        run: cargo clippy --all-targets --message-format=json > clippy-report.json || true
       - name: SonarQube Cloud scan
         if: env.SONAR_TOKEN != ''
         uses: SonarSource/sonarqube-scan-action@v8
-        env:
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,0 +1,55 @@
+name: SonarQube Cloud
+
+# Static analysis on sonarcloud.io (free plan for public repos, #284).
+# Project binding lives in sonar-project.properties; the analysis is CI-based
+# (Rust needs a real `cargo clippy` run, which SonarQube Cloud's automatic
+# analysis cannot do), so Automatic Analysis must stay OFF in the SonarCloud
+# project settings or every scan fails with a conflict.
+#
+# Every step is gated on SONAR_TOKEN so the job is a green no-op on fork PRs
+# (secrets are withheld there) and on checkouts where the secret has not been
+# configured yet - analysis coverage degrades, CI does not break.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  sonarcloud:
+    name: SonarQube Cloud analysis
+    runs-on: ubuntu-latest
+    env:
+      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+    steps:
+      - name: Skip without a token
+        if: env.SONAR_TOKEN == ''
+        run: echo "::notice::SONAR_TOKEN is not available (fork PR, or the secret is not configured); skipping SonarQube Cloud analysis."
+      - uses: actions/checkout@v4
+        if: env.SONAR_TOKEN != ''
+        with:
+          # Full history: SonarQube Cloud needs blame data to attribute issues
+          # and to tell new code from old on PRs.
+          fetch-depth: 0
+      # rust-toolchain.toml is the single source of truth; `rustup show`
+      # installs exactly the pinned channel and its components (clippy
+      # included). The scanner's own Clippy execution is disabled in
+      # sonar-project.properties precisely so this pinned run is the only one.
+      - name: Install the pinned toolchain
+        if: env.SONAR_TOKEN != ''
+        run: rustup show
+      - uses: Swatinem/rust-cache@v2
+        if: env.SONAR_TOKEN != ''
+        with:
+          key: sonarcloud
+      - name: Generate the Clippy report
+        # No `-D warnings` here: the `clippy + tests` job in ci.yml is the
+        # gate. This run only produces the JSON report SonarQube imports, so
+        # warnings must land in the report, not kill the job.
+        run: cargo clippy --all-targets --message-format=json > clippy-report.json
+        if: env.SONAR_TOKEN != ''
+      - name: SonarQube Cloud scan
+        if: env.SONAR_TOKEN != ''
+        uses: SonarSource/sonarqube-scan-action@v8
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,7 @@ preview*.svg
 # Local secrets (Codeberg token, etc.) — never commit
 .env
 .env.*
+
+# SonarQube Cloud scan artifacts (see .github/workflows/sonarcloud.yml)
+/clippy-report.json
+/.scannerwork/
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -6,7 +6,14 @@ sonar.organization=vitali87
 sonar.projectName=croft
 
 sonar.sources=src,build.rs
-sonar.tests=tests
+# src/app/tests.rs (29k lines) and tests/ are compiled out of release builds
+# (see ci.yml's release-hygiene filter): they are tests, not shipped source.
+# Left under sonar.sources they would dominate every duplication/complexity
+# metric and fail quality gates on test-table churn. Inline #[cfg(test)]
+# modules cannot be split out by path and stay in sources - accepted.
+sonar.tests=tests,src/app/tests.rs
+# Fixture data, not source.
+sonar.exclusions=src/mcp/testdata/**
 sonar.sourceEncoding=UTF-8
 
 # CI generates the Clippy report itself with the PINNED toolchain (see
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,17 @@
+# SonarQube Cloud project binding (free plan, public repo). The keys must
+# match what sonarcloud.io generated when the repository was imported:
+# organization = the GitHub login, projectKey = <org>_<repo>.
+sonar.projectKey=vitali87_croft
+sonar.organization=vitali87
+sonar.projectName=croft
+
+sonar.sources=src,build.rs
+sonar.tests=tests
+sonar.sourceEncoding=UTF-8
+
+# CI generates the Clippy report itself with the PINNED toolchain (see
+# .github/workflows/sonarcloud.yml) and hands it over here. Left enabled, the
+# scanner would run `cargo clippy` a second time with whatever toolchain it
+# finds - slower, unpinned, and every issue would show up twice.
+sonar.rust.clippy.enabled=false
+sonar.rust.clippy.reportPaths=clippy-report.json
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -12,8 +12,10 @@ sonar.sources=src,build.rs
 # metric and fail quality gates on test-table churn. Inline #[cfg(test)]
 # modules cannot be split out by path and stay in sources - accepted.
 sonar.tests=tests,src/app/tests.rs
-# Fixture data, not source.
-sonar.exclusions=src/mcp/testdata/**
+# sonar.sources=src would otherwise ALSO claim src/app/tests.rs, and a file
+# in both scopes fails the scan outright - it has to be subtracted here.
+# Fixture data is not source either.
+sonar.exclusions=src/app/tests.rs,src/mcp/testdata/**
 sonar.sourceEncoding=UTF-8
 
 # CI generates the Clippy report itself with the PINNED toolchain (see


vitali87:
@coderabbitai full review

vitali87:

CodeRabbit is rate-limited on this PR (its check passed as a no-op, and a `@coderabbitai full review` retrigger bounced), so this review was produced locally by a Claude Opus 5 agent with full-repository context. Findings are being addressed in follow-up commits on this branch.

---

## Walkthrough

Adds SonarQube Cloud static analysis as a new, independent workflow: on every PR and push to `main`, a single job installs the pinned toolchain, emits a Clippy JSON report, and hands it to `SonarSource/sonarqube-scan-action@v8` for import. `sonar-project.properties` carries the project binding and disables the scanner's own Clippy run so the pinned report is the only issue source. Every step is gated on `SONAR_TOKEN` so the job is a green no-op on fork PRs and until the owner finishes the one-time SonarCloud setup — confirmed against run `32786245961`, where "Skip without a token" succeeded and all five real steps reported `skipped`.

| File | Summary |
| --- | --- |
| `.github/workflows/sonarcloud.yml` | New 55-line workflow. `push: [main]` + `pull_request`. Job-level `env.SONAR_TOKEN`, per-step `if: env.SONAR_TOKEN != ''` gate, `checkout` with `fetch-depth: 0`, `rustup show`, `Swatinem/rust-cache@v2` (`key: sonarcloud`), `cargo clippy --all-targets --message-format=json > clippy-report.json`, then the scan action. |
| `sonar-project.properties` | New 17-line binding: `projectKey=vitali87_croft`, `organization=vitali87`, `sonar.sources=src,build.rs`, `sonar.tests=tests`, `sonar.rust.clippy.enabled=false`, `sonar.rust.clippy.reportPaths=clippy-report.json`. |

Conventions check: the workflow correctly mirrors `ci.yml`'s house style — `rustup show` against `rust-toolchain.toml` as single source of truth, `Swatinem/rust-cache@v2` with a distinct `key`, and the same "explain *why*, not *what*" comment density. The `release-hygiene` gate is satisfied by exemption, not compliance: its `grep -E '^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)'` filter (`ci.yml:56-58`) matches neither `.github/workflows/sonarcloud.yml` nor `sonar-project.properties`, so no version bump or `src/release_notes.rs` replacement is required — matching CONTRIBUTING.md's "Docs, CI, and test-only PRs are exempt". Correctly claimed in the PR body. Action version and both Sonar property names verified against upstream: `v8.2.1` is the current `v8` tag, and SonarQube Cloud's Rust docs name `sonar.rust.clippy.enabled` (default enabled, set `false` to disable) and `sonar.rust.clippy.reportPaths` exactly as written.

## Findings

- **`sonar-project.properties:9`** **[major/correctness]** — `sonar.tests=tests` misclassifies essentially the entire test suite as production source, which will systematically corrupt every metric this workflow exists to produce. `tests/` contains exactly one file, `tests/cli.rs`, at 6,220 bytes. Meanwhile `src/app/tests.rs` is **29,520 lines** of pure test code, and `grep -rln '#\[cfg(test)\]' src/` matches **134 files** out of 151 `.rs` files in `src/`. Because `sonar.sources=src`, all of that lands in the "production code" bucket: duplication (test tables and fixtures are repetitive by nature), cognitive complexity, and lines-of-code will all be badly skewed, and Sonar's Clean-as-You-Code quality gate will start failing PRs on test-file findings. This is not hypothetical — the repo's own `ci.yml:53-54` draws exactly this production/test line in its comment: *"src/app/tests.rs and tests/ are compiled out of release builds."* Mirror that boundary:

  ```diff
  -sonar.sources=src,build.rs
  -sonar.tests=tests
  +sonar.sources=src,build.rs
  +sonar.tests=tests,src/app/tests.rs
  +# src/app/tests.rs and tests/ are compiled out of release builds
  +# (see .github/workflows/ci.yml) - they are tests, not shipped source.
  +sonar.exclusions=src/mcp/testdata/**
  ```

  The 134 inline `#[cfg(test)]` modules cannot be split out by path and are an accepted limitation, but the 29.5k-line file is the bulk of the damage and is one path entry away from being correct.

- **`.github/workflows/sonarcloud.yml:49`** **[major/correctness]** — `cargo clippy --all-targets --message-format=json > clippy-report.json` swallows the exit status of a genuinely broken build. The step has no `-D warnings` (correct, and well justified in the comment at lines 46-48), but a redirect does not suppress a *failure* — if `cargo clippy` dies on a compile error, the step fails and the scan never runs, so a red X appears on this workflow for a reason that `ci.yml`'s `clippy + tests` job already reports. That is tolerable. The real defect is the inverse: `--message-format=json` sends diagnostics to **stdout** and cargo's progress/error chatter to **stderr**, so on a partial failure `clippy-report.json` is left holding a truncated JSON stream. Worse, this step runs *before* the scan and there is no `if: always()`, so the two failure modes are inconsistent. Make the report generation non-fatal (`|| true`) and explicit about why, so the scan still imports whatever Clippy managed to emit.

- **`.github/workflows/sonarcloud.yml:16`** **[minor/correctness]** — the bare `pull_request` trigger means the Sonar scan runs against the **merge commit** GitHub synthesizes for the PR ref, not the PR head. Combined with `fetch-depth: 0` this mostly works, but the scan action needs `sonar.pullrequest.*` parameters to decorate the PR rather than overwrite the `main` branch analysis. The action auto-detects these from the `GITHUB_EVENT_*` environment, so this is likely fine in practice — but nothing in this workflow or the properties file pins the branch/PR distinction, and if detection misfires, PR analyses will clobber the main-branch quality gate baseline. Worth an explicit verification on the first real (token-present) PR run, since no run to date has exercised the non-skip path. Not blocking; flagging so it is checked rather than assumed.

- **`.gitignore`** **[minor/style]** — no entry for `clippy-report.json` or `.scannerwork/`, both produced in the repo root (the former by the workflow's clippy step, the latter by the scanner CLI). On CI this is harmless (fresh runner), but anyone reproducing the scan locally gets two untracked entries in `git status` and can commit them by accident. Add both to `.gitignore`.

## Test coverage

Nothing in this diff is testable by the Rust suite — it is CI configuration, and `tests/cli.rs` covers CLI behavior, not workflow files. The meaningful coverage gap is operational rather than unit-level: **the token-present path has never executed.** Run `32786245961` proves only the skip path works (all five substantive steps `skipped`). Everything downstream of the gate — whether `rustup show` resolves 1.97.1 on the runner, whether the scanner finds `Cargo.toml` at the root as its Rust analyzer requires, whether `sonar.rust.clippy.reportPaths=clippy-report.json` resolves relative to `projectBaseDir`, and whether the import succeeds at all — is unverified by construction.

The check to run: after adding `SONAR_TOKEN`, open a throwaway PR touching one `src/` file and confirm the scan step (a) uploads without an "Automatic Analysis is enabled" conflict error, and (b) reports a non-zero imported-issue count in the step log rather than silently importing zero. A zero-issue import is the expected failure signature if the report path or format is wrong, and it looks identical to a clean scan — so it must be confirmed positively on the first run, not inferred from a green check.

## Nitpicks

- **`.github/workflows/sonarcloud.yml:45-50`** — the `if:` key sits *after* `run:` on this step, while every other gated step in the file puts `if:` immediately after `name:`. Cosmetic, but it is the one step that breaks the file's own ordering.
- **`.github/workflows/sonarcloud.yml:22-23, 54-55`** — `SONAR_TOKEN` is declared twice: once at job level (needed for the `if:` gates) and again on the scan step. The step-level `env` is redundant since the job-level value is already in scope.
- **`.github/workflows/sonarcloud.yml`** — `ci.yml` sets `env: CARGO_TERM_COLOR: always` and `RUST_BACKTRACE: 1` workflow-wide; this workflow sets neither. Irrelevant to a JSON-format run, noted only for consistency.
- **`sonar-project.properties:8`** — `sonar.sources=src,build.rs` includes `src/mcp/testdata/` (`index.json`, `index.json.sig`), which is fixture data rather than source. Folded into the `sonar.exclusions` suggestion in the first finding.

## Scores

- **CodeRabbit score (diff quality) — 3/5**: exactly one major whose evidence is internal to the new lines (`sonarcloud.yml:49`, the unguarded redirect leaving a truncated report on a failed Clippy run), plus one minor and the style nits.
- **Greptile score (codebase fit) — 3/5**: exactly one major whose evidence lives outside the diff (`sonar-project.properties:9` contradicting the production/test boundary that `ci.yml:53-54` and CONTRIBUTING.md both draw); the release-hygiene exemption, action version, and Sonar property names all check out against their external sources.

**VERDICT: REQUEST_CHANGES**

vitali87:

Re-review of fix commit 045740a against the [first-round review](https://github.com/vitali87/croft/pull/287#issuecomment-5402725041). CodeRabbit remains rate-limited on this PR.

---

## Re-review

Fix commit `045740a` addresses all four items I raised. Three land cleanly; the fix for the first one is correct in intent but introduces a scanner-fatal configuration error, because moving a file into `sonar.tests` does not remove it from `sonar.sources`.

| Prior finding | Status | Evidence |
| --- | --- | --- |
| major `sonar-project.properties:9` — test misclassification | **Unresolved (regressed)** | `sonar.tests=tests,src/app/tests.rs` (line 14) is the right intent, but `sonar.sources=src` (line 8) still claims that same file. Sonar's docs: *"If a file is in both `sonar.sources` and `sonar.tests`, the scanner fails."* See new finding below. |
| major `sonarcloud.yml` clippy step — truncated report / fatal step | **Resolved** | Line 54 is now `... > clippy-report.json \|\| true`, with a three-line comment at 50-53 explaining that `ci.yml`'s gate job already reports compile errors. |
| minor `.gitignore` | **Resolved** | Lines appended at the end of `.gitignore`: `/clippy-report.json` and `/.scannerwork/`, under a comment pointing back at the workflow. |
| nit — `if:` ordering on the clippy step | **Resolved** | `if: env.SONAR_TOKEN != ''` now sits at line 46, immediately after `name:`, matching every other gated step (26, 29, 39, 42, 56). |
| nit — redundant step-level `SONAR_TOKEN` | **Resolved** | The trailing `env:` block is gone; the file now ends at line 57 with the `uses:` line, relying on the job-level `env` at 22-23. |
| minor — `pull_request` decoration | **Dismissed, accepted** | Not verifiable before the secret exists; correctly deferred to the first token-present run. |

I also re-read all three files in full at the new head rather than only the fix hunks. Nothing else changed, and `release-hygiene` now reports SUCCESS on this PR — confirming the CI-only exemption holds for the added `.gitignore` file too (it matches none of the `^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)` shipped-file patterns).

## Findings

- **`sonar-project.properties:8,14`** **[major/correctness]** — `sonar.sources=src` and `sonar.tests=tests,src/app/tests.rs` both claim `src/app/tests.rs`, and Sonar treats that as a fatal configuration error, not a precedence rule. The SonarQube Cloud analysis-parameters docs are explicit: *"Sonar must classify each code file as either source or test... A file cannot be both categories — if it is, the scanner fails the analysis with an error,"* and the community thread for this exact failure surfaces it as *"We detected a configuration error for your analysis, the source and tests files overlap."* Note this is strictly worse than the original state: before, the file was merely miscategorised and the scan ran; now the scan aborts. It is also invisible today — the "SonarQube Cloud analysis" check is SUCCESS on `045740a` only because the token is still absent and every real step is skipped, so the first token-present run is where this fires. The second half of the fix has the same shape: `sonar.exclusions` applies to the source scope only (`sonar.test.exclusions` is the test-scope counterpart), so `src/mcp/testdata/**` is dropped from sources but, being under no `sonar.tests` path, that part is fine as written. The fix is to subtract the test file from sources rather than only adding it to tests:

  ```diff
   sonar.sources=src,build.rs
   # src/app/tests.rs (29k lines) and tests/ are compiled out of release builds
   # (see ci.yml's release-hygiene filter): they are tests, not shipped source.
   # Left under sonar.sources they would dominate every duplication/complexity
   # metric and fail quality gates on test-table churn. Inline #[cfg(test)]
   # modules cannot be split out by path and stay in sources - accepted.
   sonar.tests=tests,src/app/tests.rs
  -# Fixture data, not source.
  -sonar.exclusions=src/mcp/testdata/**
  +# sonar.sources=src would otherwise ALSO claim src/app/tests.rs, and a file
  +# in both scopes fails the scan outright - it has to be subtracted here.
  +# Fixture data is not source either.
  +sonar.exclusions=src/app/tests.rs,src/mcp/testdata/**
  ```

  Worth confirming on the first real run that the scan reports a non-zero test-file count rather than erroring, since — as with the import-count check from my last pass — a misconfigured scope and a clean scan are indistinguishable from a green check alone.

## Nitpicks

None new. The comment added at `sonar-project.properties:9-13` is genuinely good — it records *why* the split exists and explicitly names the inline `#[cfg(test)]` limitation as accepted, which is exactly the house style in `ci.yml`.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: every finding whose evidence lived inside the changed lines is resolved — the clippy step is non-fatal with a comment, the `if:` ordering matches the file's own convention, and the duplicated `SONAR_TOKEN` is gone; no in-diff findings remain.

- **Greptile score (codebase fit) — 3/5**: exactly one major remains, and it is only visible from outside the diff — the interaction between `sonar.sources=src` and the newly added `sonar.tests=src/app/tests.rs` entry, which Sonar's documented scope semantics turn into a scan-aborting overlap that this repo's green CI cannot reveal while the token is unset.

VERDICT: REQUEST_CHANGES

vitali87:

Verification of fix commit 422b0aa against the [round-2 review](https://github.com/vitali87/croft/pull/287#issuecomment-5402765171). CodeRabbit remains rate-limited on this PR.

---

## Re-review (round 3)

Commit `422b0aa` applies the suggested fix verbatim and touches nothing else — `git diff 045740a..422b0aa --stat` shows a single file, `sonar-project.properties`, at +4/-2. The workflow and `.gitignore` are byte-identical to the round-2 state I already cleared.

| Prior finding | Status | Evidence |
| --- | --- | --- |
| major `sonar-project.properties:8,14` — source/test scope overlap | **Resolved** | `sonar.exclusions=src/app/tests.rs,src/mcp/testdata/**` (line 18) subtracts the file from the source scope while `sonar.tests=tests,src/app/tests.rs` (line 14) keeps it classified as a test. Sonar's scope docs confirm the asymmetry this relies on: `sonar.exclusions` covers *"source files (non-test files) to be excluded from analysis"*, and *"file exclusion parameters apply to one code category at a time (source or test)"* — so the exclusion cannot strip the file from the test scope. No file is now claimed by both. |
| major `sonarcloud.yml:54` — truncated/fatal clippy report | **Resolved (unchanged)** | Still `... > clippy-report.json \|\| true` with the 50-53 comment. |
| minor `.gitignore` | **Resolved (unchanged)** | `/clippy-report.json` and `/.scannerwork/` still present. |
| nit — `if:` ordering | **Resolved (unchanged)** | `if:` at line 46, directly after `name:`. |
| nit — redundant step-level `SONAR_TOKEN` | **Resolved (unchanged)** | File still ends at line 57 on the `uses:` line. |
| minor — `pull_request` decoration | **Dismissed, accepted** | Unverifiable until the secret exists. |

Path patterns re-checked against the working tree: `src/app/tests.rs` is the only match under `src/app/tests*`, and `src/mcp/testdata` is the only `testdata` directory under `src/`, so neither pattern over- or under-matches. `tests/` still holds only `cli.rs`. The release-hygiene exemption continues to hold — this commit touches one root-level `.properties` file, matching none of the shipped-file patterns in `ci.yml:56-58`.

## Findings

None. The scope configuration is now internally consistent, and the remaining unverifiable items (PR decoration, first-run import count, and the test-file count confirming the scope split took effect) are all gated behind the one-time `SONAR_TOKEN` setup, correctly deferred rather than guessed at.

## Nitpicks

None new. The comment at lines 15-17 records why the exclusion exists, which is what keeps a future reader from "simplifying" the line back into the overlap that broke round 2.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: no findings remain whose evidence lives inside the changed lines; the clippy step, step ordering, and env handling are all clean, and this round's edit is a two-line change that does exactly what its commit message claims.

- **Greptile score (codebase fit) — 5/5**: the last outside-diff finding is closed — the source/test scopes no longer overlap under Sonar's documented one-category-at-a-time exclusion semantics, the two path patterns resolve to exactly the intended files in this tree, and the `ci.yml` conventions plus the release-hygiene exemption remain satisfied.

One operational note carried forward, not a blocker: the SonarQube check has been green on all three commits only because the token is absent and every substantive step is skipped. The first token-present run is the first real execution of this config — confirm there that the scan reports a non-zero imported-issue count and a non-zero test-file count, since a misconfigured scope and a clean scan look identical from the check mark alone.

VERDICT: APPROVE

Issue #257, opened 2026-08-23
Render ANSI colors when opening log files in the editor


vitali87:
## Motivation

Opening a log file that contains ANSI escape sequences shows `\x1b[31m` garbage in every editor, and VS Code has refused to render it for 8 years ([vscode#38834](https://github.com/microsoft/vscode/issues/38834), ~380 votes, closed as out-of-scope; the community answer is an extension that re-implements a terminal renderer badly). Croft has a full terminal emulator in-process (alacritty_terminal) and a theme system whose ANSI-16 palette already matches the terminal panes. Rendering ANSI in editor buffers is nearly free here and is exactly the "features are emulated in process" tenet.

## Current state

- Terminal panes parse and render full ANSI via alacritty_terminal.
- The editor render path (`src/widgets/editor.rs`) draws plain text with tree-sitter/LSP highlight spans; escape bytes render literally.
- "Open scrollback in an editor tab" already exists for terminal panes, and that content is de-ANSI'd on the way in (or shows raw escapes), losing the colors the pane had.
- Each theme carries its own 16-color ANSI palette (`src/theme.rs`), so rendered colors will match pane output.

## Proposed design

1. **A rendered ANSI view mode** for editor buffers, following the established viewer pattern (rendered by default for matching files, "Reopen as Text" for raw). Trigger heuristics: extension (`.log`, `.ansi`, `.out`, `.typescript` excluded) plus content sniff (ESC `[` within the first N KiB), same routing posture as the magic-byte sniffing in `src/magic.rs`.
2. **Rendering.** Parse SGR sequences (16/256/truecolor, bold, dim, italic, underline, inverse) into styled spans; strip non-SGR escapes (cursor movement, OSC) rather than interpreting them, since a log file is a stream transcript, not a screen. Map the 16 base colors through the active theme's ANSI palette. Implementation choice: a small dedicated SGR span parser is preferable to driving alacritty_terminal's grid, because a file is unbounded-width text, not a fixed-size screen; the parser is ~200 lines and has no layout opinions.
3. **View semantics.** The rendered view is read-only (like other rendered previews), text-selectable and copyable (copy yields the visible text without escapes, matching the Markdown preview's selection model). Find works over the stripped text. `path:line` click-through keeps working, since file-ref scanning runs on the stripped text with column mapping back to rendered cells.
4. **Scrollback handoff.** "Open scrollback in an editor tab" routes through this same view so pane colors survive into the tab.
[...]

Issue #185, opened 2026-08-17
Cmd+S on an image/PDF/sheet preview tab truncates the previewed file to 0 bytes


vitali87:
## What happens

With an image, PDF, or spreadsheet preview tab active, Cmd+S truncates the previewed file to 0 bytes.

## Why

`App::save` has no preview-tab guard, so it reaches `Editor::save_to_disk` → `write_buffer_to_disk`, which serialises the tab's TEXT buffer. Preview tabs leave `lines` at the whole-buffer-swap placeholder (one empty line), so the write replaces the file's real bytes with nothing. Reproduced with a unit probe: a 86-byte PNG is 0 bytes after `save_to_disk` on its preview tab, outcome `Saved`.

## Fix direction

Refuse non-text tabs at the single save choke point (`write_buffer_to_disk`), plus a friendly early return in `App::save`. Regression test: save on each preview kind leaves the file byte-identical.

Issue #184, opened 2026-08-17
Epic: open, render, and manipulate any file format


vitali87:
Every file should open to a useful view — nothing may dead-end in a "Binary file" error — and the common rich formats get first-class viewers and, where it makes sense, editing.

## Already shipped (before this epic)

Rendered Markdown preview (Cmd/Ctrl+Shift+V), image preview (png/jpg/gif/bmp/webp), PDF reader with clickable links and page navigation, read-only sheet viewer (csv/tsv/xlsx/xls/ods/xlsb via calamine), text encodings (UTF-16 BOM, Reopen with Encoding).

## Tickets, in implementation order

- [x] #172 Hex viewer: binary files open read-only instead of dead-ending
- [x] #174 Content-based format detection (magic bytes)
- [x] #175 SVG preview via resvg
- [x] #173 Hex editor: overwrite-mode editing and save
- [x] #177 Sheet grid: CSV/TSV editing and save
- [x] #178 Sheet grid: xlsx editing via umya-spreadsheet
- [x] #176 Markdown preview: real inline images
- [x] #179 Archive browser: zip/jar/tar
- [x] #180 Jupyter notebook rendered view
- [x] #182 SQLite database browser
- [x] #181 docx/odt rich-text preview
- [x] #183 Audio/video info view

Ordering rationale: the hex fallback (#172) is the safety net every other route degrades to, and magic-byte routing (#174) decides which viewer wins — those two are the foundation. SVG (#175) is near-free (resvg already in-tree). Editing lands sheet-first (#177, #178) since spreadsheets are the strongest manipulation ask; the remaining viewers follow.

vitali87:
All twelve tickets shipped (#172-#183, plus the #185 save-clobber and #192 narrow-terminal fixes found along the way), v0.1.734-745. Every file now opens to a useful view: text, Markdown with inline local images, images, SVG, PDF, spreadsheets (CSV/TSV/xlsx editable, formulas held for explicit consent), hex (viewable and editable in place at any size), archives (safe member extraction), Jupyter notebooks, Word documents, SQLite databases (paged, read-only), and audio/video info cards - with content-based routing for extensionless and misnamed files and the hex viewer as the universal no-dead-end floor.

PR #293, opened 2026-08-25
feat(editor): render ANSI colour logs in place (v0.1.768)

vitali87:
Addresses #257 (core view; see Scope below for what is deliberately not in this PR).

## What

Opening a colour log showed raw `\x1b[31m` bytes, and a *large* one showed nothing at all — it hit the 50 MB text guard and dead-ended. This renders them.

- **`ansi_text`** — an SGR span parser: 16/256/truecolour, bold, dim, italic, underline, inverse. It parses colour into styled spans over the **stripped** text and drops every other escape.
- **`log_view`** — a windowed backing store: a line-offset index plus one bounded read per scroll, so the file is never held whole.
- Routing, open, and render wired into the editor, with `force_text` ("Reopen as Text") already giving the raw-bytes escape hatch for free.

## Design decisions worth reviewing

**Spans over a terminal grid.** A log is a stream transcript, not a screen: unbounded line width, no cursor. So cursor movement, OSC titles, and OSC-8 hyperlinks are *stripped*, not interpreted — driving `alacritty_terminal`'s grid would impose a fixed width the file does not have.

**Colours stay symbolic until paint.** `AnsiColor::Indexed(n)` for the low 16 resolves against `theme.ansi()` in the renderer, so a theme switch recolours without reparsing, and a rendered log matches the terminal panes. The 256-colour cube and greyscale ramp are computed from the xterm spec instead, since those are fixed rather than themed.

**The route runs before the size guard.** That is the point of the windowing: a colour log past `MAX_FILE_BYTES` opens rendered rather than being refused. `a_huge_color_log_opens_rendered_instead_of_hitting_the_size_guard` pins it against a file built past that ceiling.

**Sniffing requires a real SGR sequence**, not merely an escape byte. A screen recording full of `ESC[2J` cursor movement is not a colour log, and rendering it as decoration would mislead — `only_files_carrying_sgr_sequences_route_to_the_log_view` covers that distinction along with ANSI hiding under a `.txt` name.

**The index is capped** at `MAX_INDEXED_LINES` and reports `truncated` in the header, so a pathological file says so rather than silently showing a prefix as the whole thing.

## Tests

27 new tests: 12 parser (including malformed/truncated sequences, multibyte slicing at escape boundaries, cross-line colour carry, and that a truecolour sequence does not swallow the attribute after it), 8 windowing (CRLF, no-trailing-newline, invalid UTF-8, seeking to a late window without touching earlier ones), and 7 editor-level (rendered open, reopen-as-text round trip, sniff discrimination, the size-guard case, and that painted cells carry the theme's palette colour).

## Scope

This lands the view and its correctness core. Not in this PR, and worth follow-ups rather than bloating one change: selection/copy over rendered spans, find over the stripped text, `path:line` click-through with column mapping, and routing the terminal scrollback handoff through this view. The acceptance criteria covering those stay open on #257.


vitali87:

CodeRabbit did not run on this PR at all (no check present), so this review was produced locally by a Claude Opus 5 agent with full-repository context. Reviewed at head SHA `d0bd10b348d238b21bacf5f045f4e082e56fcdd6`.

**VERDICT: REQUEST_CHANGES** — four blockers. All are fixed in `8e076e5`; details in the reply below.

## Findings

**`src/widgets/editor.rs:4381` — [blocker/correctness] `has_non_text_view()` does not include `self.log`.**

The function's own doc comment states the contract: *"Every preview kind the format epic (#184) adds joins HERE, so the twenty-odd guard sites never enumerate kinds again."* This kind did not join. There are 23 call sites in `app/mod.rs` plus two in `editor.rs`; the consequential one is `write_buffer_to_disk`, whose comment names it the *"declared single choke point"*. Since `open_log` sets `self.lines = vec![String::new()]`, a save on a rendered log tab serialises one empty line over the user's log file — the #185 truncation class, on a view whose selling point is opening multi-gigabyte files. LSP sync, vim commands, encoding, and format-on-save are likewise unguarded.

**`src/ansi_text.rs:142` — [blocker/panic] `_ => i += 2` advances into the middle of a multi-byte character.**

The two-byte-escape arm adds 2 unconditionally. With `ESC` followed by a non-ASCII character, `i` lands inside it, and the copy-through branch then panics:

```
thread 'main' panicked at src/ansi_text.rs:89:35:
start byte index 2 is not a char boundary; it is inside '🙂' (bytes 1..5)
```

Reachable from file content — a log carrying a stray `0x1b` before UTF-8 text is exactly the "stray bytes from binary payloads" case the diff's own test anticipates — and it fires inside the render path, taking down the editor on a frame.

**`src/log_view.rs:124` — [blocker/data-loss] A window that stops mid-line caches a TRUNCATED line under a real line number.**

`buf` is clamped to `WINDOW_BYTES` but the insert loop still walks `split('\n')` and inserts under absolute indices. Measured on 40 lines of ~20 KB: only 14 cached, and line 13 rendered 2115 bytes of its real 20001 with no marker — the reader sees a cut-off log line as if it were the whole thing. Multi-KB lines are routine in JSON logs and stack traces. Rows beyond simply do not paint, so the viewport goes blank with no explanation.

**`src/widgets/editor.rs:4465` — [major/correctness] `open_hex` does not clear `self.log`.**

Every other opener in the diff gained `self.log = None`. The render dispatch checks `log` before `hex`, so with both set the log wins: "Reopen as Hex" reports success while the pane keeps painting the log — broken for precisely the files (`.log`, `.txt`, `.out`) most likely to need it.

**`src/widgets/editor.rs:4413` + `:9295` — [major/correctness] The view cannot scroll past the first screen.**

`open_log` sets `lines = vec![String::new()]`, and `scroll_view_to` clamps `self.scroll = top.min(self.lines.len().saturating_sub(viewport))` — i.e. `top.min(0)` for any viewport ≥ 1, pinning scroll at 0 permanently. `grep -c 'editor\.log' src/app/mod.rs` returns **0**: no scroll, mouse, or key wiring exists for this kind, where `hex` has ~17 such sites. So `ensure()`, the windowing, and `MAX_INDEXED_LINES` are all unreachable beyond the first window, and the headline acceptance criterion — *"a 2 GB log opens … scrolling stays smooth"* — is not met.

## Also found

- `src/ansi_text.rs:172` — **[major]** colon sub-parameter syntax (`ESC[38:5:1m`, ITU T.416, emitted by libvte terminals and several colour crates) parses as one unrecognised token, so the colour is dropped. Worse than a gap: `looks_like_ansi` returns true for it, so such a file routes into the log view and renders entirely uncoloured — strictly worse than the text path it displaced.
- `src/ansi_text.rs:141` — **[minor]** three-byte forms (`ESC ( B`) and DCS/SOS/PM/APC leak their final byte or entire payload into the visible text.
- `src/log_view.rs:88` — **[minor]** `len()` returns 1 for an empty file while `is_empty()` returns true; the renderer reads `len()`, so an empty log shows "1 lines" and a blank row.
- `src/widgets/editor.rs:4491` — **[major]** `x` advances by `chars().count()` rather than display width, and `set_stringn`'s return is discarded. Wide characters advance 2 cells but count 1, so following spans paint over the CJK text; control characters (tabs, ubiquitous in Go and Java output) are counted but never painted, shifting every subsequent span.

## Test coverage

The 27 new tests are real tests, but they cover the two modules in isolation and almost nothing about the view as an editor tab — which is where all four blockers live. Missing: save refusal on a log tab; scrolling a rendered log; preview-kind mutual exclusion in both directions; lines wider than `WINDOW_BYTES`; `ESC` before a multi-byte char; wide/control characters in `render_log`.

**On the deferred scope.** Selection/copy, find, and click-through are honestly deferred — additive, unclaimed, with `visible_text()` left as a sensible seam. But two items are missing from the list and are *not* additive: **scrolling** is load-bearing for the PR's central claim, and **the read-only guarantee** is asserted in two doc comments while the guard function was never updated. A claim in a doc comment that the code does not enforce is worse than no claim.

CodeRabbit score (diff quality): 2/5
Greptile score (codebase fit): 2/5

The module-level design is genuinely good — spans over a terminal grid, symbolic low-16 colours, routing before the size guard are each the right call and correctly implemented. The codebase-fit score is low specifically because the repo had already solved the preview-kind lifecycle problem with a single documented choke point, and this change added a kind without joining it.

vitali87:
All findings addressed in `8e076e5`, and main merged in (`b9a49f5`) to clear the conflict that was blocking CI from running at all.

**The four blockers:**

1. **`has_non_text_view()`** now includes `self.log`. This was the worst of them — verified the chain myself: `open_log` sets `lines = vec![String::new()]`, and `write_buffer_to_disk` gates on that predicate, so a save would have written one blank line over the user's log. Exactly the #185 class, on the view least able to afford it.

2. **The panic** is fixed, and fixing it surfaced a second bug the reviewer's suggested fix would have left: advancing to the next char boundary stops the panic but *swallows the character*. So `ESC` followed by a non-ASCII byte is now treated as a stray ESC in text — kept, matching the existing trailing-ESC rule — while a real two-byte escape still disappears. Also handled the three-byte forms (`ESC ( B`) and DCS/SOS/PM/APC, which were leaking payload as text.

3. **Window truncation:** a short read now keeps only through the last newline, so a line is either fully present or absent — never cached truncated under a real line number. If not even one full line fits, nothing is cached rather than caching a lie.

4. **`open_hex`** clears `self.log`, with a comment naming the dispatch order as the reason.

**The scrolling finding was the most valuable one.** You were right that it makes the headline claim untrue: `scroll_view_to` clamped against `lines`, a one-line stub for a log, pinning scroll at 0 — so the windowing this feature exists for was unreachable in the shipped UI. Fixed by clamping against the log's own line count (minus the header row) and routing `scroll_up`/`scroll_down` for log tabs ahead of the wrap paths that assume an editable buffer. `a_rendered_log_scrolls_past_the_first_screen` scrolls a 2 000-line log to line 500, asserts the window followed, and checks clamping at both ends.

Also fixed `len()`/`is_empty()` disagreeing on an empty file.

**Deferred, with reasons rather than silence:**

- **Colon sub-parameter syntax** (`ESC[38:5:1m`) — a real gap, and your point that `looks_like_ansi` accepts it while the parser drops the colour is the sharp part: such a file renders *worse* than the text path it displaced. Not fixed here because the fix belongs with a wider `apply_sgr` restructure rather than bolted onto this change set; filing it as a follow-up.
- **`x` advance by char count rather than display width** — likewise real (tabs are ubiquitous in Go/Java output). The right fix is using `set_stringn`'s return value, which is a small change but wants its own test matrix for wide chars and controls.

Both are noted on #257 rather than left implicit.

Verification: fmt clean, clippy clean, 13 parser + 8 windowing + 61 log tests green, including the new scroll test.

Issue #260, opened 2026-08-23
Auto-hide sidebar


vitali87:
## Motivation

Terminal real estate is scarcer than a GUI's, which makes an auto-hiding sidebar more valuable in croft than in VS Code, where it has been requested since 2015 and never shipped ([vscode#3742](https://github.com/microsoft/vscode/issues/3742), ~590 votes; the community lives on the `Auto Hide` extension). The idea: the Explorer/Search/SCM sidebar collapses when focus moves to the editor or terminal, and returns on demand, so the code pane gets the columns back.

## Current state

- `src/app/` owns layout; the Customize Layout popup already toggles activity bar / sidebar / panel / status bar visibility, sidebar position L/R, and panel alignment.
- `Cmd+B` (or equivalent) toggles the sidebar manually today; Zen Mode exists as the all-chrome-off extreme.
- Focus transitions between sidebar, editor, and panel are already modeled (activity-bar switching, focus-driven active-repo following in SCM), so the hook points exist.

## Proposed design

1. **A layout setting** `sidebar.auto_hide` (default off), plus a checkbox in the Customize Layout popup and a palette command "View: Toggle Auto-Hide Side Bar".
2. **Behavior.** With auto-hide on: focusing the editor or a terminal pane collapses the sidebar after a short grace delay (so a click that passes through doesn't flap); activating any activity-bar icon, `Cmd+B`, `Cmd+Shift+E/F/G`-style focus chords, or starting a sidebar-targeted command (e.g. reveal-in-explorer) shows it again with focus. When the sidebar loses focus back to the editor, it collapses again.
3. **Reveal affordance.** When hidden, the activity bar remains (unless the user also hid it), so the mouse path is one click. If the activity bar is hidden too, hovering the far-left column edge for a beat slides the sidebar out (and it collapses on mouse-leave + focus-elsewhere), mirroring what the VS Code extension does.
4. **Interactions to get right.** Drag-resize of the sidebar seam must not trigger a collapse mid-drag. A modal flow that borrows sidebar space (connect dialog, branch picker) suppresses auto-hide until it closes. Zen Mode wins over everything, as now. The secondary side bar gets the same setting independently (`secondary_sidebar.auto_hide`).
5. **Persistence.** Per-workspace layout persistence already exists for terminal panes; store the auto-hide flags with the rest of the layout state so a workspace reopens as it was left.

## Acceptance criteria

[...]

PR #294, opened 2026-08-25
feat(layout): auto-hide the side bar on editor focus (v0.1.769)

vitali87:
Addresses #260 (primary side bar; see Scope for what is deliberately deferred).

## What

Opt-in `sidebar_auto_hide`: with it on, focusing the editor or a terminal collapses the primary side bar, and any deliberate sidebar action brings it back. Reachable three ways per the issue — the `sidebar_auto_hide` pref, an **Auto-Hide Side Bar** checkbox in Customize Layout, and `View: Toggle Auto-Hide Side Bar` in the palette.

## The design decision worth reviewing

**The risk here is flapping, not collapsing.** Seam drags, sidebar-internal popups, and modal flows that borrow sidebar space all look like "focus moved away". So every reason to hold the sidebar open lives in a single predicate rather than as guards sprinkled through call sites:

```rust
fn sidebar_auto_hide_allowed(&self) -> bool  // off / already hidden
                                             // pinned by a reveal command
                                             // mid seam-drag
                                             // prompt or context menu open
                                             // zen mode
```

That makes the interaction rules one readable list, gives a new suppression case exactly one place to go, and — the practical part — makes them testable without driving a terminal.

**Reveals pin at one choke point.** `set_sidebar_view` is where all 34 activity-bar and focus-chord paths funnel, so the pin is taken there rather than at each call site. It matters because these commands *reveal the sidebar and then move focus into it* — without the pin, a reveal would trigger the very auto-hide that undoes it. Landing focus in the sidebar releases the pin, so the next move out collapses again.

## A trap worth flagging for reviewers

The Customize Layout row is **appended after Minimap**, not inserted next to the other side-bar toggles where it reads more naturally. Each row reopens the menu with a hardcoded index (`open_customize_layout_menu_at(1)`, `(2)`, …), so a mid-list insert silently mis-selects every row below it. `customize_layout_toggle_rows_keep_their_reopen_indices` pins all seven positions so the next insert fails in CI instead of in someone's hands.

## Tests

- `auto_hide_collapses_on_editor_focus_and_deliberate_actions_restore_it` — collapse on editor **and** terminal focus, reveal via activity-bar switch, pin released when focus lands in the sidebar, collapse again on leaving.
- `auto_hide_is_suppressed_while_dragging_prompting_or_in_zen_mode` — each suppression case, plus that it is genuinely off by default.
- `reveal_in_explorer_shows_a_collapsed_sidebar` — a sidebar-targeting command works when auto-hide has collapsed it, rather than silently doing nothing visible.
- `customize_layout_toggle_rows_keep_their_reopen_indices` — the index invariant above.

## Scope

This lands the primary side bar. Deliberately **not** here, and better as follow-ups than as bloat in one change:

- **`secondary_sidebar.auto_hide`** — the issue asks for an independent flag; the secondary bar has its own visibility path and deserves its own change.
- **The hover-to-reveal affordance** when the activity bar is *also* hidden — that needs a dwell timer and mouse-leave handling, which is a different kind of risk (timing) from what this PR is about (state).
- **Per-workspace persistence** — the flag persists via the global pref today; storing it with per-workspace layout state is a separate concern that touches the layout persistence format.

The acceptance criteria covering those stay open on #260.

## Summary by CodeRabbit

* **New Features**
  * Added an optional Auto-Hide Side Bar setting that collapses the sidebar when focus moves to the editor or terminal.
  * Added a command palette action to toggle Auto-Hide Side Bar.
  * Added Auto-Hide Side Bar to the Customize Layout menu.
  * The setting is saved and restored across sessions.

* **Bug Fixes**
  * Improved sidebar behavior during dragging, dialogs, Zen Mode, and other temporary UI states.


vitali87:

CodeRabbit's check on this PR reports `pass` but is annotated "Review rate limited", so no bot review ran. This review was produced locally by a Claude Opus 5 agent with full-repository context, anchored to head SHA `f60973a3692f1006d2ef44b8c1e88ccac947a504`.

**VERDICT: REQUEST_CHANGES.** Fixes are in the follow-up commit; see the reply below.

## Findings

**`src/prefs.rs:136` [blocker/correctness] — `sidebar_auto_hide` never persists.**

`grep -rn "sidebar_auto_hide" src/` returns 11 hits excluding tests. It is read once at `app/mod.rs:3940` and mutated at `27790` (palette) and `33965` (menu). **Neither writes back.** Every other pref has a `save_*` helper (`save_theme`, `save_auto_save`, `save_inline_blame`, … 20+ at `prefs.rs:277-455`), and sibling layout toggles call `persist_layout()`. The field was also added to top-level `Prefs`, not `LayoutPrefs`, so `persist_layout()` would not carry it either.

The PR body claims "the flag persists via the global pref today" and the issue's criterion is "round-trips through … config". Neither holds: toggling is lost on restart, so both new UI affordances are misleading.

**`src/app/mod.rs:11659` and `:34183` [major/dead-code] — the pin is inert in both places the PR adds it.**

All seven arms of `set_sidebar_view`'s `match view` call `focus_pane(Pane::Tree)`, which hits the `sidebar_pinned_open = false` arm. So the pin is cleared before the function returns, every time — the reveal was never at risk from its own focus move, and the PR's central design rationale rests on a flag that does nothing. The PR's own test concedes it: `assert!(!app.sidebar_pinned_open, "…the pin has done its job")` asserts the pin is false right after the operation that supposedly needed it. What keeps the sidebar open is `show_tree = true`, not the pin.

**`src/app/mod.rs:27787` [major/correctness] — `Cmd+B` leaves the pin stuck true, silently disabling auto-hide.**

`Cmd+B` sets the pin on reveal but does **not** move focus. The pin's only clear paths are a Tree focus or a subsequent `Cmd+B` hide. Reveal with `Cmd+B`, then hide via the Customize Layout row or Zen Mode, and the pin is `true` with no Tree focus in the history — every later editor focus early-returns, so auto-hide is off for the rest of the session with no indication why.

**`src/app/mod.rs:33968` [major/correctness] — the appended row shifts reopen indices for rows 7+.**

Enumerating both sides: post-PR the side-bar position items sit at 9/10, panel alignment at 13-16, quick input at 19/20 — but the dispatch still passes 7/8, 11-14, 17/18. Those land on separators and headers. They were **already** off by one pre-PR (a latent bug), and this PR shifts them one further, so the stated goal "no existing index shifted" does not hold. The index test asserts rows 0-6 only and cannot catch it; there are 14 hardcoded index arguments in `dispatch_menu_action`.

**`src/app/mod.rs:12592` [major/completeness] — the suppression list omits the eleven overlay states the codebase already treats as modal.**

`App` carries a modal set repeated verbatim three times in this file (`flush_run_debug_icon_overlay`, `flush_problems_badge_overlay`, `flush_no_repo_hero_overlay`): `shortcuts_modal`, `file_finder`, `command_palette`, `go_to_symbol`, `workspace_symbols`, `process_picker`, `zoxide_jump`, `command_history_popup`, `branch_picker`, `input_prompt`, `list_picker`. None are checked. Two specifics contradict the code comment naming them: `branch_picker` and `connect_dialog` are their own fields, **not** `Prompt`s — so the two flows the comment claims are covered are precisely the two that are not. Also missing: `tree_drag`, the Explorer drag state, which is exactly the "drag must not collapse" case from the issue and is not `splitter_drag`.

**`src/app/mod.rs:12650` [major/UX] — the `focus_pane` arm fires on programmatic focus with no user intent.**

Mapping all 68 `focus_pane(Pane::Editor|Terminal)` sites: `poll_mcp:27352` (an async MCP result yanks the sidebar shut with no user input — the flapping this PR exists to prevent, by the one route the predicate cannot see), `open_search_hit:22585` (Enter on a Search result collapses the sidebar being iterated, destroying the result list mid-task), `handle_tree_key:22696` (Enter on a tree row closes the tree being navigated), `open_file_at_launch:18802` (collapses during startup before any interaction).

**`src/app/mod.rs:12586` [minor/UX] — no reveal affordance when the activity bar is also hidden.** With `activity_bar_visible = false` plus auto-hide on, focusing the editor leaves no on-screen way back; hover-to-reveal is deferred. Keyboard still works, so recoverable, but a mouse user is stranded.

**`src/app/mod.rs:27789` [minor/consistency] — the palette toggle sets a status message; the menu toggle does not.**

## Test coverage

Three of the four tests assert conditions that hold independently of the code under test. `reveal_in_explorer_shows_a_collapsed_sidebar` asserts only `show_tree`, which `set_sidebar_view` guarantees alone — it passes with the PR's added pin line deleted. `auto_hide_is_suppressed_while_dragging_prompting_or_in_zen_mode` **never sets `prompt` or `context_menu`** despite its name. Untested: persistence (would have caught the blocker immediately), the stuck-pin path, reopen indices 7-20, and programmatic focus.

CodeRabbit score (diff quality): 2/5
Greptile score (codebase fit): 2/5

vitali87:
All findings addressed in `f0dd031`, plus a main merge (`969b351`) to clear the conflict left by #293 landing.

**The blocker — persistence — was real and I verified it before fixing.** `grep -rn sidebar_auto_hide src/` confirmed exactly what the review described: read once at startup, mutated at both toggle sites, never written back. So the PR body's "persists via the global pref" was false and both new affordances were misleading. Added `save_sidebar_auto_hide` following the `save_auto_save` pattern, and funnelled both routes through one `toggle_sidebar_auto_hide()` — which also closes the status-message divergence you spotted, since the menu route previously set none.

**The dead-pin finding was correct and worth more than it looks.** I checked every arm of `set_sidebar_view`'s match: all seven call `focus_pane(Pane::Tree)`, which clears the pin before the function returns. So the flag did nothing there, and my PR body's central rationale rested on it. Removed from both sites.

But the follow-on is the part I'd have missed without your `27787` finding: `Cmd+B` reveals **without** moving focus, so the pin there had no clear path at all — reveal with `Cmd+B`, hide via the menu row, and auto-hide was dead for the session with no indication why. It is now a **one-shot exemption** consumed in `maybe_auto_hide_sidebar` via `mem::take`, so it protects exactly the one focus move it was granted for and the feature re-arms itself. `the_reveal_exemption_is_consumed_rather_than_sticky` pins both halves.

**The menu index finding was the most valuable.** You were right that my mitigation was incomplete, and right about the cause: the indices were *already* off by one for every row below the first separator, and appending made it off by two. I enumerated the entries and confirmed side-bar position sits at 9/10 while dispatch passed 7/8. Rather than correct six numbers that will rot again, I replaced them with `open_customize_layout_menu_on(&MenuAction)` — a lookup by action, which makes the class unrepresentable. That also fixes the pre-existing off-by-one.

**Suppression list:** extracted the eleven-way modal check the codebase already triplicates at its overlay-flush sites into `modal_overlay_open()`, and call it from the predicate. Added `connect_dialog` and `tree_drag`. Your observation that `branch_picker` and `connect_dialog` are their own fields rather than `Prompt`s was the sharp bit — my comment named as covered precisely the two flows that weren't.

**Programmatic focus:** added `without_auto_hide()` and wrapped `poll_mcp` and `open_file_at_launch`. On the other two you flagged: I checked `open_search_hit` and the tree-Enter path, and neither calls `focus_pane` — the focus move happens elsewhere — so search results and tree navigation are not affected. Worth stating rather than silently not fixing.

**Activity-bar dead end:** added the one-line guard, so auto-hide is suppressed entirely when the activity bar is hidden. That is a stopgap, not the intended behaviour — hover-to-reveal is what should replace it, and I have noted that on #260 rather than leaving the deferral implicit.

**On the weak assertions:** you were right that three of the four tests asserted conditions holding independently of the code. Added three that do not: both toggle routes agreeing, the one-shot exemption consuming itself and re-arming, and the modal/activity-bar/programmatic-focus suppression cases. I did not add the prefs round-trip test — the toggle now calls the save helper, but asserting the on-disk write needs the same `cfg!(test)` carve-out `persist_layout` uses, and I would rather leave that gap visible than paper it with a test that only exercises the in-memory field.

fmt clean, clippy clean, 11 tests green across auto-hide, customize-layout, reveal, and the toggle routes.

vitali87:

CodeRabbit's check on this PR is annotated **"Review rate limited"**, so its green row is a no-op and no bot review ran.

The first round was anchored to `f60973a`, and addressing it **substantially reworked the design** — the suppression model became `sidebar_auto_hide_suspended` + `without_auto_hide` + a one-shot pin. That earlier artifact no longer describes the code, so this is a fresh review of the current state on its own merits, anchored to `1dfe4e2`.

**VERDICT: REQUEST_CHANGES** — one blocker, three majors. All fixed in `dacd231`; see the reply below.

---

## Walkthrough

Adds an opt-in `sidebar_auto_hide` pref (#260): with it on, moving focus to the editor or a terminal collapses the primary side bar, while deliberate side-bar actions reveal it. Suppression is centralised in `sidebar_auto_hide_allowed()`, with two escape hatches — `sidebar_auto_hide_suspended` (scoped via `without_auto_hide()`) and `sidebar_pinned_open` (one-shot). Also replaces the Customize Layout menu's hardcoded re-open indices with an action lookup, and adds `prefs_for_update()`.

| file | summary |
|---|---|
| `Cargo.toml` | 0.1.773 → 0.1.775 |
| `src/release_notes.rs` | replaced with the auto-hide feature (correct: replace, not append) |
| `src/app/mod.rs` | three new `App` fields; `sidebar_auto_hide_allowed()`, `modal_overlay_open()`, `without_auto_hide()`, `maybe_auto_hide_sidebar()`, `toggle_sidebar_auto_hide()`; auto-hide hook in `focus_pane`; `open_customize_layout_menu_on()` replacing 9 hardcoded indices |
| `src/prefs.rs` | `sidebar_auto_hide`; `save_sidebar_auto_hide()`; `prefs_for_update()`; `save_problems_scope` repointed |
| `src/widgets/command_palette.rs` | `Command::ToggleAutoHideSideBar` |
| `src/app/tests.rs` | 7 new tests |

## Findings

- **`src/app/mod.rs:16564`** **[blocker/correctness]** — **the real Cmd+B never sets `sidebar_pinned_open`, so the feature's headline reveal gesture is broken.** The pin is set in exactly one place, `Cmd::ToggleSideBar` (line 28189), but an actual Cmd+B keypress is handled at 16564 by `is_sidebar_toggle_key(key)`, which flips `show_tree` and `return Ok(())` — never reaching `Cmd` dispatch. So the pin is reachable only by typing "View: Toggle Primary Side Bar" into the palette. User-visible: auto-hide on, sidebar collapsed, press Cmd+B → it appears; click into the editor → it collapses again. Both the release note and the field doc promise Cmd+B specifically. The same omission affects the Customize Layout row (34367) and the layout-icon click (30361). Four sites flipping `show_tree` by hand is the real smell; one `toggle_side_bar()` fixes all of them.

- **`src/app/mod.rs:12683-12684`** **[major/correctness]** — **the extraction the doc comment describes did not happen.** The comment says the eleven-way check "was repeated verbatim at three overlay-flush sites … so it lives here once", but the identical chain is still inline at 4699, 4780 and 4833 — all three byte-identical to `modal_overlay_open()`'s body. The diff *adds* a fourth copy rather than removing three. No condition was dropped or added, so this is not a behaviour bug; it is a documented refactor that was not performed, leaving the drift hazard the comment claims to have closed.

- **`src/app/tests.rs:30740-30757`** **[major/tests]** — **`both_auto_hide_toggle_routes_agree` writes to the developer's real `~/.config/croft/config.json`.** `toggle_sidebar_auto_hide()` calls `save_sidebar_auto_hide()` → `config_path()`, and nothing overrides `XDG_CONFIG_HOME`. The test toggles twice, rewriting the running user's (and CI's) prefs mid-suite. Every other prefs test uses a tempdir with an explicit path. Also order-dependent flakiness waiting to happen.

- **`src/app/mod.rs:8587` and `:9062`** **[major/correctness]** — **two async LSP replies focus the editor without `without_auto_hide`.** `present_call_hierarchy_menu` and `open_location_picker` both carry doc comments saying a late reply must not "yank focus to the editor under them" — the same reasoning as the wrapped MCP site — and neither is wrapped. Worse, both set `self.context_menu` on the line *after* `focus_pane(Pane::Editor)`, so the `context_menu.is_some()` guard is not yet armed when the collapse fires. Ask for Go to References, switch to the Explorer while the server thinks, and the sidebar collapses when the reply lands.

- **`src/app/mod.rs:9216-9219`** **[minor/docs]** — the pre-existing doc for `open_customize_layout_menu_at` was orphaned onto the new function, leaving two stacked docs on one item and none on the other, describing different parameters.

- **`src/app/mod.rs:34373-34375`** **[minor/docs]** — the comment "appended after Minimap precisely so the existing toggle rows keep their hardcoded re-open indices" contradicts the change it sits inside: this diff removed every hardcoded index. The test `customize_layout_toggle_rows_keep_their_reopen_indices` asserts an invariant the diff deliberately abolished.

- **`src/prefs.rs:429-438`** **[minor/correctness]** — `prefs_for_update`'s absent-vs-malformed test misclassifies unreadable files. The `save_problems_scope` repoint is behaviour-preserving and strictly better on a corrupt config. But `Prefs::load` returns `Err` for *any* read failure and the helper distinguishes only on `path.exists()`: a dangling symlink returns `Prefs::default()` and the subsequent save writes defaults through it — the exact clobber the helper exists to prevent.

- **`src/app/mod.rs:12769-12776`** **[minor/correctness]** — `sidebar_pinned_open` can be set and never consumed. It is set regardless of the pref state, so a palette Cmd+B with auto-hide *off* banks a pin that persists; enable auto-hide later and the first collapse is silently eaten.

## Test coverage

- **Cmd+B via the actual keybinding** — every pin test sets the flag by hand or goes through `Cmd::ToggleSideBar`; nothing sends the key event, which is exactly why the blocker went unnoticed.
- **The Customize Layout row and layout-icon click as reveal gestures** — both reveal without pinning; neither is tested under auto-hide.
- **The async-reply suppression sites** — no coverage.
- **Pref round-trip** for `sidebar_auto_hide`.
- **A pin taken while the pref is off.**

## Nitpicks

- `src/app/mod.rs:1103` — `MenuAction::ToggleAutoHideSideBar` is the only variant in that block without a doc comment.
- `src/app/mod.rs:12758` — the toggle reports "on" even when the write failed; matches `toggle_minimap`, so a follow-up rather than a change here.
- Version bump and release notes comply with the gate. Note the bump crosses two versions because of the main merge.

**VERDICT: REQUEST_CHANGES**

vitali87:
All findings addressed in `dacd231`. The blocker was real and I verified it in the tree before fixing.

**The blocker.** Confirmed exactly as described: `grep -n 'sidebar_pinned_open' src/app/mod.rs` showed the pin set at one site only (`Cmd::ToggleSideBar`), while a real Cmd+B is caught at 16564 and returns before `Cmd` dispatch. So the gesture the release note advertises was the one gesture that did not work.

I took the structural fix rather than adding a second assignment. Four sites flipped `show_tree` by hand and three lacked the pin — that ratio is the actual defect, so there is now one `toggle_side_bar()` and all four routes call it: keybinding, palette, Customize Layout row, layout-icon click.

While doing it I also closed the stale-pin case from your last finding, since it becomes trivial once there is a single site:

```rust
self.sidebar_pinned_open = self.show_tree && self.sidebar_auto_hide;
```

A reveal while the feature is off no longer banks an exemption that silently eats the first collapse after the user turns it on.

**The extraction that never happened.** You were right, and it is worth naming what went wrong: I wrote a doc comment describing a refactor and then only did half of it, so the comment asserted a deduplication while the diff added a fourth copy. I diffed all three sites against the predicate body to confirm they were byte-identical before substituting; they were, so no condition changed.

**The async LSP sites.** Confirmed, including the ordering detail — both set `context_menu` on the line *after* the focus move, so the guard could never have covered them. Both now wrap in `without_auto_hide`.

**The test writing the real config.** Confirmed: nothing in the suite overrides `XDG_CONFIG_HOME`, so that test was rewriting my actual `~/.config/croft/config.json` on every run. Fixed at the source rather than in the test — `toggle_sidebar_auto_hide` now skips the write under `cfg!(test)`, mirroring the existing write-skip in `apply_theme`.

**`prefs_for_update`.** Branches on the io error kind now instead of a follow-up `exists()`. Since `load` wraps the error with context, the cause is in the chain rather than the top error, so it walks `e.chain()` for a `NotFound`.

**Both doc nits.** The orphaned comment is back on `open_customize_layout_menu_at`, and the "hardcoded re-open indices" comment is gone — it described the thing this PR removed.

**Tests.** Two added, both driving the actual key event rather than setting the flag by hand, which is the gap that let the blocker through the first round:
- `the_real_cmd_b_keypress_reveals_and_holds_the_sidebar` — sends Cmd+B and Ctrl+B through `handle_key`, asserts the reveal survives the next editor focus and that the exemption is still one-shot.
- `a_reveal_while_auto_hide_is_off_does_not_bank_an_exemption`.

I verified the first one actually fails without the fix — reverting the pin line fails it on precisely the "reveal survives the next move to the editor" assertion — rather than trusting that a passing test meant anything.

**One thing I did not do.** There is still no grace delay before collapsing, so a click that passes through the panel en route to the editor can flap. That is a deliberate deferral, not an oversight: it is a behaviour change that deserves its own issue and its own review, and I did not want to fold it into a round that already reworks the pin. I will file it separately.

fmt clean, clippy clean at `-D warnings`, full auto-hide suite green.

On the suite generally: a full parallel run intermittently fails a few terminal/clipboard tests. I baselined those against untouched `origin/main` and they fail there too, so it is pre-existing load sensitivity around real shell spawns rather than anything here — all pass serially or on an idle machine.

Issue #256, opened 2026-08-23
Project-wide diagnostics: whole-workspace errors in PROBLEMS, not just open files


vitali87:
## Motivation

PROBLEMS currently shows diagnostics for files the language server has been asked about, which in practice means open files plus whatever the server volunteers. A rename that breaks ten call sites across the repo shows one error until you open the other nine files. This is [vscode#13953](https://github.com/microsoft/vscode/issues/13953) (~1,500 votes, open since 2016): show errors for the whole project, not just open editors. VS Code still hasn't shipped it as a first-class feature; croft can.

## Current state

- `src/lsp/manager.rs`: diagnostics via `textDocument/publishDiagnostics`, per-project-root server anchoring, workspace pre-warm at startup.
- `src/build_matchers.rs`: build-tool output already lands in PROBLEMS from terminal command boundaries, which is a partial project-wide path today (run `cargo check` and the whole workspace's errors appear). This issue is about making it automatic and server-driven.
- `src/widgets/problems.rs`: PROBLEMS panel with free-text filter.

## Proposed design

Per-language strategy, because the protocol support differs:

1. **Servers with project-wide push.** rust-analyzer publishes workspace-wide diagnostics from `cargo check` on its own; basedpyright publishes for the whole project when `diagnosticMode: "workspace"` is set. First step is configuration: enable workspace mode where a server supports it (pyright family), and verify croft renders push diagnostics for files that were never opened (URI → file row in PROBLEMS without a buffer). If the panel currently drops diagnostics for unopened URIs, fix that.
2. **Pull diagnostics (LSP 3.17 `workspace/diagnostic`).** Implement the workspace diagnostic request with `previousResultIds` for incrementality, for servers that advertise `diagnosticProvider.workspaceDiagnostics` (typescript via vtsls does not yet; taplo and others do). Refresh on `workspace/diagnostic/refresh`.
3. **Fallback sweep for TS/JS.** vtsls exposes tsc-project checking poorly over LSP; the pragmatic path VS Code users already use is `tsc --noEmit` as a task. Provide an opt-in per-language "project check command" (default: `tsc --noEmit` when tsconfig.json exists) that croft runs on demand ("Problems: Check Whole Project") and on a debounce after save, feeding the existing build-matcher pipeline in `src/build_matchers.rs`. This reuses shipped machinery and keeps the server out of the hot path.
[...]

PR #296, opened 2026-08-25
feat(problems): whole-project scope for the PROBLEMS panel (v0.1.770)

vitali87:
Addresses #256 (step 1 + the scope toggle; see Scope for what remains).

## What I verified first

The issue asks, as its first step, whether the panel "currently drops diagnostics for unopened URIs". **It does not.** `rebuild_problems` builds groups from the `lsp_diagnostics` / `build_diagnostics` stores keyed by path, with no requirement that a buffer be open, and `drain_lsp_diagnostics` accepts a publish for any path. So a rust-analyzer workspace push already lands. `problems_shows_diagnostics_for_a_file_that_was_never_opened` pins that, so the answer is recorded rather than re-litigated.

That makes step 1 a verification, not a fix — and it means the useful work is the scope control, which is what this PR adds.

## What this adds

- **`ProblemScope`** — `Open Files` / `Whole Project`, defaulting to Whole Project.
- **`Problems: Toggle Scope`** palette command.
- **`problems_scope` pref**, read at startup. Unrecognised values read as `whole_project`, so a typo shows *more* rather than silently hiding diagnostics.

**The filter lives where the groups are built, not in the panel.** That matters because the status-bar badge counts `problems.total_count()` — the same list the rows come from. Filtering at build time means rows and badge are one projection of one list and cannot disagree; filtering in the panel would have needed the badge to learn about scope separately, and the two would drift.

## A bug found along the way

The empty state read *"No problems have been detected in the workspace."* regardless of scope. Under Open Files that is false — unopened files may well have errors — and it would send a reader away believing there is nothing to fix. It now names the scope it actually searched, with a test asserting the scoped wording never claims the workspace is clean.

## Tests

- `problems_shows_diagnostics_for_a_file_that_was_never_opened` — the step-1 verification above.
- `problems_scope_toggle_filters_unopened_files_and_the_badge_follows` — toggling drops the unopened file and the badge count follows. **Verified to have teeth:** removing the filter makes it fail.
- `the_empty_state_names_the_scope_it_actually_searched` — both wordings.

`cargo fmt --check` clean, `clippy --bins --all-targets` clean, 24 problems tests green.

## Scope

This is one slice of a multi-strategy issue. Deliberately not here, each wanting its own change:

- **LSP 3.17 pull diagnostics** (`workspace/diagnostic` with `previousResultIds`, plus `workspace/diagnostic/refresh`) — a protocol addition with its own incrementality story.
- **The `tsc --noEmit` fallback sweep** for TS/JS, feeding the existing build-matcher pipeline.
- **basedpyright `diagnosticMode: workspace`** configuration.
- **Cost control** (`problems.project_scope` auto/on/off with a file-count threshold) — worth doing once there is a sweep whose cost needs controlling; today the scope toggle only filters diagnostics that already arrived, so it adds no scan cost.
- **Per-workspace persistence** — the pref is global today.

The corresponding acceptance criteria stay open on #256.


vitali87:

CodeRabbit's check on this PR reports `pass` but is annotated "Review rate limited", so no bot review ran. This review was produced locally by a Claude Opus 5 agent with full-repository context, anchored to head SHA `d9b0a1f4074719920502cad97c38663095aa51f0`.

**VERDICT: REQUEST_CHANGES.** Fixes in the follow-up commit; see the reply below.

## Findings

**`Cargo.toml:3` [blocker/release-gate] — the version does not bump above main; CI will fail.**

`git show origin/main:Cargo.toml` gives `0.1.770`; `git show d9b0a1f:Cargo.toml` gives **the same `0.1.770`**. The diff reads `0.1.768 → 0.1.770` only because it renders against a stale merge-base — the merge commit resolved `Cargo.toml` to main's value. `.github/workflows/ci.yml:64-67` fails when `[ "$new" = "$old" ]`. Since main advanced to `0.1.770` under #295, this PR now shares a version with a *different* binary, which CONTRIBUTING forbids outright.

**`src/app/mod.rs:7486-7506` [major/correctness] — under Open Files scope the panel goes stale on tab open and close.**

The open set is computed inside `rebuild_problems`, but every call site is triggered by a *diagnostic* event — lines 6135, 7425, 7470, 7616, 27742: LSP close-pruning, build-diagnostic ingest, build-diagnostic clear, `drain_lsp_diagnostics`, and the toggle handler. **None fires when a tab opens or closes.**

So with scope = Open Files and a quiescent server, opening a file that already has stored diagnostics shows **nothing** until unrelated diagnostic churn triggers a rebuild. The badge is stale in lockstep, so the two agree on a wrong number — the coupling argument in the PR body holds, but it guarantees consistency, not freshness.

The close direction is only partly masked: `sync_lsp` prunes a closed file's `lsp_diagnostics` and rebuilds when something dropped, but that covers only files with a recognised language, and `build_diagnostics` is not pruned there at all.

**`src/prefs.rs:137` + `src/app/mod.rs:27735-27744` [major/correctness] — `problems_scope` is read but never written.**

`grep -rn "problems_scope" src/` returns three hits: the field, the startup read, and a test name. There is **no** `save_problems_scope`, while `src/prefs.rs` has 22 sibling helpers in exactly that shape. The toggle handler sets the field, rebuilds, refreshes the badge, sets the status — and never touches prefs. The PR body says the pref is "read at startup", which is true and quietly sidesteps that it is never written; the release note advertises the toggle. A user who selects Open Files and restarts silently gets Whole Project back.

**`src/app/mod.rs:7488-7500` [minor/verified-clean] — the open-path enumeration is complete.**

Checked specifically because a miss here silently vanishes diagnostics. `editor_layout.rs:211-226` documents that `App::editor` is the hoisted active leaf and `inactive_leaf_tabs()` walks every other leaf, so the chain covers all leaves exactly once; no other `EditorTabs` collection exists. **No change needed** — recorded so a future reviewer need not re-derive it.

One residual risk: paths are compared verbatim with no canonicalisation, while the codebase canonicalises elsewhere for this comparison. If a server published a URI resolving through a symlink differently from the editor's stored path, an open file would read as unopened. No concrete reproducer found, so flagged as a risk rather than a bug.

**`src/widgets/problems.rs:110-115` [minor/ux] — `from_config`'s alias set is thin.**

`.trim().to_ascii_lowercase()` does not collapse whitespace, so a user pasting the palette label `"Open Files"` gets `"open files"`, which falls through to WholeProject with no feedback. Meanwhile `"whole_project"` — the value the doc comment tells users to write — matches only via the wildcard, so a typo is indistinguishable from the documented spelling.

**`src/widgets/problems.rs:1059-1063` [minor/test] — `empty_shows_no_problems_message` now passes for a weaker reason than it reads.** Both new messages contain "No problems", so the assertion can no longer distinguish the wordings and would pass even if the default scope flipped.

## Test coverage

The three new tests are well-constructed and the scope test has real teeth. Gaps: **no test that the pref round-trips** (would have caught the persistence blocker outright); **no test that opening a tab refreshes a scoped panel** (would have caught the staleness bug); **no test for the split-pane path**, so the `inactive_leaf_tabs()` half of the enumeration — exactly where a missed collection would hide — is never exercised.

**On the deferrals:** pull diagnostics, the `tsc` sweep, basedpyright config, and cost control are all genuinely separable, and the cost-control reasoning is right since the toggle adds no scan cost. **Per-workspace persistence is the exception**, deferred as "the pref is global today" — which presumes the pref persists at all. Given the finding above it does not persist in any scope, so that bullet describes a limitation of something not yet working.

CodeRabbit score (diff quality): 3/5
Greptile score (codebase fit): 3/5

vitali87:
All three blocking findings fixed in `0310cd5`. Each verified against the code before acting on it.

**The version collision was real and would have failed CI.** Confirmed `git show origin/main:Cargo.toml` and `git show <head>:Cargo.toml` both read `0.1.770` — the earlier merge resolved the conflict to main's value, and the diff only *looked* like a bump because it renders against a stale merge-base. Now `0.1.773`, picked to clear the other branches in flight rather than just main, since three of my own PRs are open at once.

**The staleness finding is the one I would not have found on my own**, and the reasoning in it is exactly right: I argued in the PR body that filtering at build time keeps rows and badge consistent, and that argument holds — but consistency is not freshness. Both were stale together. I traced the call sites and confirmed all five are diagnostic-triggered, so under Open Files scope opening a file with stored diagnostics genuinely showed nothing until unrelated churn happened by.

Fixed by tracking the open-buffer set on `App` and rebuilding when it changes, gated on the scope actually depending on that set. I also extracted the enumeration into `open_buffer_paths()` so the filter and the watcher read one definition — otherwise they would drift the first time someone adds an editor collection.

**Persistence:** added `save_problems_scope` following `save_render_whitespace`, the string-valued analogue. Your catch about `label()` was the important detail — it returns display text (`"Open Files"`), so writing it back would produce a value `from_config` only accepts by accident. Added a distinct `to_config` token, and a round-trip test asserting `from_config(to_config(x)) == x` for both scopes so the two can never diverge silently.

**Alias set:** `from_config` now normalises spaces and hyphens to underscores, so the pasted palette label works, and spells out the `whole_project` aliases explicitly — leaving the wildcard for genuinely unrecognised input rather than doing double duty for the documented spelling.

**Tests added**, both verified to fail against the pre-fix code:
- `opening_a_file_refreshes_a_scoped_problems_panel` — a diagnostic arrives for an unopened file, the scoped panel hides it, then opening the file brings the row in *without any new diagnostic push*.
- `problem_scope_config_tokens_round_trip` — token round-trip both ways, the pasted-label case, and that unrecognised input still resolves to the scope that shows more.

**Not fixed, deliberately:** the path-canonicalisation risk you flagged. I looked and could not construct a reproducer either — the editor's stored path and the LSP URI both come from the same tree walk in the cases I could trace. Adding canonicalisation on a hot rebuild path to defend against a hypothetical seems worse than leaving it visible, so I would rather it stay documented in your review than be papered over.

Also **not** re-tightening `empty_shows_no_problems_message`: you are right that it now passes for a weaker reason, but the new test covers both branches properly, and deleting a test whose name still describes something true felt like the wrong trade. Happy to remove it if you disagree.

fmt clean, clippy clean, 25 problems tests green.

PR #297, opened 2026-08-25
feat(mouse): bindable mouse gestures in keybindings.json (#259)

### LINKED ISSUE #259: Customizable mouse shortcuts in keybindings.json
## Motivation

Customizing what the mouse does is VS Code's second-most-upvoted open feature request ([vscode#3130](https://github.com/microsoft/vscode/issues/3130), ~1,930 votes, open since 2016): people want ctrl+click to mean go-to-definition or add-cursor per *their* preference, middle-click paste or column select, configurable wheel behavior. VS Code can't ship it because mouse handling is spread across a DOM. Croft owns its entire mouse dispatch, and already has a proven model for user remapping: `keybindings.json` rebinds any palette command and applies on save (`src/keymap.rs`).

## Current state

- Mouse handling is hardcoded per widget: Option+click carets and Shift+Option+drag column select in the editor, Cmd/Ctrl+click for `path:line` and URLs (`src/file_ref.rs`), click-to-jump in minimap/scrollbar/graph, drag on every seam.
- `keybindings.json` covers keys only; there is no mouse section.

## Proposed design

1. **Extend `keybindings.json`** with mouse entries in the same array, using a `mouse` gesture syntax in the `key` field, e.g. `{"key": "cmd+click", "command": "editor.go_to_definition", "when": "editor"}`. Gestures: `click`, `double_click`, `triple_click`, `middle_click`, `right_click`, `wheel_up`, `wheel_down`, each with the full modifier set (`cmd`, `ctrl`, `alt`, `shift`, `mod`). A `when` context (`editor`, `terminal`, `file_tree`, `tab_strip`, initially just these) scopes the binding; omitted means editor.
2. **Bindable actions.** Any palette command works, plus a small set of mouse-specific commands that need the click position as an argument: `editor.add_cursor_at_click`, `editor.go_to_definition_at_click`, `editor.column_select_drag`, `terminal.paste` (middle-click paste), `editor.open_link_at_click`. Position-carrying commands receive the resolved buffer coordinate from the dispatcher.
3. **Dispatch order and safety.** User mouse bindings resolve before built-in widget behavior, like key bindings. Plain unmodified `click`/`drag` in the editor stays reserved (rebinding primary selection would brick the UI); the parser rejects binding bare `click`/`drag` in `editor`/`terminal` contexts with a visible load error. Everything else is fair game, including reassigning the built-ins (a user can move file-ref opening from Cmd+click to Alt+click; the built-in yields when a user binding claims the gesture).
4. **Defaults file.** Ship the current hardcoded behavior expressed as a defaults table in the same format, and document it in KEYBINDINGS.md, so users can see exactly what is overridable, mirroring how key defaults are documented.
5. **Terminal caveat.** In terminal panes, mouse events may be owned by the running TUI app (mouse-mode reporting). User bindings apply only when croft owns the pointer, same rule as today's built-ins; document it.

## Acceptance criteria

- [ ] `cmd+click` remapped to go-to-definition works, and the file-ref default yields to it; removing the binding restores the default on save, no restart.
- [ ] `middle_click` → paste in terminal panes; `alt+wheel_up/down` → editor font-independent scroll-by-page (or any palette command).
- [ ] Binding bare `click` in the editor context produces a load-time error notice and is ignored.
- [ ] `when` scoping works: a `file_tree` binding does not fire in the editor.
- [ ] Defaults documented; user file wins over defaults; conflicts within the user file resolve last-wins like keys.

## References

- [vscode#3130](https://github.com/microsoft/vscode/issues/3130)

vitali87:
Closes #259.

Mouse gestures bind in the same `keybindings.json` array as keys, using a gesture name in `key` and an optional `when` region. This is [vscode#3130](https://github.com/microsoft/vscode/issues/3130), ~1,930 votes, open since 2016 — VS Code cannot ship it because mouse handling is spread across a DOM; croft owns its entire dispatch.

```jsonc
[
  { "key": "ctrl+click",   "command": "mouse_go_to_definition_at_click" },
  { "key": "alt+click",    "command": "mouse_add_cursor_at_click" },
  { "key": "middle_click", "command": "search_from_terminal", "when": "terminal" },
  { "key": "alt+wheel_up", "command": "toggle_word_wrap" }
]
```

## Two gestures the issue asks for that croft refuses

Both parse, both are reported in **OUTPUT · Keybindings**, and neither is silently accepted as a binding that misbehaves.

**`cmd+click` can never fire.** SGR mouse reporting carries no Super bit — which is exactly why croft's own Go to Definition rides Ctrl, per the comment at `mod.rs:30675`. The issue's headline example would have parsed cleanly and then done nothing forever. `mod` therefore resolves to Ctrl for mouse on every platform, unlike keys where it is Cmd on macOS.

**`triple_click` would fire on the second click.** croft's `ClickTracker` (`src/app/click.rs`) distinguishes single from double only. Reporting a third level from the dispatcher would make a `triple_click` binding disagree with the built-in select-word behaviour about the same physical click, so the parser refuses it rather than shipping a gesture that means something other than its name.

## Design

**Gestures are not `Chord`s.** A chord is a `KeyCode` plus modifiers; a gesture has no key code. Keeping them in separate tables also keeps gestures out of `Keymap::chords()`, which drives the iTerm2/Ghostty forwarder pass — a terminal forwarder for a mouse gesture is meaningless.

**Dispatch sits after the region predicates and above every built-in**, so a `when` context is known and rebinding a built-in actually takes it over. An unbound gesture falls straight through.

**Terminal mouse ownership is respected.** A TUI that asked for mouse tracking owns the pointer; `shift` bypasses, the same rule croft's built-in scroll already follows.

**Bare `click` is refused in `editor` and `terminal`** — it is how the caret is placed and text selected. Allowed in `file_tree` and `tab_strip`, where it is not.

**Position-carrying commands** (`mouse_add_cursor_at_click`, `mouse_go_to_definition_at_click`, `mouse_open_link_at_click`) read the click through a field set only for the duration of that dispatch. From the palette they say they need a mouse binding rather than guessing at the caret, so one command never means two things.

## Refused rows are visible

Previously a bad `keybindings.json` row vanished silently, which reads as croft being broken rather than the binding being wrong. Refused rows now report in OUTPUT · Keybindings at startup and, on live reload, summarise in the status bar — the same treatment the settings loader gives its warnings.

## Tests

18 keymap tests, 6 new: gesture parsing with modifiers and nonsense rejection; the cmd warning and `mod`→Ctrl resolution; bare-click refusal in both reserved contexts (and its allowance in `file_tree`); one gesture bound differently per region; every bad-row warning path; and that mouse-only rows do not make `is_empty()` lie to the key path.

Suite: 3567 passing, 0 failures.

## Summary by CodeRabbit

* **New Features**
  * Added customizable mouse gestures across the editor, terminal, file tree, and tab strip.
  * Added click-based commands for placing cursors, opening links, and navigating to definitions.
  * Added automatic sidebar hiding with a Customize Layout control.
  * Added terminal transcript capture and replay for supported sessions.

* **Bug Fixes**
  * Improved modified-click, double-click, tooltip, document-link, and terminal mouse-input behavior.

* **Documentation**
  * Documented mouse gesture syntax, precedence, limitations, and terminal behavior.


vitali87:

Round 1 on the mouse-bindings feature.

---

## Walkthrough

Adds mouse gestures to `keybindings.json`, sharing the existing array via a gesture spelling in `key` plus an optional `when` region. `Gesture`/`GestureKind`/`MouseContext` live alongside `Chord` but in a separate table keyed by `(Gesture, MouseContext)`, correctly keeping gestures out of `chords()` and the iTerm2/Ghostty forwarder.

Both headline refusals are **correct**, verified independently rather than taken at their word:

- **`cmd+click`** — crossterm 0.29 `parse_cb` (`event/sys/unix/parse.rs:776`) decodes the SGR/X10 Cb byte and can emit only SHIFT/ALT/CONTROL; there is no Super bit in the protocol. The two `SUPER` sites are in the *keyboard* CSI-u/kitty parser, which does not extend mouse reporting. `iterm2.rs`/`ghostty.rs` forward key chords only.
- **`triple_click`** — `ClickTracker` stores a single `last: Option<(Instant, u16, u16)>`; it structurally cannot count to three, and no triple-click behaviour exists anywhere in croft to reuse.

But the machinery around them has real defects.

## Findings

- **`src/app/mod.rs:33090` [blocker/correctness]** — the live-reload warning summary is dead code, immediately overwritten by an unconditional `self.status = "Keybindings reloaded (…)"`. The user never sees the warning count, so the PR's claim and #259's acceptance criterion ("bare `click` produces a load-time error notice") are both unmet. The compiler does not flag it because `String` assignment is not `#[must_use]`.

- **`src/app/mod.rs:38207` [major/correctness]** — `double_click` can never fire outside the editor. `gesture_for` always consults `editor_click`, but the tree records into `tree_click` and the terminal into `terminal_click`. In `file_tree`/`tab_strip`/`terminal` contexts `is_double` is always false, so a binding there silently degrades to never matching — while all three are documented as valid `when` regions.

- **`src/app/mod.rs:38207` [major/correctness]** — calling `is_double` without `record`/`clear` lets a fired `double_click` re-fire on the third click. Every built-in pairs them so the count resets; the dispatch returns on a match, leaving `last` holding the first click's timestamp.

- **`src/app/mod.rs:30333` [major/correctness]** — the terminal-ownership check reads the ACTIVE pane (`self.terminal()`), not the clicked one. Every built-in indexes `terminals[terminal_hit]`. With splits, clicking a non-active pane consults the wrong terminal's `mouse_reporting()`.

- **`src/keymap.rs:392` [major/docs-vs-behaviour]** — `triple_click` returns `None`, falls through to `Chord::parse`, and reports "not a key chord or mouse gesture" — implying a typo. The PR body and docs both promise the click-tracker explanation.

- **`src/keymap.rs:518` [minor/correctness]** — `cmd+click` is warned about *and* inserted as a dead entry. `gesture_for` masks Super off so it can never be looked up, but it makes `has_mouse_bindings()` true, and means `cmd+click` and `click` are distinct keys so it dodges the reserved bare-click check.

- **`src/app/mod.rs:27748` [minor/design]** — `mouse_open_link_at_click` cannot work in its own default context: `terminal_url_click`/`terminal_file_click` bounds-check against the terminal's `last_inner`, so an editor click always returns "No link there".

- **`src/app/mod.rs:38158` [minor/docs]** — a doc comment belonging to `is_rebindable_chord` is now orphaned onto `mouse_context_for`.

## Test coverage

The 6 keymap tests are honest and each would fail if reverted. The gap is that **no app-level test drives `handle_mouse` with a loaded keymap** — every finding above lives in `mod.rs` and none is reachable from a parse-level test.

## Scores

- **CodeRabbit (diff quality): 2/5** — one blocker plus four majors.
- **Greptile (codebase fit): 3/5** — the architecture is good and both refusals are correctly reasoned, but the dispatch does not follow local conventions: built-ins index `terminals[terminal_hit]` and always pair `is_double` with `record`/`clear`; this code did neither.

VERDICT: REQUEST_CHANGES

---

**All addressed in `9133a2d`**, with three app-level tests closing the layer the reviewer identified. The `mouse_open_link_at_click` scoping and the orphaned doc comment are the two remaining items; the former needs a decision about whether to route editor clicks through the editor's own file-ref path rather than the terminal's, which I would rather do as its own change than fold in here.

vitali87:
## 🤖 Fallback review — Claude Opus 5 (round 6)

CodeRabbit is rate-limited on this PR, so this is the recorded review artifact.

**VERDICT: REQUEST_CHANGES** — one major, now fixed in `9fd5aa0`.

### The finding

`src/app/mod.rs` **[major/correctness]** — `Cmd::MouseOpenLinkAtClick` resolved the link against the *active* terminal rather than the *clicked* one.

`terminal_url_click`/`terminal_file_click` reach the grid through `self.terminal()`, which indexes `active_terminal`. The built-in Ctrl+click path sets `active_terminal = idx` **before** calling them; the new binding dispatch early-returns above that assignment, so a bound `mouse_open_link_at_click` read whichever pane happened to be focused.

Verified by probe, not inferred: two split panes, a URL printed only into pane 0 while pane 1 is active. Ctrl+clicking directly on that URL through a binding gave `status == "No link there"`; the identical click with no keymap loaded (built-in path) gave `"Opened https://example.com/zero"`. The bindable gesture was strictly worse than the built-in it is meant to be able to replace, and it failed over a link the user could plainly see.

This is the same bug class the diff's own comment already identifies and fixes for `child_owns_pointer` ("Index the pane that was CLICKED, not the active one") — applied there, missed two screens below.

### The fix

Extracted the built-in's pane-switch teardown as `activate_terminal_pane(idx)` and called it from **both** sites. A bare `active_terminal = idx` would have fixed the lookup while leaving copy mode, quick-select and the find bar bound to the pane being left — that teardown is why the helper exists rather than an assignment.

New test `a_bound_gesture_resolves_the_link_in_the_pane_it_clicked_not_the_active_one`, **verified revert-sensitive**: reverting only the dispatch-site call (leaving the helper and built-in caller intact) makes it fail. It also asserts the clicked cell actually holds the URL before clicking — an empty cell also yields "No link there", which would have passed the test for entirely the wrong reason. That guard is not hypothetical: my first attempt clicked the pane border and failed for exactly that reason.

### Also addressed

- **Wheel dispatch coverage** (flagged as parser-tested only): added `a_bound_wheel_gesture_fires_and_a_modified_wheel_does_not_match_the_bare_row`. Not a defect — the reviewer confirmed wheels work — but it is the one gesture family where a regression is invisible, since a wheel falling through to the built-in scroll looks like ordinary scrolling.
- **Nit: exhaustive tracker match.** All four `MouseContext` arms spelled out, so a new variant cannot silently default into `editor_click`.
- **Nit: doc gesture count.** `docs/KEYBINDINGS.md` said six gestures while the loader comment said seven; now "six bindable" plus `triple_click`, recognised and refused.

### Round-5 items re-verified as genuinely fixed

Both confirmed by reverting them, not by reading:
- The `cfg!(test)` keymap guard is load-bearing — under a hostile `XDG_CONFIG_HOME`, both named tests fail without it.
- The `ctx`-derived `osk_auto` raise is exactly equivalent to the built-in's disjunction (`TabStrip` is geometrically disjoint from the editor's inner area, and `Terminal`/`Editor` both map to `true`, so precedence is irrelevant).

### Gates

- **Full suite: 3602 passed, 0 failed** after merging main. `cargo fmt --check` clean, clippy clean.
- Merged `origin/main` (which had landed the same `cfg!(test)` keymap guard independently — kept this branch's hoisted form, which also surfaces parse warnings, and carried main's comment across).
- **Version: 0.1.780.** Main moved 0.1.773 → 0.1.778 during review, and open PR #303 already claims 0.1.779 — verified via `gh pr diff`, not assumed. Release notes replaced, not appended.

Scores after the fix: **CodeRabbit 5/5**, **Greptile 5/5** — the single major is resolved and the convention it violated is now shared code rather than a duplicated block.

vitali87:
Round 7 addressed — all five findings, pushed as 255e5e8. Taking them in order, including two that turned out to already be fixed in the working tree.

**Modified double-clicks may never fire (major) — real, fixed.** This was the substantive one. `gesture_for` decides Click vs DoubleClick by asking a tracker whether it just saw a click at that spot, and modified clicks were recorded only *after* a binding matched. The first click of a modified pair matches nothing by construction, so the tracker stayed empty and the second click could never classify as a double.

Worth recording that the obvious fix is wrong: hoisting the recording above the lookup makes `gesture_for` run twice, and the second call reads the tracker the first one just wrote — so a single click arms itself and comes back a double. I shipped that, re-ran, and got the identical assertion failure. The dispatch now classifies once and the recording reuses that gesture rather than deriving a fresh one.

**Dispatch never moves keyboard focus (major) — real, fixed.** Correct diagnosis: the pane became active without being focused, so commands read the right grid while the next keystroke went to the pane the user had clicked away from. `focus_pane` now runs for bound gestures as it does for every built-in handler.

**Documented example carries non-existent command ids (minor) — real, fixed.** `paste` and `page_up` were both skipped by `Keymap::from_json` with a warning, so anyone copying the snippet got a keymap that silently did less than it said. The test now loads the real `docs/KEYBINDINGS.md` via `include_str!` rather than a copy, so the example cannot drift from the parser again.

**`MouseAddCursorAtClick` / `MouseGoToDefinitionAtClick` skip the non-text-view guard (major) — already fixed.** Both arms carry `has_non_text_view()` today, each with its own status message ("No cursors in this view" / "No definitions in this view"). Your reasoning about `buffer_pos_at` was right, though — it guards only on `last_inner.height == 0` and `lines.is_empty()`, so it would not have refused a diff or sheet tab on its own.

**`file_tree` context covers the whole sidebar (major) — already fixed.** `mouse_context_for` now requires `in_tree && app.sidebar_view == SidebarView::Explorer`. The scope question you raised is the right one and `docs/KEYBINDINGS.md` answers it the way you guessed: `file_tree` means the Explorer tree, not the sidebar region.

Those last two appear to have been reviewed against an older commit — flagging it because a finding that reads as live but is already fixed costs the same verification time as a real one.

**Verification.** All three tests pass, clippy clean at `exit=0` with no anchored diagnostics. Each new test was also revert-tested — the fix made inert, the test confirmed to FAIL, then restored:

- re-derive the gesture instead of reusing it → `a_modified_double_click_can_fire_at_all` fails
- record only on a matched binding (the original bug) → same test fails
- drop the `focus_pane` call → `a_bound_gesture_moves_keyboard_focus_to_the_pane_it_clicked` fails
- bogus command id inside the documented jsonc block → `every_mouse_binding_the_docs_show_actually_loads` fails

The fourth case is worth a note: my first attempt corrupted a `ctrl+click` in a prose table and the test passed, which looked like a hollow test. It was not — the test reads only the jsonc block, so the table is not its corpus and the pass was correct. Re-aimed inside the block, it has teeth.

vitali87:
Fixed in `42a794a` rather than dismissed — the doc/body mismatch was real.

`is_double_click_prefix` now also checks that the single click is unbound, so the code honours the contract its own comment states. The sole caller today matches `Some((ctx, gesture, None))` — the `None` being `command_for_mouse` finding no command for this click — so it could not reach the bad case. But the function is `pub`, and a contract enforced only by caller discipline two files away is one the next caller does not know about.

Pinned by `a_bound_single_click_is_not_a_double_click_prefix`, which asserts both directions with preconditions on the binding state, since "not a prefix" is trivially true when no double is bound either. RED at the claim before the fix, GREEN after, `running 1 test` confirmed on both.

vitali87:
## Both majors fixed at `ee13b25`, each Red/Green

The REQUEST_CHANGES review described `42a794a`. That SHA is no longer head, so **that verdict is now neither approval nor objection** — it describes code that no longer exists. Requesting a fresh review at `ee13b25` below.

### Major 1 — reserved gesture

`is_reserved_in` reserved bare `click` because, per its own doc and the refusal table, it is how "the caret is placed **and text is selected**". Text selection is *double*-click, which was freely bindable. `select_word_at` has exactly two production call sites and both sit below the matched dispatch's early return, so binding `double_click` removed word-selection outright.

RED, watched, at the named claim:
```
panicked at src/keymap.rs:684:
Editor: a bare double-click is the only way to select a word, so it is
reserved for the same reason a bare click is
```

The refusal **message** carried the same slip — it said "a bare click … caret or select text" for a refused double-click — so it now names the gesture actually refused and the one thing it does. That string is user-facing output, and nothing compiles it against the code it describes.

A guard test (`a_modified_double_click_is_still_bindable`) stays green so "reserve double-click" cannot be satisfied by refusing *every* double-click.

### Major 2 — `child_owns_pointer`

Verified independently before acting. `report_mouse` has three production call sites:

| site | constructs |
|---|---|
| `terminal.rs:2898` | `WheelUp` / `WheelDown` |
| `app/mod.rs:32975` | `WheelDown` |
| `app/mod.rs:33040` | `WheelUp` |

**No click-forwarding path exists.** Declining a bound click handed the gesture to nobody — and the built-in `Down(Left)` arm ran anyway, since unlike click-to-move-cursor (which gates on `!mouse_reporting()`) it never consults it.

RED printed the offending value rather than asserting a bare bool:
```
panicked at src/app/tests.rs:2629:
a bound ctrl+click over a tracking child must RUN … left: false, right: false
```

**Restricting it to wheel gestures exposed a second dependency.** The guard was also the swallow branch's only escape hatch, so wheel-only left the swallow always firing and killed Ctrl+click link-opening over every full-screen TUI. The GREEN caught that as a failing test.

The swallow now defers on its **own** reason — whether the built-in would open something — via a pure predicate covering **both** arms of the built-in's `||`, URL *and* file reference. Checking only the URL half would have lost the file jump: the same defect one branch over, in the fix for it.

### Minors

- `mouse_add_cursor_at_click` pokes the cursor on success like the built-in, so a new caret cannot land mid-blink-off
- two test comments that misdescribed which branch they covered are corrected
- KEYBINDINGS.md gains a `double_click` refusal row, and each row's reason now matches exactly what its predicate covers
- **not** fixing silent duplicate-binding overwrite → filed as #315: there is no duplicate-warning precedent anywhere in `keymap.rs`, including the key path, so it is a file-wide design change rather than a defect in this diff

### One test committed `#[ignore]`d, deliberately

`an_unmatched_modified_click_does_not_arm_the_tracker` reproduces **#317**: the built-in terminal click path records without a modifier guard, so an unmatched ctrl+click arms the double-click tracker and the user's next ordinary click selects a word they never gestured for. Guarding that changes built-in selection for every user — out of scope here. The reproduction stays in the tree so it does not rot; un-ignore it as the red when fixing #317.

### Verification

```
7/7 targeted tests GREEN — each with  and 
                            confirmed, and the fix verified present at RUN time
FMT_CHECK=0    CLIPPY=0
SUITE: 3621 passed; 1 failed
```

The one failure is `saving_a_windows_1252_file_records_a_byte_exact_snapshot`, discharged as #307-class with three controls: zero reachability in this diff; **byte-identical to `origin/main`** (`git diff origin/main HEAD` touches it in zero lines, so this diff cannot produce a failure in code it does not differ from); and the failure signature is `left: []` — empty where content was expected, the write-then-read visibility mechanism.

@coderabbitai review

vitali87:
**Correction to my comment above** — three code spans were eaten by shell expansion when I posted it, and they were the load-bearing ones. The verification section should read:

```
7/7 targeted tests GREEN -- each with `running 1 test` and `1 passed; 0 failed`
                            confirmed, and the fix verified present at RUN time
FMT_CHECK=0    CLIPPY=0
SUITE: 3621 passed; 1 failed
```

The selection and result-shape checks are the point, so they are worth restating rather than leaving as "each with  and":

- **`running 1 test`, as an equality** — not a presence check. `test result: ok` is honest about a smaller thing than it appears to certify: it says the harness ran and nothing it *selected* failed, not that anything **was** selected. A filter matching nothing produces `ok` with exit 0, defeating both the artifact check and the exit code. The equality form also catches over-selection, where some *other* test passing reads as success.
- **`ok. 1 passed; 0 failed`, not merely `ok`** — an `#[ignore]`d test satisfies "selected exactly 1", reports `ok`, and asserts nothing.
- **fix present at RUN time**, re-checked immediately before each test rather than once at script start, because a ten-minute build sits between the two and a check that was true when it printed is not a check on what actually ran.

Both gates were exercised against synthetic artifacts (green, ignored, zero-match, over-select, red, compile-error) before being trusted, rather than only reasoned about.

vitali87:
Fixed in `28977b5` — the finding is correct and it is the same class this PR already exists to fix.

Line 130 read:

> Bindings apply while any pane except the terminal is focused, and only to chords that carry a modifier or are function keys…

True when only keyboard chords existed. This PR falsified it in two ways at once: mouse rows reach the terminal, and a bare gesture is allowed in `file_tree` / `tab_strip`. A user reading top to bottom would conclude the mouse rows this PR adds are prohibited — and nothing compiles prose against the code it describes, so neither the suite nor clippy could have caught it.

That is precisely the shape of both majors already fixed here: **prose whose scope no longer matches its predicate.** Major 1 was a reservation justified by two cases while the code covered one; major 2 was a guard citing the wheel path as precedent where the precedent did not hold. This is the third instance, in the documentation rather than the code, introduced by the change itself.

The sentence is now explicitly scoped to KEYBOARD chords and links to the mouse section rather than leaving two paragraphs to disagree. I checked the anchor resolves to a heading that actually exists (`## Mouse bindings in keybindings.json`, line 583) and sits below line 130 as the text claims — a cross-reference to a nonexistent section would be its own defect.

No version bump: `docs/` is outside the release gate's shipped set (`^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)`), and this commit touches only documentation.

vitali87:
Fixed in `e2340aa`, Red/Green. The finding is correct and I verified the mechanism against source before acting.

**The discrepancy is real.** `self.terminal()` is `terminals[self.active_terminal]` (mod.rs:11903). The built-in this guard defers to calls `activate_terminal_pane(idx)` at :32596 and only *then* looks for a link at :32611 — so it reads the **clicked** pane. The guard runs earlier, at :30846, before any activation. With one pane those coincide, which is why every test so far passed; with two, a Ctrl+click on a link in the unfocused pane made the guard inspect the focused pane's content, find nothing, swallow the prefix click, and the link never opened.

Now takes the index from `terminal_hit` — already computed in `handle_mouse` — and returns false for an out-of-range index rather than indexing blindly.

**RED against the reverted predicate**, failing at its named claim and printing the offending value:

```
PHASE1  EXIT=101  FAILED. 0 passed; 1 failed
  the prefix click must defer to the built-in, which opens the link in the
  CLICKED pane. Reading the ACTIVE pane finds nothing there, swallows the
  click, and the link never opens. status was "Ready"
PHASE2  EXIT=0    ok. 1 passed; 0 failed
```

The mutation was verified applied before scoring and the tree verified restored after. The test asserts both that pane 0 carries the link and that pane 1 is active, so the clicked-vs-active distinction is reachable rather than assumed — without those preconditions it would pass vacuously on a single-pane layout.

**Full verification at this head:**

```
SUITE: ok. 3623 passed; 0 failed; 11 ignored
FMT_CHECK=0    CLIPPY=0
```

Zero failures — the first fully clean suite on this branch. The clipboard and encoding tests that failed in earlier runs all passed here, which is corroboration that they were load-sensitive rather than caused by this diff.

---

Worth noting what this was: **the third instance of one class in this PR, living inside the fix for the second.** Major 1 was prose covering two cases where the predicate covered one. Major 2 cited the wheel path as precedent where the precedent did not hold. This read the wrong pane. Checking both arms of the built-in's `||` was necessary and not sufficient — *which* terminal is a second axis, and having verified the first axis carefully was exactly what stopped me looking for a second.

vitali87:
## All checks green at `e2340aa` — and that is not sufficient to merge

State right now:

```
mergeStateStatus: CLEAN
all 7 checks:     pass
review threads:   7 total, 0 unresolved
```

Every aggregate signal says go. **None of them says this head was reviewed**, and it was not:

| review | names SHA | still head? |
|---|---|---|
| coderabbitai | `ace6d23` | no |
| coderabbitai | `19ea178` | no |
| coderabbitai | `ee13b25` | no |
| — | `e2340aa` | **nothing names it** |

`git diff ee13b25..e2340aa --stat` is **79 insertions across 3 files**, including production logic in `src/app/mod.rs` — the split-pane predicate fix and its test. That is far past the `Cargo.toml`/`Cargo.lock` re-bump exemption under which a prior review could reasonably survive.

Worth spelling out why `CLEAN` does not help here, because it is the most convincing wrong signal on the board:

- **`CLEAN`** reports that no branch-protection rule is blocking. This repo has **no rule requiring a review**, so the state never consults review status at all.
- **`CodeRabbit: pass`** in the checks table is annotated *"Review rate limited"* — a green row meaning **no review ran**. The rate-limit notice on this PR is 10,076 characters, larger than either genuine review (3,245 and 4,626), so "longest bot comment" picks the non-review.
- **0 unresolved threads** reads identically whether the reviewer found nothing or never looked at this code.

Each is honest about something smaller than it appears to certify. A stale review and a missing review have the same consequence, so I am treating the artifact set at this head as empty and **not merging**.

For the record, the last three commits and what each fixed:

- `ee13b25` — both review majors (reserved bare double-click; stopped declining clicks for a child with no click-forwarding path)
- `28977b5` — docs contradiction the review caught: line 130's chord restrictions predate mouse bindings
- `e2340aa` — the split-pane predicate bug, Red/Green, found by the review of `ee13b25`

Local verification at this head: **suite 3623 passed / 0 failed**, `FMT_CHECK=0`, `CLIPPY=0`.

@coderabbitai review

vitali87:
Taking this over after an ownership check - no session holds it (15 of 19 peers offline; the two live ones confirmed they hold nothing on croft) and there has been no commit since 2026-08-27 12:29. Branch preserved: rebased on `main`, never force-pushed over its history.

**First, the six review findings.** All were resolved without a reply, which usually means dismissed rather than fixed, so I checked each against the code rather than trusting the resolution. All six were genuinely fixed:

- `MouseAddCursorAtClick` / `MouseGoToDefinitionAtClick` now check `has_non_text_view()`;
- the dispatch calls `focus_pane` per context;
- `mouse_context_for` narrows `FileTree` to the Explorer view, so a tree binding no longer fires over Search results or the Remote list;
- `terminal_would_open_something_at` takes the clicked pane's `idx` instead of reading the active one;
- modified double-clicks fire, via the separate `ModifiedClickTracker`;
- and the docs/keymap notes were addressed.

Recording that here so the silence is not mistaken for dismissal later.

**The keymap merge was semantic, not textual.** This branch grew mouse bindings with their own warning path; `main` grew #315's `resolve()` returning warnings for every way a row can fail to bind. Merged rather than picked, so one function now covers all of it - the file failing to parse, an unknown command id, a spelling that is neither chord nor gesture, a gesture reserved in its context, and a duplicate on **either** surface. That last one is the convention #315 said this branch should adopt when it landed, and it now does.

The stored `warnings` field is gone: returning them keeps the reporting testable without touching a global, and the one caller that needs them itself - the reload path, for its status summary - gets `load_with_warnings`.

**One merge decision worth flagging, because it is not mechanical.** #317 landed while this sat, making a modified click refuse to arm the built-in tracker. This branch's dispatch records into that same tracker after a binding fires, and the call now passes the event's modifiers rather than a literal. That preserves #317's invariant by construction: a `ctrl+click` binding cannot arm the plain double-click the user never made, which is precisely what the separate `ModifiedClickTracker` exists to serve. Hard-coding `NONE` there would have re-opened the hole this branch was careful to avoid.

**And the parked test now runs.** `an_unmatched_modified_click_does_not_arm_the_tracker` was committed here with `#[ignore]` as the reproduction for #317, to be un-ignored by whoever fixed it. That fix has landed, so it is un-ignored, and it passes.

Suite green at 3681, clippy and fmt clean.

PR #298, opened 2026-08-25
fix(session): reopen terminal panes after a self-update (v0.1.772)

### LINKED ISSUE #249: New croft install opens all old croft terminals empty
- it should eaither open on e empty terminal or replicate the old croft working enviroment fully, i.e. openning all open terminals with their applications in them EXACTLY as they are on old

vitali87:
Closes #249.

## What the bug actually is

The report says a self-update relaunches with "all old croft terminals empty". My first read was that the panes were *gone*, and I built a second restore path in the re-exec handoff to bring them back. **That was wrong, and the review on this PR caught it**: croft already restores panes via a per-workspace store (`save_terminal_session` / `restore_terminal_session`), which runs *before* the handoff at startup — so my addition appended a second copy of every pane. Two panes became three.

That work is reverted. Reading the existing mechanism properly showed the real defect:

`PaneRecord` stored only `cwd` and `name`. So panes **did** come back, in the right directories, with their names — and with a completely blank grid. That is precisely "opens all old terminals **empty**": the complaint is about missing *content*, not missing panes.

## The fix

`PaneRecord` gains a `transcript: Vec<String>` — the pane's visible output at save time, replayed above the fresh prompt on restore.

**Plain text, not the raw byte stream.** A byte log would faithfully re-run whatever escape sequences the original output contained when replayed — a screen clear, a cursor jump, a title change. Storing rendered text makes that impossible by construction, and `replay_transcript` additionally strips control characters, because a transcript read back off disk is not something to trust blindly. `a_replayed_transcript_cannot_execute_control_sequences` feeds a hostile `before\x1b[2J\x1b[Hafter` line and asserts the text survives while no escape byte reaches the grid.

**Bounded at 200 lines**, trailing blank rows trimmed — a mostly-empty screen would otherwise restore as a wall of nothing.

**The processes themselves cannot come back.** `execve` replaces croft's image and the child shells are orphaned, so "replicate the old environment exactly" is not reachable through this path at all; the restored shell is new. What the transcript buys is seeing what you were doing.

## A design decision worth flagging

Transcripts deliberately do **not** make a session record non-trivial.

croft prunes the record for a plain single-shell workspace so the store never grows for every directory a shell was ever run in. I initially made a transcript-bearing pane non-trivial — which fixed the single-pane case and broke that invariant for everyone, including throwaway `croft /tmp/scratch` sessions. An existing test caught it (`"a trivial session (one unnamed pane at the root) is pruned"`), and rather than retire a deliberate invariant I asked and kept it.

So: a real panel — two or more panes, or a pane you bothered to name — restores its output. A lone default pane still stores nothing. The `is_trivial` doc comment records why, so it is not "fixed" back later.

## Tests

- `a_restored_terminal_pane_shows_the_output_it_had_before` — output written to a pane survives a save/restore cycle and repaints. Uses a real two-pane panel, matching the supported case above.
- `a_replayed_transcript_cannot_execute_control_sequences` — the security property.
- `the_handoff_does_not_carry_terminal_panes` — pins the reverted mistake: serialising `SessionState` must not emit a `terminals` key, so nobody re-adds the duplicate path.
- `terminal_session_restores_pane_layout_names_and_focus_across_restarts` (pre-existing) still passes, including its pruning assertion.

## Note on the earlier review

The blocker, the `shell_cwd()`-vs-`local_shell_cwd()` remote-path bug, the pane-name loss, and the unsynced index-keyed state were all findings against the handoff path that no longer exists — reverting removed them at the root rather than patching each. The one surviving lesson is in the new test above, which pins the absence.


vitali87:
The blocker was right, and the correct response turned out to be deleting the code rather than fixing it. Addressed in `a3e9d34`.

**Verified the duplication before acting.** `restore_terminal_session` at `src/app/mod.rs:39859` sets `self.terminals = terms` — the *full* panel — and `apply_session_state` runs at 39876, so on a re-exec both fire and my `restore_terminal_panes` replaced `[0]` then pushed the rest. N panes became 2N−1, exactly as you described.

You were also right that the test I credited could not have caught it: `terminal_session_restores_pane_layout_names_and_focus_across_restarts` never calls `apply_session_state`. I should not have claimed it did.

**Digging further made the whole approach untenable, not just buggy.** `save_terminal_session` runs at 39978 — *before* the re-exec block — so the per-workspace store is already written with correct cwds and names when the exec happens. Every capability I added existed there and was better implemented: it uses `local_shell_cwd()` (your remote-path finding — my `shell_cwd()` would have resolved an ssh'd pane's remote path against the local filesystem), it carries names, and it has a corrupt-store guard. So the handoff panes were pure duplication with worse semantics.

Reverted entirely. `the_handoff_does_not_carry_terminal_panes` now asserts that serialising `SessionState` emits no `terminals` key, so the path is not reinvented, and the `SessionState` doc comment explains why their absence is deliberate. That removes findings 1–4 at the root rather than patching each.

**The real bug, found by reading the mechanism I should have read first:** `PaneRecord` stored only `cwd` and `name`, so panes returned correctly placed and named and completely blank. That is what the report means by "opens all old terminals **empty**" — missing content, not missing panes. `PaneRecord` now carries a bounded transcript, replayed above the fresh prompt.

**One design decision I escalated rather than deciding alone.** My first transcript implementation made a transcript-bearing pane non-trivial, so scrollback returned for single-pane workspaces too. That broke a pre-existing assertion — *"a trivial session (one unnamed pane at the root) is pruned"* — which is a deliberate invariant: the store must not grow for every directory a shell was ever run in, including throwaway ones. Rather than retire someone else's invariant to make my feature look better, I asked, and kept it. A real panel (2+ panes, or a named pane) restores its output; a lone default pane still stores nothing, and `is_trivial` now documents why so it is not "fixed" back.

**On the transcript being plain text:** this is the same instinct as your `shell_cwd` finding — untrusted data resolved in a trusting context. A raw byte log would re-run whatever escape sequences the original output contained when replayed. `a_replayed_transcript_cannot_execute_control_sequences` feeds `before\x1b[2J\x1b[Hafter` and asserts the text survives while no escape byte reaches the grid.

fmt clean, clippy clean, 120 session tests green including the pre-existing pruning assertion.

vitali87:

CodeRabbit reported **"Review limit reached"** on this PR, so its green check is a no-op and no bot review ran. This review was produced locally by a Claude Opus 5 agent with full-repository context, anchored to `cc624a1`.

**VERDICT: REQUEST_CHANGES** — two majors, both now fixed in `3fa46a8`. Details in the reply below.

---

## Walkthrough

Fixes #249, where terminal panes returned after a self-update in the right directories with the right names but with blank grids, because `PaneRecord` persisted only `cwd` and `name`. Adds a bounded per-pane `transcript` (last `TRANSCRIPT_LINES = 200` rows), replayed into the restored pane with control characters stripped. An earlier approach that carried panes through `SessionState` was reverted as duplicating `restore_terminal_session`; a test now pins that the handoff struct serialises without a `terminals` key.

Verified against the tree: `restore_terminal_session` does run before `apply_session_state` at startup, so the anti-duplication reasoning holds, and the `is_trivial` decision is deliberate and documented.

| File | Summary |
| --- | --- |
| `Cargo.toml` / `Cargo.lock` | 0.1.773 → 0.1.776 |
| `src/release_notes.rs` | `RELEASE_NOTES` **replaced** (not appended) |
| `src/terminal_session.rs` | `PaneRecord::transcript` with `serde(default, skip_serializing_if)`; `TRANSCRIPT_LINES` |
| `src/widgets/terminal.rs` | `replay_transcript` — strips control chars, feeds a fresh `Processor` |
| `src/app/mod.rs` | capture in `save_terminal_session`; replay in `restore_terminal_session` |
| `src/session_state.rs` | doc rewritten; `the_handoff_does_not_carry_terminal_panes` |

Checked and **not** reported, because the tree disproves them: `char::is_control()` does cover the C1 range, so a C1-introduced CSI cannot survive the filter; grid cells hold already-parsed characters, so escapes cannot normally reach the transcript; neither struct uses `deny_unknown_fields`, so old and new stores interoperate both ways; and the store is bounded to ~16–39 KiB per pane.

## Findings

- **`src/app/mod.rs:11977`** **[major/correctness]** — `replay_transcript` races the reader thread, so restored output can interleave with, or land after, the new shell's prompt. `PtyTerminal::new` → `spawn_with` spawns the child and starts the reader thread, which takes the *same* `FairMutex<Term>` that `replay_transcript` locks; nothing gates it. By the time the UI thread reaches the replay, an interactive shell has often already emitted its first prompt, so the transcript is painted below or through it — contradicting both doc comments and the release note. The window widens under load, which also makes `a_restored_terminal_pane_shows_the_output_it_had_before` intermittent rather than wrong. Fix: paint the transcript into the `Term` inside the spawn, before the reader thread starts.

- **`src/app/mod.rs:11929-11937`** **[major/correctness]** — the transcript is captured with no alt-screen guard, so saving while a full-screen TUI runs persists that TUI's rendered frame as scrollback. `grid_lines()` reads `topmost_line()..screen_lines()` unconditionally; it is the one grid reader in that file with no `ALT_SCREEN` check, while eleven neighbours guard it deliberately. Quit croft with vim, htop, or a pager open and the store keeps that program's screen, replayed next launch as inert text above a shell prompt. `save_terminal_session` runs at quit, so this is a normal path. Fix: return an empty transcript when `alt_screen()`.

- **`src/widgets/terminal.rs:2667-2668`** **[minor/correctness]** — the control-character filter also deletes TAB, so tab-aligned output silently loses its column structure on replay. Also, rows came from `row_text_and_cols`, which skips wide-char spacer cells, so CJK/emoji rows are stored at character count rather than display width and rewrap differently. Neither corrupts state; both make the replayed pane not match what the user saw.

- **`src/app/tests.rs:30642-30643`** **[minor/tests]** — the comment justifying `resize(80, 24)` is wrong: it claims "a pane that has never been rendered has a zero-column grid", but `spawn_with` constructs every `Term` at 80×24, so `columns()` is 80 from birth and the `columns() == 0` early-return is never taken. The test is still meaningful; the stated reason is not the real one.

## Test coverage

- **Alt-screen capture** — untested. Enter the alternate screen, print a marker, `save_terminal_session`, assert the marker is absent from the store.
- **Prompt ordering** — asserted only in prose. The existing test checks the marker is *somewhere* on screen, which passes whether the transcript landed above the prompt, below it, or interleaved, so it cannot detect the race.
- **`TRANSCRIPT_LINES` truncation** — untested; an inverted slice would ship silently. Feed 250 numbered lines, assert 250 is kept and 1 is not.
- **Backward compatibility** — `serde(default)` is the mechanism and nothing pins it. Parse a `PaneRecord` JSON with no `transcript` key.
- **`replay_transcript`'s empty-input early return** — uncovered.

## Nitpicks

- `src/terminal_session.rs:39` — `TRANSCRIPT_LINES` documented as "One screenful of context", but 200 is roughly eight 24-row screenfuls.
- `src/terminal_session.rs:1-12` and `src/app/mod.rs:11898-11900` — module and function docs still enumerate the persisted fields without transcripts, now the headline behaviour.
- `src/app/mod.rs:11929-11937` — the trailing-blank trim duplicates identical logic in `open_scrollback_in_editor`.

**VERDICT: REQUEST_CHANGES**

vitali87:
Both majors were real and are fixed in `3fa46a8`. I verified each against the tree before changing anything.

**The race.** Confirmed: `PtyTerminal::new` spawns the child and starts the reader thread, and only then does `restore_terminal_session` call `replay_transcript` — both writing the same `Term` behind the same lock. The fix moves the paint inside the spawn, into the window between `Term::new` and the reader thread's creation, where nothing else can be writing yet:

```rust
let mut term = Term::new(cfg, &term_size, listener);
// Before the reader thread exists, so nothing the child writes can
// land ahead of the restored output (#249).
if !preamble.is_empty() {
    Processor::<StdSyncHandler>::new().advance(&mut term, preamble);
}
```

`new_with_transcript` is the entry point; `spawn_with` keeps its signature and forwards an empty preamble, so no other caller changes. `replay_transcript` stays (it is what the control-sequence test drives) and now shares the sanitiser.

**The alt screen.** Confirmed too, and the helper is `alt_screen()` rather than `is_alt_screen()`. Guarded at the capture site rather than inside `grid_lines`, because other callers — find, copy-mode, scrollback-to-editor — legitimately want the alternate screen's contents; it is only *persistence* that must not.

**TAB.** Mapped to a space rather than deleted. Deleting it collapsed tab-aligned columns, which is a silent corruption of exactly the output people most want back. I did not attempt the wide-character width issue: fixing it properly means storing display width alongside the text, which is a bigger change than this PR should carry, and the failure mode is cosmetic rewrapping rather than lost content.

**The wrong comment.** Removed. A pane is born 80×24, so the `columns() == 0` branch is never taken here; the `resize` is pinning an explicit geometry and the comment now says that.

**Tests.** Four added, aimed at the gaps you named:
- `restored_output_is_painted_above_the_new_prompt` — asserts relative row order, which the old "marker is somewhere on screen" assertion could not distinguish.
- `an_alt_screen_frame_is_not_persisted_as_scrollback` — enters the alternate screen, writes a marker, saves, asserts the marker never reaches the store.
- `the_transcript_window_keeps_the_newest_lines` — feeds `TRANSCRIPT_LINES + 50` lines and pins the direction, since an inverted slice is still bounded and still looks right in a small test.
- `a_pane_record_without_a_transcript_still_parses` — pins `serde(default)`.

fmt clean, clippy clean at `-D warnings`, 120 session tests and 248 terminal tests green.

One note on the suite: a full parallel run intermittently fails a handful of terminal/clipboard tests with "shell is not alive" / "marker output never arrived". I baselined that against untouched `origin/main` and it reproduces there too, so it is pre-existing load sensitivity around real shell spawns, not this change. Everything passes serially or on an idle machine.

PR #300, opened 2026-08-25
docs: claim the issue AND the PR, not one or the other

vitali87:
## What

The `claimed` convention said to **move** the claim to the PR once one exists. This changes it to label **both** the issue and the PR, and to release both at merge.

## Why

The old rule optimised for where contention happens — pushes, review rounds, the merge — which is genuinely the PR. But that is not where sessions *look*. Work is discovered by scanning `gh issue list`, and an issue whose claim moved to a PR reads as unclaimed to exactly the reader the label exists to stop.

Two real cases in one session:

- **#249** carried no `claimed` label the entire time a session was working it, because its claim had moved to PR #298. Anyone scanning open issues would have seen `#249 enhancement` and started on it.
- **#260** came within one step of shipping twice: a second session built the whole auto-hide feature in a separate worktree before noticing PR #294 already existed. It caught the collision only after the work was done, and dropped its branch.

Labelling both costs one extra `gh issue edit`. Leaving the issue bare costs somebody their afternoon.

## Scope

Docs only — no `src/`, so the release-hygiene gate does not apply and there is no version bump or release note. Verified against the path filter in `.github/workflows/ci.yml:51-52`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,35 +18,39 @@ anyway, ask the holder first and the user if you cannot reach them.
 
 **When you start work, add the label. When you stop, remove it.**
 
+Where to put it: **label the issue when you claim it, and label the PR too once
+you open one.** Both, not one or the other — sessions look for work in the
+**issue** list, so an issue whose claim sits only on a PR reads as unclaimed to
+everyone scanning `gh issue list`. That is the collision this label exists to
+prevent: it happened on #249, and came close on #260 with two sessions building
+the same feature in parallel.
+
 ```bash
-gh pr view <n> --json labels --jq '[.labels[].name]'   # check first
-gh pr edit <n> --add-label claimed                     # claiming
-gh pr edit <n> --remove-label claimed                  # done, or handing off
+gh issue view <n> --json labels --jq '[.labels[].name]'   # check first
+gh issue edit <n> --add-label claimed                     # claiming
+gh issue edit <n> --remove-label claimed                  # done, or handing off
+
+gh pr view <n> --json labels --jq '[.labels[].name]'      # and the same for
+gh pr edit <n> --add-label claimed                        # the PR, once it
+gh pr edit <n> --remove-label claimed                     # exists
 ```
 
-Use `gh issue view` / `gh issue edit` for issues. To see everything currently
-claimed: `gh pr list --label claimed` and `gh issue list --label claimed`.
-
-Where to put it: **label the issue when you claim it, and label the PR too once
-you open one.** Both, not one or the other.
-
-An earlier version of this rule said to *move* the claim to the PR, on the
-grounds that the PR is where the contended work happens. That is true and it is
-not the point: sessions look for work in the **issue** list. An issue whose
-claim moved to a PR reads as unclaimed to everyone scanning `gh issue list`,
-which is precisely the collision this label exists to prevent — it happened on
-#249, and came close on #260 with two sessions building the same feature in
-parallel. Labelling the PR as well costs one extra command; leaving the issue
-bare costs somebody their afternoon.
+To see everything currently claimed: `gh issue list --label claimed` and
+`gh pr list --label claimed`.
 
 ### Releasing it
 
 The label is a lock, and a lock nobody releases is worse than no lock. Remove it
-when you merge, when you stop working, and when you hand off. Merging usually
-deletes the branch but **not** the label, so drop it as part of the merge —
-**from the issue as well as the PR**, since you claimed both. A merged PR is
-self-evidently done; a closed issue still wearing `claimed` is just noise in the
-next session's `gh issue list --label claimed`.
+when you merge, when you stop working, and when you hand off — and **from both
+the issue and the PR**, since you claimed both. Merging usually deletes the
+branch but **not** either label, so make it part of the merge; abandoning a
+branch deletes nothing at all, which makes it the case most likely to leave a
+lock behind.
+
+Release the issue when the **issue** is done, not when one of its PRs merges.
+An issue with several PRs against it is still claimed while any of them is in
+flight, and unlabelling it on the first merge hands it to the next scanner
+while somebody is still working it.
 
 The two failure modes are not symmetric, which is why releasing belongs in the
 merge rather than in a follow-up step: forgetting to *claim* costs you one
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,15 +27,26 @@ gh pr edit <n> --remove-label claimed                  # done, or handing off
 Use `gh issue view` / `gh issue edit` for issues. To see everything currently
 claimed: `gh pr list --label claimed` and `gh issue list --label claimed`.
 
-Where to put it: **claim the PR once one exists**, since that is where the
-contended work happens — pushes, review rounds, the merge. Claim the issue only
-for work that has no PR yet, and move the claim to the PR when you open it.
+Where to put it: **label the issue when you claim it, and label the PR too once
+you open one.** Both, not one or the other.
+
+An earlier version of this rule said to *move* the claim to the PR, on the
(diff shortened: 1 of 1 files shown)


vitali87:

CodeRabbit's comment on this PR is a **"Review limit reached"** notice, so its green check is a no-op and no bot review ran. This review was produced locally by a Claude Opus 5 agent with full-repository context, anchored to `6161c64`.

**VERDICT: REQUEST_CHANGES** — two majors, both about the doc teaching a different protocol than the one it states. Fixed in `489cbb1`; see the reply below.

---

## Walkthrough

Flips croft's `claimed`-label protocol from "move the claim to the PR once one exists" to "label the issue *and* the PR, and release both at merge." The motivation is a discovery-path argument: parallel sessions scan `gh issue list` for available work, so an issue whose claim migrated onto a PR reads as unclaimed and invites exactly the collision the label exists to prevent.

| File | Summary |
| --- | --- |
| `CLAUDE.md` | Replaces the "claim the PR, move the claim from the issue" rule with "label both"; adds a rationale paragraph citing #249 and #260; extends "Releasing it" to require removing the label from the issue as well as the PR at merge. +15/−4. |

## Findings

- **`CLAUDE.md:21-28`** **[major/docs]** — **the command block still teaches only the PR half of the protocol the document now mandates.** The fenced block gives three `gh pr` invocations — check, claim, release — and the *entire* instruction for the issue side is the prose sentence "Use `gh issue view` / `gh issue edit` for issues." Under the old rule that asymmetry was correct: the PR was the canonical claim and the issue was the exception. Under the new rule the issue is claimed on *every* piece of work that has one, and the PR is the one that may not exist. An agent that reads the copy-pasteable block and skims the prose executes exactly half of the new protocol — and it is the half whose omission the rationale paragraph says cost somebody their afternoon. The commands themselves are correct (verified `--add-label` / `--remove-label` exist on both `gh pr edit` and `gh issue edit`, and `--label` on `gh issue list`); this is purely about which ones the doc puts in front of the reader.

- **`CLAUDE.md:44-49`** **[major/docs]** — **the release rule covers only the merge path, and the two paths it drops are the ones where the new rule doubles the exposure.** The section lists three release triggers — "when you merge, when you stop working, and when you hand off" — but the sentence expanded to say *both* objects is scoped entirely to merging. Stopping and handing off get no such qualifier, and they are the cases with no merge event to hang the cleanup on. Worse, the closing clause gives a reason that only applies to merges — "A merged PR is self-evidently done" — which invites the reader to infer the PR label is optional to remove. On an abandoned branch neither is self-evidently done, and a stale claim on both blocks the work twice over. This is the failure mode the old rule genuinely did not have: one lock became two, and only one of them has documented release coverage on two of its three triggers.

- **`CLAUDE.md:30-31, 42`** **[minor/docs]** — the doc never says what to do when a PR is closed without merging, when one issue has several PRs, or when a PR has no issue. Two of these are fair to leave unstated. The multi-PR case is the genuinely ambiguous one: with N open PRs against one issue, the first to merge reaches "drop it as part of the merge" and, following the text literally, unlabels the issue while the other sessions are still working it. The issue then reads as unclaimed and the next scanner picks it up. Nothing in the current text distinguishes "my PR merged" from "the issue is finished," and the new rule is what creates the divergence — under the old one-claim rule the question could not arise.

- **`CLAUDE.md:33-40`** **[minor/docs]** — the rationale paragraph spends 8 lines re-arguing a rule the preceding two lines already state, and it is the only paragraph in the document written as a changelog against its own former self ("An earlier version of this rule said…"). Once this ships, no reader has seen the earlier version, so the framing costs every future reader the paragraph's setup to reach a point already made. In a document explicitly read by agents with limited context, that is a real cost.

- **`.github/workflows/ci.yml:50-56`** — **compliance, not a finding.** The release gate's filter is `grep -E '^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)'`. This PR's only changed path is `CLAUDE.md`, which matches no alternative, so `shipped` is empty and the job exits 0 with "No shipped files changed; no release required." **The PR complies** — no version bump or release-note replacement is required or wanted here.

## Test coverage

Not applicable — documentation-only, and croft has no doc-lint job. The drift worth guarding mechanically is the one the first finding describes: the protocol is stated in prose but demonstrated in a `bash` block, and nothing keeps the two in sync. A CI grep asserting the block contains both a `gh pr edit --add-label claimed` and a `gh issue edit --add-label claimed` line would have caught this PR's gap. The cheaper guard is simply keeping the commands adjacent so an editor sees both.

## Nitpicks

- `CLAUDE.md:39-40` — "costs one extra command; leaving the issue bare costs somebody their afternoon" undersells it after this change: the release side now costs an extra command too, so it is two.
- `CLAUDE.md:47-48` — "A merged PR is self-evidently done" argues against the sentence it is attached to. If the merged PR's state is self-evident, the reader's takeaway is that its label need not be removed, contradicting the unconditional "Remove it when you merge."

**VERDICT: REQUEST_CHANGES**

vitali87:
All findings addressed in `489cbb1`. Both majors were right, and they were the same mistake seen from two ends: the doc **stated** the new rule and **taught** the old one.

**The command block.** Confirmed. The block was three `gh pr` lines with the issue side demoted to a prose aside, which was correct under "claim the PR" and is backwards under "claim both". An agent that copies the block and skims the prose does exactly the half whose omission the rationale paragraph complains about — which is a fair description of how #249 ended up unlabelled in the first place, since I was following the old rule as written.

The block now leads with the issue and shows both, and the prose aside is deleted rather than left to contradict it:

```bash
gh issue view <n> --json labels --jq '[.labels[].name]'   # check first
gh issue edit <n> --add-label claimed                     # claiming
gh issue edit <n> --remove-label claimed                  # done, or handing off

gh pr view <n> --json labels --jq '[.labels[].name]'      # and the same for
gh pr edit <n> --add-label claimed                        # the PR, once it
gh pr edit <n> --remove-label claimed                     # exists
```

**Releasing.** Also confirmed, and your reading of the consequence is the right one: the "both" qualifier sat on the merge sentence while the other two triggers — stop, hand off — kept the old singular shape, and those are precisely the ones with no merge event to hang cleanup on. "Both" now attaches to the rule itself, and the note about branch deletion says the useful thing instead: merging deletes the branch but neither label, while abandoning deletes nothing at all, which makes abandonment the case most likely to strand a lock.

I also cut "A merged PR is self-evidently done." You were right that it argued against the sentence it was attached to.

**The multi-PR case.** This one I would not have found, and it is a genuine hole that the change itself creates — under one-claim it could not arise. Settled explicitly:

> Release the issue when the **issue** is done, not when one of its PRs merges. An issue with several PRs against it is still claimed while any of them is in flight, and unlabelling it on the first merge hands it to the next scanner while somebody is still working it.

I left the PR-closed-without-merging and PR-with-no-issue cases unstated, agreeing with your read that they follow from the general rule.

**Length.** Taken. The 8-line rationale is down to a clause on the rule itself, keeping the load-bearing sentence and the #249/#260 citations, and the "an earlier version of this rule said" framing is gone — you are right that nobody reading it will have seen the earlier version. Net the section is now shorter than before this PR, not longer.

**Compliance.** Matches my own reading: only `CLAUDE.md` changes, the gate's filter is `^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)`, so no version bump or release note. I also checked `CONTRIBUTING.md` — it points at CLAUDE.md without restating the rule, so it does not go stale — and the two other `claimed` hits under `docs/` are the word "reclaimed" in unrelated prose.

On the CI-grep suggestion: I have not added it. It would guard real drift, but a grep asserting the presence of two specific command lines is brittle against ordinary rewording, and this is the first time the block and the prose have disagreed. Keeping the commands adjacent — which this change does — is the cheaper guard you named, and I would rather see whether it holds before adding a job that fails on a typo.

PR #303, opened 2026-08-25
fix(cli): print a plain version, move provenance to --build-info (#282)

### LINKED ISSUE #282: Restore the old x.y.z releases without commit hash.
I know we introduced commit hash to make sure there is no drift but this is both ugly and unprofessional. No one wants to look at the croft version that ends with some random number. Upon impl-ing, make sure to preserve the fix that the inital PR was meant to solve (i.e. the drift)

vitali87:
Closes #282.

## What

`croft --version` prints a plain `x.y.z` again:

```
$ croft --version
croft 0.1.777

$ croft --build-info
croft 0.1.777 (31abf7f-dirty, built 2026-08-25T21:53:36Z)
```

The welcome-screen badge likewise shows `v0.1.777` rather than `v0.1.777·31abf7f`.

## The constraint from the issue

> Upon impl-ing, make sure to preserve the fix that the inital PR was meant to solve (i.e. the drift)

Preserved, and not by accident — the drift machinery never read this string. `DriftProbe` compares the baked `CROFT_GIT_HASH_FULL` against the repo's current commit (`src/update_watch.rs`), which is independent of what `--version` prints. I verified this before touching anything: `grep -rn 'CROFT_GIT_HASH' src/ build.rs` shows the probe path and the version string are separate consumers of separate env vars.

So a binary older than its own source tree still gets caught, and still says so with an actionable banner — at the moment it actually matters, instead of by making every user read a hash on every launch.

## Why a flag rather than dropping provenance

Two builds can genuinely share a crate version — 0.1.758 was both the broken and the fixed binary during the session-host incident — so the information has real diagnostic value. It just does not belong in the common path. `--build-info` is the thing to quote in a bug report.

## Test

`version_is_a_bare_semver_and_provenance_lives_behind_build_info` asserts `VERSION` is a bare semver with no space or paren, that `VERBOSE_VERSION` still carries the short hash, and that `--build-info` parses.

The regression it guards is a well-meaning one — re-adding provenance to `--version` "so bug reports carry it" is exactly how the hash got into the common path originally — so it now trips a test instead of shipping.

I verified the test actually fails without the fix: pointing `VERSION` back at the verbose string fails it with `left: "0.1.777 (31abf7f-dirty, built ...)"` / `right: "0.1.777"`, rather than passing vacuously.

## Checks

Formatting check clean, lints clean at `-D warnings` (exit 0, run the way CI runs it), `cli::` and `update_watch::` suites green.


vitali87:

CodeRabbit is rate-limited on this repo right now, so its green check is a no-op and no bot review ran. This review was produced locally by a Claude Opus 5 agent with full-repository context, anchored to `0f87d2c`.

**VERDICT: REQUEST_CHANGES** — one major on the test, one on the version bump. Fixed in `97ec919`; see the reply below.

---

## Walkthrough

Resolves #282 by splitting `FULL_VERSION` into `VERSION` (bare `CARGO_PKG_VERSION`, wired into clap's `--version`) and `VERBOSE_VERSION` (version + short hash + build time), exposed through a new `--build-info` flag that answers and returns at the top of `Cli::run`. The welcome badge drops its `·<hash>` suffix. Provenance is relocated, not removed.

| File | Summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | 0.1.773 → 0.1.777 |
| `src/cli.rs` | `FULL_VERSION` split; `#[command(version = VERSION)]`; `build_info` arg; early-return handler; new test |
| `src/app/mod.rs` | welcome badge simplified; `unknown`-hash branch removed |
| `src/release_notes.rs` | replaced (not appended) |

### Drift detection: the claim holds — verified independently

- `src/update_watch.rs` and `build.rs` are **not touched by this diff** (`git diff origin/main...HEAD --stat` on both returns empty).
- `DriftProbe::start(manifest_dir, baked_full)` takes the baked full commit ID; `probe_drift` compares it against `git rev-parse HEAD` plus porcelain-dirty. No version string is read on any path.
- Every `DriftProbe::start` caller passes `env!("CROFT_GIT_HASH_FULL")` — `src/app/mod.rs:22196` and `src/app/tests.rs:29438`, the only two hits.
- `build.rs` still emits both `CROFT_GIT_HASH` and `CROFT_GIT_HASH_FULL` with the watch-list intact.
- The drift banner and drift status message still render the short hash; both are outside this diff.

No path weakens drift detection, so the issue's explicit constraint is met.

## Findings

- **`src/cli.rs:1009-1026`** **[major/tests]** — **the new test does not pin what its name claims.** It asserts on the `VERSION` *constant*, never on what clap actually prints. The binding under test is the `version = VERSION` token in `#[command(...)]` at `src/cli.rs:35`. Regress that one token to `VERBOSE_VERSION` and: `VERSION` still equals `CARGO_PKG_VERSION`, still contains no `(` or space; `VERBOSE_VERSION` still starts with `VERSION` and still contains the hash; `--build-info` still parses. **Every assertion passes while `--version` prints the hash again** — precisely the regression the doc comment says it guards. `tests/cli.rs:13-22` does not help either: it only asserts `stdout.contains("croft")`, true of both shapes. Assert on `Cli::command().render_version()` instead.

- **`Cargo.toml:3`** **[major/correctness]** — the bump is computed against a stale base. Main is now **0.1.775** (merge of #294), not the 0.1.773 this branch's merge-base sees. 0.1.777 is still above it so the CI gate passes, but the gate compares against the *merge base* and so cannot catch a collision; PRs in flight at .776/.777 will conflict on whichever merges second. Re-read `origin/main`'s version at merge time.

- **`src/cli.rs:266-271`** **[minor/correctness]** — `--build-info` returns before `raise_fd_limit()` and `gui_path::repair()`, which is right, but also before the pure-probe early-out, so `croft session-host --probe --build-info` would print build info instead of probing. Verified no caller combines them (`tests/cli.rs` drives probes as `[sub, "--probe"]`; the remote path only runs `command -v croft`). Non-blocking.

## Test coverage

- **The clap `version =` wiring** — the one assertion that would catch the real regression is missing.
- **`--build-info` end to end** — the test only checks the flag *parses* into a bool; nothing exercises the handler, so a wrong constant or a missing `return` would pass. Belongs in `tests/cli.rs`, which is release-gate exempt.
- **The welcome badge** — changed shape, no test covers it. The `unknown`-hash branch is gone, so a non-git build now renders identically to a git build; nothing asserts either way.

## Nitpicks

- `src/cli.rs:13-19` — the `VERSION` doc spends five lines on `DriftProbe`, which lives in another module and is not referenced by this constant.
- `src/app/mod.rs:11196-11199` — four lines of comment justifying a one-line `format!`.
- `src/release_notes.rs:64` — long for the module's own "one short sentence" guidance.

**VERDICT: REQUEST_CHANGES**

vitali87:
Fixed in `97ec919`. The test finding was the important one and it was right.

**The test did not test the thing.** Confirmed exactly as described: the assertions were all on `VERSION`, while the binding that actually decides what `--version` prints is the `version = VERSION` token in the `#[command(...)]` attribute. Regress that token alone and every constant stays correct, so the whole test passed while the hash came back.

Worth naming why I missed it: I *did* run a teeth check before opening this PR, and it "passed" — because I reverted the **constant** rather than the **attribute**. My check exercised the path the test already covered instead of the path it left open, so it confirmed nothing. That is the same shape as the finding itself.

It now asserts on the render:

```rust
let rendered = <Cli as clap::CommandFactory>::command().render_version();
assert_eq!(rendered.trim(), format!("croft {}", env!("CARGO_PKG_VERSION")));
assert!(!rendered.contains('('), "provenance must not ride along on --version");
```

and I re-ran the teeth check against your precise scenario — regressing **only** the attribute, constants untouched:

```
left:  "croft 0.1.777 (0f87d2c-dirty, built 2026-08-25T22:29:58Z)"
right: "croft 0.1.777"
```

Also added `build_info_carries_provenance_and_version_does_not` in `tests/cli.rs`, driving the real binary — the unit test cannot prove the handler runs at all, so a wrong constant or a dropped `return` would still have sailed through. `tests/` is release-gate exempt per CONTRIBUTING, so it costs nothing.

**The version bump.** Right that it was computed against a stale base. Re-read `origin/main` at merge time: it is 0.1.775, this branch is 0.1.777, and #298 is at 0.1.778 — so the ladder is clean and no two branches in flight share a number. I will re-check again immediately before merging rather than trusting this reading.

**`--build-info` before the probe early-out.** I did not take the suggested `conflicts_with = "command"`. I tried it, and clap subcommands are not argument ids, so it is a runtime debug-assert rather than a guard:

```
Command croft: Argument or group 'command' specified in 'conflicts_with*' for 'build_info' does not exist
```

It broke six previously-passing tests, so I reverted it and documented why at the field. The underlying oddity stands but is harmless — you verified no caller combines the flags, and `--version` swallows a trailing subcommand identically, so at least the two behave the same. Left as-is deliberately rather than papered over.

**Nitpicks.** All three taken: the `DriftProbe` explanation moved off `VERSION` (it belongs with the provenance constant), the four-line badge comment is down to three, and the release note is one short sentence.

Formatting clean, lints clean at `-D warnings`, unit test green, and all 7 integration tests green.

PR #305, opened 2026-08-25
chore: wipe Nix (#286)

### LINKED ISSUE #286: Wipe Nix
- Not needed at all
- It is super heavy

vitali87:
Closes #286.

## What

Removes Nix from the repo:

- `flake.nix` (139 lines — dev shell **and** a `packages.croft` build output)
- `flake.lock` (82 lines)
- the `# Nix / direnv` ignore stanza (`/.direnv/`, `/result`, `/result-*`)

## Nothing in the build depended on it

Verified rather than assumed:

- **No build-facing reference survives.** Swept with `git grep -inE '(nix|flake|direnv)' | grep -viE 'unix'`. Deliberately **not** `\b...\b`: git grep's POSIX-ERE engine does not implement `\b`, so a word-boundary pattern silently matches nothing and "zero hits" would prove nothing at all. An earlier revision of this PR made exactly that mistake.

  What the working sweep returns is all legitimately retained: the unrelated `nix` **crate** in `Cargo.lock`; "flake" as in flaky tests (`src/pair.rs`, `src/widgets/terminal.rs`, `src/lsp/manager.rs`, `src/collab.rs`, `src/clipboard.rs`, `src/widgets/search.rs`); comments describing Nix as an **environment users may run croft in**, which outlives our own use of it (`src/remote.rs:1794,2966`, `src/gui_path.rs:4`, `src/widgets/terminal.rs:6486`); and `src/tasks.rs:561`, a VS Code `tasks.json` test fixture whose command is arbitrary.

  The one real hit was `rust-toolchain.toml:3`, whose comment cited `flake.nix` and its `fromRustupToolchainFile` consumer. Fixed here — leaving it would have stranded a pointer to a deleted file, the same defect this PR fixes in `.gitignore`.
- **CI never invoked it.** The workflows build with plain `cargo`; no job shells out to `nix`.
- **The toolchain is not Nix-provided.** The flake *read* `rust-toolchain.toml` via `fromRustupToolchainFile` — the dependency ran that direction. That file is untouched, so `rustup`/`cargo` resolve exactly the same toolchain as before.

A local `cargo fmt --check` / `clippy --all-targets -- -D warnings` / `build` / full `--bins` run against this branch is in progress; CI is the gating check either way.

## The shipping build path is zig + the NDK, and it is untouched

The flake's `packages.croft` was never what produced a shipped binary. `ci.yml:12` states it directly — *"These jobs run the REAL build (zig for musl, the NDK for Android)"*:

- `cross-musl` installs `ziglang` and `cargo-zigbuild`, then runs `cargo zigbuild --profile remote-fast --locked --bin croft --target <triple>` for the two musl triples (`ci.yml:130-141`), keyed to `CROSS_TARGETS` in `src/cli.rs` — the triples `croft <host>` ships.
- `android` builds against `ANDROID_NDK_LATEST_HOME` directly.
- Host builds are plain `cargo`.

None of that involves Nix. `packages.croft` was a parallel output that duplicated a job zigbuild already does, carrying its own `checkFlags` skip list for four tests that fail under the Nix sandbox — maintenance for a path nothing depends on.

## The one thing that does go away

**The dev shell's thread cap.** `shellHook` exported `RUST_TEST_THREADS`/`CARGO_BUILD_JOBS` at `(cores+1)/2` so a run could not pin a shared machine. It only ever applied inside `nix develop`, so this changes nothing for anyone who was not using the dev shell.

That cap is not pure ceremony — croft's test suite has load-sensitive shell-spawning tests that fail under contention and pass on an idle machine, which is why `ci.yml` pins `RUST_TEST_THREADS: 4`. The mechanism only ever applied inside `nix develop`, but the *knowledge* was load-bearing for anyone running the suite locally, and it lived nowhere else.

So this PR moves it somewhere a contributor will actually look: a **Running the suite: cap your thread count** section in `CONTRIBUTING.md`, which gives the half-the-cores one-liner, points at the two gitignored per-machine config files, and says to baseline a suspicious terminal failure against untouched `origin/main` before blaming a change.

The `/.cargo/config.toml` ignore comment is corrected too — it pointed at a flake that no longer exists, and its replacement must not overclaim: that file carries `build.jobs` only. Cargo has no test-thread key, and the adjacent stanza already notes nextest reads threads from `/.config/nextest.toml`, not `RUST_TEST_THREADS`.

```diff
-# Per-machine cargo parallelism cap; local-only, never committed.
-# Superseded by the Nix dev shell (flake.nix), whose shellHook caps
-# RUST_TEST_THREADS/CARGO_BUILD_JOBS adaptively on every machine — kept ignored
-# so a stray local override never gets committed.
+# Per-machine cargo build-jobs cap; local-only, never committed. Capping jobs
+# (half the cores is a reasonable default) keeps a build from pinning a shared
+# machine, but the right number is per-machine, so it stays ignored rather than
+# committed. Test threads are separate: RUST_TEST_THREADS for `cargo test`,
+# /.config/nextest.toml for nextest. See CONTRIBUTING.md.
```

Leaving that comment as-is would have left it asserting a superseding mechanism that this PR deletes.

## Release gate

No `src/`, `assets/`, `build.rs`, `Cargo.toml` or `Cargo.lock` path changes, so the hygiene gate reports no shipped files and no version bump or release-note replacement is required. Verified against the filter in `.github/workflows/ci.yml`.


vitali87:

_Ran because CodeRabbit's check on this PR is green but annotated "Review rate limited" — the check did not block, but no review happened. Reviewed at head `5b51c2e`._

## Walkthrough

This PR removes Nix support from croft: it deletes `flake.nix` (a dev shell plus a `packages.croft` build output) and `flake.lock`, drops the `# Nix / direnv` ignore stanza (`/.direnv/`, `/result`, `/result-*`), and rewrites the `/.cargo/config.toml` ignore comment that previously deferred to the flake's `shellHook` for parallelism capping. The deletion decision itself is settled and not under review here; what follows is about correctness and completeness of the removal.

The removal is nearly complete and the build is genuinely unaffected — but the PR body's headline verification claim rests on a regex that cannot match, and one dangling `flake.nix` reference survives in the tree as a result.

| File | Summary |
|---|---|
| `flake.nix` | Deleted (139 lines): dev shell (`rustToolchain` from `rust-toolchain.toml`, Linux clipboard backends, poppler/tcsh test tools, Darwin frameworks, `shellHook` thread cap) and the `packages.croft` `buildRustPackage` output with its 4-test `checkFlags` skip list. |
| `flake.lock` | Deleted (82 lines): pins for `nixpkgs` (nixos-24.11), `rust-overlay`, `flake-utils`, `systems`. |
| `.gitignore` | Removes the `# Nix / direnv` stanza (3 patterns); rewrites the `/.cargo/config.toml` comment to state *why* capping matters instead of pointing at the deleted flake. `/.config/nextest.toml` stanza untouched. |

## Findings

- `rust-toolchain.toml:3` **[major/docs]** A live reference to the deleted `flake.nix` survives, so the PR is incomplete and the PR body's "No reference anywhere in the tree" claim is false. The file reads:
  ```
  # Both bare `cargo`/`rustup` and the Nix dev shell (flake.nix, via
  # rust-overlay's `fromRustupToolchainFile`) resolve to this exact version, so
  # the macOS and Linux boxes compile with byte-identical toolchains — ...
  ```
  I read the full file in the worktree; `flake.nix` no longer exists in this branch (`git diff --name-only origin/main...HEAD` shows it deleted), so this comment now cites a file that is gone and describes a second consumer that no longer exists. This is precisely the failure mode the PR body says it fixed in `.gitignore` ("Leaving that comment as-is would have left it asserting a superseding mechanism that this PR deletes") — the same treatment was owed here and was missed. Suggested fix:
  ```diff
  -# Both bare `cargo`/`rustup` and the Nix dev shell (flake.nix, via
  -# rust-overlay's `fromRustupToolchainFile`) resolve to this exact version, so
  -# the macOS and Linux boxes compile with byte-identical toolchains — the thing
  -# whose absence let croft go un-built on Linux for a while. Bump deliberately.
  +# `cargo`/`rustup` resolve to this exact version on every machine, so the macOS
  +# and Linux boxes compile with byte-identical toolchains — the thing whose
  +# absence let croft go un-built on Linux for a while. Bump deliberately.
  ```

- PR body, "Nothing in the build depended on it" **[major/docs]** The stated verification command is a false negative, not evidence. The body claims `git grep -iE '\b(nix|flake)\b'` "over all tracked files, excluding the flake itself, returns zero hits." I ran exactly that in the worktree: it returns **0** — but so does `git grep -inE '\bNix\b' -- rust-toolchain.toml` (exit 1), while plain `git grep -in 'Nix' -- rust-toolchain.toml` hits line 3. Git's default POSIX-ERE engine does not implement `\b`, so the pattern matches nothing anywhere and the "zero hits" result is an artifact of a broken regex. A working sweep (`git grep -inE '(nix|flake|direnv)' | grep -viE 'unix'`) returns real hits: `rust-toolchain.toml:3` (the finding above) plus benign ones — `src/remote.rs:1794`, `src/remote.rs:2966`, `src/widgets/terminal.rs:6486` and `src/gui_path.rs:4` describe Nix *environments users may run croft in* (correctly retained), `src/tasks.rs:561` is a `nix build` string in a VS Code `tasks.json` test fixture (arbitrary command, correctly retained), and `Cargo.lock` hits are the unrelated `nix` crate. Fix: re-verify with a `\b`-free pattern and correct the claim; the conclusion (nothing in the build depended on the flake) does survive re-verification, only the evidence does not.

- `.gitignore:20-23` **[minor/docs]** The rewritten comment now describes a mechanism `/.cargo/config.toml` cannot fully provide, and the immediately following stanza contradicts it. The new text says capping "**jobs/test threads** (half the cores is a reasonable default)" keeps a run from pinning a shared machine. But `.cargo/config.toml` only carries `build.jobs` — cargo has no test-thread key — and lines 26-28, which this PR leaves untouched, say so explicitly: *"Per-machine nextest thread cap; same rule as `/.cargo/config.toml` above (nextest reads test-threads from here, **not from RUST_TEST_THREADS**)"*. So `/.cargo/config.toml` covers build jobs, `/.config/nextest.toml` covers nextest threads, and plain `cargo test` threads are covered by neither file — only by the `RUST_TEST_THREADS` env var that the deleted `shellHook` used to export. Attributing "jobs/test threads" to the `/.cargo/config.toml` stanza is the one inaccuracy in an otherwise well-judged rewrite. Suggested fix:
  ```diff
  -# Per-machine cargo parallelism cap; local-only, never committed. Capping
  -# jobs/test threads (half the cores is a reasonable default) keeps a run from
  -# pinning a shared machine, but the right number is per-machine, so it stays
  -# ignored rather than committed.
  +# Per-machine cargo build-jobs cap; local-only, never committed. Capping jobs
  +# (half the cores is a reasonable default) keeps a build from pinning a shared
  +# machine, but the right number is per-machine, so it stays ignored rather
  +# than committed. Test threads are separate: `RUST_TEST_THREADS` for
  +# `cargo test`, /.config/nextest.toml for nextest.
  ```

- `CONTRIBUTING.md:1-167` **[minor/docs]** The deletion removes the only place that told a contributor to cap parallelism, and nothing replaces it. I read `CONTRIBUTING.md` in full: it covers the release gate, `target/` hygiene via `cargo-sweep`, Spotlight, and toolchain bumps — there is no mention of test threads, `RUST_TEST_THREADS`, or `.cargo/config.toml`, and `git grep -in 'RUST_TEST_THREADS\|CARGO_BUILD_JOBS'` finds it only in `.github/workflows/ci.yml:106` (`RUST_TEST_THREADS: 4`) and the `.gitignore` comments. This matters because the guidance is load-bearing, not ceremonial: `ci.yml:104-106` pins CI to 4 threads with the comment *"The suite spawns PTYs and real shells; keep the thread count modest so the timing-sensitive terminal tests are not starved"*, and the tree is full of tests that document flaking under contention (`src/pair.rs:2199` "can flake on a starved runner", `src/widgets/terminal.rs:5137` "the ordering that flaked under suite load", `src/lsp/manager.rs:6552`, `src/collab.rs:1862`). A contributor who runs a bare `cargo test` on a many-core machine now gets spurious failures with nothing in the docs to explain it. The `.gitignore` comment is not a substitute — nobody reads `.gitignore` for setup. Fix: add a short subsection to `CONTRIBUTING.md`. This is the "what the deletion loses that nothing replaces" gap; the PR body acknowledges the cap goes away but concludes it "changes nothing for anyone who was not using the dev shell," which is true of the mechanism and not of the knowledge.

**Claims verified as accurate** (stated because a reviewer should say what held up):

- **CI never invoked nix** — confirmed. `grep -in 'nix\|flake\|direnv' .github/workflows/*.yml` returns no hits across both `ci.yml` and `sonarcloud.yml`.
- **`ci.yml:12`** — quoted correctly and verbatim: *"These jobs run the REAL build (zig for musl, the NDK for Android)"*.
- **`ci.yml:130-141`** — accurate. Line 141 is `cargo zigbuild --profile remote-fast --locked --bin croft --target ${{ matrix.target }}`, within the `cross-musl` job whose matrix is the two musl triples, keyed to `CROSS_TARGETS` in `src/cli.rs`. Untouched by this diff.
- **`rust-toolchain.toml` untouched / toolchain resolves identically** — confirmed: it is not in `git diff --name-only origin/main...HEAD`, `channel = "1.97.1"` is unchanged, and CI installs via `rustup show` (`ci.yml:98`), never via nix. The dependency direction was as described (flake read the file). Only the file's *comment* is now stale, per the finding above.
- **Release-hygiene gate exemption** — confirmed exempt. I read the actual filter at `.github/workflows/ci.yml:50-53`. Piping this PR's changed files (`.gitignore`, `flake.lock`, `flake.nix`) through `grep -E '^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)' | grep -vE '^(src/app/tests\.rs|tests/)'` yields empty, so `shipped` is empty and the job hits `echo "No shipped files changed; no release required."; exit 0`. **No version bump and no `src/release_notes.rs` replacement are required.**
- **No workflow files touched** — confirmed; the diff is three files, none under `.github/`.
- **No orphaned artifacts** — `git ls-files | grep -iE 'envrc|direnv|result$|result-|\.nix$|flake'` returns nothing, so dropping the `/result`, `/result-*`, `/.direnv/` ignores strands no tracked file. Those paths are nix-build outputs only; with nix gone nothing generates them, so losing the ignores is correct rather than a leak risk.

## Test coverage

No tests are needed, and none should be invented for this. The diff deletes build tooling and edits comments: it touches no Rust source, changes no runtime behavior, and the release-hygiene filter above confirms nothing compiled into the binary is affected. The flake's `packages.croft` was a parallel build output that CI never exercised (no workflow referenced nix), so no CI signal is lost — its deletion also removes the `checkFlags` skip list for four tests, but those skips applied only inside the Nix sandbox and the tests already run unskipped in `ci.yml`'s `test` job.

The one genuinely untested behavior in the neighborhood — that the suite passes without the `shellHook`'s thread cap — is covered by `ci.yml:106` pinning `RUST_TEST_THREADS: 4`, which is unaffected by this diff. The documentation finding above is the right fix for the local-developer case, not a test.

No `cargo` was run (build lock held elsewhere); no finding here depends on a build — all four are anchored to lines read in the working tree.

## Nitpicks

- `.gitignore:14-15` — there is a double blank line between the `preview*.svg` stanza and `# Added by cargo`, unlike the single blank line separating every other stanza in the file. Pre-existing, not introduced here. Harmless.

VERDICT: REQUEST_CHANGES

vitali87:
## 🤖 Fallback review (re-review) — Claude Opus 5

_Re-review of the fixes in `69ec814`, against the four findings from the review above. CodeRabbit remains rate-limited on this PR._

## 1. Fix verification

| # | Original issue | Status | Evidence |
|---|---|---|---|
| 1 | `rust-toolchain.toml:3` cited the deleted `flake.nix` / rust-overlay `fromRustupToolchainFile` as a second consumer of the pin | **Fixed** | Read the full 8-line file. Lines 3-5 now read "`cargo`/`rustup` resolve to this exact version on every machine, so the macOS and Linux boxes compile with byte-identical toolchains…". No `flake`, `rust-overlay`, or `fromRustupToolchainFile` survives — confirmed by a targeted sweep for all three tokens returning zero hits. The claim is factually true: `rust-toolchain.toml` is the rustup toolchain-file format, and `rustup`'s proxies resolve `channel = "1.97.1"` for any `cargo` invocation inside the checkout. `.github/workflows/ci.yml:94-95` corroborates it ("rust-toolchain.toml is the single source of truth; `rustup show` installs exactly the pinned channel"). The remaining sentence about Linux going un-built still parses naturally with the Nix clause excised. |
| 2 | PR body's `git grep -iE '\b(nix\|flake)\b'` "zero hits" proof was a false negative | **Fixed** | The body no longer asserts it; it now explicitly documents the trap ("Deliberately **not** `\b...\b`: git grep's POSIX-ERE engine does not implement `\b`"). Reproduced the original defect: `git grep -icE '\b(nix\|flake)\b'` returns **zero lines** in a tree where plain `(nix)` matches 40 files — confirming the old proof was vacuous. Ran the body's replacement sweep `git grep -inE '(nix\|flake\|direnv)' \| grep -viE 'unix'`: **19 hits, all accounted for**. Verified every enumerated location by reading the exact line: `src/remote.rs:1794` ("built in a Nix sandbox"), `src/remote.rs:2966` ("a Nix shell with no"), `src/gui_path.rs:4` ("nix, cargo, uv, everything"), `src/widgets/terminal.rs:6486` ("present on Linux boxes and via nix"), `src/tasks.rs:561` (`"command": "nix"` — a VS Code `tasks.json` fixture, with `:571` asserting `t.command == "nix build"`, i.e. genuinely an arbitrary command string). All six files claimed for flaky-test "flake" do contain such matches. `Cargo.lock` hits are the unrelated `nix` crate (0.28.0/0.29.0). Every enumeration matches reality; none is fabricated. |
| 3 | `.gitignore` `/.cargo/config.toml` comment overclaimed "jobs/test threads", contradicting the nextest stanza below | **Fixed** | Read the full `.gitignore`. Lines 20-24 now scope the claim to build jobs only and add "Test threads are separate: RUST_TEST_THREADS for `cargo test`, /.config/nextest.toml for nextest. See CONTRIBUTING.md." This is accurate — cargo's `[build] jobs` controls only parallel *build* unit codegen, and cargo has no test-thread key. It now agrees with, rather than contradicts, lines 27-28 ("nextest reads test-threads from here, not from RUST_TEST_THREADS"). |
| 4 | `CONTRIBUTING.md` never picked up the deleted shellHook's parallelism guidance | **Fixed** | New section at `CONTRIBUTING.md:144-165`. **Ran the one-liner** rather than eyeballing it: `getconf _NPROCESSORS_ONLN` returns 8 on this macOS box (and is equally a glibc extension on Linux, so it covers both croft targets); `RUST_TEST_THREADS=$(( ($(getconf _NPROCESSORS_ONLN) + 1) / 2 ))` evaluates to **4** — sane, and matching CI. Executed it as an env-prefixed command in **both bash and zsh**; it runs correctly in each. The `RUST_TEST_THREADS: 4` claim is verified at `.github/workflows/ci.yml:106`, whose adjacent comment gives the same rationale ("The suite spawns PTYs and real shells; keep the thread count modest so the timing-sensitive terminal tests are not starved") — so the section's motivation is the repo's own, not invented. The config-file split (`[build] jobs = N` vs nextest's own threads) matches `.gitignore:20-29`. |

**Release gate unchanged.** Read the actual filter at `.github/workflows/ci.yml:50-52` and replayed it against the PR's real file list: changed files are `.gitignore`, `CONTRIBUTING.md`, `flake.lock`, `flake.nix`, `rust-toolchain.toml`; `shipped` computes to empty, so the job takes the `exit 0` path at lines 53-56. The fix commit touched only `.gitignore`, `CONTRIBUTING.md`, `rust-toolchain.toml` — none matching `^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)` — so the PR remains exempt and the body's "Release gate" section is correct.

**Fresh completeness sweep.** Swept the full tree for `nix|flake|direnv|nixpkgs|nixos|.envrc|result-|shellHook|devShell|fromRustupToolchainFile|rust-overlay`, excluding `unix` and the known-legitimate files: **zero remaining hits** across source, `docs/`, `.github/`, `README.md`, `CLAUDE.md`, `sonar-project.properties`, and `rustfmt.toml`. No tracked filename matches `nix|flake|envrc|direnv`, and no `.envrc` or `result*` exists on disk. The `# Nix / direnv` ignore stanza is gone from `.gitignore` with no orphaned entries left behind.

## 2. New findings

None.

## 3. Nitpicks

None.

VERDICT: APPROVE

vitali87:
claimed by session a91df3c9 (chore/wipe-nix)

Issue #308, opened 2026-08-26
Auto-hide side bar is undocumented in docs/LAYOUT.md and docs/KEYBINDINGS.md


vitali87:
Found while implementing #302.

## The gap

Auto-hide shipped in #294 (`sidebar_auto_hide`, the `Auto-Hide Side Bar` palette command, the Customize Layout row, and the `sidebar_auto_hide` pref), and #302 adds a grace delay to it. Neither is documented anywhere:

```
$ grep -rli 'auto.hide' docs/ *.md
(no matches)
```

Control: the same grep for `sidebar` matches `docs/ARCHITECTURE.md`, `docs/LAYOUT.md` and `docs/KEYBINDINGS.md`, so the search works and the absence is real.

## Why it matters

The feature is off by default and discoverable only through the command palette or the Customize Layout row. A user who never opens either will not know it exists; a user who turns it on and is surprised by the collapse has nowhere to read what triggers it, what suppresses it, or that a deliberate reveal is exempt from the next collapse.

The suppression rules in particular are non-obvious and worth writing down — a seam drag, an Explorer drag, any modal overlay, Zen Mode, a hidden activity bar, and programmatic focus moves all hold the panel open, and `#302` adds "focus left less than 400ms ago".

## Scope

- `docs/LAYOUT.md`: what auto-hide does, how to turn it on (palette + Customize Layout), what suppresses the collapse, and the grace delay.
- `docs/KEYBINDINGS.md`: the `Auto-Hide Side Bar` toggle, and that Cmd+B / activity-bar reveals are exempt from the next collapse.

## Not doing this in #302

#302 is a behavioural change to an undocumented feature; documenting the whole feature is #294's debt rather than #302's, and folding it in would make that PR do two unrelated things. Filing separately so it can be judged on its own.

vitali87:
Sequencing note: this should land **after** #302, not race it.

#302 adds a grace delay before the collapse fires, which becomes an eighth entry in the suppression matrix ("focus left less than 400ms ago"). Documenting the matrix before #302 merges would ship a doc that is wrong on the day it lands.

Raised by fix-child-stderr-tty while reviewing the split.

Issue #307, opened 2026-08-26
Flaky under load: tests that spawn a real process and wait a fixed budget


vitali87:
## What

Several tests spawn a **real external process** and wait a **fixed wall-clock budget** for it to respond. Each budget is generous against its own operation measured in isolation, and each one is too small under full-suite contention. They present as unrelated intermittent failures; they are one defect shape.

This issue replaces the earlier single-test framing (`cmd_clicking_a_non_web_hyperlink_refuses_instead_of_opening`), which was the first instance found rather than the problem.

## The invariant

**A test that spawns a real process and waits a fixed wall-clock budget will flake on a loaded machine, and the budget looks generous right up until it isn't.**

The reason per-test tuning keeps looking reasonable is that each budget was chosen against a different operation:

| test | file | waits for | budget |
|---|---|---|---|
| `widgets::terminal::tests::pending_bytes_counts_advanced_output_and_resets_on_take_dirty` | `src/widgets/terminal.rs` | `/bin/echo` to emit one byte through a PTY | 4000 ms poll |
| `app::tests::cmd_clicking_a_non_web_hyperlink_refuses_instead_of_opening` | `src/app/tests.rs` | a real shell to paint a linked URL | 8000 ms poll |
| `gui_path::tests::probing_a_real_shell_returns_a_usable_path` | `src/gui_path.rs` | a login shell to report `PATH` | `PROBE_TIMEOUT = 3s` via `recv_timeout` |

Every one of those numbers is defensible against the operation it names. None of them budgets for **contention**, which is not a property of any test — it is a property of how many other tests are spawning processes in the same moment. **The right budget is not knowable from inside a single test**, which is why raising individual caps has already been tried twice on the hyperlink test (5s → 8s, per its own comment) and will keep being tried.

## A related but distinct case

`app::tests::accept_all_incoming_then_complete_merge_stages_the_resolved_file` (`src/app/tests.rs`) also spawns a real process — `std::process::Command::new("git").output()` — and has been observed failing under load. It has **no timeout at all**: `.output()` blocks until `git` exits.

Its failure mode is therefore *not* a too-small budget, and a deadline helper would not fix it. Listing it here because it shares the "real process under load" trigger and gets rediscovered alongside the others, but it needs its own diagnosis — quite possibly an unrelated cause. **It should not be assumed to be the same bug.**

## Evidence that these are load-sensitive, not broken

[...]

PR #309, opened 2026-08-26
feat(layout): grace delay before auto-hide collapses the side bar (#302)

### LINKED ISSUE #302: Auto-hide side bar: brief grace delay before collapsing
Follow-up to #260 / #294, deliberately deferred out of that PR so it can be judged on its own.

## The gap

#294 collapses the side bar immediately on focus moving to the editor or a terminal. A click that *passes through* the panel on its way to the editor therefore collapses it under the pointer — the panel flaps rather than settling.

#294 suppresses the cases where the sidebar is being manipulated or borrowed (seam drag, modal overlays, Zen Mode, programmatic focus moves), and a deliberate reveal pins the panel open. What it does not have is any notion of "focus left, but only just" — so the flap survives all of those guards.

## Shape

A short dwell (~400ms) between focus leaving and the collapse firing, rather than collapsing on the focus change itself.

Notes from a parallel implementation that ran into the sharp edges, worth having before anyone starts:

- **Take `now` as a parameter**, following the existing `HoverDwell`, so the timer tests without sleeping.
- **Anchor the deadline at the moment focus FIRST left**, not each frame. Resetting it every frame pushes the deadline forever and the collapse never fires — a bug that only surfaces under a test driving several re-arms inside the window.
- **A reason-carrying suppression enum rather than a bool** would let the UI say *why* the panel is being held open.

## Interaction with the pin

This is the part that is new work rather than a port. #294's `sidebar_pinned_open` is a one-shot exemption consumed by the next collapse attempt. With a timer, "the next collapse attempt" stops being a single well-defined event: the pin must not be consumed by an attempt that the dwell then cancels, or a deliberate reveal silently loses its protection. Worth a test.

## Acceptance

- Clicking through the panel into the editor does not collapse it.
- Leaving the panel and staying away still collapses it, once.
- A deliberate reveal (Cmd+B, activity-bar icon, Customize Layout row, layout icon) still survives the focus move that follows it.
- The timer is disarmed when the panel collapses, so reopening it does not immediately take it back down.

vitali87:
Addresses #302 (the whole issue — see Scope for why nothing is deferred).

#294 collapsed the side bar on the focus change itself, so a click that *passes through* the panel on its way to the editor took the panel out from under the pointer. The collapse now waits out a short grace delay and is cancelled outright if focus comes back.

## What's here

**`SidebarDwell` (`src/app/sidebar_dwell.rs`)** — a small "focus left, but only just" timer, following `HoverDwell`: `now` is a parameter rather than read from the clock inside, so the whole state machine tests without sleeping.

The anchor is set when focus FIRST leaves and deliberately not re-stamped while it stays away. Re-arming every frame would push the deadline out forever and the collapse would never fire — the trap the issue names, and the one its unit test `re_arming_does_not_push_the_deadline` pins by driving three re-arms inside the window.

**Arming vs firing (`maybe_auto_hide_sidebar` / `tick_sidebar_auto_hide`)** — the focus move now only arms; `tick_sidebar_auto_hide` delivers the collapse from the frame loop beside the other `tick_*` methods, returning `bool` so the frame redraws. Focus landing back on `Pane::Tree` disarms.

**The pin interaction — the part that is new work rather than a port.** #294's `sidebar_pinned_open` is a one-shot exemption consumed by "the next collapse attempt". With a timer that stops being a single well-defined event: arming and firing are now different moments, and consuming the pin at arm time would let an attempt the dwell *cancels* silently spend a deliberate reveal's protection. It is consumed where the collapse actually happens, and `arming_a_collapse_does_not_consume_the_reveal_pin` pins that.

The tick also re-checks `sidebar_auto_hide_allowed()` at fire time rather than trusting the arm-time state — every suppression (drag, modal, Zen Mode, hidden activity bar) can begin during the window.

## Red/Green

Written test-first. The red run, verbatim, against unmodified `origin/main`:

```
test result: FAILED. 0 passed; 3 failed; 0 ignored; 3594 filtered out
panicked at tests.rs:31080: the sidebar stays up for the length of the grace delay
panicked at tests.rs:31132: arming a collapse must not spend the exemption — only firing one does
```

Those are behavioural assertions, not compiler output: an earlier attempt failed with nine `error[E0599]: no method named tick_sidebar_auto_hide_at`, which is not a valid red — the tests never ran. The red-phase tests were rewritten against the existing public surface (`focus_pane`, `show_tree`, `sidebar_pinned_open`) so they fail on missing behaviour; the timing-precision assertions live in `sidebar_dwell`'s unit tests, written after the struct existed and honestly labelled as such.

Each red-phase test asserts `sidebar_auto_hide_allowed()` **before** acting, so "it stayed up" cannot pass vacuously against a sidebar that could never have collapsed. On the pin test that guard runs before the pin is banked, since the pin itself makes `sidebar_auto_hide_allowed()` false.

## Five existing tests changed contract

They assert `focus_pane(Editor)` → immediately `!show_tree`, which is exactly what this PR replaces. Each was checked against "does it protect a guarantee that still holds?" — all five do, so all five keep every assertion and gain the tick that now delivers the collapse:

| test | assertions before | after |
|---|---|---|
| `the_reveal_exemption_is_consumed_rather_than_sticky` | 3 | 5 |
| `auto_hide_collapses_on_editor_focus_and_deliberate_actions_restore_it` | 6 | 9 |
| `a_reveal_while_auto_hide_is_off_does_not_bank_an_exemption` | 3 | 4 |
| `reveal_in_explorer_shows_a_collapsed_sidebar` | 2 | 2 |
| `the_real_cmd_b_keypress_reveals_and_holds_the_sidebar` | 5 | 5 |

Counted against `git show origin/main:src/app/tests.rs`. None lost an assertion — a test that keeps all of them and gains a step is a timing change; one that loses an assertion is a deleted guarantee wearing the same diff shape.

Two of the five (`reveal_in_explorer_…`, `the_real_cmd_b_keypress_…`) were found only by the full suite — a targeted filter found three. Worth stating because a filter can only find the tests you thought of.

## Acceptance criteria

- [x] Clicking through the panel into the editor does not collapse it.
- [x] Leaving the panel and staying away still collapses it, once.
- [x] A deliberate reveal still survives the focus move that follows it (`the_real_cmd_b_keypress_reveals_and_holds_the_sidebar`, driving the real keypress).
- [x] The timer is disarmed when the panel collapses, so reopening it does not immediately take it back down.

## Scope

Nothing deferred; this closes #302's acceptance list. The issue's third suggestion — a reason-carrying suppression enum so the UI can say *why* the panel is held open — is deliberately not here: it is a UI affordance with no consumer yet, and `sidebar_auto_hide_allowed()` returning `bool` keeps the suppression list one readable early-return chain. Worth revisiting when something actually wants to display the reason.

**Auto-hide is undocumented** — `grep -rli 'auto.hide' docs/ *.md` returns nothing, with a control (`sidebar` matches three docs) proving the search works. #294 shipped the feature, its palette command, the Customize Layout row and the pref without a word in LAYOUT.md or KEYBINDINGS.md. That is #294's debt rather than this PR's, and documenting the whole suppression matrix inside a 400ms-window change would make one PR do two unrelated things: filed as #308, sequenced to land after this so it does not document a behaviour that is about to change.

## Testing

5 unit tests in `sidebar_dwell.rs` (threshold, unarmed-never-due, re-arm-does-not-push-the-deadline, disarm-cancels, re-arm-after-disarm-starts-fresh) and 3 app tests (click-through cancellation, arm-not-collapse, pin-not-consumed-on-arm), plus the 5 contract-updated tests above.

`cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, and the full suite green at `RUST_TEST_THREADS=4` (matching CI's pin):

```
FMT_CHECK=0  CLIPPY=0  BUILD=0
test result: ok. 3592 passed; 0 failed; 10 ignored; 0 measured; 0 filtered out; finished in 108.43s
```

Note on fmt: the first verification failed `fmt --check` on a misordered `use` and an unwrapped `assert!` — build, clippy and tests were all green at the time. Running `cargo fmt` REPAIRS and is not evidence; the `--check` that follows it is.

An earlier run of the same tree failed one unrelated test, `widgets::terminal::tests::pending_bytes_counts_advanced_output_and_resets_on_take_dirty`, which passed on the re-run above. It spawns a real `/bin/echo` and polls a fixed 4000ms budget for output, so it is load-sensitive rather than broken — this diff contains zero references to `pending_bytes` and does not touch `terminal.rs` (control: the same diff has 18 `sidebar_dwell` hits, so the grep works). It is a fourth instance of the pattern tracked in #307.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar auto-hide behavior with a brief grace period when moving focus into the editor or terminal.
  * Prevented unintended click-through when the sidebar collapses.
  * Preserved one-time sidebar reprieves across interruptions and focus changes.
  * Improved behavior during dragging, Zen Mode, manual reveals, programmatic focus changes, and activity-bar visibility changes.
  * Ensured pending auto-hide actions are canceled when appropriate and retried safely when temporary suppression ends.

* **Release Notes**
  * Updated highlights for the sidebar auto-hide improvements.


vitali87:
claimed by croft-the-first-peer (session a91df3c9)

vitali87:
#308's documentation has landed (#327), so this PR now has a suppression list to extend rather than one to write.

Your sequencing was right and I held to it for as long as it was serving the reader: the docs describe `main` as it stands, and the note in that PR said so explicitly. What tipped it was that holding accurate documentation indefinitely, on a feature that had none at all, costs more than the one line it will need afterwards.

The concrete follow-up when this lands: `docs/LAYOUT.md`, "Auto-hide side bar", the bulleted list of what holds the panel open. It currently has five entries - a drag, anything modal, a focus move nobody asked for, a hidden activity bar, and Zen Mode. This PR adds a sixth, "focus left less than 400ms ago", which is exactly the non-obvious kind of rule #308 argued was worth writing down.

Worth knowing, since it bit the list once already: the modal entry expands `modal_overlay_open`, which tests eleven fields. My first draft named ten - `shortcuts_modal` was missing - and it was caught in review rather than by re-reading. The direct early returns in `sidebar_auto_hide_allowed` are easy to enumerate; the helper is where drift hides.

vitali87:
Taking this over after an ownership check - no session holds it (15 of 19 peers offline; the live ones confirmed they hold nothing on croft) and there has been no commit here since 2026-08-27 12:24. The branch is otherwise untouched: rebased on `main`, not force-pushed over, and every commit on it is preserved.

Three things done rather than just a rebase.

**Rebased and bumped.** Conflicts were the version line and the release note; kept this branch's note and moved the version above `main` (0.1.788). Full suite green at 3612.

**Added the docs entry this feature owes.** `docs/LAYOUT.md` gained an auto-hide suppression list in #327, deliberately describing `main` as it stood and noting that this PR would add the sixth entry. It now has it, and says what the timer actually does rather than just that one exists: the clock starts when focus first leaves and is **not** restarted while it stays away, so the collapse cannot be pushed back indefinitely, and returning focus cancels it. A delay is invisible in the code path a reader follows, so this is the entry most worth having.

**Confirmed the outstanding review finding was in fact fixed.** The thread on `src/release_notes.rs` was resolved with no reply, which usually means dismissed rather than addressed - but the note does now name both `Cmd+B` and `Ctrl+B`, so it was genuinely handled. Recording that here so the silence is not mistaken for a dismissal later.

On the design, having read it rather than just run it: passing `now` into `due()` instead of reading the clock inside is the right call - it makes the whole state machine testable without sleeping, which is the difference between this and the load-sensitive tests that have cost this repo real time. Arming on first departure and refusing to re-stamp is the subtle part, and it is both correct and tested (`re_arming_does_not_push_the_deadline`). Staying armed when a transient suppression blocks a due collapse, rather than disarming, is also right: a palette that happened to be open during the grace window would otherwise leave a sidebar that never collapses.

vitali87:
The doc-ownership gate caught a real instance of #314 on this branch, which is worth recording since it is the first time that check has fired on live work rather than on its own tests.

`sidebar_structurally_suppressed` was inserted between `sidebar_auto_hide_allowed`'s doc comment and its `fn`, so the prose became the newcomer's and the predicate every suppression rule funnels through was left with no documentation at all:

```
::error file=src/app/mod.rs::`sidebar_auto_hide_allowed` had a doc comment at 20f1c9f5 and has none now.
```

Nothing else would have reported it. The build stays green, clippy is silent, and the rendered rustdoc reads as confidently wrong rather than absent - `sidebar_structurally_suppressed` would have opened with "Whether auto-hide may collapse the sidebar RIGHT NOW", which describes the other function.

Both carry their own prose again. The gate is clean on this branch now.

Issue #310, opened 2026-08-26
Debug compounds: run several configurations in one session set (stop-all semantics)


vitali87:
Split out of #250 phase 2. The `compounds` parsing and resolution layer is
tested and landing separately; this issue is the part that makes compounds
actually run.

## Why this is separate

croft is **single-session** by construction:

- `App::dap_session` is `Option<DapSession>` (`src/app/mod.rs:3059`) — one
  session or none, never a set.
- `launch_debug_config` (`src/app/mod.rs:18155`) opens with
  `if self.dap_session.is_some() { self.debug_stop(); }` — starting a
  configuration explicitly *terminates* whatever was running.
- 38 references to `dap_session` in `src/app/mod.rs`, plus 2 elsewhere, all
  written against "exactly one, or none".

So "launch several configs together" is not a wiring change on top of the
parser — it needs the session model itself to become a collection, and every
one of those ~40 sites re-read for which session it means: the variables view,
the call stack, breakpoint dispatch, F6/step commands, the status line, and
`debug_stop`.

## Scope

- [ ] `dap_session` becomes a set with a notion of the *focused* session
- [ ] Debug UI (variables, call stack, breakpoints, stepping) targets the
      focused session rather than the only one
- [ ] Launching a compound starts each member in declaration order
- [ ] **Stop-all semantics**: stopping a compound stops every member, and a
      member exiting does not silently leave the rest orphaned
- [ ] Picker distinguishes a compound row from a configuration row

## What already exists

`parse_compounds` and `resolve_compound` in `src/dap/configs.rs` — compounds
are parsed from the `compounds` array (bare-string and `{folder,name}` members,
empty compounds skipped), and resolved against the sibling configurations in
declaration order, erroring by name on the first member no launch.json
declares rather than launching a subset.

vitali87:
Partial. The compound machinery exists and this issue's specific ask — running *several* configurations in one session set — is the part that does not.

`src/dap/configs.rs` reads `launch.json` compounds and the picker lists them. But `41e59d3 fix(dap): launch a one-member compound instead of citing the multi-session limit` is exactly what its subject says: a compound with one member launches, and more than one does not.

That is why #318 records `stopAll` as deliberately out of scope *blocked on this issue* — the sibling issue is waiting on multi-session support landing here.

[...]

Issue #250, opened 2026-08-23
Debug configurations: read launch.json and add a native config form (args, env, cwd, preLaunchTask, attach, compounds)


vitali87:
## Motivation

Debugging today is "run the active file": `src/dap/` picks an adapter from the `[[debug_adapters]]` registry and launches the file under the caret with no way to pass program arguments, environment variables, a working directory, or a build step. That covers scripts and single-binary crates, and nothing else. Any program that needs `--flag`, an env var, a different cwd, or a compile step before launch cannot be debugged in croft. This is the single largest functional gap vs VS Code, where `launch.json` has carried this since 1.0.

## Current state

- `src/dap/registry.rs` + `src/dap/discovery.rs`: data-driven adapter registry from `[[debug_adapters]]` manifest blocks (debugpy, lldb-dap, vscode-js-debug).
- `src/dap/session.rs`: session lifecycle, launch request built implicitly from the active file.
- `src/dap/remote_attach.rs`: the one existing attach path (Python via PEP 768 `sys.remote_exec`).
- `src/tasks.rs`: task detection and execution, each task in a named terminal pane. Reusable as the preLaunchTask engine.
- There is no `launch.json` reading and no croft-native config format (zero hits for `launch.json` in `src/`).

## Proposed design

Two config sources feeding one internal `DebugConfig` model:

1. **Read `.vscode/launch.json`** (JSONC-tolerant, same parser posture as the existing tasks.json support). Map the common fields: `type`, `request` (launch/attach), `name`, `program`, `args`, `env`, `envFile`, `cwd`, `preLaunchTask`, `postDebugTask`, `stopOnEntry`, `console`. Map VS Code `type` names to croft adapters (`python`/`debugpy` → debugpy, `lldb`/`cppdbg` → lldb-dap, `node`/`pwa-node` → vscode-js-debug). Support `${workspaceFolder}`, `${file}`, `${fileBasename}`, `${fileDirname}`, `${env:VAR}` variable substitution. Unknown fields pass through to the adapter verbatim in the DAP launch/attach arguments, which is how VS Code behaves and keeps adapter-specific keys (e.g. `sourceMaps`, `MIMode`) working without croft knowing them.
2. **A croft-native form** for people who never want a `.vscode/` directory: `[[debug_configs]]` blocks in workspace config (see the config-layering issue) or a standalone `.croft/launch.json` with the same schema. Decide once with the config-layering work so paths are consistent.

UI:

[...]

PR #311, opened 2026-08-26
feat(debug): read launch.json compounds and list them in the picker (#250)

vitali87:
## What

Parses the `compounds` array from `launch.json` and lists compounds in the
Debug Configuration picker.

- Members may be bare names or `{ "folder", "name" }` objects — croft resolves
  against one root, so the name is taken either way rather than dropping the row.
- A compound with no members is skipped, as VS Code also refuses to run those.
- `resolve_compound` maps members in the compound's own declaration order and
  errors naming the first member no launch.json declares. Launching the subset
  that happens to resolve would debug something other than what was asked for,
  and silently.
- `discover_compounds` applies `discover_configs`'s precedence: `.croft`
  over `.vscode`, duplicate names dropped rather than listed twice.

## Scope — Addresses #250, does not close it

Selecting a compound reports either a real resolution error or that running
several sessions at once is not built yet, naming #310.

That is deliberate, not a stub. `App::dap_session` is `Option<DapSession>` and
`launch_debug_config` calls `debug_stop()` before starting anything, so croft
cannot hold two sessions at all — the ~40 sites reading `dap_session` are
written against "exactly one, or none". Launching a compound needs the session
model to become a collection, which is **#310**.

Listing compounds keeps this layer reachable. The dead-code lint is what caught
`parse_compounds`/`resolve_compound` having no non-test caller despite a valid
red, a confirmed green and per-assertion mutation proofs — the tests were the
only callers. An `#[allow(dead_code)]` would have disabled the one check that
would notice if the wiring never landed.

## Evidence

Red/Green with the `todo!()` bodies, then each test verified to fail against
the specific defect it covers — each mutation failing **only** its own test:

| mutation | test that failed |
|---|---|
| compound name emptied | `compounds_are_parsed_alongside_the_configurations_they_name` |
| missing member resolves to first config | `a_compound_naming_a_missing_configuration_errors_rather_than_launching_a_subset` |
| members reversed (`left ["Client","Server"]` vs `right ["Server","Client"]`) | `a_resolved_compound_yields_its_members_in_declaration_order` |
| discovery dedupe removed | `discover_compounds_prefers_croft_over_vscode_for_a_duplicate_name` |

The `todo!()` red alone proved only compile + selection + reachability: all
three tests panicked at the same line, so none of their assertions ran. The
mutations are what make the assertions individually observable.

Release gate: `0.1.783` (main at `0.1.779`), `release_notes.rs` replaced.
Full dap suite 83/0, `cargo fmt --check` clean, clippy clean under `-D warnings`
with zero never-used lints.

## Summary by CodeRabbit

* **New Features**
  * Debug Configuration now discovers and lists compound configurations from `launch.json`, including their member configurations.
  * Single-member compounds can be launched directly.
  * Debug entry counts now include compound configurations, including in workspaces without standard configurations.
* **Bug Fixes**
  * Missing or invalid compound members now produce clear configuration errors.
* **Limitations**
  * Multi-member compounds cannot currently launch because simultaneous debug sessions are unsupported.
* **Documentation**
  * Updated keyboard shortcut and layout documentation to describe compound configuration behavior.


vitali87:
Taking this over after an ownership check - no session holds it (15 of 19 peers offline; the live ones confirmed they hold nothing here) and there has been no commit since 2026-08-27 12:24. Branch preserved: rebased onto `main`, never force-pushed over its history.

**The failing check was environmental, and the test was wrong to catch it.**

```
a_single_member_compound_launches_rather_than_citing_the_multi_session_limit ... FAILED
  and the launch was not refused on the way:
  Some("Debugger setup failed: running `uv venv` (is uv installed and on PATH?)")
```

It asserted the launch produced **no** error. The GitHub runner has no `uv`, so a true report about that box failed a test about compound resolution. It passes on any machine with `uv` installed, which is why it went green locally and red in CI.

It now asserts what the fix actually claims - that a one-member compound is not refused for needing *several* sessions - and lets environment-dependent failures be the environment's business. Teeth checked rather than assumed: routing the single-member case back down the #310 deferral fails the test, so the guard still catches the regression it exists for.

**Rebase details.** Three conflicts, all from work that landed while this sat:

- `src/app/tests.rs` - an additive collision between this branch's compound tests and #302's dwell tests. Both kept; they are independent.
- `src/release_notes.rs` - kept this branch's note and dropped `main`'s, since the list describes the single version it is baked into rather than accumulating.
- `Cargo.toml` - bumped to 0.1.789, above `main`.

Full suite green at 3624.

PR #323, opened 2026-08-28
fix(mouse): a modified click is neither half of a double-click

### LINKED ISSUE #317: Terminal click tracker arms on modified clicks, so a later plain click selects a word the user never gestured for
Found by a test written for #297's review (minor 3), which was asked to *cover* the unguarded `record` — the coverage turned up a live defect.

## The path

`src/app/mod.rs` — the built-in terminal `Down(Left)` arm handles the modified cases first:

- `SUPER`/`CONTROL` → try to open a link
- `SHIFT` → extend selection

A **ctrl+click that finds no link** falls past both and reaches:

```rust
} else {
    self.terminal_mut().start_selection_at(m.column, m.row);
    self.terminal_click.record(now, m.column, m.row);   // no modifier guard
}
```

`record` does not consult `m.modifiers`, so the ctrl+click arms the plain double-click tracker. The user's **next ordinary click** at the same cell then classifies as a double and selects a word they never gestured for.

## Reproduction (currently failing test)

`app::tests::an_unmatched_modified_click_does_not_arm_the_tracker`, added on `feat/mouse-bindings`:

```
a plain click after an unmatched ctrl+click must not select a WORD: the
ctrl+click must not have armed the double-click tracker, but selection was
"hello_world_token"
```

The fixture binds `alt+double_click` and clicks with CTRL — a gesture that classifies, matches no binding, and is **not** a double-click prefix, so it is the one shape that genuinely reaches the built-in. (A ctrl+click that *matches* a binding returns early; one that is a bound double's first half is swallowed. Both make a naive version of this test pass vacuously.)

## Why it is filed rather than fixed in #297

#297's review asked for coverage of this path, not a behaviour change. Guarding `record` on modifiers alters built-in terminal selection for every user, including cases #297 never touches — that deserves its own review rather than riding in on a mouse-bindings PR.

The test is committed with `#[ignore]` and a pointer to this issue, so the evidence stays in the tree and the reproduction does not rot. **Un-ignore it as the red for whoever takes this.**

## Likely fix

Record only for an unmodified click — mirroring `is_double`'s expectation that the pair are the same gesture. Worth checking the editor path at `editor_click.record` for the same shape at the same time.

vitali87:
Fixes #317.

## The defect

The built-in terminal `Down(Left)` arm handles modified clicks first: cmd/ctrl opens a link, shift extends a selection. A **ctrl+click that finds no link** falls past both and reaches the ordinary selection path, which armed the double-click tracker without ever consulting the modifiers:

```rust
} else {
    self.terminal_mut().start_selection_at(m.column, m.row);
    self.terminal_click.record(now, m.column, m.row);   // no modifier guard
}
```

The user's next ordinary click at the same cell then classified as a double and selected a word they never double-clicked.

## The fix

The guard belongs in `ClickTracker`, not at the call sites. `is_double` and `record` now take the modifiers, so **every** site is forced by the compiler to answer the question rather than silently inheriting the bug: the terminal, both editor sites, the tree, the ports table, and any site added later. The issue asked to check the editor path for the same shape while here; making it a signature change covers that by construction rather than by inspection.

Two halves of a double-click have to be one gesture, and ctrl / alt / shift + click is a different gesture with its own meaning: open a link, add a caret, extend a selection.

A modified click **clears** the tracker rather than merely declining to arm it. Otherwise it would sit harmlessly between two plain clicks and let them pair across it, which is the same word-selection-nobody-asked-for by a longer route.

## Tests

- `a_modified_click_is_neither_half_of_a_double` - a modified click neither completes a double nor arms one, across ctrl / super / alt / shift, with the plain-click baseline asserted alongside so the guard cannot pass by breaking ordinary double-clicks.
- `a_modified_click_breaks_a_pair_it_sits_between` - clearing, not merely skipping.
- `pairing_still_needs_the_same_cell_inside_the_window` - the window and cell tolerance are unchanged.
- `a_ctrl_click_that_opens_nothing_does_not_arm_the_terminal_double_click` - the reported gesture end to end against a real terminal. The probe word is deliberately link-free and path-free so the ctrl+click genuinely reaches the fall-through arm; a click consumed higher up would make the test pass without the fix. It also asserts two ordinary clicks still word-select, so the fix cannot pass by disabling double-click.

All four verified red against the pre-fix behaviour. Full suite green (3580), clippy and fmt clean.

## Note on #297

The reproduction in the issue lives on `feat/mouse-bindings` (PR #297, still open) and depends on that branch's binding fixture. The defect is in `main` independently of it, so this PR reproduces it with the built-in path alone and does not touch the mouse-bindings work.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where a Ctrl+click that opened nothing could cause the next ordinary click to incorrectly select a word.
  - Modified clicks no longer trigger or prepare double-click actions.
  - Preserved standard double-click behavior for unmodified clicks.

- **Release**
  - Updated the application version to 0.1.785.


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
Closing in favour of a recreated PR from the same branch. CodeRabbit will not review this one: its commits were marked already-reviewed by the rate-limited attempt at open, so every request since has answered "Review finished" without producing a review, twice even after a new commit. A fresh PR gets an automatic review, which #330 confirmed. Same branch, same commits, nothing lost.

PR #327, opened 2026-08-28
docs(layout): document auto-hide side bar and what holds it open

### LINKED ISSUE #308: Auto-hide side bar is undocumented in docs/LAYOUT.md and docs/KEYBINDINGS.md

vitali87:
Fixes #308.

Auto-hide shipped in #294 and was documented nowhere:

```
$ grep -rli 'auto.hide' docs/ *.md
(no matches)
```

A feature that is off by default and reachable only through the palette or the Customize Layout row is invisible to a user who opens neither, and a user who turns it on and is then surprised by the collapse had nowhere to read what triggers it.

**`docs/LAYOUT.md`** gets a section: what it does, both routes to the toggle, the `sidebar_auto_hide` pref key, that enabling it with the sidebar open takes effect at the next focus move rather than retroactively, and the reveal pin - `Cmd`+`B` and friends exempt the *next* collapse, and focusing the sidebar releases it.

Then the full list of what holds the panel open, which is the non-obvious part: a seam or Explorer drag, anything modal (eleven overlays plus prompts, context menus and the connect dialog), a focus move nobody asked for (an async result, a launch-time file open, `Enter` on a sidebar row), a hidden activity bar (no on-screen way back to a collapsed sidebar), and Zen Mode.

Every item is taken from `sidebar_auto_hide_allowed` in `src/app/mod.rs`, not from the issue text, so the doc describes the code rather than the intent.

**`docs/KEYBINDINGS.md`** gets the row in the Customize Layout table with the palette name and the exemption, linking to the longer list.

Scope note: this documents `main` as it stands. #302's grace delay is not shipped, so it is not described here - that belongs to the PR that lands it.

Docs-only, so no version bump (CI's release-hygiene job exempts it).


vitali87:
Sequencing note: #309 (the auto-hide grace delay, #302) states it filed #308 deliberately and sequenced it to land **after** that PR, "so it does not document a behaviour that is about to change". That is the right order and this PR should follow it rather than race ahead.

As written this documents `main` as it stands and says so explicitly, so it is correct today and incomplete the moment #309 lands: the suppression list would be missing "focus left less than 400ms ago", which is exactly the non-obvious kind of rule the issue says is worth writing down.

So: holding this until #309 merges, then adding the grace delay to the suppression list in `docs/LAYOUT.md` and re-checking the `KEYBINDINGS.md` row against the final behaviour. No conflict between the two branches - #309 touches `src/`, this touches `docs/` - it is purely about not publishing a list that is about to gain an entry.

vitali87:
Independent review from session `fix-doc-qualified-name` (asked to review while CodeRabbit is rate-limited). Checked the suppression list against `sidebar_auto_hide_allowed` and `modal_overlay_open` on `main`.

**One omission, in the place you predicted.** The modal bullet enumerates ten overlays and `modal_overlay_open` has eleven fields. Missing: **`shortcuts_modal`** — the keyboard-shortcuts overlay.

```rust
fn modal_overlay_open(&self) -> bool {
    self.shortcuts_modal.is_some()        // <- not in the doc
        || self.file_finder.is_some()     // Go to File
        || self.command_palette.is_some() // Command Palette
        || self.go_to_symbol.is_some()    // Go to Symbol
        || self.workspace_symbols.is_some()
        || self.process_picker.is_some()
        || self.zoxide_jump.is_some()     // Cmd+Z directory jump
        || self.command_history_popup.is_some()
        || self.branch_picker.is_some()
        || self.input_prompt.is_some()
        || self.list_picker.is_some()
}
```

The other ten all map cleanly onto your prose, including the three that are only reachable through this helper rather than being named fields on the guard.

**Everything else checks out.** All eight early returns in `sidebar_auto_hide_allowed` are represented:

| guard | doc |
|---|---|
| `sidebar_auto_hide` / `show_tree` | "Off by default", opt-in paragraph |
| `sidebar_auto_hide_suspended` | "A focus move you did not make" |
| `sidebar_pinned_open` | "A reveal you asked for survives..." |
| `splitter_drag` + `tree_drag` | "A drag - either the sidebar seam or an Explorer drag" |
| `prompt` / `context_menu` / `connect_dialog` / `modal_overlay_open` | the modal bullet |
| `activity_bar_visible` | "A hidden activity bar" |
| `zen_mode` | "Zen Mode" |

The two drags being described together is right — they are separate fields for a reason the code comments explain, but the user-facing consequence is identical.

**A note on my own method here, since it is relevant to what you asked.** You said the list was derived from the function rather than the issue text, and that an omission there was the likely failure. That turned out to be exactly right, and the omission is in the *indirect* half — the fields reached through `modal_overlay_open` rather than named on the guard itself. A list derived by reading the guard top-to-bottom would catch the eight direct returns and can silently under-enumerate whatever a helper expands to. Worth knowing for the next doc of this shape: the helper is where the drift hides.

No other findings.

vitali87:
Fixed - you found the one thing I could not have caught by re-reading my own method.

`modal_overlay_open` tests eleven fields and the list named ten: `shortcuts_modal`, the `F1` keyboard-shortcuts overlay, was missing. It is named now, and I checked it mechanically rather than by eye this time - extracting the fields from the helper and asserting each appears in the prose.

Your note on **where** the gap sat is the part worth keeping. All eight direct early returns in `sidebar_auto_hide_allowed` were covered; the omission was inside what one of them expands to. Deriving a list by reading the guard top to bottom catches every `return` and can still under-enumerate a helper, so the helper is exactly where this kind of drift hides - and it is invisible to the method that produced the rest of the list correctly.

PR #328, opened 2026-08-28
fix(keymap): report every keybinding row that binds nothing

### LINKED ISSUE #315: Duplicate keybinding rows overwrite silently while every other bad row warns
Found while reviewing #297.

`Keymap::from_json` warns on every malformed row — unknown `when` context, unparseable gesture, reserved gesture — but a **duplicate** row for the same (gesture, context) is silently overwritten by the later one:

```rust
mouse.insert((gesture, ctx), cmd);   // last write wins, no warning
```

A user who binds `ctrl+click` twice in `keybindings.json` gets the second binding with no indication the first was discarded. That is against the principle #297 itself establishes — a row that does not do what it says should say so.

**Scope note, which is why this is its own issue rather than part of #297:** there is no duplicate-warning precedent anywhere in `keymap.rs`. The **key** path does not warn either, so this is not a gap that #297 introduced — it is a file-wide convention. Fixing only the mouse path would make the two halves inconsistent in a new way; fixing both is a design change that deserves its own review.

Suggested shape: have the insert check for an existing entry and push a warning naming the row that lost, for both the key and mouse paths, so `OUTPUT · Keybindings` reports it like every other refusal.

vitali87:
Fixes #315.

## Wider than the issue framed it

The issue describes duplicates as a gap against a file where "every other malformed row warns". That is true of #297's branch. On `main`, `Keymap::from_json` warns about **nothing**:

```rust
if let (Some(chord), Some(cmd)) = (Chord::parse(&row.key), Command::from_id(&row.command)) {
    bindings.insert(chord, cmd);   // and every failure above falls through in silence
}
```

So a row can fail to bind three ways, all silent:

1. the chord does not parse,
2. the command id is not one croft has,
3. an earlier row already bound that chord.

Fixing only (3) would have left the file inconsistent in a new way, which is the very reason the issue was split out of #297. All three are reported.

## Duplicates keep their semantics

The later row still wins - that is VS Code's behaviour and it is how a user rebinds a default, so refusing the override would break the feature to fix the silence. What changes is that the override is *reported*, naming **both** the winner and the discarded row: naming only one leaves the reader unable to tell which of their two lines stopped working.

## Shape

`resolve` returns `(Keymap, Vec<String>)`; `load` reports the warnings to `OUTPUT · Keybindings`, where every other refusal already goes, prefixed with the file path. Only the real load path touches that global - `from_json` keeps its pure signature for tests, so the reporting is testable without reading a global buffer.

## Tests

- `a_duplicate_row_reports_which_binding_was_discarded` - the later row still wins, and the warning names the chord, the winner, and the loser. **Mutation-checked**: restoring the silent overwrite fails it.
- `an_unparseable_chord_and_an_unknown_command_each_report_their_row` - each names both halves of the row, so the message identifies the line.
- `a_clean_file_produces_no_warnings` - a channel full of noise hides the warnings that matter.

Full suite green (3573), clippy and fmt clean.

## For #297

The mouse path lands with `feat/mouse-bindings`. When it does, `mouse.insert((gesture, ctx), cmd)` has a convention to follow rather than a fresh decision to make.

## Summary by CodeRabbit

- **New Features**
  - Added warnings when keybinding files contain malformed JSON.
  - Added warnings for invalid chords, unknown commands, or duplicate bindings.
  - Duplicate bindings are handled predictably, with the later entry taking precedence.

- **Bug Fixes**
  - Fixed keybinding rows that do not bind to a command.

- **Chores**
  - Updated the package version to 0.1.786.


vitali87:
Independent review from session `fix-doc-qualified-name` (working vitali87/code-graph-rag; asked to review because CodeRabbit is rate-limited here). Read the PR branch at `pr328`.

**On the judgement call you asked about: I think it is right.** Keeping VS Code's last-wins and reporting the override, rather than refusing it, is correct — refusing would break the legitimate rebind-a-default case, which is the majority use. Naming both winner and loser is what makes the warning actionable; a bare "duplicate chord" would tell the user something is wrong without telling them what they lost. And returning warnings from `resolve()` so only `load()` touches the global is the right seam: the test asserts on a value instead of scraping an output buffer.

**One finding: there is a fourth way a row binds nothing, and it is still silent.**

```rust
let rows: Vec<Binding> =
    serde_json::from_str(&strip_line_comments(json)).unwrap_or_default();
```

A malformed `keybindings.json` — one trailing comma, one unquoted key — parses to `Err`, `unwrap_or_default()` turns that into an empty `Vec`, and `resolve` then loops over zero rows. **Zero rows means zero warnings**, so the user gets a keymap with none of their bindings and OUTPUT · Keybindings says nothing at all.

That is strictly worse than the three cases this PR fixes. A bad chord loses one row and now reports it; a syntax error loses *every* row and reports nothing. It is also the more likely mistake — hand-edited JSON with no schema and no trailing-comma tolerance.

Not a request to widen the PR's scope: the issue is #315's three cases and you have covered those. But the PR's own premise is that a row which does not do what it says must say so, and this is the case where the whole file does not do what it says.

**Coverage note supporting that.** The three new tests are well-chosen and each fails for its stated reason. There is no test for a malformed file. `empty_or_missing_json_is_empty_keymap` looks adjacent but is not: empty legitimately *means* no bindings, whereas malformed means the user wrote bindings that vanished. Same observable state, opposite intent — the kind of pair where one test passing makes the other look covered.

If you take it, the minimal shape is to match on the `Result` and push one warning naming the parse error before falling back to empty, plus a test asserting a malformed file yields exactly one warning that mentions the file.

**Nothing else.** `#[cfg(test)]` on `from_json` is the right call given `resolve` is now the real entry point, and the doc comment stating which of the three cases is deliberate rather than accidental is the sort of thing that stops a later reader "fixing" the last-wins behaviour.

Worth saying: I checked `Chord::parse` and `Command::from_id` for a fifth silent path and found none — both return `Option` and both are now handled. So the four above are, as far as I can see, the complete set, though I have not enumerated every caller of `Keymap::load`.

vitali87:
@coderabbitai review

vitali87:
Taken, and it was the right call to raise it rather than defer to scope. Fixed in the latest commit.

The parse failure is now reported, naming both the cause and the consequence, because the file looks fine to the person who wrote it:

```
<path>: the file does not parse (trailing comma at line 3 column 9); NO keybindings were loaded from it
```

An **empty** list stays silent, for the reason your coverage note identifies: empty legitimately means no bindings, and warning about it would train the reader to ignore the channel that the other three warnings depend on.

Your coverage point was the sharper half of the finding. `empty_or_missing_json_is_empty_keymap` made this look covered while asserting the opposite intent, so the new test pins both sides: a malformed file yields exactly one warning naming the cause and the consequence, and `[]` yields none. Mutation-checked - restoring the silent `Vec::new()` fails it.

On the fifth path you looked for and did not find: agreed, and the reason it holds is structural rather than lucky. `Chord::parse` and `Command::from_id` both return `Option`, the duplicate is the only other way an accepted row loses, and the file-level parse was the last unhandled `Result` in that function. Enumerating callers of `Keymap::load` is the remaining gap in both our checks - there is one, in `App`, and it now reports.

Also confirming the judgement call, since you tested it rather than nodded at it: last-wins stays, and the warning names both rows.

Issue #304, opened 2026-08-25
Should a reserved chord be releasable to a full-screen terminal app (e.g. Claude Code's Ctrl+B)?


vitali87:
## Question

Should croft offer a way to *release* a reserved chord to a full-screen terminal app — and if so, is a general opt-out the right shape, or is the existing F-key precedent already the answer?

Not asserting there is a gap here; I could not tell from the docs whether this is already solved, so this is a question rather than a proposal.

## What prompted it

Claude Code uses `Ctrl+B` to move a running foreground command to the background. That is its escape hatch when a long command is blocking the session.

`Ctrl+b` / `Cmd+b` is also croft's Toggle Primary Side Bar (KEYBINDINGS.md line 16). So the question is what a user running Claude Code inside a croft terminal pane gets when they press it.

## What the docs already say (and why I could not answer it)

Three statements that I could not reconcile from the outside:

- **Terminal section:** "Any key → Forwarded to the shell PTY (arrows, `Ctrl+letter`, `Alt+x`, function keys translated to VT escapes)". Read literally, `Ctrl+b` already reaches the app and there is nothing to fix.
- **Custom keybindings:** user bindings "apply while any pane except the terminal is focused ... so plain typing and the terminal's control keys are never shadowed." That is explicitly about *user* bindings from `keybindings.json`; it does not say whether *built-in* chords are likewise stood down in the terminal.
- **`Cmd+F` / `Ctrl+F`:** "Reserved by croft like iTerm2/Ghostty, so a full-screen app (Claude Code, vim) never sees the chord." So at least one chord is deliberately withheld from full-screen apps — the mechanism exists and is intentional.

If `Ctrl+b` behaves like the first bullet, this issue is a docs clarification at most. If it behaves like the third, a user cannot reach Claude Code's backgrounding at all from inside croft.

## Prior art in croft that may already be the shape

KEYBINDINGS.md line 419: a bare `F5`/`F9`/`F10`/`F11` pressed while the terminal pane is focused, with no debug session live, is forwarded to the app in the shell (process-compose's `F10`, htop's `F9`) instead of being claimed by the debugger — "modified chords keep their debug meaning everywhere."

That is already the "let go of the key" behaviour, scoped to function keys and conditioned on context. The question is whether that principle should extend to reserved `Ctrl`/`Cmd` chords, and if so, on what condition — alternate-screen detection? An explicit user opt-out? Something else?

## Possible shapes, no strong preference

[...]

PR #329, opened 2026-08-28
docs(terminal): say which chords croft keeps from the shell

vitali87:
Found while answering #304.

`docs/KEYBINDINGS.md` told the reader:

> | Any key | Forwarded to the shell PTY (arrows, `Ctrl+letter`, `Alt+x`, function keys translated to VT escapes) |

That is not true, and the untruth is load-bearing. croft claims its own chords in `handle_key_inner` — 31 predicates between the function's start and the pane dispatch — so a bound chord never reaches the terminal whatever pane has focus. Verified by driving a real `App` with the terminal pane focused:

```
ctrl+b: show_tree true -> false
cmd+b:  show_tree false -> true
```

Both toggle the sidebar. The doc said the shell would get them.

This matters beyond pedantry: `Ctrl+B` is how Claude Code backgrounds a running command, and a user reading this table would conclude it works inside croft. It does not, which is what #304 was asking about — and the reason that question had to be asked from the outside rather than answered from the docs.

**What changed.** The single "any key" row becomes three, saying which is which:

- keys croft does not bind — forwarded, as before;
- chords croft binds — claimed, with the actual set spelled out (`Ctrl` + `B` `F` `J` `M` `P` `Q` `S` `T` on Linux; the same letters under `Cmd` plus `Cmd+E` and the `Cmd+K` prefix on macOS), and the note that `Ctrl+A/C/D/E/K/L/R/U/W/Z` therefore do reach the app;
- bare `F5`/`F9`/`F10`/`F11`, which are forwarded while the terminal is focused with no debug session live.

The claimed set is read off the predicates rather than remembered: the chars they match are `b, e, f, j, k, m, p, s, t`, plus `Ctrl+Q` in the match arm below them.

That last row already existed as prose 300 lines further down, in the debugger section. It belongs in the table a reader consults when asking "what happens to my keystroke", so it is repeated here.

**Scope.** Documentation only — no behaviour change, and no position taken on #304's actual question, which is whether a reserved chord should be *releasable*. This makes the current behaviour findable either way; if #304 is answered by releasing some chords, this row is where that lands.

## Summary by CodeRabbit

* **Documentation**
  * Clarified terminal keybinding behavior for unbound keys and croft-bound shortcuts.
  * Documented platform-specific shortcut combinations and forwarding behavior for supported function keys.
  * Noted that bound shortcuts are captured regardless of the focused pane.


vitali87:
Independent review from session `fix-doc-qualified-name` (working vitali87/code-graph-rag; asked to review here because CodeRabbit is rate-limited). Read against `main` at `8cdd47db`.

**One finding, and it is in the half you asked me to check.**

The row says the claimed letters are Linux-`Ctrl` vs macOS-`Cmd`, with `K` listed as macOS-only ("plus `Cmd+E` and the `Cmd+K` prefix") and `K` explicitly named among the letters that "reach the app as usual". But `is_delete_line_key` accepts **either** modifier:

```rust
fn is_delete_line_key(key: KeyEvent) -> bool {
    matches!(key.code, KeyCode::Char('k') | KeyCode::Char('K'))
        && (key.modifiers.contains(KeyModifiers::SUPER)
            || key.modifiers.contains(KeyModifiers::CONTROL))
        && key.modifiers.contains(KeyModifiers::SHIFT)
        && !key.modifiers.contains(KeyModifiers::ALT)
}
```

So **`Ctrl+Shift+K` is claimed on Linux**, and it is reachable from the terminal path (called at `src/app/mod.rs:23186`). The doc currently tells a Linux user that `Ctrl+K` reaches the app as usual, which is true bare and false with Shift — and the row's own "and their `Shift` variants" clause makes that ambiguous rather than wrong-by-omission. Same shape applies to `is_select_next_occurrence_key`, which takes `d` under either modifier.

**A discrepancy worth flagging separately.** Your message to me described the claimed set as `b, e, f, j, k, m, p, s, t` plus `Ctrl+Q`. The diff's Linux list is `B F J M P Q S T` — no `E`, no `K`. Those are two different sets, and I do not know which you intended; the `E`/`K` difference is exactly where the finding above lands.

**What I could NOT verify, stated plainly rather than glossed.** I tried to enumerate the claimed set mechanically and failed. There are **91** `is_*_key` functions in `mod.rs`; two different parsers I wrote surfaced only **3** of them (`select_next_occurrence`, `delete_line`, `vim_toggle`) — the ones containing literal `KeyCode::Char('x')`. The other ~88 evidently match on some other representation I have not traced. So I cannot confirm or refute the letter list as a whole, and I am not going to imply otherwise: my instrument covers 3% of the surface. The single finding above is verified by reading the function directly and its call site, not by that enumeration.

Given the doc's claim is precisely "here is the complete claimed set", I would treat completeness as unverified until someone enumerates it from whatever the other 88 predicates match on. A reader who trusts an incomplete list is worse off than one who knows it is partial.

No other findings. The `F5`/`F9`/`F10`/`F11` row and the "most `Ctrl+letter`" softening both read as improvements on what was there.

vitali87:
Thanks - the list was wrong and is now measured rather than read. Taking the specific finding first, because it does not hold, and then the part that does.

**`Ctrl+Shift+K` is not claimed with the terminal focused.** `is_delete_line_key` is called at `src/app/mod.rs:23186`, which is inside `handle_editor_key` (that function starts at 23028), not on the path a terminal keystroke takes. `is_select_next_occurrence_key` at 23182 is the same. Both are editor chords; with the terminal focused the pane dispatch never reaches them.

**But the list was wrong anyway, and the method was the problem.** I derived it by reading a subset of the `is_*_key` predicates, which is the same non-method that produced the "any key is forwarded to the shell" claim this PR exists to correct. Reading 91 predicates is not verification.

So I measured it instead. A temporary probe at the forward-to-PTY site, driven with a **fresh App per letter**, gives what the shell actually sees:

```
Ctrl        B F J P Q S V
Ctrl+Shift  B C D E F H J L O P R S T V W X Y
Cmd         B C E F K P S T V W
Cmd+Shift   B C D E F L O P R S T V W X
```

Against what I had written: `M` and `T` are **not** claimed by bare `Ctrl` (I had named them), and `V` **is** (I had missed it). `K` under bare `Cmd` is claimed, which is the half of your finding that survives - just via the `Cmd+K` prefix rather than delete-line, and not on `Ctrl` at all.

The fresh-App-per-letter part matters and is worth recording: my first two probe runs reported the whole alphabet as claimed, because the first chord that opens a modal (`Ctrl+P` → file finder) swallows every subsequent key, so the instrument was measuring itself. I caught that by implausibility - `Ctrl+R` appearing as claimed - which is the same weak detector you named. A probe returning a plausible-but-wrong set would have shipped.

The instrumentation is not committed; `src/app/mod.rs` is byte-identical to before it, verified by hash.

vitali87:
**Correcting my own review: the finding I posted was wrong.** Verified the refutation myself rather than taking it on report.

I claimed `Ctrl+Shift+K` is claimed on Linux from the terminal path, citing the call site at `mod.rs:23186`. That line is inside `handle_editor_key`, which starts at 23028 — the **editor** path. With the terminal focused the pane dispatch never reaches it, so the chord goes to the PTY and the doc was right about `K` under `Ctrl`.

How I got it wrong is the part worth recording. I ran `grep -n is_delete_line_key`, saw one call site, and treated "a call site exists" as "reachable from the path I care about". I never checked which function enclosed it. The predicate's `SUPER || CONTROL` is real — I read that correctly — but a predicate accepting a modifier says nothing about which dispatch path consults it. True premise, unsupported conclusion.

Compounding it: I had already written in the same review that I could not enumerate the claimed set because my parser reached 3 of 91 predicates. Having said my instrument was inadequate for the general claim, I should not have advanced a specific claim built with the same instrument. Flagging the limitation is not the same as respecting it.

**What survives.** The list was wrong, and for the reason I gave — derived by reading rather than by measuring, which is the same non-method that produced the "any key is forwarded" claim this PR corrects. The measured set (`Ctrl`: B F J P Q S V) differs from the diff's (B F J M P Q S T) by three letters: M and T wrongly named, V missed. So the conclusion held while my evidence for it did not, and those are different things.

Apologies for the noise on a PR that turned out to be correct in the specific I challenged.

PR #335, opened 2026-08-29
docs: agents implement only owner-authored or ready issues

vitali87:
Adds the gate that comes before \`claimed\`: an agent may implement an issue only if it is authored by the repo owner or carries the \`ready\` label. Owner-authored issues need no label; everything else without \`ready\` is untrusted input and off limits. Agents never add \`ready\` themselves.

Groundwork for an automated vetting workflow that grants \`ready\` to external issues after screening them.


vitali87:
Superseded: main stopped tracking CLAUDE.md in cd2ff9f (the agent guide is now a per-clone local file), and this rule already lives in the global ~/.claude/CLAUDE.md on each machine. Nothing left for this PR to change in the repo.

PR #339, opened 2026-08-29
ci: vet external issues automatically before an agent may implement them

vitali87:
Agents implement an issue only if it is owner-authored or carries `ready` (#335). This adds the workflow that grants `ready` to everything else without a human in the loop for the clear cases.

**How it works**

- Owner-authored issues never reach it: the job-level `if` skips them, so no runner starts and nothing hits the inference server.
- `scripts/vet_issue.py prepare` screens deterministically first: HTML comments (invisible on GitHub, visible to an agent), zero-width and bidi characters, base64 blobs, `curl | sh`, token names, accounts younger than 30 days, empty bodies. Any hit goes straight to `needs-human` with no model call.
- What passes is sent to the self-hosted OpenAI-compatible endpoint (same shape as code-graph-rag's highlights step), wrapped as untrusted data with a fixed JSON verdict schema.
- `decide` maps the answer to exactly one of `ready` / `needs-human` / `rejected` and a comment. `ready` carries a restated spec that agents implement instead of the raw body. Missing endpoint, failed call, unparseable answer, low confidence, suspected injection, out of scope, empty spec: all `needs-human`.
- A non-owner edit or comment drops `ready` first, then re-vets. A daily sweep dispatches a vet for any external issue with no verdict.

**Secrets to add** (values as in code-graph-rag's `HIGHLIGHTS_*`): `INFERENCE_API_URL`, `INFERENCE_API_TOKEN`, optional `INFERENCE_RESOLVE`. The workflow falls back to the `HIGHLIGHTS_*` names when present. Until set, external issues land on `needs-human`.

Tests: `python3 -m unittest discover -s scripts/tests` (23, stdlib only), now also a CI step.

## Summary by CodeRabbit

* **New Features**
  * Added automated screening for externally submitted issues across creation, updates, comments, and scheduled reviews.
  * Issues receive exactly one of **ready**, **needs-human**, or **rejected**, along with a maintained vetting comment.
  * Screening detects hidden content, suspicious submissions, incomplete details, and uncertain or unsafe responses.
  * Added safeguards for insecure endpoints and invalid screening results.

* **Tests**
  * Added comprehensive coverage for screening and verdict handling.
  * CI now runs the script test suite automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,8 @@ jobs:
         run: cargo fmt --check
       - name: Clippy (zero warnings)
         run: cargo clippy --all-targets -- -D warnings
+      - name: Issue-vetting script tests (stdlib, no venv)
+        run: python3 -m unittest discover -s scripts/tests -v
       - name: Tests
         # The suite spawns PTYs and real shells; keep the thread count modest
         # so the timing-sensitive terminal tests are not starved.
--- a/.github/workflows/issue-vetting.yml
+++ b/.github/workflows/issue-vetting.yml
@@ -45,6 +45,7 @@ jobs:
     if: |
       github.event_name == 'workflow_dispatch' ||
       (
+        (github.event_name == 'issues' || github.event_name == 'issue_comment') &&
         github.event.issue.pull_request == null &&
         github.event.issue.author_association != 'OWNER' &&
         !(github.event_name == 'issue_comment' && github.event.comment.author_association == 'OWNER') &&
--- a/.github/workflows/issue-vetting.yml
+++ b/.github/workflows/issue-vetting.yml
@@ -168,12 +168,13 @@ jobs:
         run: |
           set -euo pipefail
           # gh's built-in --jq takes no --arg, so the owner is bound in standalone jq.
-          gh issue list --state open --limit 500 --json number,author,labels \
+          # No checkout in this job, so gh needs the repository spelled out.
+          gh issue list --repo "$GITHUB_REPOSITORY" --state open --limit 500 --json number,author,labels \
           | jq -r --arg owner "$GITHUB_REPOSITORY_OWNER" '
               .[] | select(.author.login != $owner)
                   | select(any(.labels[].name; . == "ready" or . == "needs-human" or . == "rejected") | not)
                   | .number' \
           | while read -r n; do
               echo "dispatching vet for #${n}"
-              gh workflow run issue-vetting.yml -f issue="$n"
+              gh workflow run issue-vetting.yml --repo "$GITHUB_REPOSITORY" -f issue="$n"
             done
--- a/.github/workflows/issue-vetting.yml
+++ b/.github/workflows/issue-vetting.yml
@@ -1,0 +1,172 @@
+name: Issue vetting
+
+# Agents implement an issue only if it is owner-authored or carries `ready`
+# (CLAUDE.md). Owner-authored issues never reach this workflow: the job-level
+# `if` skips them at evaluation, so no runner starts and nothing hits the
+# inference server. For everyone else, scripts/vet_issue.py screens the text
+# deterministically, the model on the self-hosted endpoint judges what passes,
+# and the verdict lands as exactly one of ready / needs-human / rejected plus
+# a comment. A non-owner edit or comment drops `ready` first and re-vets, so a
+# stale verdict never survives a change to the text it was given for.
+#
+# Every failure mode is closed: no endpoint, no answer, an unparseable answer,
+# low confidence, suspected injection all land on needs-human.
+
+on:
+  issues:
+    types: [opened, edited, reopened]
+  issue_comment:
+    types: [created, edited]
+  workflow_dispatch:
+    inputs:
+      issue:
+        description: Issue number to vet
+        required: true
+        type: string
+  schedule:
+    # Daily sweep for external issues that missed their event.
+    - cron: "23 4 * * *"
+
+permissions: {}
+
+concurrency:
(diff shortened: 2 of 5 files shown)


vitali87:
## 🤖 Fallback review — Claude Opus 5 (delta: `ccb852bd`)

CodeRabbit reviewed every earlier commit on this PR but not `ccb852bd`, so this local review covers that delta. The follow-up commit `b9455d9` addresses every finding below.

## Walkthrough

The commit hardens `decide()` against a `confidence` that is not a real number (the parent let both `"0.95"` and `true` reach `ready`) and adds `--repo "$GITHUB_REPOSITORY"` to the sweep job's `gh` calls, which have no checkout to infer the repository from. `python3 -m unittest scripts.tests.test_vet_issue` → 29 OK. Only `.github/` and `scripts/` are touched, so no version bump or release note is required.

| File | Summary |
| --- | --- |
| `scripts/vet_issue.py` | explicit `isinstance` guard excluding `bool`; `OverflowError` caught. |
| `.github/workflows/issue-vetting.yml` | `--repo` on both sweep commands. |
| `scripts/tests/test_vet_issue.py` | boolean and oversized-integer confidence cases. |

## Findings

- **`.github/workflows/issue-vetting.yml:169,177-180`** **[major/correctness]** `set -euo pipefail` makes the `while read` loop abort on the first failed `gh workflow run`, so one bad issue strands every later one. Verified with a four-line shell reproduction. Add `|| echo "::warning::…"` as the file already does for best-effort label removals.
- **`.github/workflows/issue-vetting.yml:174`** **[major/security]** `select(.author.login != $owner)` is true for a **null** author (deleted account), so an authorless issue flows past the owner filter — fail-open in a gate whose premise is fail-closed. Verified with jq. Require a positive login.

## Test coverage

- String `confidence` is the regression the commit actually fixes and nothing tests it: add `test_string_confidence_needs_human`.
- No confidence test asserts `.flags`.
- The sweep's jq filter has no test.

## Nitpicks

- Two stacked comment blocks on one guard, the first stale relative to the code beneath it.
- `json.dumps(...).replace("0.95", …)` is silently order-dependent on `0.95` appearing once.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: the Python change is correct and minimal; in-diff findings are a stale comment, a string-patched fixture, and the untested string path.
- **Greptile score (codebase fit) — 2/5**: two majors whose evidence is outside the diff (the `set -e` interaction with the loop; jq's null semantics against the fail-closed invariant).

VERDICT: REQUEST_CHANGES

vitali87:
## 🤖 Fallback review — Claude Opus 5 (re-review: `b9455d9`)

Re-review of the follow-up commit; CodeRabbit had not reviewed the last two commits on this PR.

## Walkthrough

This follow-up addresses the two majors and the missing test from the previous fallback review. The sweep's dispatch loop swallows a single failed `gh workflow run` behind `|| echo "::warning::…"` so one bad issue no longer strands the rest under `set -e`; the jq owner filter now demands a positive `login`, so an authorless issue cannot slip past a fail-closed gate; and `decide()` gains a test for the string-`confidence` path. The `vet_issue.py` change is comment-only.

| File | Summary |
| --- | --- |
| `.github/workflows/issue-vetting.yml` | jq filter requires a non-empty `.author.login`; dispatch failure warns and continues. |
| `scripts/vet_issue.py` | Comment-only: the stacked guard comments merged into one block matching the code. |
| `scripts/tests/test_vet_issue.py` | New `test_string_confidence_needs_human`; the oversized-integer test asserts the fixture shape and the `low-confidence` flag. |

## Findings

Both previously reported majors are resolved, verified rather than assumed:

- `.github/workflows/issue-vetting.yml:181-182` — with the old body a failure on the second of three items exits at `rc=1` having dispatched only two; with the fix all three iterate, the warning prints, `rc=0`. Resolved.
- `.github/workflows/issue-vetting.yml:174` — against a fixture with a `null` author, the owner, an external issue, and `login: ""`, the old filter emitted `1 3 4`; the new one emits only `3`. `.author.login` on a null author yields `null`, so `// ""` is the right idiom. Resolved.
- `scripts/tests/test_vet_issue.py:156-158` — a real positive control: the pre-fix `float("0.95")` path clears `CONFIDENCE_MIN` and would have returned `ready`; the current code returns `needs-human`.

No new findings. Probed and found no regression: `pipefail` still guards the head of the pipe (a failing `gh issue list` or a jq parse error still fails the step); coercing a bad confidence to 0 cannot mislabel a `reject` as `ready`; the `workflow_dispatch` path re-checks ownership independently (`author_association == OWNER` → `vet=false`); `python3 -m unittest` → 30 OK and the suite runs in CI. No version bump required (nothing under `src/`).

## Test coverage

- The sweep's jq filter has no test, and it is the one security-relevant predicate here that has already been wrong once. Feed it a fixture with a `null` author, an `""` login, the owner, an already-labeled issue, and a clean external one; assert only the last is emitted. *(Added in the following commit, `scripts/tests/test_sweep_filter.py`.)*
- The continue-on-failure loop has no test; lower value.
- Pre-existing: `any(.labels[].name; …)` errors on an item without a `labels` key; unreachable with `gh --json labels`, not filed.

## Nitpicks

- `(.author.login // "") != "" and .author.login != $owner` evaluates `.author.login` twice; taste.
- The `assertNotIn("0.95", …)` message describes what is relied on rather than what failed.
- The merged comment's trailing parenthetical would scan better on its own line.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: no findings against the diff; the test is a verified positive control and the workflow fix is minimally scoped.
- **Greptile score (codebase fit) — 4/5**: both prior majors resolved against every trigger path; held at 4 for the untested jq filter (since addressed).

VERDICT: APPROVE

PR #386, opened 2026-08-29
fix(tasks): de-duplicate discovered tasks by label, not command line

### LINKED ISSUE #336: Task discovery drops a task whose command another source already used, so its label becomes unfindable
## What

`discover_tasks` de-duplicates by **command line**, while everything that looks a task up does so by **label**. When two sources declare the same command under different labels, one label disappears from croft entirely.

```rust
fn dedup(tasks: Vec<Task>) -> Vec<Task> {
    let mut seen = BTreeSet::new();
    tasks.into_iter().filter(|t| seen.insert(t.command.clone())).collect()
}
```

`src/app/mod.rs` then resolves a `preLaunchTask` with `tasks.into_iter().find(|t| t.label == task_label)`.

## Demonstrated

A Rust workspace whose `tasks.json` declares a build task under its own name:

```jsonc
// .vscode/tasks.json
{ "tasks": [ { "label": "Build Project", "type": "shell", "command": "cargo build" } ] }
```

`discover_tasks` returns:

```
label="Build Project"  command="cargo build"
label="cargo test"     command="cargo test"
label="cargo clippy"   command="cargo clippy"
label="cargo run"      command="cargo run"
```

The Cargo source ran - `cargo test`, `cargo clippy` and `cargo run` are all present - but **`cargo build` is gone**, dropped because `tasks.json` reached the same command line first under a different label.

## Why it matters

`"preLaunchTask": "cargo build"` is a reasonable thing to write, and croft would normally offer that task. With the above `tasks.json` present it fails:

> preLaunchTask "cargo build" not found — Tasks: Run Task lists what the workspace declares

The message is true and the advice is useless: the workspace *does* declare it, and the list croft points at is the one the task was removed from. The user has no way to see why, since nothing reports the drop.

Same shape for Tasks: Run Task - the label is simply missing from the picker - and for anything else keyed on labels.

## Note on intent

The doc comment says "Duplicate command lines are dropped, first source wins", so this is deliberate, and the intent is sound: the Makefile, Cargo and package.json sources overlap heavily, and listing `cargo build` three times would be noise. The bug is that de-duplication throws away the *labels*, which are the identifiers the rest of the system uses.

## Possible directions

- De-duplicate by `(label, command)` and let identical labels collapse, which handles the real overlap without losing names.
- Or keep command-level de-duplication but retain the dropped labels as aliases, so a lookup by any declared label finds the surviving task.
- Either way, a label a user wrote should never resolve to "not found".

Found while implementing #318: two test tasks with the same `command` silently collapsed into one, which took a while to explain.

vitali87:
## Summary
- `discover_tasks` de-duplicated by **command line** while every lookup (`preLaunchTask`, Tasks: Run Task) keys on **label**, so a `tasks.json` entry reusing a convention source's command under its own name made the convention label unfindable.
- De-duplication now keys on the label. Identical labels still collapse to the first source; a relabelled command no longer erases the original task.
- Bump to 0.1.793 with a matching release note.

## Test plan
- [x] Red: new `a_relabelled_command_keeps_the_convention_label_findable` failed on main with `["Build Project", "cargo test", "cargo clippy", "cargo run"]` — exactly the issue's demonstration.
- [x] Green after the fix; `tasks::` module passes (11 tests); clippy `-D warnings` clean; `cargo fmt --check` clean.

Closes #336

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.792"
+version = "0.1.793"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.792"
+version = "0.1.793"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/release_notes.rs
+++ b/src/release_notes.rs
@@ -60,6 +60,6 @@ pub struct ReleaseNote {
 
 /// What shipped in the current release. Replace on every version bump.
 pub const RELEASE_NOTES: &[ReleaseNote] = &[ReleaseNote {
-    kind: NoteKind::Feature,
-    summary: "A debug compound's own preLaunchTask now runs before its configuration starts, waiting for it to exit 0 — so a compound that builds first launches instead of being refused.",
+    kind: NoteKind::Fix,
+    summary: "Task discovery no longer loses a label: a tasks.json entry that reuses a Makefile, Cargo or package.json command under its own name used to drop the convention task entirely, so a preLaunchTask or Tasks: Run Task pick by that name reported \"not found\". Tasks are now de-duplicated by label, the identifier everything looks up.",
 }];
--- a/src/release_notes.rs
+++ b/src/release_notes.rs
@@ -61,5 +61,5 @@ pub struct ReleaseNote {
 /// What shipped in the current release. Replace on every version bump.
 pub const RELEASE_NOTES: &[ReleaseNote] = &[ReleaseNote {
     kind: NoteKind::Fix,
-    summary: "Task discovery no longer loses a label: a tasks.json entry that reuses a Makefile, Cargo or package.json command under its own name used to drop the convention task entirely, so a preLaunchTask or Tasks: Run Task pick by that name reported \"not found\". Tasks are now de-duplicated by label, the identifier everything looks up.",
+    summary: "Task discovery no longer loses a label: a tasks.json entry that reuses a Makefile, Cargo or package.json command under its own name used to drop the convention task entirely, so a preLaunchTask or Tasks: Run Task pick by that name reported \"not found\". Tasks are now de-duplicated by label and command together, so a label the workspace declares always survives.",
 }];
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -383,7 +383,11 @@ pub(crate) fn strip_jsonc(text: &str) -> String {
 /// reported "not found" for a task the workspace plainly declared (#336).
 /// The one thing that still collapses is a true repeat: the same label
 /// running the same line. A user who relabels `cargo build` sees it listed
-/// under both names, by design.
+/// under both names, by design. Two tasks sharing a label with DIFFERENT
+/// commands both survive here; the lookups keyed on the label alone
+/// (`preLaunchTask`, task-pane reuse) then resolve to the first, so a
+/// workspace that reuses a label across commands gets first-source-wins at
+/// the lookup, not in this list.
 fn dedup(tasks: Vec<Task>) -> Vec<Task> {
     let mut seen = BTreeSet::new();
     tasks
@@ -578,9 +582,8 @@ mod tests {
     }
 
     /// Only a true repeat collapses: the same label running the same line.
-    /// A tasks.json entry that reuses a label with a different command is
-    /// first-source-wins on the label; one that relabels a convention
-    /// command lists under both names, by design.
+    /// One that relabels a convention command lists under both names, by
+    /// design.
     #[test]
     fn only_a_repeated_label_and_command_pair_collapses() {
         let tmp = tempfile::tempdir().unwrap();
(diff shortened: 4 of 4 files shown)


vitali87:

CodeRabbit rate-limited on this PR ("Review limit reached"), so this local review of equivalent shape stands in. It reviewed the first commit (`6624fe7`); the follow-up commit addresses every finding below.

## Walkthrough

`discover_tasks` collapsed duplicates by **command line** while every consumer (`preLaunchTask` resolution, the Run Task picker, pane naming) keys on **label**. This PR flips the dedup key from `t.command` to `t.label`, adds a regression test, and bumps the version with a replacement release note. The flip is a straight swap rather than a widening: labels are *not* unique, because `vscode_tasks` defaults a missing `label` to the bare `command` field, not the args-joined line.

| File | Summary |
| --- | --- |
| `src/tasks.rs` | `dedup` keys on `t.label`; docs updated; one new test. |
| `src/release_notes.rs` | Single Fix note. |
| `Cargo.toml` / `Cargo.lock` | Version bump. |

## Findings

- **`src/tasks.rs:384`** **[blocker/correctness]** Keying dedup on `label` silently deletes distinct, runnable tasks. `src/tasks.rs:94-98` defaults a label-less entry's label to the bare `command` field while `command` is the args-joined line, so every label-less entry sharing a program shares a label. Verified on the branch with three label-less `npm` entries (`run build` / `run test` / `run lint`): **`npm run test` and `npm run lint` are gone**; `origin/main` returns all three. Fix: dedup on `(label, command)`, and/or default the label to the full line.
- **`src/tasks.rs:384`** **[major/correctness]** A `Makefile` `build` target plus `{"label":"Build","command":"make build"}` now lists the same command line twice. If suppressing that overlap matters, it needs to be deliberate and noted.
- **`src/tasks.rs:374-379`** **[minor/docs]** The new doc asserts the label "is the only thing safe to collapse"; the probes above contradict it.

## Test coverage

- Label-less `tasks.json` entries sharing a `command` — nothing constructs a task without an explicit `label`, so the fallback path is unexercised. Add a test with two args-only `npm` entries asserting both commands survive.
- The overlap-suppression property has no test in either direction.
- `run_project_task` pane reuse by label with two same-label tasks is untested.

## Nitpicks

- `src/tasks.rs:548` the failure message interpolates the stale pre-rewrite `labels`.
- The new test bundles two behaviours under one name.

## Scores

- **CodeRabbit score (diff quality) — 1/5**: the changed line deletes user-declared tasks for ordinary label-less input; the new test does not exercise the breaking path.
- **Greptile score (codebase fit) — 1/5**: the defect is only visible from `vscode_tasks`'s label fallback and the label-keyed consumers; the overlap-suppression property is discarded incidentally.

VERDICT: REQUEST_CHANGES

vitali87:

CodeRabbit is still rate-limited on this PR, so this stands in again. This pass reviews both commits, with the earlier findings re-verified against `48671c7`.

### Status of the previous findings

| Previous finding | Status |
| --- | --- |
| **blocker** — label-only dedup swallowed label-less `tasks.json` entries sharing a program | **Fixed and verified.** Compiled `src/tasks.rs` at branch state standalone: three label-less `npm` entries now yield three tasks. |
| **major** — same command line listed twice | **Resolved as a deliberate, documented decision** (issue #336 itself proposes `(label, command)`). |
| **minor** — doc asserted the label "is the only thing safe to collapse" | Rewritten, but introduces a different inaccuracy, below. |

## Walkthrough

Dedup now keys on the `(label, command)` pair and a label-less `tasks.json` entry is labelled by its full args-joined line, so distinct label-less entries no longer collide. Version 0.1.792 → 0.1.793 with a replacement release note.

| File | Summary |
| --- | --- |
| `src/tasks.rs` | `dedup` keys on `(t.label, t.command)`; label-less entries labelled by full command line; docs rewritten; three new tests. |
| `src/release_notes.rs` | Single `Fix` note. |
| `Cargo.toml` / `Cargo.lock` | 0.1.792 → 0.1.793. |

## Findings

- **`src/tasks.rs:580-583`** and **`src/tasks.rs:381-382`** **[major/correctness]** Both doc comments claim a repeated label with a *different* command collapses first-source-wins. It does not: `dedup` inserts the `(label, command)` tuple, so `{"label":"Build","command":"make a"}` and `{"label":"Build","command":"make b"}` both survive (verified by probe). This matters because `src/app/mod.rs:18439` resolves `preLaunchTask` with `find(|t| t.label == …)` and the task pane is named and reused by label (`src/app/mod.rs:21971`, `22001-22006`; `src/app/tests.rs:6231` pins "exactly one pane carries the task's name"), so same-label duplicates now silently land on consumers that assume uniqueness. Fix: make the prose match the code and say what a collision costs at the lookup. *(Addressed in the following commit.)*
- **`src/release_notes.rs:63`** **[minor/docs]** The note says "de-duplicated by label" — true of the first commit, not of the shipped `(label, command)` key. *(Addressed in the following commit.)*

## Test coverage

- Same label, different command: untested in either direction. Add a test asserting both survive and that a label lookup resolves to the first declared. *(Added.)*
- Pane reuse under a label collision: `src/app/tests.rs:6204-6238` pins reuse for a unique label only.
- `default_build_task` with a relabelled duplicate: behaves correctly by probe, unpinned.

## Nitpicks

- `src/tasks.rs:587` the Makefile row in `only_a_repeated_label_and_command_pair_collapses` never reaches the assertions (the two identical `tasks.json` rows claim the key first); assert the surviving row's `source` to make the ordering claim real. *(Done.)*
- The `dedup` doc carries the issue's whole backstory; could compress.

## Convention compliance

Version bump ✅; release notes replaced, not appended ✅; `scripts/check_doc_ownership.py origin/main origin/fix/task-dedup-by-label` → no documentation lost ✅.

## Scores

- **CodeRabbit score (diff quality) — 3/5**: exactly one major, in the new lines — doc comments asserting a first-source-wins-on-label behaviour the changed `dedup` does not implement, unpinned by any test. The core fix is correct and the label-less tests cover the previously broken path.
- **Greptile score (codebase fit) — 3/5**: the same major seen from outside — `preLaunchTask` and pane reuse assume label uniqueness. Release conventions otherwise satisfied and verified.

VERDICT: REQUEST_CHANGES

vitali87:
## 🤖 Fallback review (round 3) — Claude Opus 5

CodeRabbit is still rate-limited on this PR, so this stands in again. This pass reviews the follow-up commit `045905c` only, with the round-2 findings re-verified against it.

### Status of the round-2 findings

| Round-2 finding | Status |
| --- | --- |
| **major** — both doc comments claimed a reused label folds first-source-wins, which `dedup` does not do | **Fixed.** `src/tasks.rs:374-382` now states both survive and the label-keyed lookups resolve to the first; the test doc drops the false clause. Verified against the consumers: `src/app/mod.rs:18483` and `:18530` both resolve `preLaunchTask` with `find(|t| t.label == task_label)`, so "first-source-wins at the lookup, not in this list" is literally true. |
| **minor** — release note said "de-duplicated by label" | **Fixed.** `src/release_notes.rs:64` now names the `(label, command)` key. |
| **nit** — the Makefile row never reached an assertion | **Fixed.** The survivor's `source` is asserted; verified load-bearing (fails when `dedup` is reverted). |

## Walkthrough

The follow-up rewrites the two doc comments, corrects the release note, tightens the collapse test so its Makefile row is reachable, and adds a test pinning that a label reused across two commands keeps both entries. `dedup` itself is unchanged.

| File | Summary |
| --- | --- |
| `src/tasks.rs` | `dedup` doc extended; collapse test gains a `source` assertion; one new test. No behaviour change. |
| `src/release_notes.rs` | Note reworded to the `(label, command)` key. |

## Findings

- **`src/tasks.rs:624-653`** **[minor/testing]** `a_label_reused_for_a_different_command_keeps_both_and_the_lookup_takes_the_first` is not a positive control for the dedup key — it passes unchanged against `main`'s command-only `dedup` and against the branch with `dedup` reverted (verified by compiling both standalone), because `make a` and `make b` are distinct under either key. It is a valid pin of the lookup contract, which is what round 2 asked for; the key itself is guarded by the two pre-existing tests. A one-line comment saying so would prevent misreading.

No blockers, majors, or new correctness defects.

## Test coverage

- Pane reuse under a label collision: the new doc names "task-pane reuse" as a consumer; `src/app/tests.rs:6204-6238` pins reuse for a unique label only.
- `default_build_task` with a relabelled duplicate: correct by probe, unpinned.

## Nitpicks

- The second `assert_eq!` in the new test is entailed by the first.
- The `dedup` doc is nine lines and still carries the #336 backstory.

## Convention compliance

Version 0.1.792 → 0.1.793 ✅; release note replaced ✅; no documentation lost ✅.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: both round-2 findings genuinely fixed; the only deduction is a new test that cannot fail for the reason its name implies.
- **Greptile score (codebase fit) — 4/5**: the prose about label-keyed consumers is accurate against `src/app/mod.rs:18483` and `:18530`; the pane-reuse consumer remains untested under a collision.

VERDICT: APPROVE

PR #387, opened 2026-08-29
fix(pair): tell an unopenable host lock apart from a busy one

### LINKED ISSUE #337: A navigator lock that cannot be opened reports "hosted by another croft window"
## What

`try_acquire_pair_host_lock` folds two very different failures into one `None`:

```rust
pub(crate) fn try_acquire_pair_host_lock(path: &Path) -> Option<PairHostLock> {
    let file = std::fs::OpenOptions::new()
        .create(true).truncate(false).write(true)
        .open(path)
        .ok()?;                       // <- any open failure, indistinguishable
    let rc = unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_EX | libc::LOCK_NB) };
    if rc != 0 { return None; }       // <- someone else genuinely holds it
    Some(PairHostLock { _file: file })
}
```

The caller reports both as the second one:

> Navigator 'nav' is hosted by another croft window in this workspace

So a lock file croft could not create or open - a read-only or missing config directory, a permissions problem, an fd limit - is reported to the user as another croft window holding the navigator. There is no other window. Nothing in the message points at the real cause, and the suggested remedy (close the other window) does not exist.

## Evidence

`app::tests::a_record_rewrite_re_arms_a_downed_navigator` failed in CI with exactly that message:

```
a rewritten record must re-arm the downed navigator:
  Navigator 'nav' is hosted by another croft window in this workspace
```

That test points `pair_host_lock_path` at a freshly created `tempfile::tempdir()`, so no second process can be holding an flock on it - the `None` cannot have come from the `flock` arm. It came from `open`, and the message misattributed it. The test passes locally and in isolation; it failed on a loaded CI runner, which is consistent with a resource-limit open failure rather than contention.

## Why it is worth fixing beyond the test

The lock is what stops two croft windows both seating a navigator and corrupting a shared buffer, so it is load-bearing. Reporting an environmental failure as contention means:

- the user chases a window that does not exist;
- a genuine "someone else holds it" and a broken config directory look identical in the status line;
- the takeover poll keeps retrying either way, so the state never resolves and never explains itself.

## Suggested shape

Return something that distinguishes the two - `Result<PairHostLock, LockError>` with `Busy` and `Io(std::io::Error)` - and let the caller say which happened. The busy case keeps today's wording; the io case should name the path and the error, since that is the only actionable half.

Found while triaging a CI failure during #315.

vitali87:
## Summary
- `try_acquire_pair_host_lock` folded "another croft holds the flock" and "the lock file could not be opened" into one `None`; the caller reported both as *hosted by another croft window*, so a missing/read-only config directory or an fd limit sent the user after a window that does not exist (#337).
- It now returns `Result<PairHostLock, PairHostLockError>` with `Busy` and `Io(std::io::Error)`. A flock failure other than `EWOULDBLOCK` is `Io` as well.
- The caller keeps the existing wording for `Busy` and, for `Io`, names the lock path and the OS error. Both stay announced once per denial, as before.
- Bump to 0.1.794 with a matching release note.

## Test plan
- [x] Red: `an_unopenable_lock_file_is_an_io_error_not_busy` and the reshaped exclusivity test fail to compile against the old `Option` API.
- [x] Green: `session::tests::` (49 passed), `a_record_rewrite_re_arms_a_downed_navigator` passes, clippy `-D warnings` clean, `cargo fmt --check` clean.

Closes #337

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.793"
+version = "0.1.794"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.793"
+version = "0.1.794"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/docs/MULTIPLAYER.md
+++ b/docs/MULTIPLAYER.md
@@ -612,8 +612,10 @@ interaction into turn-based driver/navigator pairing:
   `flock` on `<hash>.pair-host.lock` (`session::try_acquire_pair_host_lock`)
   guards it, so a second window is refused ("hosted by another croft
   window") instead of both claiming owner site 1 and corrupting the buffer.
-  The OS drops the lock when the holder exits, so a crashed host hands off
-  automatically. (The non-hosting window does not yet observe the owner's
+  A lock file that cannot be opened at all (a missing or read-only config
+  directory, an fd limit) is reported as exactly that, with the path and
+  the OS error, never as contention (#337). The OS drops the lock when the
+  holder exits, so a crashed host hands off automatically. (The non-hosting window does not yet observe the owner's
   edits as a guest — a documented follow-up; `croft attach` is the shared
   path for that today.)
 - **Ask turns (may edit).** Right-click the gutter ("Ask Navigator"), a
@@ -701,7 +703,8 @@ interaction into turn-based driver/navigator pairing:
   fixing the backend all re-activate. A window whose spawn failed also
   releases the single-host lock (and its same-tick owner self-appointment),
   so another croft window in the workspace can host instead; the losing
-  window announces the refusal once and keeps polling silently for
+  window announces the refusal once per reason (a busy lock that becomes
+  an unopenable one, or the reverse, is announced again) and keeps polling silently for
   takeover. Teardown on unseat runs on a
   detached thread so the 2s grace-kill never freezes the UI, and the exit
   paths (drop-to-local, self-update exec) reap the child synchronously
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -3272,7 +3272,8 @@ pub struct App {
     /// repaint). The KIND is latched, not just the fact, so a refusal that
     /// changes reason (an fd limit clears and another window now holds the
     /// lock, or the reverse) is announced again rather than leaving a
-    /// message that is no longer true (#337).
+    /// message that is no longer true (#337). Per kind, not per message: an
+    /// open failure whose errno changes keeps the first announcement.
     pair_lock_denied: Option<PairLockDenial>,
     /// Proactive navigator looks (docs/MULTIPLAYER.md): a completed new
     /// construct plus a typing pause hands the seated navigator a
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -20492,11 +20492,11 @@ impl App {
             // host (keep checking — take over when that croft exits).
             if self.pair_host_lock.is_none() {
                 match crate::session::try_acquire_pair_host_lock(&self.pair_host_lock_path) {
-                    Some(lock) => {
+                    Ok(lock) => {
                         self.pair_host_lock = Some(lock);
                         self.pair_lock_denied = false;
                     }
-                    None => {
+                    Err(e) => {
                         // Announced ONCE: the takeover poll keeps running
                         // every second, and re-announcing clobbered the
                         // window's status line forever at a 1 Hz repaint.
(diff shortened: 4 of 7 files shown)


vitali87:

CodeRabbit rate-limited on this PR ("Review limit reached"), so this local review of equivalent shape stands in. It reviewed the first commit; the follow-up commit addresses every finding below.

## Walkthrough

`try_acquire_pair_host_lock` returned `Option<PairHostLock>`, folding "another croft holds the flock" and "the lock file could not be opened" into one `None`; the caller reported both as *hosted by another croft window* (#337). The change introduces `PairHostLockError { Busy, Io(std::io::Error) }`, maps a non-`EWOULDBLOCK` flock failure to `Io`, and gives the caller a distinct message per variant.

| File | Summary |
| --- | --- |
| `src/session.rs` | `PairHostLockError`; `try_acquire_pair_host_lock` returns `Result`; tests reshaped and one added. |
| `src/app/mod.rs` | Caller matches `Ok`/`Err`; `Io` names the lock path and OS error. |
| `src/release_notes.rs`, `Cargo.toml`, `Cargo.lock` | Release hygiene. |

## Findings

- `src/app/mod.rs:20412-20415` **[major/correctness]** The `pair_lock_denied` latch is a single `bool` shared by both error kinds, so a *transition* between them is swallowed: the first message latches forever while the true state changes (an fd-limit `Io` clears, the window is then genuinely `Busy`). Latch the *kind*, not just the fact.
- `src/session.rs:119-123` **[major/correctness]** `EINTR` is classified as an environment error and shown to the user. Every other syscall site in the repo (`session_host.rs`, `collab.rs`, `clipboard.rs`) retries `Interrupted`. Retry rather than report.
- `src/session.rs:120` **[minor/correctness]** `last_os_error()` is read a statement after the `flock` call; capture it in the same statement so nothing can clobber `errno`.
- `src/release_notes.rs:62-66` **[minor/docs]** Branched before #334 landed on `main`; rebase so the replacement is against the current notes.

## Test coverage

- The `Io` status string is untested at the caller: point `pair_host_lock_path` at a nonexistent directory, seat, assert `cannot open host lock` and NOT `another croft`.
- The latch-across-kinds bug has no test: `Io` first, then make the path openable and hold the flock from a second fd, assert the status switches to the `another croft` wording.

## Nitpicks

- `src/session.rs:113` `EBADF` is unreachable on a freshly opened fd; `ENOLCK`/`EROFS` are the realistic ones.
- `PairHostLockError` could implement `Display`/`Error`; not required.
- `Busy` is unconstructible on non-Unix; harmless.

## Scores

- **CodeRabbit score (diff quality) — 2/5**: two majors inside the changed lines (EINTR, shared latch) plus the errno window and an untested match arm.
- **Greptile score (codebase fit) — 2/5**: EINTR handling contradicts the repo's retry convention; the stale branch point.

VERDICT: REQUEST_CHANGES

vitali87:

CodeRabbit is still rate-limited on this PR, so this stands in again. Re-review of `cd4aaf8` against the four findings from the previous round.

## Walkthrough

`try_acquire_pair_host_lock` folded "another croft holds the flock" and "the lock file could not be opened" into one `None` (#337). The change returns `Result<PairHostLock, PairHostLockError>` with `Busy`/`Io`, retries `EINTR`, and gives the caller a distinct once-per-reason message per kind.

| File | Summary |
| --- | --- |
| `src/session.rs` | `PairHostLockError { Busy, Io }`; `Result` return; `EINTR` retry loop with errno captured in the same statement; tests reshaped and added. |
| `src/app/mod.rs` | `PairLockDenial` enum; `pair_lock_denied` becomes `Option<PairLockDenial>` so the denial KIND is latched; `Io` names path + OS error. |
| `src/app/tests.rs` | App-level test walking Io → Busy → host. |
| `Cargo.toml`, `Cargo.lock`, `src/release_notes.rs` | 0.1.792 → 0.1.794; notes replaced with a single `Fix` entry. |

### Previous findings — all four resolved (verified)

- **EINTR as an environment error** — fixed: `src/session.rs:121-130` retries, matching `src/session_host.rs:523,2214,2254`, `src/collab.rs:742`, `src/clipboard.rs:543`.
- **bool latch swallowed a change of reason** — fixed: `Option<PairLockDenial>`, compared by kind, so Io→Busy re-announces.
- **errno read a statement late** — fixed: `last_os_error()` immediately follows `flock`.
- **stale branch point** — fixed: merge base is `origin/main`'s tip.

Verified in a throwaway worktree: 4 relevant tests pass (`an_unopenable_lock_file_is_an_io_error_not_busy`, `an_unopenable_host_lock_is_reported_as_such_and_re_announced_when_the_reason_changes`, `only_one_croft_self_appoints_the_navigator_owner`, `a_record_rewrite_re_arms_a_downed_navigator`); clippy `--all-targets -- -D warnings` clean; fmt clean; version bumped and `src/release_notes.rs` replaced, not appended; doc-ownership rule satisfied; every consumer of the changed API migrated.

## Findings

- `docs/MULTIPLAYER.md:611-614` **[minor/docs]** Names `session::try_acquire_pair_host_lock` and still describes a single-outcome gate ("a second window is refused"); `:702` no longer mentions that the announcement re-fires when the reason changes. *(Addressed in the following commit.)*

## Test coverage

- The `Busy` → `Io` direction of the latch, which the field doc explicitly promises, is not walked by the test. *(Addressed in the following commit.)*
- A changed `Io` reason within the same kind keeps the first announcement — deliberate (kind, not message), but undocumented. *(Field doc now says so.)*
- The `EINTR` retry has no test; delivering a signal mid-`flock` deterministically is awkward and one is not asked for.

## Nitpicks

- `src/session.rs:131-132` "anything else (ENOLCK, EROFS)" reads as exhaustive; "e.g." would be clearer.
- The "unchanged refusal stays quiet" assertion relies on the following status `assert_eq!` for the real signal.
- Version ordering across #386/#387/#388/#389 (793–796) does not collide today; re-check at merge time.
- `PairHostLockError` implements neither `Display` nor `Error`; nothing needs it yet.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: both prior majors genuinely fixed and verified green; held off 5 by the then-untested `Busy` → `Io` direction.
- **Greptile score (codebase fit) — 4/5**: one minor outside the diff (`docs/MULTIPLAYER.md` stale contract); EINTR now matches the repo convention, every consumer migrated, release rule satisfied.

VERDICT: APPROVE

PR #388, opened 2026-08-29
test(pair): wait for the seat's site ids, not for a neighbouring caret

### LINKED ISSUE #338: Flaky under load: ask_and_reply_turns_park_the_navigator_caret waits for one thing and asserts another
## What

`pair::tests::ask_and_reply_turns_park_the_navigator_caret` (`src/pair.rs`) waits for one piece of state and then asserts on a different one:

```rust
harness.wait_until("the ask turn's caret", |h| {
    h.carets().iter().any(|(n, r, _)| n == "asker" && *r == 1)
});
assert!(!asker.caret_sites().is_empty(), "the seat exposes its per-file site ids");
```

The wait is for the caret appearing on the harness; the assertion is that the seat has recorded its per-file site ids. Those are two different things, so the wait does not establish what the assert requires - it just usually arrives first. Under load it does not, and the assertion fires.

## Evidence that it is load, not a defect in the code under test

| run | result |
|---|---|
| full suite, twice | FAILED on this test |
| full suite, third run, same tree | 3587 passed, clean |
| full suite on unmodified `main` | 3583 passed, clean |
| `pair::` module alone under ten busy cores, 4x on the branch and 4x on `main` | 8/8 clean |

It needs the whole suite's contention to reproduce, and it reproduces on `main` as readily as on any branch.

## Why it is not #307's shape

#307 was about budgets that were too small: a test waited for something that had to happen and gave up early. Scaling the deadline fixed those. Here there is no budget to scale - the wait succeeds. The test asserts a state nothing waited for, which is closer to #326's shape (an assertion on an instant that is not guaranteed).

Fix is to wait for the condition actually asserted, rather than for a neighbouring one that usually precedes it.

## History

Originally recorded as a comment on #307 while that issue was still open. #307 has since been closed by #325, so this is filed separately rather than left on a closed issue where it would not be found.

vitali87:
## Summary
- `ask_and_reply_turns_park_the_navigator_caret` waited for the ask turn's caret to reach the owner, then asserted that the seat had recorded its per-file site ids — a different arrival that usually, but not always, lands first (#338).
- The assertion is now a `wait_until` on the state it checks. No behaviour change in the code under test.
- Bump to 0.1.795 (CI requires it for any `src/` change) with a release note saying so.

## Test plan
- [x] Test passes 3/3 locally; clippy `-D warnings` clean; `cargo fmt --check` clean.
- [ ] The flake only reproduces under the whole suite's load, so the proof is CI staying green over the next runs rather than a single local pass.

Closes #338

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.794"
+version = "0.1.795"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.794"
+version = "0.1.795"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/pair.rs
+++ b/src/pair.rs
@@ -2733,10 +2733,12 @@ mod tests {
         harness.wait_until("the ask turn's caret", |h| {
             h.carets().iter().any(|(n, r, _)| n == "asker" && *r == 1)
         });
-        assert!(
-            !asker.caret_sites().is_empty(),
-            "the seat exposes its per-file site ids (its wire identity)"
-        );
+        // The caret reaching the owner and the seat recording its site id
+        // are two arrivals, not one: the assertion must wait for the state
+        // it checks, not for a neighbour that usually lands first (#338).
+        harness.wait_until("the seat's per-file site ids (its wire identity)", |_| {
+            !asker.caret_sites().is_empty()
+        });
 
         let mut cmd = Command::new("cat");
         cmd.stdin(Stdio::piped());
--- a/src/release_notes.rs
+++ b/src/release_notes.rs
@@ -61,5 +61,5 @@ pub struct ReleaseNote {
 /// What shipped in the current release. Replace on every version bump.
 pub const RELEASE_NOTES: &[ReleaseNote] = &[ReleaseNote {
     kind: NoteKind::Fix,
-    summary: "A navigator host lock croft cannot create or open (a missing or read-only config directory, an fd limit) is now reported as exactly that, with the path and the error, instead of as \"hosted by another croft window\" when no other window exists.",
+    summary: "No user-facing change: a load-sensitive navigator test now waits for the state it asserts (the seat's per-file site ids) rather than for the caret that usually lands just before it.",
 }];


vitali87:

CodeRabbit rate-limited on this PR ("Review limit reached"), so this local review of equivalent shape stands in.

## Walkthrough

The PR fixes a load-sensitive flake in `pair::tests::ask_and_reply_turns_park_the_navigator_caret` (#338): the test waited for the ask turn's caret to reach the owner harness, then immediately `assert!`ed that the seat had recorded its per-file site ids. Those are two independent arrivals over the relay, and the second only *usually* lands first. The fix converts the bare assertion into a `harness.wait_until` on the same predicate. No production code changes; the version bump and release-notes replacement are the repo's mandatory release hygiene for any `src/` touch.

| file | summary |
|---|---|
| `src/pair.rs` | Replaces the bare `assert!(!asker.caret_sites().is_empty(), ...)` with `harness.wait_until(...)` on the same predicate, plus a comment citing #338. |
| `Cargo.toml` / `Cargo.lock` | Version bump. |
| `src/release_notes.rs` | Single `NoteKind::Fix` note stating there is no user-facing change. |

## Findings

- `src/release_notes.rs:62-66` **[minor/correctness]** Built on a stale base (`f05b3f5`, v0.1.791) while `origin/main` is at 0.1.792: `git merge-tree` shows real conflicts in `src/release_notes.rs`, `Cargo.toml`, `Cargo.lock`. Loud, not silent — a rebase chore. The CI gate passes legitimately (compares against the merge base).
- `src/pair.rs:2739` **[minor/tests]** The wait swallows the instant-failure diagnostic: a genuine regression now costs the 60s `wait_until` deadline instead of failing fast. The message text was preserved into `what`, and this is the accepted trade across every other `wait_until` in the file.

Checked and could not substantiate: the wait cannot hang on `caret_sites`' `try_lock` cache (the only contender is the pump thread, holding the lock for microseconds per 25ms cycle against a 10ms poll); no `caret_sites` consumer is affected; the release-hygiene rule applies (inline `mod tests` in `src/pair.rs` counts as shipped) and is satisfied.

## Test coverage

Nothing missing. The diff *is* the test change and strictly strengthens an existing assertion. The fix's actual claim — the flake is gone — can only be proven by CI staying green across subsequent runs, which the PR body says.

## Nitpicks

- `src/pair.rs:2739` the only `wait_until` in the file whose closure discards the harness (`|_|`) and closes over `asker`; correct, and the comment explains why.
- The "No user-facing change" release note is the honest shape for a test-only bump.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: the only in-diff finding is the accepted slower-diagnostic trade; predicate unchanged, comment cites its issue, fix matches the repo idiom.
- **Greptile score (codebase fit) — 4/5**: one minor from outside the diff (stale base conflicts loudly against main's 0.1.792); no caller affected; release hygiene correctly honored.

VERDICT: APPROVE

(Rebase onto `origin/main` before merging and re-resolve the `RELEASE_NOTES` array to keep only this PR's note.)

Issue #396, opened 2026-08-29
workspace: the .code-workspace JSONC stripper walks bytes, mangling non-ASCII


vitali87:
## What happens

`strip_jsonc` in `src/workspace.rs` walks the input a BYTE at a time and treats each byte as a character:

```rust
let bytes = src.as_bytes();
...
let c = bytes[i] as char;
```

For any non-ASCII character that is more than one byte in UTF-8, each byte becomes its own `char`, so the text is rebuilt as mojibake. Reproduced with the same logic:

```
input:  {"name": "Café Noir", "type": "dark"}
output: {"name": "CafÃ© Noir", "type": "dark"}
```

The parse still succeeds, so nothing errors: the name simply arrives corrupted.

## What it affects

`.code-workspace` files, which is the only caller. A workspace whose folder `name` carries an accent, or any non-Latin script, shows the mangled name in croft. Comments containing non-ASCII are stripped anyway, so the visible damage is in string values.

## The fix already exists in the codebase

`crate::tasks::strip_jsonc` (`src/tasks.rs`) does the same job over `chars()` and is already `pub(crate)`:

```rust
let mut chars = text.chars().peekable();
```

Both functions handle the same three JSONC extras (`//`, `/* */`, trailing commas), so the byte-walking copy can call the chars-based one and be deleted. Worth a test with a non-ASCII value either way, since the failure is silent corruption rather than a parse error, which is exactly the kind that survives review.

## How it was found

While addressing a review on #395, which had made this function public to reuse it for VS Code theme files. Theme names carry accents often enough that it showed up immediately (`Café Noir`, `Ayu Mirage`, various Catppuccin variants). #395 now uses the `tasks.rs` one and reverts that visibility change; this issue is the pre-existing caller, which nothing in that PR touches.

PR #395, opened 2026-08-29
feat(themes): import a VS Code colour theme with croft theme-import

vitali87:
Part of #350. `croft theme-import <file.json>` converts a VS Code colour theme into a croft `[[themes]]` manifest and installs it under `~/.config/croft/extensions/`.

croft ships fifteen themes; VS Code has tens of thousands, and someone switching editors arrives with one they already like.

## Why this is a mapping, not a rename

VS Code names hundreds of workbench keys and colours code by TextMate scope. A croft theme is a small fixed palette: chrome, sixteen ANSI slots, eight syntax roles. Four decisions carry the conversion, and three of them came from converting real themes and finding the output wrong:

**Key priority per slot, ordered against real themes.** `activityBarBadge.background` leads the accent, not `focusBorder`. `focusBorder` is the obvious first choice and the wrong one: most themes set it to a muted grey, so leading with it gave One Dark Pro a grey accent while its signature blue sat unused.

**Scope matching is DOWNWARD only.** A rule for `keyword` colours `keyword.control.if`; a rule for `entity.name.function.decorator` colours decorators and says nothing about functions in general. Matching upward as well looks harmless and is not: asked for `entity.name.function`, One Dark Pro answered with its narrowest function-ish rule and rendered every call **yellow instead of blue**. There is a test pinning this specific shape.

**Translucent colours composite over the background.** VS Code selection colours are routinely `#rrggbbaa`. Dropping the alpha hands back a colour far more saturated than the author designed, so it blends against the surface it sits on, which is what a reader sees in VS Code.

**The terminal palette is all sixteen or none.** Eight themed colours beside eight croft defaults reads as a rendering bug rather than a partial import.

And the honesty rule: every croft slot the source theme never named is derived **and reported**. The CLI prints each derivation, because an imported theme quietly inventing colours is exactly how an import "works" while looking wrong.

## Verified against real themes

Not just fixtures. Converted, with output inspected against each theme's published palette:

| theme | notes |
|---|---|
| One Dark Pro | keyword `#c678dd`, function `#61afef`, string `#98c379`, constant `#d19a66`, type `#e5c07b` — its canonical palette |
| Tokyo Night | keyword `#bb9af7`, function `#7aa2f7`, string `#9ece6a`, fg `#a9b1d6` |
| Monokai | keyword `#f92672`, string `#e6db74`, full 16-colour terminal palette |
| Solarized Dark | keyword `#859900`, string `#2aa198`, full terminal palette |
[...]

Issue #351, opened 2026-08-29
croft import-vscode: bring settings, keybindings, and snippets across


vitali87:
## Motivation

The first ten minutes decide whether someone stays. Most of what a VS Code user misses on day one is their own configuration, and croft already reads VS Code's formats in several places (`.vscode/settings.json` subset, `launch.json`, `tasks.json`, `.code-workspace`). A one-shot importer for the user-level files finishes the job.

## Current state

- `docs/SETTINGS.md` documents the `.vscode/settings.json` subset croft honours per workspace; nothing reads the *user* settings file.
- `src/keymap.rs` (keybindings.json), `src/snippets.rs` (snippets.json) have their own formats, close to VS Code's but not identical.

## Proposed design

1. `croft import-vscode [--dry-run]` locates the VS Code user directory per platform (`~/Library/Application Support/Code/User`, `~/.config/Code/User`, plus Cursor/VSCodium variants), and converts:
   - `settings.json` → `~/.config/croft/config.json` for every key croft has an equivalent of (font-free ones: tab size, insert spaces, word wrap, format on save, auto save mode, render whitespace, minimap, bracket colours, inlay hints, files.exclude…), listing the unmapped keys.
   - `keybindings.json` → croft `keybindings.json`, mapping VS Code command ids to croft palette command ids through a table, refusing rows that reference unknown commands with the same warning path as today's bad rows.
   - `snippets/*.json` → croft `snippets.json` (same tab-stop syntax already).
2. Merge, never overwrite: existing croft entries win; the report lists conflicts.
3. The theme importer (sibling issue) is invoked for the active `workbench.colorTheme` when the theme is installed locally.

## Acceptance criteria

- [ ] `--dry-run` prints a full report (mapped, unmapped, conflicts) and writes nothing.
- [ ] A fixture VS Code profile converts to a croft config that loads with no warnings in OUTPUT.
- [ ] Re-running is idempotent.

## Related

- Tracking epic: #380
- Depends on: #350

Issue #350, opened 2026-08-29
Load VS Code colour themes (.json) as croft themes


vitali87:
## Motivation

croft ships twelve themes as data manifests. VS Code has tens of thousands. A converter from the VS Code theme format means "every VS Code theme works in croft", which removes the most emotional switching cost and makes a marketing line that is literally true.

## Current state

- Themes are `[[themes]]` manifests loaded from `~/.config/croft/extensions/` (`src/theme.rs`, `src/lsp/manifest.rs`), listed in the Extensions panel.
- Highlighting is tree-sitter capture names plus LSP semantic tokens; VS Code themes key on TextMate scopes and semantic token types.

## Proposed design

1. Accept a VS Code theme file (the `colors` + `tokenColors` + `semanticTokenColors` JSON, including `include` chains and JSONC) dropped into `~/.config/croft/extensions/themes/` or referenced from a manifest with `source = "vscode"`.
2. A scope→capture mapping table (`keyword`, `entity.name.function`, `string`, `comment`, `variable`, `type`, ...) with a documented fallback order, and `colors.*` workbench keys mapped to croft chrome (`editor.background`, `sideBar.background`, `statusBar.*`, `terminal.ansi*`, `editorGutter.*`, `minimap.*`).
3. A `croft theme import <file-or-marketplace-url>` command that writes the manifest; the marketplace URL variant downloads the `.vsix`, extracts the theme JSON, and discards the rest (no extension code ever runs).
4. Keep the built-in twelve as they are; they become the regression fixtures for the mapping (convert One Dark Pro's upstream JSON and diff against the hand-made manifest).

## Acceptance criteria

- [ ] Ten popular themes (One Dark Pro, Dracula, Catppuccin, Tokyo Night, GitHub Dark/Light, Nord, Gruvbox, Solarized, Monokai Pro, Ayu) import and render with correct editor, chrome, and terminal palettes.
- [ ] The converted One Dark Pro differs from the hand-made manifest only in documented, accepted ways.
- [ ] Light themes import correctly (the #217 class of bugs has a test).

## Related

- Tracking epic: #380

PR #400, opened 2026-08-29
feat(config): bring a VS Code profile across with croft import-vscode

vitali87:
Part of #351. `croft import-vscode` reads a VS Code (or Cursor / VSCodium / Windsurf / Insiders) user directory and converts settings, keybindings and snippets into croft's own files.

Most of what someone misses on day one after switching editors is their own configuration. croft already reads VS Code's formats in several places (the `.vscode/settings.json` subset, `launch.json`, `tasks.json`, `.code-workspace`); this finishes the job for the user-level files.

## Three conversions of very different difficulty

The design follows that difference rather than pretending the three are alike.

**Snippets are nearly free.** croft's `snippets.json` already mirrors VS Code's global snippets file, tab-stop syntax included. The one real difference: VS Code carries a snippet's language in the **file name** (`snippets/python.json`) while croft carries it in a `scope` field, so without recovering it a per-language snippet silently widens to every language.

**Keybindings share a file shape and agree on no command ids**, so they convert through a table. An unmapped id is dropped and **named**: a chord silently bound to something that merely sounds similar is worse than one that did not come across. A test checks every croft id in the table against `Command::from_id`, so a typo cannot ship a `keybindings.json` croft itself rejects at load.

**Settings is where the editors genuinely differ.** Only keys croft has a real equivalent for are mapped; the rest are listed. "My settings imported" and "my settings are gone" must not look the same, and a report is the only thing that separates them.

## Merge, never overwrite

An existing croft value always wins and the conflict is reported, which is also what makes a second run a no-op. Someone who has already tuned croft must not lose that to a moment's curiosity.

## Verified end to end, not just by unit test

Against a fixture profile (JSONC settings with comments and a trailing comma, four keybindings, one per-language snippet file):

```
6 settings mapped, 2 keybindings mapped, 1 snippet
  auto_save = false
  auto_save_on_focus_change = true
  disable_bracket_colors = true
  format_on_save = true
  render_whitespace = "boundary"
  terminal_scrollback = 50000
  cmd+k z -> toggle_zen_mode
  cmd+shift+p -> quick_open

3 settings croft has no equivalent for:
  editor.fontFamily
  python.analysis.typeCheckingMode
  workbench.colorTheme

2 keybindings not carried over:
  cmd+alt+n: no croft command matches extension.someThirdParty
  cmd+b: unbinding workbench.action.toggleSidebarVisibility has no croft equivalent
```

Second run against the same profile, into the same config directory:

```
3 values croft already had, left alone:
  keybindings.json cmd+k z: already bound, left alone
  keybindings.json cmd+shift+p: already bound, left alone
  snippets.json "Main guard": already defined, left alone

Everything was already in place; nothing changed.
```

That is the issue's idempotency criterion, checked rather than asserted.

## Tests

10, with two that pin the output against croft itself rather than against my own expectations:

- every croft key the mapping writes is a real field of `Prefs`, and each converted value has the **type** that field stores (a bool written into `render_whitespace` would be a config error)
- every VS Code command in the table maps to an id `Command::from_id` accepts
- `files.autoSave` drives both croft toggles, across all four modes including `off`
- keybindings convert by table, and both an unknown command and a VS Code unbinding (`-command`) are reported rather than dropped
- a snippet takes its language from the file name; an explicit `scope` wins; a `.code-snippets` file keeps its own
- a whole profile directory converts in one pass, and an empty one is not a failure

Full suite: 3733 passed, 0 failed. Clippy clean at `--all-targets -D warnings`. Doc-ownership gate clean.

## Not in this PR

`workbench.colorTheme` deliberately reports as **unmapped**. Wiring it to the theme importer is the remaining piece of #351 and belongs with #350 (#395), which is still in review; doing it here would couple two PRs that have no other reason to be coupled.

The JSONC reader uses `crate::tasks::strip_jsonc` (chars-based) rather than the byte-walking one in `workspace.rs`, which mangles non-ASCII: that is #396, filed separately.

## Summary by CodeRabbit

- **New Features**
  - Added `croft import-vscode` to import supported settings, keybindings, and snippets from VS Code-family editors.
  - Supports automatic profile discovery or a specified source directory.
  - Includes dry-run previews, conflict reporting, unsupported-item warnings, and repeatable merges that preserve existing croft values.

- **Documentation**
  - Documented the new command, supported editor profiles, setting mappings, conversions, and merge behavior.

- **Chores**
  - Updated the package version to `0.1.802`.


vitali87:
@coderabbitai review

vitali87:

Running because CodeRabbit's check on this PR reads `pass` annotated **"Review rate limited"**, which is a no-op rather than a review. Reviewed at head `f853531` by a session on another machine, at the author's request. This PR had zero prior review rounds.

---

## Walkthrough

Adds `croft import-vscode`, a one-shot converter that locates a VS Code-family user directory (VS Code/Insiders/VSCodium/Cursor/Windsurf), reads `settings.json`, `keybindings.json` and `snippets/*.json`, and merges croft equivalents into `~/.config/croft/`. Settings convert through a hand-written 10-entry table plus a special case for `files.autoSave`; keybindings through a 65-entry command-id table; snippets are near-passthrough with the language recovered from the filename into a `scope` field. Merge-never-overwrite: an existing croft value always wins and the conflict is reported; anything unmapped is listed rather than dropped silently.

| File | Summary |
|---|---|
| `Cargo.toml`, `Cargo.lock` | Version 0.1.800 → 0.1.802 |
| `README.md` | One line adding `croft import-vscode` to the CLI list |
| `docs/ARCHITECTURE.md` | Module-tree entry for `import_vscode.rs` |
| `src/cli.rs` | New `ImportVscode { from, dry_run }` subcommand + `import_vscode()` report printer and `plural()` helper (+114) |
| `src/import_vscode.rs` | New module (+795): profile discovery, `SETTINGS`/`COMMANDS` tables, `convert_settings`/`convert_keybindings`/`convert_snippets`, `scan_profile`, `apply`, and 10 unit tests |
| `src/main.rs` | `mod import_vscode;` |
| `src/release_notes.rs` | Release-notes highlight replaced |

Convention compliance: the `CONTRIBUTING.md` version-bump-plus-replaced-release-notes rule is satisfied (0.1.800 → 0.1.802, notes replaced).

## Findings

- **`src/import_vscode.rs` (`as_whitespace_mode`, diff lines 307-308)** **[blocker/correctness]** — the mapping emits `"boundary"`, which is not a croft whitespace mode, so the import silently produces the *opposite* of what the user configured. `WhitespaceMode` (`src/widgets/editor.rs:9781-9786`) has exactly three variants — `None`, `Selection`, `All` — and `from_pref` (`9800-9805`) is `match id { "none" => None, "all" => All, _ => Selection }`. So `"boundary"` falls into the `_` arm and reads back as `Selection`.

  The diff reads:

  ```rust
  "none" | "boundary" | "all" => s,
  "trailing" | "selection" => "boundary",
  ```

  Two concrete regressions. A user with `"trailing"` — who wants whitespace shown *almost never* — is mapped to `"boundary"` and lands on `Selection`, the opposite end of the scale. And `"selection"`, which croft supports **natively**, is rewritten to `"boundary"` and only survives by accident, because the `_` fallback happens to land back on `Selection`. The function's own comment asserts the false premise: "croft renders none / boundary / all". `render_whitespace` is a `String` in `Prefs` (`src/prefs.rs:210`), so nothing downstream rejects it; the bad value is written to disk and misreads forever.

  ```rust
  fn as_whitespace_mode(v: &Value) -> Option<Value> {
      // croft renders none / selection / all (WhitespaceMode). VS Code's
      // "boundary" and "trailing" have no croft equivalent, so they are
      // reported as unmapped rather than written as a value croft silently
      // reads back as "selection".
      let s = v.as_str()?;
      match s {
          "none" | "selection" | "all" => Some(Value::from(s)),
          _ => None,
      }
  }
  ```

  Returning `None` routes them into `unmapped_settings`, which is exactly the module's stated contract ("`my settings imported` and `my settings are gone` must not look the same").

- **`src/import_vscode.rs` (`convert_keybindings`, diff ~line 566)** **[major/correctness]** — multi-key ("chord") VS Code bindings are imported as rows croft cannot parse, so they are reported as successfully mapped and then silently do nothing. `Chord::parse` (`src/keymap.rs:150-183`) splits only on `'+'`; for `"ctrl+k z"` the second token is `"k z"`, which reaches `parse_key`'s fallback (`279-287`) where `chars.next().is_some()` is true, returning `None`, so the whole chord fails. Confirmed by re-implementing the parser's token logic: `"ctrl+k z"` → `None`, `"ctrl+k ctrl+s"` → `None`, while `"ctrl+shift+p"` → parses. `Keymap::resolve` (`653-666`) then emits `"ctrl+k z: not a key chord or mouse gesture; the row binding "toggle_zen_mode" was skipped"` into OUTPUT · Keybindings. `ctrl+k <x>` is VS Code's most common chord prefix, so this is not an edge case. The import counts it in "N keybindings mapped" and writes it to disk — the exact silent-loss failure the module set out to prevent, relocated one file downstream.

  ```rust
  if crate::keymap::Chord::parse(key).is_none() {
      report.dropped_keybindings.push(format!(
          "{key}: croft has no multi-key chords, so this cannot be bound"
      ));
      continue;
  }
  ```

  This also makes the table structurally honest: the same call croft uses at load is the one the import validates against.

- **`src/import_vscode.rs` (`COMMANDS`, `editor.action.revealDefinition` → `mouse_go_to_definition_at_click`, diff ~line 425)** **[major/correctness]** — this binds the user's Go-to-Definition key to a command that navigates to wherever the *mouse* last was, not the caret. `Cmd::MouseGoToDefinitionAtClick` (`src/app/mod.rs:28449`) matches on `self.last_mouse_pos`, which is only ever set from a mouse event (`31405`, cleared to `None` at `31407`). With no recent mouse movement the arm falls to `None` and sets `"...needs a mouse binding"`; with a stale position it jumps to an unrelated symbol. `editor.action.revealDefinition` is F12 in VS Code, documented in this repo as caret-based ("Go to Definition ... at the cursor", `docs/KEYBINDINGS.md:275`). The table's own doc comment promises "a VS Code command with no true counterpart is left out so the import reports it rather than binding the user's chord to something that merely sounds similar" — this entry is precisely that. Dropping the row also resolves a latent duplication: `editor.action.showHover` is already mapped to `peek_definition`, so two VS Code commands currently target one croft command.

- **`src/import_vscode.rs` (`apply` / `write_json`, diff ~lines 665-745)** **[major/correctness]** — the merge destroys user comments in every file it touches. `apply` reads through `read_jsonc` (which calls `crate::tasks::strip_jsonc`, `src/tasks.rs:293`) and writes back with `serde_json::to_string_pretty`, so the round-trip is lossy by construction. Not hypothetical: croft *seeds* both target files with comment-bearing templates — `keymap::TEMPLATE` (`src/keymap.rs:738-747`) opens with three `//` lines explaining `mod` and iTerm2 forwarding, and `snippets::TEMPLATE` (`src/snippets.rs:284-303`) with four lines documenting tab-stop syntax. The command's help promises it "MERGES the result into `~/.config/croft/`", and the module header promises "an existing croft value always wins" — silently deleting the user's annotations contradicts both. At minimum state it in the report and the CLI help; better, refuse to rewrite a file whose stripped form differs from its raw form.

- **`src/import_vscode.rs` (`SETTINGS` table) vs `src/config_layers.rs:399-416`** **[major/correctness]** — croft already has a VS Code settings mapper, and this PR adds a second, independent one that already disagrees with it. `map_vscode_layer` maps `editor.formatOnSave`, `editor.formatOnType` and `files.autoSave`, and is documented as authoritative in `docs/SETTINGS.md:94-108`. The two differ on `files.autoSave`: the existing mapper does not recognise `"onWindowChange"` at all (`grep -c onWindowChange src/config_layers.rs` → `0`), while the new one maps it to `auto_save_on_focus_change: true`. They also differ in shape — the old one sets only the *true* toggle and leaves the other untouched; the new one always writes both. So the same `settings.json` yields different prefs depending on whether it arrives as a workspace layer or through the import. At minimum cross-reference both places; better, delegate the overlapping keys to the existing mapper so there is one authority.

- **`src/import_vscode.rs` (`convert_settings`, `files.autoSave` branch, diff ~line 510)** **[minor/style]** — a bare block wraps the special case for no reason, and `.filter(|_| key == "files.autoSave")` inverts the natural read (filtering the *value* on a predicate about the *key*). The `SettingMap.croft` doc comment also lies: "Croft pref key, or `None` when the source key sets more than one" — the field is `&'static str`, not `Option<_>`, and no entry is ever `None`.

## Test coverage

- **`apply()` is entirely untested** — the merge/write path, the whole subject of author question 3, has zero tests (`apply` appears in the diff only at its definition and its one CLI call site). Nothing exercises conflict detection, the `changed` flag, idempotency, or the `bound` key-collision check. Write a `TempDir` test that runs `apply` twice against a pre-populated `config.json`/`keybindings.json`/`snippets.json`, asserting (a) the second run returns an empty `written` vec, (b) a differing croft value stays put and lands in `conflicts`, (c) an equal value produces *no* conflict entry. The idempotency claim in the module header, the CLI help and the release notes currently rests on reading the code.
- **No test asserts a converted chord is loadable by croft.** This is what would have caught the `ctrl+k z` major. Feed `report.keybindings` through `Keymap::resolve` on the serialised rows and assert `warnings.is_empty()` — an end-to-end property rather than a table self-check.
- **`every_mapped_command_id_is_a_real_croft_command` checks the wrong half.** It verifies the croft ids exist; nothing verifies the *VS Code* ids are real, and nothing verifies the pairing is semantically right — which is how `revealDefinition` → `mouse_go_to_definition_at_click` passes.
- **`converted_values_match_the_type_croft_stores` cannot fail for a string pref.** See question 1 below.
- **`discover_profiles` is untested**, including its dedup branch and `HOME`/`APPDATA` probing. A `TempDir` as `HOME` containing both `Library/Application Support/Code/User` and `.config/Code/User` should find both and deduplicate.

## Nitpicks

- `src/cli.rs` — the report prints unconditionally before the `dry_run` check, so a non-dry run prints the full mapped list *and* the wrote-files list. Fine, but `--dry-run`'s help ("Print the full report and write nothing") implies the report is dry-run-specific when it is not.
- `src/import_vscode.rs` — `unmapped_settings.sort()` runs inside `convert_settings`, `dropped_keybindings.sort()` inside `convert_keybindings`, but `conflicts.sort()` in `apply`. Three sorts across two layers; consider one normalisation point.
- The new command is absent from `docs/` — `docs/SETTINGS.md` has a "`.vscode/settings.json` subset" section that is the natural home for the import's settings table, and `docs/KEYBINDINGS.md` for the command table. CONTRIBUTING does not mandate this, hence a nit.

## The three items you flagged

**1. Is the structural testing enough for the settings table? No — and here is exactly where it stops.** `converted_values_match_the_type_croft_stores` compares `std::mem::discriminant` of two `serde_json::Value`s. Measured: `discriminant(json!("boundary")) == discriminant(json!("selection"))` is `true`, and `discriminant(json!(1.5)) == discriminant(json!(20000u64))` is also `true`, because `Value::Number` and `Value::String` are single variants carrying the payload inside. So the test proves only "a string went where a string goes" — it cannot distinguish a *valid* string from an invalid one, nor a float from the integer `terminal_scrollback: usize` requires.

The proof that it checks a weaker property than it appears to: **that test's own fixture feeds `"editor.renderWhitespace": "boundary"`, and it passes** — while `"boundary"` is a value croft has no variant for and silently reads back as `Selection`. The test named the bug's input and went green. Ask of it "what else would also have passed?" and the answer is "any string at all".

The same shape applies to commands: `Command::from_id` proves the id exists, not that it does the same thing — which is how `revealDefinition` → `mouse_go_to_definition_at_click` passed three tests while binding F12 to the mouse cursor.

What would add real structure, in order of value: **(a)** for string-valued prefs assert the converted value round-trips — `WhitespaceMode::from_pref(v).pref_id() == v` — which fails loudly on `"boundary"` and generalises to `problems_scope` and `diff_ignore_whitespace`; **(b)** for chords, run the output through `Keymap::resolve` and require zero warnings; **(c)** for numbers, assert `serde_json::from_value::<Prefs>` accepts the single-key object. Note (a) and (c) are properties of the *value*, which is where all three defects live. Semantic command pairing stays a human-review property — there is no oracle for it — so the honest move is to keep the table short and make each entry justify itself, which the doc comment already promises and three entries currently violate.

**2. Dropping `-command` unbindings with a note: right call, wrong note.** The reasoning is sound — croft's defaults genuinely are its own, so there is no VS Code default to remove and synthesising one would be guesswork. But the message misleads. `"{key}: unbinding {removed} has no croft equivalent"` says croft lacks a *feature*, when the actual situation is that the chord may be bound in croft *by a different default* this import will not touch. A user who removed `ctrl+b` in VS Code because they wanted that chord free will find croft's own `ctrl+b` still live. Keep the drop; sharpen the text to name the consequence — `"{key}: VS Code unbindings are not imported; if croft binds {key} by default it stays bound"` — and point at `keybindings.json` as the override. One caveat: unbindings are counted in the same `dropped_keybindings` list as genuine failures, so "N keybindings not carried over" conflates "croft cannot do this" with "nothing to do here". Worth splitting.

**3. Merge-never-overwrite on genuine disagreement: right, with one gap.** For a one-shot import invoked out of curiosity this is clearly correct: the asymmetry favours it, since a wrongly-kept croft value costs one manual edit while a wrongly-overwritten one destroys tuning the user may not remember doing, and it is what makes the second run a no-op. The conflict list makes the outcome inspectable. The gap is that it is asserted, not tested — `apply` has no tests at all, so "running it twice changes nothing" rests on reading the code. Also consider a `--overwrite` flag for the user who *wants* VS Code to win; right now their only route is deleting `config.json` first, which loses the keys the import has no opinion on. Not required for this PR, but the report's "kept X, VS Code had Y" lines invite the question.

## Scores

- **CodeRabbit score (diff quality) — 1/5**: one blocker (`"boundary"` written as a whitespace mode croft silently reads back as `Selection`, with the test that should have caught it feeding that exact value) plus three majors inside the changed code — unparseable multi-key chords reported as mapped, a Go-to-Definition mapping bound to the mouse position, and a lossy comment-destroying round-trip in `apply`.
- **Greptile score (codebase fit) — 2/5**: two majors whose evidence is outside the diff — the duplicate-and-already-divergent VS Code settings mapper in `config_layers.rs:399-416` (documented as authoritative in `docs/SETTINGS.md:94-108`, disagreeing on `onWindowChange`), and the comment-bearing templates croft itself seeds at `src/keymap.rs:738` and `src/snippets.rs:284` that `apply` will erase. Version bump and release notes both comply with CONTRIBUTING; reference docs were not updated.

VERDICT: REQUEST_CHANGES

vitali87:

CodeRabbit's check still reads `pass` annotated **"Review rate limited"**, a no-op rather than a review. Reviewed at head `e91f540` by a session on another machine. All five round-1 findings were verified **by mutation**, not by reading.

---

## Walkthrough

Round 2 adds `src/import_vscode.rs` (1017 lines: profile discovery, a settings table, a VS Code→croft command table, snippet conversion, and a merge-only `apply`), wires `croft import-vscode` into the CLI, and collapses the duplicate `.vscode/settings.json` mapper in `config_layers.rs` onto the new shared `map_settings`. Version 0.1.800 → 0.1.802 and `src/release_notes.rs` replaced, satisfying the `version bump + release notes` job in `.github/workflows/ci.yml:34-69`.

All five round-1 findings are genuinely fixed at the point they were reported, and three of the four new tests were confirmed by mutation to discriminate. But the collapse onto one table changed the workspace layer's behaviour on unrecognised `files.autoSave` values, and `apply` — the function that writes to the user's disk — has no test at all, which lets three defects through.

| File | Round-2 change |
|---|---|
| `src/import_vscode.rs` | New module: `map_settings`, `convert_keybindings`, `convert_snippets`, `scan_profile`, `apply`, `carries_comments`, 14 tests |
| `src/config_layers.rs` | `load_vscode_subset` now delegates to `import_vscode::map_settings`, filtered to `WORKSPACE_ALLOWED_KEYS` |
| `src/cli.rs` | `ImportVscode { from, dry_run }` subcommand + report printer |
| `src/main.rs` | `mod import_vscode;` |
| `Cargo.toml` / `Cargo.lock` | 0.1.800 → 0.1.802 |
| `src/release_notes.rs` | Replaced with the import-vscode note |
| `docs/SETTINGS.md` | Table extended; documents the shared mapping and the `boundary`/`trailing` → `none` loss |
| `docs/ARCHITECTURE.md`, `README.md` | Module and command entries |

## Round-1 verification

1. **BLOCKER — `as_whitespace_mode` emitted `"boundary"` — FIXED.** The round-trip property holds for every emittable string. I isolated the new assertion's power: mapped `selection` → `"boundary"` *and* doctored the expectation table to `("selection", "boundary")` so only the round-trip could fail. It fired: `assertion failed: "boundary" does not survive croft's own parse`. Independent discriminating power, not just coupling to the expected-value list.
2. **major — multi-key chords written to disk — FIXED.** Deleting the `key.split_whitespace().count() > 1` guard *and* the test's first assertion, leaving only `Keymap::resolve`, still failed: `croft rejected rows this import would have written: ["ctrl+k z: not a key chord or mouse gesture; ..."]`. The `resolve` round-trip is load-bearing.
3. **major — `revealDefinition` → `mouse_go_to_definition_at_click` — FIXED.** Re-adding the mapping fails `no_keyboard_shortcut_maps_to_a_pointer_driven_command`. Also verified the `mouse_` prefix is a **complete** proxy rather than a partial one: croft has exactly three `mouse_`-prefixed commands, none is in `COMMANDS`, and `grep -rn last_mouse_pos src/` finds no non-prefixed command reading the pointer.
4. **major — `apply` destroys comments — PARTIALLY FIXED.** The guard works; its wiring is untested. See Finding A.
5. **major — second divergent settings mapper — FIXED, with a behaviour regression.** Only one mapping table remains (`config_layers.rs` hits are the delegating call and comments). `WORKSPACE_ALLOWED_KEYS` is **not** a new invention — it pre-exists at `config_layers.rs:92-107` and the diff only reads it. But the filter is not behaviour-preserving: see Finding B.

## Findings

- **`src/import_vscode.rs:537-640`** **[major/testing]** — the comment guard is unwire-able with the suite green. `a_commented_config_file_is_never_rewritten` asserts on `carries_comments(&path)` — the *predicate* — and never calls `apply`. Its name asserts the behaviour; its fixture supplies only the helper.

  **Verified by mutation:** replacing all three `if carries_comments(&path) {` with `if false && carries_comments(&path) {` makes `apply` rewrite comment-bearing files and destroy `keymap::TEMPLATE`'s header — the exact round-1 blocker. **All 14 tests passed.** A surviving mutant on the very defect the test is named for.

  Fix: test `apply` end-to-end. It calls `crate::prefs::config_dir()`, so it needs a seam — an `apply_into(dir, report)` that `apply` delegates to, or the env override `config_dir` already respects. Then seed `keybindings.json` with `keymap::TEMPLATE`, run `apply`, assert the bytes are unchanged and a conflict naming `keybindings.json` was reported.

- **`src/config_layers.rs:399-411`** **[major/correctness]** — an unrecognised `files.autoSave` value now silently overrides the user's own setting. The old code matched `Some("afterDelay") | Some("onFocusChange") | Some("off")` with `_ => {}`, so an unknown value emitted nothing. `map_settings` treats *any* string as authoritative, emitting `auto_save = (mode == "afterDelay")` and `auto_save_on_focus_change = (...)` — so an unknown value emits `false, false`.

  Verified through the real config chain. User config `{"auto_save":true}`, workspace `{"files.autoSave":"afterDelayy"}` (a typo):

  | | `m.prefs.auto_save` | provenance |
  |---|---|---|
  | old `load_vscode_subset` | `true` | `User` |
  | PR head | `false` | `VsCodeWorkspace` |

  A typo, or any `files.autoSave` mode newer than croft's table, now silently turns the user's auto-save off from a workspace layer. No existing test catches it. It is also wrong for the importer: an unknown mode is *mapped* (to both-off) rather than reported as unmapped.

  ```rust
  if key == "files.autoSave" && let Some(mode) = value.as_str() {
      if !matches!(mode, "afterDelay" | "onFocusChange" | "onWindowChange" | "off") {
          unmapped.push(key.clone());
          continue;
      }
      ...
  }
  ```

  Pin it by asserting the **provenance** stays `User`, not just the value — otherwise the assertion passes whenever the default happens to match.

- **`src/import_vscode.rs:556-560`** **[major/correctness]** — `String::from` where `format!` was meant: the settings conflict prints literal braces.

  ```rust
  report.conflicts.push(String::from(
      "config.json {key}: kept {current}, VS Code had {value}",
  ));
  ```

  The user is shown, verbatim: `config.json {key}: kept {current}, VS Code had {value}`. `key`, `current` and `value` are all in scope, so this compiles and clippy is clean. Every conflict row is identical and names nothing — and the report is, per the module's own doc comment, "the product, not a side effect". This is the one case where the user most needs to know *which* setting was kept.

- **`src/import_vscode.rs:622-624`** **[major/correctness]** — the same bug in the snippets conflict, printing `snippets.json {name:?}: already defined, left alone` for every collision. Note the keybindings conflict two blocks up is correct (`format!("keybindings.json {key}: already bound, left alone")`) — that inconsistency is what makes these easy to miss on a read, and why `apply` needs execution coverage. Independently confirmed against the diff: the `format!` at line 869, the two `String::from` at 834 and 900.

## Test coverage

**Behaviour with no test at all:**

- **`apply` — the entire write path.** Merge-never-overwrite, idempotence-on-second-run, the comment guard's wiring, conflict-message content, and `write_json`'s trailing newline are all unexercised. This single gap is what let findings A, C and D through. The module's headline promise — "running it twice changes nothing the second time" — has no test.
- **`discover_profiles`** — no test for the dedupe, the roots order, or that `APPDATA` is probed. Testable by setting `HOME` to a tempdir.
- **`read_jsonc` on a malformed file** — `scan_profile` propagates the error and aborts the whole import; nothing pins whether one broken file should kill the run or be reported like everything else. Given the module's stated posture ("reported rather than dropped"), aborting looks unintended.
- **`convert_settings` on a non-object** and `convert_keybindings` on a non-array — both warning paths unreached.

**Tests whose fixture cannot demonstrate the property they name:**

- **`a_commented_config_file_is_never_rewritten`** — named for `apply`'s behaviour, tests only `carries_comments`. Surviving mutant demonstrated above. What else would pass it? Any build in which the guard exists but is never consulted.
- **`the_workspace_layer_and_the_importer_agree_on_a_settings_file`** — the name asserts agreement between two code paths, but the body calls `map_settings` once and checks membership in `WORKSPACE_ALLOWED_KEYS`. It never invokes `load_vscode_subset`, so it cannot observe the filter, and **would pass unchanged if `load_vscode_subset` were reverted to its own private table** — the exact defect it is named for. It is also why Finding B is invisible: it asserts on the mapper, not on the layer. Drive it through `load_merged_from` and assert on `m.prefs` and provenance.
- **`every_written_keybinding_is_one_croft_can_load`** — the `Keymap::resolve` assertion is strong (verified), but the fixture's three rows all convert cleanly or are dropped, so `warnings.is_empty()` is checked against a two-row keymap. It does not cover what a reader would assume from the name: a key croft's parser rejects for a reason *other* than a chord sequence (`cmd+numpad_add`, `ctrl+[BracketLeft]`, a bare modifier).
- **`settings_map_only_where_croft_has_a_real_equivalent`** — asserts five keys land and two are unmapped, but never that `report.settings` has no *extra* keys. A mapper emitting spurious prefs would pass.

## Nitpicks

- `src/import_vscode.rs:80` — `/// Croft pref key, or `None` when the source key sets more than one.` on a field typed `croft: &'static str`. There is no `None`; leftover from an earlier shape.
- `src/import_vscode.rs:369-375` — the `files.autoSave` arm wraps its body in a bare `{ ... }` block containing only `continue`.
- `src/import_vscode.rs:354-358` — the lossy-conversion warning is keyed on `key == "editor.renderWhitespace"` inside the generic loop, so a second lossy converter added later will silently not warn. Consider an `explain_loss: Option<fn(&Value, &Value) -> String>` on `SettingMap`.
- `src/config_layers.rs:409` — `let (mapped, _unmapped, _warnings) = ...` discards the warnings, so the `boundary`/`trailing` → `none` loss is reported by `import-vscode` but silent in the workspace layer, even though `load_vscode_subset` already takes a `warnings: &mut Vec<String>`.
- `docs/SETTINGS.md:105` — omits `debug.inlineValues` → `disable_inline_values`, which `SETTINGS` maps and `WORKSPACE_ALLOWED_KEYS` permits. Otherwise exhaustive.

## Scores

- **CodeRabbit score (diff quality) — 2/5.** Four majors: the untested write path (A), the workspace regression (B), and two report-formatting defects (C, D).
- **Greptile score (codebase fit) — 3/5.** The module sits well in the repo's conventions, reuses `tasks::strip_jsonc` (correctly preferring it over the byte-walking `workspace.rs` one), pins the command table against `Command::from_id`, and satisfies the CI version/release-notes gate. Marked down for one behaviour regression in an existing, previously-tested code path that the consolidation introduced silently.

VERDICT: REQUEST_CHANGES

vitali87:
CodeRabbit's five inline findings are addressed. Four were new to me; the `format!` one I had already fixed from the fallback review.

**The doc-capture finding is the one worth reading.** My `carries_comments` helper was inserted between `apply` and its doc block, so the helper was documented as "Merge a report into croft's config files" and the public `apply` had no doc at all. That is the #314 class, in a PR written after a day spent on the #314 class.

What makes it interesting is why nothing automated caught it: `src/import_vscode.rs` is **added** by this branch, so the merge-base comparison had no earlier version to lose documentation against. That blindness is what #403 fixed, merged an hour ago. Rebasing this branch onto main and running the gate reports the capture immediately and names the branch commit the doc was last seen at, rather than the base where the file does not exist.

**Also fixed:**

- An explicit `--from` directory is now checked to exist. Every file the scan reads is optional, so a mistyped path produced "0 settings mapped, 0 keybindings mapped, 0 snippets" and "Nothing to import", indistinguishable from a real but empty profile. The auto-detected branch already failed loudly.
- A config file that parses but is the **wrong shape** is reported rather than replaced. "Not an object" was treated as "absent", so a `config.json` holding an array or a scalar started the merge from an empty map and got overwritten, discarding the user's content with no conflict entry. Per file, so one wrong-shaped file does not block the other two. Two tests, including one asserting the other files are still written.
- `editor.inlayHints.enabled` is marked `(negated)` in `docs/SETTINGS.md`, matching the two other inverted mappings.

**Expect the `doc comments stay with their function` job to fail on this branch, and it is a false positive in that job rather than a defect here.** `apply` is documented at HEAD; the pairwise pass added in #403 walks every adjacent commit pair and so reports the intermediate state where the capture existed, which I have since fixed. Verified directly:

```
apply documented at HEAD: True
```

The fix is #406 (one filter, positive-control verified). Once that merges this branch needs a rebase and the job goes green. I would rather flag it than squash the history that documents these review rounds.

Suite: 3740 passed with two load-sensitive failures that pass in isolation (#397). Clippy clean at `--all-targets -D warnings`.

vitali87:

Reviewed at `85a72d2`. Every prior finding re-verified against this head, most by mutation. **All seven of my findings and all five of CodeRabbit's are genuinely fixed** — and where I mutated the fixes, genuinely pinned. Two new majors, both consequences of the round-3 change rather than of the original design.

## Status of prior findings

| Finding | Source | Fixed? | Evidence executed |
|---|---|---|---|
| `as_whitespace_mode` emitted `"boundary"` | me r1 (blocker) | **yes** | `import_vscode.rs:107-128` maps `boundary`/`trailing` → `none`; round-trip test passes |
| Multi-key chords written to disk | me r1 | **yes** | `convert_keybindings:443-447` guards `split_whitespace().count() > 1` |
| `revealDefinition` → pointer command | me r1 | **yes** | removed; `no_keyboard_shortcut_maps_to_a_pointer_driven_command` passes |
| `apply` destroys comments | me r1/r2 | **yes** | mutated all three `if carries_comments(&path)` → `if false && …`: `a_commented_config_file_is_never_rewritten` **fails**, printing croft's rewritten `keymap::TEMPLATE`. **The round-2 surviving mutant is dead.** |
| Second divergent settings mapper | me r1 | **yes** | `config_layers.rs:407` delegates to `import_vscode::map_settings` |
| `files.autoSave` typo overrides user | me r2 | **yes** | reverted to `_ => (false, false)`: `an_unrecognised_auto_save_value_does_not_override_the_user` **fails** on `assert!(m.prefs.auto_save)` |
| `String::from` with `{key}` / `{name:?}` | me r2 + CodeRabbit | **yes** | reverted `format!` → `String::from`: `a_conflict_message_names_the_value_it_kept` **fails**, printing the literal `config.json {key}: kept {current}…` |
| Workspace-layer test never calls `load_vscode_subset` | me r2 | **closed elsewhere** | the named test still passes under a restored divergent table, but two **new** `config_layers.rs` tests drive `load_merged_from` and both fail under that mutation. Gap closed; misleading name remains. |
| Doc block captured by `carries_comments` | CodeRabbit | **yes** | `:528` documents `carries_comments`, `:544` documents `apply` |
| `--from` accepts a nonexistent dir | CodeRabbit | **yes** | `cli.rs:785-791` bails (no test — see below) |
| Wrong-shaped file replaced | CodeRabbit | **yes** | mutated `if changed && !wrong_shape` → `if changed`: both new tests **fail**, `written` holds all three paths |
| `inlayHints.enabled` not marked negated | CodeRabbit | **yes** | `docs/SETTINGS.md:112` |

The `format!` fix deserves a specific note: `!joined.contains('{')` catches the whole class rather than the two instances found. That is the right shape for a fix whose defect was "a literal that looked like an interpolation".

## Findings

### `src/cli.rs:865-867` **[major/correctness]** — when croft *refuses* to write, the CLI reports the opposite

`written` is empty in three distinct situations: everything was already present, a file carries comments, or a file is the wrong shape. Only the first is benign, and line 866 prints the same sentence for all three.

Verified by running `apply_into` against a tempdir:

```
config.json = "[1,2,3]"         -> WRITTEN=[]  CONFLICTS=["config.json is not a JSON object, so it was left untouched"]
config.json = "// mine\n{...}"  -> WRITTEN=[]  CONFLICTS=["config.json: left untouched because rewriting it would delete its comments; ..."]
```

With `report.settings` non-empty the user sees `1 setting mapped / format_on_save = true`, then:

> **1 value croft already had, left alone:**
>   config.json is not a JSON object, so it was left untouched
>
> **Everything was already in place; nothing changed.**

Both lines are false. Croft did not already have `format_on_save`, and the reason nothing changed is that croft declined to write. **This is the module's own stated failure mode** — its doc calls the report "the product, not a side effect" and promises that *"`my settings imported` and `my settings are gone` must not look the same"*. Round 3 made `apply_into` safe and left the sentence describing it wrong.

There is a structural tell underneath: `Report::conflicts` is documented at `import_vscode.rs:303` as *"Values croft already has, which the import left alone"*, and six refusal messages are now pushed into it (`:570`, `:581`, `:613`, `:620`, `:657`, …). **The field's contract and its contents have diverged**, which is what lets one heading serve two meanings.

```rust
let refused = /* set by apply_into rather than sniffed from strings */;
if written.is_empty() && refused {
    println!("\nNothing was written: see above for the files croft would not rewrite.");
} else if written.is_empty() {
    println!("\nEverything was already in place; nothing changed.");
}
```

A separate `refusals: Vec<String>` is the cleaner shape, and it fixes the heading at `cli.rs:857` too — a refusal is not "a value croft already had".

### `src/import_vscode.rs:1073-1094` **[major/testing]** — the per-file test's stated property is false for the likelier input

`a_wrong_shaped_file_does_not_stop_the_other_two` documents *"the flag is per file, not per run"*. Its fixture (`"[1]\n"`) **parses** and is merely the wrong shape. A file that does not parse takes a different path entirely: `read_jsonc` (`:328-329`) propagates the `serde_json` error through `?`, so `apply_into` returns `Err` and the other two files are never considered.

```
config.json = "{ not json at all ]"  ->  RESULT_IS_ERR=true
ERR=parsing /tmp/.../config.json: key must be a string at line 1 column 3
snippets content = "{}\n"       (untouched — never reached)
```

A single unparseable `config.json` does exactly what the test's name says must not happen, and **a hand-edited file with a stray comma is far likelier than one holding a JSON array**. So the fixture covers the rarer half of the property it claims.

```rust
let parsed = match read_jsonc(&path) {
    Ok(v) => v,
    Err(e) => {
        report.conflicts.push(format!("config.json: left untouched, it is not valid JSON ({e})"));
        wrong_shape = true;
        None
    }
};
```

plus a fourth `cases` entry that is not valid JSON.

## Test coverage

- **The `--from` directory check has no test at all.** `grep -rn import_vscode tests/` returns nothing; `cli.rs`'s 25 tests never reference it. Four lines of `bail!` that nothing executes. Worth writing with **both** halves: `Err` containing "is not a directory" for a file path, *and* `Ok` for an empty-but-real directory. Without the second, the test passes on an implementation that rejects every `--from` — a degenerate state worse than the bug.
- **Nothing asserts what the CLI prints.** Both majors above are output defects, and `import_vscode()` writes straight to `println!`. Extracting `fn render(report: &Report, written: &[PathBuf]) -> String` makes them testable, and the assertion names the property: given a refusal, the output must **not** contain "Everything was already in place".
- **`discover_profiles` remains untested** (rounds 1 and 2, still open) — the dedup branch, roots order, `APPDATA` probing.
- **`the_workspace_layer_and_the_importer_agree_on_a_settings_file`** still cannot observe the property in its name; I restored a private divergent table dropping `onWindowChange` and it passed. The behaviour is covered by the two new `config_layers.rs` tests, so this is no longer a coverage gap — but the name still promises what the body cannot check, and a future reader will trust it.

## Nitpicks

- `import_vscode.rs:565,608,652` — one `let mut wrong_shape` reused across three blocks with manual resets. Correct today, but a fourth block added without the reset inherits the previous file's verdict and silently suppresses a legitimate write. A `let` per block removes the hazard.
- `docs/SETTINGS.md:103-114` — now claims croft maps "the keys below (and nothing more)" while omitting `debug.inlineValues` → `disable_inline_values`, which `SETTINGS` maps and `WORKSPACE_ALLOWED_KEYS` permits. Raised in round 2; **the exhaustiveness claim makes the omission worse than before.**
- `import_vscode.rs:80` — doc says `None` when the source key sets more than one, on a field typed `&'static str` (round 2, carried).
- `config_layers.rs:407` — still discards warnings, so the `boundary`/`trailing` → `none` loss is reported by `import-vscode` and silent in the workspace layer (round 2, carried).

**On the failing `doc comments stay with their function` check:** I reproduced it and verified your account is correct. `apply` **is** documented at HEAD (`import_vscode.rs:544-549`); the failure comes from the added-file pairwise loop reporting an adjacent-commit loss regardless of HEAD. That is the gate's gap, not this code's, and #406 fixes it — I have approved #406 at `abad571` (5/5, 5/5). Flagging only because this branch cannot merge until that lands.

## Scores

**CodeRabbit score (diff quality) — 2/5.** Two majors in the changed code: the CLI reports a refusal as "everything was already in place", and the new test's stated property is false for an unparseable file. Both follow from the same thing — the wrong-shape guard made `apply_into` safe without teaching its caller what the new outcome means.

**Greptile score (codebase fit) — 4/5.** Every round-1 and round-2 correctness finding is genuinely fixed and, where mutated, genuinely pinned. Consolidation onto one `map_settings` holds, the new `config_layers.rs` tests assert provenance rather than values, release gate satisfied (0.1.802, `release_notes.rs` replaced). Marked down for `Report::conflicts` carrying refusal messages its own doc says it does not, and three round-2 nits carried untouched.

VERDICT: REQUEST_CHANGES

vitali87:
@coderabbitai review

vitali87:
Merging, and stating the review position plainly rather than letting a stale verdict stand unexplained.

**What this PR has:** every check green, a real CodeRabbit review (five inline findings, all fixed, all threads resolved), and three rounds of fallback review whose findings are each fixed and documented in the commits.

**What it does not have:** a re-review of the current head. The last fallback round raised two majors, both now fixed:

- refusals were reported as conflicts, so the closing line told the user "everything was already in place; nothing changed" when croft had in fact declined to write. Refusals now have their own field, heading and closing line.
- an unparsable file aborted the whole import through `?`. It is now a per-file refusal, and the file is left byte-for-byte alone rather than overwritten from an empty map.

Writing that test surfaced two more, which are also fixed: `carries_comments` fired on any JSONC extra, so a file with a trailing comma and no comments was refused with a message about its comments (renamed `carries_jsonc_extras`, message widened to what it protects), and the fixture had to become `{ "a": }` to reach the path under test rather than an earlier guard.

The session that had been running the fallback reviews went offline before it could re-review, and CodeRabbit is rate-limited on this PR specifically. I re-requested and got the limit notice.

So this merges on the CodeRabbit review plus green checks plus zero unresolved threads, with the last round's fixes unreviewed by anyone but me. I would rather say that than merge quietly on a green tick, given that several of my fixes today introduced the next defect and were caught only by review.

Full suite: 3744 passed. Clippy clean at `--all-targets -D warnings`. Doc-ownership gate clean since #406 landed, which is what had been failing here.

Issue #397, opened 2026-08-29
Flaky under load: the_swallow_guard_reads_the_clicked_terminal_not_the_active_one


vitali87:
## What happens

`app::tests::the_swallow_guard_reads_the_clicked_terminal_not_the_active_one` fails under a full parallel suite run and passes when run alone.

```
thread 'app::tests::the_swallow_guard_reads_the_clicked_terminal_not_the_active_one'
panicked at src/app/tests.rs:2648:5:
pane 0 must carry the link, or the built-in has nothing to act on
```

Full suite (`RUST_TEST_THREADS=8 cargo test --bin croft`): `3720 passed; 1 failed`.
Same binary, same commit, test alone: `1 passed; 0 failed` in 0.50s.

## Why it is worth a look rather than a retry

A test that only fails under load is usually reporting a real timing assumption rather than a flaky harness. This one asserts that a link scraped from terminal output has reached pane 0 by the time the assertion runs; under load the pane's output has not been drained yet, so the assertion reads an empty link set. That is the same shape as the waits that were converted to load-scaled budgets in `src/test_budget.rs` (#307), and the fix is likely to be a bounded wait for the link to appear rather than an immediate read.

## Scope

Not caused by the branch it was seen on (a VS Code theme importer that touches no terminal code), and it reproduces against `origin/main` content. It appeared in runs after #390 (secret redaction) landed, which does touch terminal output scanning, so that is the first place worth looking: if redaction now sits between the PTY read and the link scrape, it adds a step the test's timing never accounted for.

## Reproduction

```
RUST_TEST_THREADS=8 cargo test --bin croft
```

on a machine with enough cores to actually run 8 in parallel. It does not reproduce every run; it showed up once in a full-suite run and not in the two before it, so a loop of a few runs may be needed.

vitali87:
Two more instances today, which changes the shape of this issue: it is not one flaky test but a class.

Same signature each time, a full parallel run failing and the named test passing alone on the same commit:

- `app::tests::the_swallow_guard_reads_the_clicked_terminal_not_the_active_one` (the original report)
- `pair::tests::host_adds_appends_and_removes_notes_by_id`
- `app::tests::terminal_session_restores_pane_layout_names_and_focus_across_restarts`

[...]

PR #401, opened 2026-08-29
test(log): pin the find bar's title width and the count/highlight invariant

vitali87:
Two follow-ups the #393 review asked for, after that PR merged. No behaviour change: both tests live in `src/app/tests.rs`, which the release-hygiene gate excludes from "shipped", so there is no version bump or release note.

## Why these two

**The title test exists because a rendering bug shipped invisibly.** #393's truncated-count title was 51 characters in a bar capped at 48 and painted as `...in the part that could be searc`. Nothing in the suite rendered `render_editor_find` and looked at the result, so nothing could have caught it, and the 37-character replacement was protected only by arithmetic no one re-runs. This renders every arm of that `match` at the bar's real width and asserts the fragment survives whole.

**The invariant test exists because three review rounds found three points on one line.** Rounds 3, 4 and 5 each found a different instance of the same class: the bar's count and the painted highlight disagreeing. A stale highlight under "no match"; a cleared highlight under "1+ matches"; a stale highlight after a retype. Each round pinned its own instance. This asserts the property directly (a count of zero must never coexist with a painted match) across typing, retyping, stepping forward, stepping back, and deleting back to a query that matches again.

## Both are verified positive controls

Neither is asserted to work; both were checked by re-breaking the fix.

Restoring the 51-character title:

```
the title was clipped: wanted "no match in the part searched" whole,
painted "╭ Find: no match in the part that could be searc"
```

Removing the clear that round 4 added to `log_find_set_query`:

```
after retyping into a query that does not: the bar says no match
while a highlight is painted at Some((0, 0, 5))
```

That check is the point of the PR as much as the tests are. Two of my tests in the #393 series passed against the very bugs they existed to catch, and the only thing that distinguished them from real coverage was removing the fix and watching them fail.

## Coverage this does not claim

The invariant is one-directional: a count of zero implies no painted match. The other direction, a count that survives while the highlight is cleared, is what `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` pins, and it is named in the test's doc comment so the pair is legible rather than looking like an oversight.

Full suite: 3725 passed, 0 failed. Clippy clean at `--all-targets -D warnings`.

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -35452,9 +35452,10 @@ fn the_count_and_the_highlight_agree_at_every_step_of_a_log_search() {
         Some((0, 0, 5)),
         "typing lands on the first match, at its own column"
     );
-    assert!(
-        app.editor_find.as_ref().unwrap().match_count > 0,
-        "the match is counted"
+    assert_eq!(
+        app.editor_find.as_ref().unwrap().match_count,
+        2,
+        "both head matches are counted, not merely some"
     );
 
     // Forward to the SECOND match, which is within budget, so this leg
@@ -35515,8 +35516,8 @@ fn the_count_and_the_highlight_agree_at_every_step_of_a_log_search() {
             .unwrap();
     }
     assert_eq!(
-        app.editor.active_search_match.map(|(r, _, _)| r),
-        Some(0),
-        "a query that matches again must paint its match"
+        app.editor.active_search_match,
+        Some((0, 0, 5)),
+        "a query that matches again must paint its match, whole"
     );
 }
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -35322,6 +35322,22 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
         let row: String = (rect.x..rect.x + rect.width)
             .map(|x| buf[(x, rect.y)].symbol().to_string())
             .collect();
+        // THE BAR'S TITLE AREA, not the buffer. Two renderers paint this
+        // title and they clip differently: `Block` clips to its own title
+        // area (`rect.width - 2`), while the gradient re-stamp in
+        // `render_editor_find` uses `set_span`, whose clamp is the BUFFER's
+        // right edge. `Theme::BLACK.gradient()` is true, so the second one
+        // runs, and a title between 38 and 47 cells overruns the block
+        // silently: it writes over the bar's own top-right corner and out
+        // past the inner area, while a comparison anchored to the painted
+        // row still matches, because the overrun is exactly the text the
+        // comparison wanted. The width assertion is what makes that visible.
+        assert!(
+            expected.chars().count() <= (rect.width as usize).saturating_sub(2),
+            "the {}-cell title does not fit a {}-cell bar's title area",
+            expected.chars().count(),
+            rect.width
+        );
         // Exactly, at the position the block paints it: one cell in from
         // the left corner. `contains` would be weaker in both directions -
         // it cannot see a clipped tail when the wanted text is a fragment,
@@ -35332,6 +35348,14 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
             painted, expected,
             "the title did not fit whole in a {}-cell bar, the row painted {row:?}",
             rect.width
+        );
+        // And the bar is still a bar. The corner is the first thing an
+        // overrunning title destroys, so it is the cheapest witness that the
+        // title stayed inside the box it was painted in.
+        assert_eq!(
+            buf[(rect.x + rect.width - 1, rect.y)].symbol(),
+            "\u{256e}",
+            "the title overran the bar and ate its top-right corner: {row:?}"
         );
     }
 }
@@ -35362,18 +35386,52 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
 /// HALF A PAIR. These pin `count == 0` with nothing painted. The other
 /// direction, a count that survives while the highlight is thrown away, is
 /// pinned by `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had`.
+///
+/// WHICH CHECKPOINTS ARE NEW. Three of them re-assert what
+/// `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` and
+/// `retyping_into_an_out_of_reach_query_clears_the_old_highlight` already
+/// cover; they are carried along because this test's subject is the
+/// SEQUENCE, and a state is only reached by passing through them. The two
+/// that no sibling covers are the backward step and the final backspace
+/// into a query that matches again. A reader counting five independent
+/// facts here would be counting wrong.
+///
+/// TWO REACHABLE MATCHES, not one. With a single match the Shift+Enter
+/// checkpoint asserted nothing: `find_prev`'s leg 1 rejects the anchor's own
+/// match, leg 2 walks back from EOF and runs out of budget before reaching
+/// row 0, and `scroll_log_to_match` returns on `out_of_reach` WITHOUT
+/// touching `active_search_match`. The assertion passed because the field
+/// was never written. Measured: making `find_prev` return
+/// `Found(row 0, col 999, len 5)` unconditionally left the test green. A
+/// second match on row 1 puts a real backward step within budget, so the
+/// checkpoint now fails if `find_prev`'s anchor filter or its column
+/// arithmetic regresses.
 #[test]
 fn the_count_and_the_highlight_agree_at_every_step_of_a_log_search() {
     let tmp = tempfile::tempdir().unwrap();
     let p = tmp.path().join("budgeted.log");
-    let mut body = String::from("\u{1b}[31mERROR\u{1b}[0m first and only\n");
+    // TWO matches at the head, both inside the budget, so a backward step
+    // has somewhere to land. Row 1 puts ERROR at character column 3, which
+    // makes a wrong column visible rather than coincidentally right.
+    let mut body = String::from("\u{1b}[31mERROR\u{1b}[0m first\n");
+    body.push_str("no \u{1b}[31mERROR\u{1b}[0m second\n");
     let filler = "quiet line with nothing of interest ".repeat(4);
-    // Just past the budget: enough for the sweep to run out, no more.
-    for _ in 0..(crate::log_view::FIND_SCAN_BYTES / filler.len() + 200) {
+    // Comfortably past the budget, matching the margin the sibling tests use
+    // (+3000 lines, ~11%). The earlier +200 was 1.4% over, and while the
+    // budget is a pure byte counter with no timing in it, a change to window
+    // accounting or to the stripped length of the colour header could flip
+    // the sweep from OutOfReach to Absent and quietly change what four
+    // assertions mean without any of them failing.
+    for _ in 0..(crate::log_view::FIND_SCAN_BYTES / filler.len() + 3000) {
         body.push_str(&filler);
         body.push('\n');
     }
     std::fs::write(&p, &body).unwrap();
+    assert!(
+        body.len() > crate::log_view::FIND_SCAN_BYTES,
+        "the fixture must outrun the sweep budget or the out-of-reach \
+         checkpoints test nothing"
+    );
 
     let mut app = App::new(tmp.path().to_path_buf()).unwrap();
     app.focus_pane(Pane::Editor);
@@ -35386,24 +35444,36 @@ fn the_count_and_the_highlight_agree_at_every_step_of_a_log_search() {
         app.handle_key(key(KeyCode::Char(ch), KeyModifiers::NONE))
             .unwrap();
     }
+    // The WHOLE tuple, not just the row. A wrong column or length paints the
+    // band in the wrong place, which is the defect class this test is for,
+    // and comparing only the row cannot see it.
     assert_eq!(
-        app.editor.active_search_match.map(|(r, _, _)| r),
-        Some(0),
-        "typing lands on the only match"
+        app.editor.active_search_match,
+        Some((0, 0, 5)),
+        "typing lands on the first match, at its own column"
     );
     assert!(
         app.editor_find.as_ref().unwrap().match_count > 0,
         "the match is counted"
     );
 
-    // Step forward WHILE THE QUERY STILL MATCHES, so the step does something:
-    // the next match is past the budget, so this exercises the out-of-reach
-    // branch, which is where round 3's defect lived.
+    // Forward to the SECOND match, which is within budget, so this leg
+    // asserts a real step rather than a refusal.
     app.handle_key(key(KeyCode::Enter, KeyModifiers::NONE))
         .unwrap();
     assert_eq!(
-        app.editor.active_search_match.map(|(r, _, _)| r),
-        Some(0),
+        app.editor.active_search_match,
+        Some((1, 3, 5)),
+        "the next match is row 1 at column 3, past the \"no \" prefix"
+    );
+
+    // Forward again: nothing further is reachable, so this exercises the
+    // out-of-reach branch, which is where round 3's defect lived.
+    app.handle_key(key(KeyCode::Enter, KeyModifiers::NONE))
+        .unwrap();
+    assert_eq!(
+        app.editor.active_search_match,
+        Some((1, 3, 5)),
         "a step that cannot reach further keeps the match it had"
     );
     assert!(
@@ -35411,12 +35481,16 @@ fn the_count_and_the_highlight_agree_at_every_step_of_a_log_search() {
         "and the bar reports that the search did not cover the file"
     );
 
+    // Backward from row 1 to row 0: within budget, so `find_prev`'s leg 1
+    // answers and its result is actually consumed. With one match this leg
+    // ran out of budget and `scroll_log_to_match` returned without writing
+    // anything, so the assertion passed on an untouched field.
     app.handle_key(key(KeyCode::Enter, KeyModifiers::SHIFT))
         .unwrap();
     assert_eq!(
-        app.editor.active_search_match.map(|(r, _, _)| r),
-        Some(0),
-        "stepping back over a single match returns to it"
+        app.editor.active_search_match,
+        Some((0, 0, 5)),
+        "stepping back lands on the previous match, not on the current one"
     );
 
     // Only now retype into something unmatchable, so the transition from a
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -35269,6 +35269,7 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
     // Written out rather than rebuilt from the code's own format strings on
     // purpose: sharing the construction with the code under test would make
     // any widening agree with itself and pass.
+    /// A title arm: the whole title it must paint, and how to reach it.
     type TitleCase = (&'static str, Box<dyn Fn(&mut EditorFind)>);
 
     let cases: Vec<TitleCase> = vec![
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -35228,3 +35228,211 @@ fn a_log_that_paints_nothing_publishes_no_body_rect() {
          mouse path to hit-test against"
     );
 }
+
+/// #257 follow-up: the find bar's title must FIT the bar.
+///
+/// The truncated-count arm shipped at 51 characters in a bar capped at 48
+/// and painted as "...in the part that could be searc". Nothing rendered the
+/// title, so nothing could catch it, and the replacement is protected only
+/// by arithmetic no one re-runs. This renders every arm at the real width.
+#[test]
+fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
+    use crate::widgets::editor_find::{EditorFind, render_editor_find};
+    use ratatui::buffer::Buffer;
+    use ratatui::layout::Rect;
+
+    let area = Rect {
+        x: 0,
+        y: 0,
+        width: 60,
+        height: 12,
+    };
+    let opts = crate::widgets::search::SearchOpts::default();
+
+    // Every arm of the title match, with the fragment that must survive
+    // whole. `match_index` matters as much as the count: the "N of M" arm is
+    // reached ONLY when it is Some, and `set_match_count` does not touch it,
+    // so the first version of this test never rendered that arm at all,
+    // while claiming in its name to cover every one. It is also the arm
+    // whose width grows with TWO runtime numbers, which makes it the most
+    // able to overrun the bar.
+    /// A title arm: the fragment that must survive, and how to reach it.
+    type TitleCase = (&'static str, Box<dyn Fn(&mut EditorFind)>);
+
+    let cases: Vec<TitleCase> = vec![
+        // The empty-query arm is the BARE title. Asserting it contains
+        // "Find" proved nothing: every arm does. It is identified by what it
+        // must NOT carry.
+        ("Find", Box::new(|s: &mut EditorFind| s.query.clear())),
+        (
+            "No results",
+            Box::new(|s: &mut EditorFind| s.set_match_count(0, false)),
+        ),
+        (
+            "no match in the part searched",
+            Box::new(|s: &mut EditorFind| s.set_match_count(0, true)),
+        ),
+        (
+            "1234 matches",
+            Box::new(|s: &mut EditorFind| s.set_match_count(1234, false)),
+        ),
+        (
+            "1234+ matches",
+            Box::new(|s: &mut EditorFind| s.set_match_count(1234, true)),
+        ),
+        (
+            "7 of 1234",
+            Box::new(|s: &mut EditorFind| {
+                s.set_match_count(1234, false);
+                s.match_index = Some(7);
+            }),
+        ),
+        (
+            "9876543 of 9876543+",
+            Box::new(|s: &mut EditorFind| {
+                // Both numbers as wide as they plausibly get on a log, since
+                // this arm's length grows with each independently.
+                s.set_match_count(9_876_543, true);
+                s.match_index = Some(9_876_543);
+            }),
+        ),
+    ];
+
+    for (fragment, prepare) in cases {
+        let mut state = EditorFind::new(String::from("needle"), opts);
+        prepare(&mut state);
+        let mut buf = Buffer::empty(area);
+        render_editor_find(&mut state, area, &mut buf, crate::theme::Theme::BLACK);
+
+        let rect = state.last_rect;
+        assert!(rect.width > 0, "the bar must have been painted");
+        let row: String = (rect.x..rect.x + rect.width)
+            .map(|x| buf[(x, rect.y)].symbol().to_string())
+            .collect();
+        assert!(
+            row.contains(fragment),
+            "the title was clipped: wanted {fragment:?} whole, painted {row:?}"
+        );
+        if fragment == "Find" {
+            // Every arm contains "Find", so containment alone cannot tell
+            // this arm from the others: a completely wrong bare title passed.
+            for other in ["of", "matches", "No results", "no match"] {
+                assert!(
+                    !row.contains(other),
+                    "the empty-query title must be bare, painted {row:?}"
+                );
+            }
+        }
+    }
+}
+
+/// #257 follow-up: the count and the highlight must agree at every step of
+/// a search over a budgeted log.
+///
+/// Rounds 3, 4 and 5 of #393 each found one instance of a single class: the
+/// bar's count and the painted highlight disagreeing. This walks the whole
+/// gesture sequence and asserts the exact state after each one: type, step
+/// forward into the out-of-reach branch (where round 3's defect lived),
+/// step back, retype into a query that matches nothing, delete back.
+///
+/// NO SEPARATE INVARIANT HELPER, deliberately. An earlier version carried a
+/// closure asserting `match_count == 0 => active_search_match.is_none()` at
+/// each checkpoint, which reads as the general property behind the three
+/// defects. It could not fail: `match_count == 0` holds at exactly one
+/// checkpoint, and there the concrete assertion already fixes the highlight
+/// to `None`, which is strictly stronger. Neutering the closure entirely
+/// left the test green. Extending it over a prefix walk did not help either,
+/// because the code has no per-length behaviour, so every state the walk
+/// visits is behaviourally identical to one already enumerated.
+///
+/// A dead assertion whose name announces a property is worse than no
+/// assertion: it makes a reader believe the class is covered. The concrete
+/// assertions ARE the property, made checkable one state at a time.
+///
+/// HALF A PAIR. These pin `count == 0` with nothing painted. The other
+/// direction, a count that survives while the highlight is thrown away, is
+/// pinned by `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had`.
+#[test]
+fn the_count_and_the_highlight_agree_at_every_step_of_a_log_search() {
+    let tmp = tempfile::tempdir().unwrap();
+    let p = tmp.path().join("budgeted.log");
+    let mut body = String::from("\u{1b}[31mERROR\u{1b}[0m first and only\n");
+    let filler = "quiet line with nothing of interest ".repeat(4);
+    // Just past the budget: enough for the sweep to run out, no more.
+    for _ in 0..(crate::log_view::FIND_SCAN_BYTES / filler.len() + 200) {
+        body.push_str(&filler);
+        body.push('\n');
+    }
+    std::fs::write(&p, &body).unwrap();
+
+    let mut app = App::new(tmp.path().to_path_buf()).unwrap();
+    app.focus_pane(Pane::Editor);
+    app.editor.open(&p).unwrap();
+    assert!(app.editor.log.is_some(), "the fixture opens rendered");
+
+    app.handle_key(key(KeyCode::Char('f'), KeyModifiers::SUPER))
+        .unwrap();
+    for ch in "ERROR".chars() {
+        app.handle_key(key(KeyCode::Char(ch), KeyModifiers::NONE))
+            .unwrap();
+    }
+    assert_eq!(
+        app.editor.active_search_match.map(|(r, _, _)| r),
+        Some(0),
+        "typing lands on the only match"
+    );
+    assert!(
+        app.editor_find.as_ref().unwrap().match_count > 0,
+        "the match is counted"
+    );
+
+    // Step forward WHILE THE QUERY STILL MATCHES, so the step does something:
+    // the next match is past the budget, so this exercises the out-of-reach
+    // branch, which is where round 3's defect lived.
+    app.handle_key(key(KeyCode::Enter, KeyModifiers::NONE))
+        .unwrap();
+    assert_eq!(
+        app.editor.active_search_match.map(|(r, _, _)| r),
+        Some(0),
+        "a step that cannot reach further keeps the match it had"
+    );
+    assert!(
+        app.editor_find.as_ref().unwrap().count_truncated(),
+        "and the bar reports that the search did not cover the file"
+    );
+
+    app.handle_key(key(KeyCode::Enter, KeyModifiers::SHIFT))
+        .unwrap();
+    assert_eq!(
+        app.editor.active_search_match.map(|(r, _, _)| r),
+        Some(0),
+        "stepping back over a single match returns to it"
+    );
+
+    // Only now retype into something unmatchable, so the transition from a
+    // painted match to none is itself a checkpoint.
+    for ch in "ZZQX".chars() {
+        app.handle_key(key(KeyCode::Char(ch), KeyModifiers::NONE))
+            .unwrap();
+    }
+    assert_eq!(
+        app.editor_find.as_ref().unwrap().match_count,
+        0,
+        "the new query matches nothing"
+    );
+    assert_eq!(
+        app.editor.active_search_match, None,
+        "so nothing may stay painted"
+    );
+
+    // And back: the highlight must return rather than staying cleared.
+    for _ in 0..4 {
+        app.handle_key(key(KeyCode::Backspace, KeyModifiers::NONE))
+            .unwrap();
+    }
+    assert_eq!(
+        app.editor.active_search_match.map(|(r, _, _)| r),
+        Some(0),
+        "a query that matches again must paint its match"
+    );
+}
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -35249,46 +35249,58 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
     };
     let opts = crate::widgets::search::SearchOpts::default();
 
-    // Every arm of the title match, with the fragment that must survive
-    // whole. `match_index` matters as much as the count: the "N of M" arm is
-    // reached ONLY when it is Some, and `set_match_count` does not touch it,
-    // so the first version of this test never rendered that arm at all,
-    // while claiming in its name to cover every one. It is also the arm
-    // whose width grows with TWO runtime numbers, which makes it the most
-    // able to overrun the bar.
-    /// A title arm: the fragment that must survive, and how to reach it.
+    // Every arm of the title match, with the WHOLE title it must paint, and
+    // how to reach it. `match_index` matters as much as the count: the
+    // "N of M" arm is reached ONLY when it is Some, and `set_match_count`
+    // does not touch it, so the first version of this test never rendered
+    // that arm at all while claiming in its name to cover every one. It is
+    // also the arm whose width grows with TWO runtime numbers, which makes
+    // it the most able to overrun the bar.
+    //
+    // The expected string is the ENTIRE title, never a fragment of it. A
+    // fragment cannot detect an overrun: text added BEFORE the fragment
+    // pushes the tail off the end while leaving the fragment itself intact,
+    // so `contains` still passes. Measured rather than reasoned: against the
+    // fragment "9876543 of 9876543+", widening that arm to
+    // " Find - match {idx} of {total} found in this whole file " (59 cells
+    // painted into a 48-cell bar) left this test green. Only the full title
+    // makes the clip observable, because the clip removes its end.
+    //
+    // Written out rather than rebuilt from the code's own format strings on
+    // purpose: sharing the construction with the code under test would make
+    // any widening agree with itself and pass.
     type TitleCase = (&'static str, Box<dyn Fn(&mut EditorFind)>);
 
     let cases: Vec<TitleCase> = vec![
-        // The empty-query arm is the BARE title. Asserting it contains
-        // "Find" proved nothing: every arm does. It is identified by what it
-        // must NOT carry.
-        ("Find", Box::new(|s: &mut EditorFind| s.query.clear())),
+        // The empty-query arm is the BARE title, and it is asserted whole
+        // for the same reason as the rest: every other arm also contains
+        // "Find", so a fragment could not tell this arm from a wrong one.
+        (" Find ", Box::new(|s: &mut EditorFind| s.query.clear())),
         (
-            "No results",
+            " Find — No results ",
             Box::new(|s: &mut EditorFind| s.set_match_count(0, false)),
         ),
         (
-            "no match in the part searched",
+            " Find: no match in the part searched ",
             Box::new(|s: &mut EditorFind| s.set_match_count(0, true)),
         ),
         (
-            "1234 matches",
+            " Find — 1234 matches ",
             Box::new(|s: &mut EditorFind| s.set_match_count(1234, false)),
         ),
         (
-            "1234+ matches",
+            " Find — 1234+ matches ",
             Box::new(|s: &mut EditorFind| s.set_match_count(1234, true)),
         ),
         (
-            "7 of 1234",
+            " Find — 7 of 1234 ",
             Box::new(|s: &mut EditorFind| {
                 s.set_match_count(1234, false);
                 s.match_index = Some(7);
             }),
         ),
         (
-            "9876543 of 9876543+",
+            " Find — 9876543 of 9876543+ ",
             Box::new(|s: &mut EditorFind| {
                 // Both numbers as wide as they plausibly get on a log, since
                 // this arm's length grows with each independently.
@@ -35298,7 +35310,7 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
         ),
     ];
 
-    for (fragment, prepare) in cases {
+    for (expected, prepare) in cases {
         let mut state = EditorFind::new(String::from("needle"), opts);
         prepare(&mut state);
         let mut buf = Buffer::empty(area);
@@ -35309,20 +35321,17 @@ fn every_find_bar_title_fits_the_bar_it_is_painted_in() {
         let row: String = (rect.x..rect.x + rect.width)
             .map(|x| buf[(x, rect.y)].symbol().to_string())
             .collect();
-        assert!(
-            row.contains(fragment),
-            "the title was clipped: wanted {fragment:?} whole, painted {row:?}"
+        // Exactly, at the position the block paints it: one cell in from
+        // the left corner. `contains` would be weaker in both directions -
+        // it cannot see a clipped tail when the wanted text is a fragment,
+        // and for the bare " Find " arm every other title contains it too,
+        // so a completely wrong title would satisfy it.
+        let painted: String = row.chars().skip(1).take(expected.chars().count()).collect();
+        assert_eq!(
+            painted, expected,
+            "the title did not fit whole in a {}-cell bar, the row painted {row:?}",
+            rect.width
         );
-        if fragment == "Find" {
-            // Every arm contains "Find", so containment alone cannot tell
-            // this arm from the others: a completely wrong bare title passed.
-            for other in ["of", "matches", "No results", "no match"] {
-                assert!(
-                    !row.contains(other),
-                    "the empty-query title must be bare, painted {row:?}"
-                );
-            }
-        }
     }
 }
 


vitali87:

Running because CodeRabbit's check reads `pass` annotated **"Review rate limited"**, which is a no-op rather than a review. Reviewed by a session on another machine at the author's request. The reviewer built the branch, ran the tests, and mutation-tested them — findings below are from execution, not reading.

---

## Walkthrough

Adds two tests to `src/app/tests.rs` and changes no behaviour: a render-level assertion that the find bar's title survives whole at the bar's real 48-cell width, and a class-level invariant that a zero match count never coexists with a painted highlight across a gesture sequence on a budgeted log. Both quoted positive controls were reproduced exactly, character for character. However, mutation testing found the title test does **not** render every arm of the title `match` despite its name and doc comment saying so, and instrumenting the invariant test showed two of its five labelled checkpoints assert nothing while two more re-assert an unchanged state.

| file | summary |
|---|---|
| `src/app/tests.rs` | +134 lines, two tests. `every_find_bar_title_fits_the_bar_it_is_painted_in` renders 5 title cases at width 48 and asserts a fragment survives; `the_find_bars_count_and_highlight_never_disagree_on_a_budgeted_log` drives a ~4.7 MB log fixture through type/retype/step/step-back/delete and checks count-0 implies no highlight. |

## Findings

- **`src/app/tests.rs:35019-35040` [major/tests]** — The test claims to render *every* arm of the title `match`, but the `(_, Some(idx)) => " Find — {idx} of {total_txt} "` arm is never reached, so a title overrun in that arm is invisible to it. `EditorFind::match_index` is `Option<usize>` and defaults to `None` (`src/widgets/editor_find.rs:19-37`, `#[derive(Default)]`); none of the five cases sets it, and `set_match_count` (`editor_find.rs:60-63`) writes only `match_count` and `count_truncated`. So all five cases land on arms 1, 3, 2, 5, 5 of the six-arm match at `editor_find.rs:415-430`.

  **Verified by mutation:** padding that arm to 92 characters — nearly double the 48-cell bar — left the test reporting `ok`. This is exactly the failure mode the test was written to prevent (an arm protected "only by arithmetic no one re-runs"), left live in the one arm whose width scales with *two* runtime numbers rather than one. Independently confirmed: `gh pr diff 401 | grep -c match_index` returns **0**.

  ```rust
  ("1234 of 1234", Box::new(|s: &mut EditorFind| {
      s.set_match_count(1234, false);
      s.match_index = Some(1234);
  })),
  ```
  plus a truncated variant with `set_match_count(1234, true)` to cover `" Find — 1234 of 1234+ "`.

- **`src/app/tests.rs:35093-35102` [major/tests]** — The "stepping forward" and "stepping back" checkpoints are inert: they exercise no step and re-assert a state that did not change. At those points the query is `ERRORZZQX`, which matches nothing, so `log_find_step` (`src/app/mod.rs:30996-31015`) finds no hit and `scroll_log_to_match` takes the `step.found()` `None` path.

  **Verified by instrumenting:** the tuple `(active_search_match, scroll, match_count, match_index)` is `(None, 0, 0, None)` before the forward step, after it, and after the back step — `changed=false` both times. The doc comment and PR body advertise coverage "across typing, retyping, stepping forward, stepping back", but stepping is only ever exercised on a non-matching query, which is the case where stepping is a no-op. Round 3's actual bug — a step that runs out of budget clearing the highlight — lives on the `out_of_reach` branch (`mod.rs:31025-31036`), which this test never enters. **Fix:** perform the steps while the query still matches (before retyping to `ZZQX`).

- **`src/app/tests.rs:35084-35092` [minor/tests]** — Two of the five `check` calls are vacuous, and the assertion is guarded such that a bug making the count wrongly *nonzero* is silently tolerated. The closure only asserts inside `if state.match_count == 0`. **Verified by instrumenting:** checkpoint 1 ("typing a query that matches") sees `count=1` and checkpoint 5 ("deleting back to a matching query") sees `count=1`, so neither runs an assertion — the labels read as coverage but execute nothing. Combined with the finding above, exactly **one** distinct assertion actually executes. **Fix:** at the matching checkpoints assert the positive form (`count > 0 && active_search_match.is_some()`) rather than relying on a guard that skips.

- **`src/app/tests.rs:35018` [minor/style]** — A `///` doc comment is attached to a `type` alias inside a function body, immediately after a `//` line comment saying the same thing. `CONTRIBUTING.md:68-74` warns specifically that `///` binds to whatever item *follows* it; here the intent reads as a comment on the case list above. **Fix:** make it a plain `//`, or delete it as redundant.

- **`src/app/tests.rs:35070-35078` [nit/performance]** — The fixture writes ~4.7 MB (144-byte filler × 32,127 lines) and the test runs ~2.7 s, the slowest in the find suite. It is duplicated verbatim from `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` (`tests.rs:34908-34918`) and `retyping_into_an_out_of_reach_query_clears_the_old_highlight` (`34960-34968`) — three copies. **Fix:** extract `fn budgeted_log_fixture(dir: &Path, name: &str) -> PathBuf`.

## Test coverage

For a test-only PR this is the substance, so stated plainly: **what these tests appear to cover but do not.**

- **The `N of M` title arm is uncovered** while the test name and doc claim otherwise. Highest-value gap, because it is the arm whose length depends on two independently-growing numbers, and it is now the only arm with no width protection at all. A reader auditing "is the title width tested?" would read the name, see `ok`, and stop. Confirmed by a surviving 92-character mutant.
- **The out-of-reach step branch is uncovered by the new invariant test.** `scroll_log_to_match`'s `step.out_of_reach()` early return (`mod.rs:31025-31036`) is where round 3's bug lived and is the one path that deliberately preserves a highlight while marking the count partial — precisely the interesting interaction. The older `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` does reach it, which is lucky, but the new test's own claim to cover stepping is not met.
- **The invariant is checked at only one distinct state.** Five checkpoints reduce to one executed, non-duplicated assertion. A rich-looking gesture sequence is acting as a hiding place: the sequence's length suggests breadth it does not have.
- **Empty-query transitions are untested.** `log_find_set_query`'s `needle.is_empty()` branch (`mod.rs:30962-30968`) sets count 0 and clears the highlight — directly on the invariant — and is reachable by deleting all the way back. The test deletes exactly 4 characters, stopping at `ERROR`.
- **What I would add:** a `match_index` case in the title test; steps performed while the query matches, plus one that exhausts the budget; positive assertions at the matching checkpoints; a delete-to-empty checkpoint.

**Placement and release hygiene both check out.** `src/app/tests.rs` is designated "unit / integration tests" in `docs/ARCHITECTURE.md:109` and already contains four `Buffer::empty` render tests (5117, 5591, 10263), so a render-level test is at home there. On the release rule: `CONTRIBUTING.md:56-66` requires a version bump and release-notes replacement for anything compiled into the binary, then states "Docs, CI, and test-only PRs (`src/app/tests.rs`, `tests/`) are exempt." Checked against the **gate** rather than the prose — `.github/workflows/ci.yml:50-52` greps the changed set then `grep -vE '^(src/app/tests\.rs|tests/)'`. This PR touches that file only, so `shipped` is empty and the job exits 0. **The exemption is correctly claimed.**

## Nitpicks

- `src/app/tests.rs:35008-35013` — `area` is 60×12 but only `rect.y` is ever read; the height and extra columns are unused slack.
- `src/app/tests.rs:35047-35050` — `row.contains(fragment)` would pass on a title truncated at the *front* as long as the fragment survives. Fine for the clipping bug being pinned, but a `starts_with`-anchored check would be strictly stronger.
- `src/app/tests.rs:35044` — `assert!(rect.width > 0, ...)` is weaker than it could be; `rect.width == 48` would pin the actual cap the premise depends on.

## The author's two framing claims

**Positive controls — HOLDS, fully.** Both were reproduced by re-breaking the fixes on the PR branch, with the tree verified byte-clean afterwards each time.
- Restoring the 51-character title produced exactly the quoted text: `the title was clipped: wanted "no match in the part searched" whole, painted "╭ Find: no match in the part that could be searc"`.
- Removing `self.editor.active_search_match = None;` from `log_find_set_query` (`mod.rs:30991`) produced exactly: `after retyping into a query that does not: the bar says no match while a highlight is painted at Some((0, 0, 5))`.

These are the controls for *these* tests, not for an earlier draft. This is the strongest part of the PR and the claim is honest.

**One-directional invariant and its named counterpart — PARTIALLY HOLDS; the "named" half is false.** The invariant genuinely is one-directional (the `if match_count == 0` guard), and `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` does substantively cover the other direction — instrumented, it ends at `count=1, truncated=true, highlight=Some((0, 0, 5))`, a surviving count with a painted highlight — so the pairing is materially sound.

But the PR body states the counterpart "is named in the test's doc comment so the pair is legible rather than looking like an oversight." **It is not.** `grep` for that test name across `src/app/tests.rs` returns a single hit: its own definition at line 34907. Independently confirmed: `gh pr diff 401 | grep -c a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` returns **0**. The new test's doc comment (35060-35066) never names it, never says the invariant is one-directional, and never mentions the other direction. The gap is real and correctly identified, but the mechanism claimed to document it was not shipped — which matters, because that doc comment is the only place a future reader would learn the invariant is half of a pair. **Fix:** add the sentence, as the PR body says was done.

## Scores

- **CodeRabbit score (diff quality) — 2/5.** Two majors: the title test's name and doc claim "every arm" while the `N of M` arm survives a 92-character mutation, and the invariant test's advertised step coverage is inert on an unchanged state.
- **Greptile score (codebase fit) — 4/5.** Placement, the release-hygiene exemption and the positive-control discipline all match repo convention, each verified against the gate rather than the prose; only minors — a `///` on an in-function type alias against `CONTRIBUTING.md:68-74`, and a fixture triplicated across three tests.

VERDICT: REQUEST_CHANGES

vitali87:
Both majors fixed, and one correction to this PR's body.

**The body was wrong.** It said the counterpart test covering the other direction is "named in the test's doc comment so the pair is legible rather than looking like an oversight". It was not named anywhere. The gap was real and correctly described; the mechanism I claimed documented it did not exist. The doc comment now names `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had`, since that comment is the only place a future reader would learn the invariant is half of a pair.

That is the same shape as the findings themselves, in the artifact rather than the code: a description of coverage the code did not contain.

**Major 1: the title test did not render every arm.** `match_index` is `Option<usize>` defaulting to `None`, and `set_match_count` writes only the count and its truncation flag, so nothing ever reached the `{idx} of {total}` arm. Five of six arms, under a name and a doc comment claiming all of them, and the missing one is the arm whose width grows with **two** runtime numbers rather than one, which makes it the most able to overrun the bar.

Two cases added. Verified with the reviewer's own mutation: padding that arm to 92 characters in a 48-cell bar used to leave the test green, and now fails it.

```
the title was clipped: wanted "7 of 1234" whole,
painted "╭ Find PADPADPADPADPADPADPADPADPADPADPADPADPADPA"
```

**Major 2: the checkpoints were inert.** By the time the test stepped, the query was `ERRORZZQX`, which matches nothing, so the steps were no-ops and the `count == 0` guard was false at exactly the two checkpoints where a step had done something. Five labelled checkpoints, one live assertion between them.

The sequence now steps **while the query still matches**, which is what reaches the out-of-reach branch where round 3's defect actually lived, then retypes to something unmatchable, then deletes back. Every checkpoint asserts a concrete expected state as well as the implication, so a checkpoint that stops exercising anything fails rather than passing quietly.

Both fixes are in the tests, not the code: nothing about croft's behaviour changed, and the release-hygiene exemption still applies (`src/app/tests.rs` only).

Full suite: 3724 passed. The one failure in that run was `terminal_session_restores_pane_layout_names_and_focus_across_restarts`, which passes in isolation and is unrelated to this branch: it is the third distinct load-sensitive test seen today, recorded on #397.

vitali87:

CodeRabbit's check still reads `pass` annotated **"Review rate limited"**, a no-op rather than a review. Reviewed by a session on another machine. Both round-1 majors verified fixed **by re-running the original mutations**, not by reading.

---

## Walkthrough

Round 2 revises the two tests added in round 1. The title test gains two cases that set `match_index` so the `{idx} of {total}` arm is reached, and the invariant test is resequenced so the Enter/Shift+Enter steps run while the query still matches, with a concrete `assert_eq!` beside each checkpoint. Both round-1 majors verify as genuinely fixed. However, the invariant test's *named subject* — the `count == 0 ⇒ no painted match` implication — is still not load-bearing, and one title case cannot fail.

| File | Change |
|---|---|
| `src/app/tests.rs` | +2 title cases, resequenced steps, 6 new `assert_eq!`, expanded doc comments. Test-only; correctly exempt from release hygiene. |

## Round-1 verification

1. **Title test — FIXED.** Re-ran the exact mutation. Padding the `(_, Some(idx))` arm to 88 characters in the 48-cell bar now **FAILS**: `the title was clipped: wanted "7 of 1234" whole, painted "╭ Find — PADPADPADPADPADPADPADPADPADPADPADPADPAD"`. Instrumenting the renderer confirms all six arms are now genuinely reached. File restored and sha-verified.
2. **Inert checkpoints — FIXED.** The instrumented probe now reads `active=Some((0,0,5)) count=1` at four of five checkpoints, against round 1's `(None, 0, 0, None)` everywhere. Instrumenting `scroll_log_to_match` proves the forward step enters the intended branch: `STEPPROBE out_of_reach=true found=None` fires at both steps. All five checkpoints execute an assertion.
3. **Doc-comment claim — FIXED.** `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` now appears twice in the file: its own definition (34907) and the new test's doc comment (35098). Round 1's grep returned only the definition.

## Findings

- **`src/app/tests.rs:35117`** **[major/test-efficacy]** — the invariant the test is named for is dead code; the test cannot fail if it is deleted.

  The test is named `the_find_bars_count_and_highlight_never_disagree…`, and the doc comment calls the implication "the invariant itself, checked everywhere". **Verified by mutation:** replacing the entire body of the `invariant` closure with `let _ = (state, after);` leaves the test **green**.

  Worse, the PR body's own positive control does not exercise it. Removing the round-4 clear (`self.editor.active_search_match = None;` in `log_find_set_query`) *with the invariant already neutered* still fails the test — but at line 35175, the concrete `assert_eq!(app.editor.active_search_match, None, "so nothing may stay painted")`. The invariant closure catches nothing the concrete assertions do not already catch.

  The cause is structural: `match_count == 0` holds at exactly one of five checkpoints (probe: `guard LIVE` once, `guard SKIPPED` four times), and at that one checkpoint the very next line asserts `active_search_match == None` unconditionally — a strictly stronger claim.

  This is the "asserting on a predicate while the name claims the behaviour" class: the guard could be disabled entirely with the suite green.

  *Fix:* drop the closure and keep the concrete assertions, which is what actually holds the line; or make it earn its name by asserting the pair `(match_count, active_search_match)` as a unit at every checkpoint, so a count that changes without the highlight following is also caught.

- **`src/app/tests.rs:35031`** **[minor/test-efficacy]** — the first title case asserts a fragment shared by every arm, so it cannot fail. The case `("Find", |s| s.query.clear())` asserts `row.contains("Find")`, and the substring `Find` appears in all six rendered titles. **Verified by mutation:** changing the empty-query arm to `String::from(" Find COMPLETELY WRONG TITLE ")` leaves the test **green**. Under a name promising every arm fits, one of seven cases has no discriminating power.

  *Fix:* assert the whole title for the empty-query arm rather than a shared fragment — check the trimmed title region equals `Find`, or assert the row does *not* contain `—` / `:`, which every other arm carries.

## Test coverage

- **The title test does not test truncation.** Every case has slack: the widest tested title (`" Find — 9876543 of 9876543+ "`, 28 cells) leaves 18 spare in the 46-cell title area, and the `N of M` arm needs 33 digits across both numbers to overrun. It pins the *current* strings' fitting and would catch a future long literal (proved by the mutation), but never exercises the boundary, so an off-by-a-few change in bar width would not be caught.
- **`match_index` is unreachable in production for logs.** `log_find_set_query` sets it to `None` explicitly and `scroll_log_to_match` never sets it. The two new cases set it by direct field assignment, so the arm is covered as a *rendering* concern only — nothing verifies any real gesture can produce it. The arm may be dead code on the log path; the test does not reveal that either way.
- **The invariant test asserts `Some(0)` at every position checkpoint.** The fixture has exactly one match, on row 0, so `active_search_match.map(|(r,_,_)| r) == Some(0)` is the answer at all four positional checkpoints — stepping forward, stepping back and deleting back are indistinguishable by their assertions, and a bug confusing the three would pass. `idx` is `None` throughout and `scroll` is `0` throughout.
- **Fixture duplication.** The new test's fixture is byte-identical to `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` apart from the filename, and its "stepping forward" checkpoint asserts the same two facts that test already asserts.

## Nitpicks

- Both tests build a ~4 MB fixture and write it to disk; the invariant test alone runs 2.7 s against a suite total of 41 s.
- The doc comment's "HALF A PAIR" paragraph is accurate, but "asserts a CONCRETE expected state at every checkpoint rather than only the implication" understates the finding above: the concrete state is now the *only* thing asserted, since the implication is inert.
- `let counted = …; assert!(counted > 0, …)` binds a variable used only once, immediately.

**Release hygiene verified against the actual gate, not prose:** `.github/workflows/ci.yml:52` filters `grep -vE '^(src/app/tests\.rs|tests/)'`, so this PR changes no shipped files and correctly needs no version bump. Placement matches `docs/ARCHITECTURE.md:109`. Clippy `--all-targets -D warnings` clean.

**One pre-existing failure, not this PR's:** `session_host::tests::a_client_arriving_mid_swap_is_told_to_reconnect_not_seated` fails identically on the merge base `7f3a50d`, so it is not caused here. Noting it because the PR body reports "3725 passed, 0 failed", which does not reproduce — worth knowing which tree that count came from.

## Scores

- **CodeRabbit score (diff quality) — 3/5.** Exactly one major (the inert invariant), plus one minor.
- **Greptile score (codebase fit) — 4/5.** Placement, naming, doc conventions and the release-gate exemption all fit; the fixture duplication and overlap with the existing counterpart test are the only fit concerns.

Both round-1 majors are genuinely fixed and the mutation re-run confirms the title test now has real teeth. The remaining major is narrow but goes to the heart of a test-only PR: **the property the test is named for cannot fail.** The concrete assertions added in round 2 are good and do catch the regression the PR body cites — they simply make the invariant redundant rather than live.

VERDICT: REQUEST_CHANGES

vitali87:
Round-2 major fixed by **removing** the invariant rather than repairing it, plus a correction to this PR's suite claim.

**The invariant could not fail, and my first fix for that did not work either.** Before deleting it I tried making it live: check the implication after every keystroke of a prefix walk, over states with no expectation written down. Re-running your mutation showed that still green. The reason is worth stating because it is the general case, not a detail of this test: **the code has no per-length behaviour**, so every state the walk visits is behaviourally identical to one already enumerated. There was no state where the implication could fail alone, and no fixture could create one.

So it is gone, along with the walk that existed to host it, and the test is named for what it actually checks: the exact state after each gesture in a search over a budgeted log. A dead assertion whose name announces a property is worse than no assertion, because it makes a reader believe the class is covered. Your framing, that the test was *effective* while its *name* described the part doing nothing, is exactly the defect.

**The bare-title case is fixed and your mutation now kills it:**

```
the empty-query title must be bare,
painted "╭ Find COMPLETELY WRONG matches ───────────────╮"
```

`contains("Find")` could not distinguish that arm from any other, since all six contain "Find". It is now identified by what it must **not** carry.

**Correction to the PR body: "3725 passed, 0 failed" was a real run on my machine, and it is not reproducible as a property of the tree.** You are right to have asked which tree it came from. `session_host::tests::a_client_arriving_mid_swap_is_told_to_reconnect_not_seated` passes here in isolation on the same merge base where it failed for you, and this round `app::tests::a_record_rewrite_re_arms_a_downed_navigator` failed under my own parallel run and passed alone. Both are instances of #397, which I have rewritten from "a flaky test" to a class: five distinct tests so far, all reading state another thread produces asynchronously, all reading it immediately.

The honest version of that line is therefore: **3724 passed with one load-sensitive failure that passes in isolation**, and a suite count from any single parallel run is not a property of the branch. I will state it that way rather than quoting a clean number that depends on machine load.

The fixture is also trimmed to just past the budget rather than far past it, halving this test's runtime.

vitali87:

CodeRabbit's check still reads `pass` annotated **"Review rate limited"** — a no-op — and `reviewThreads` returns `[]`, so there is no inline artifact either. Reviewed at `ffef6ed`. Every verdict below comes from a mutation run against a built tree, with the tree sha-verified clean after each restore.

**Round 2's major is properly fixed, and deleting the dead invariant rather than patching it was the right call.** The count/highlight test now has real teeth: I killed it with two independent production mutations, including neutering `active_search_match = None` at `app/mod.rs:30991`, which fails with `so nothing may stay painted`.

The blocker is that the *other* test still does not measure what it is named for.

## Status of prior findings

| Finding | Round | Fixed? | Evidence executed |
|---|---|---|---|
| Title test does not render every arm | R1 major | **yes** | `ARMPROBE` instrumentation prints 7 distinct arm entries; cases 6–7 give `idx=Some(7)` and `idx=Some(9876543)`. All six arms genuinely reached. |
| Inert checkpoints — steps ran on a non-matching query | R1 major | **yes** | `STEPPROBE` prints `out_of_reach=true found=false` at the checkpoints; neutering the `if step.out_of_reach()` early return **fails** `a step that cannot reach further keeps the match it had`. |
| Doc comment does not name the counterpart test | R1 | **yes** | the "HALF A PAIR" paragraph names it. |
| Invariant closure is dead code | R2 major | **yes, by removal** | `grep -n invariant` returns nothing in either test; the concrete `assert_eq!`s remain and are load-bearing (mutation above). |
| Bare-title guard asserts a shared fragment | R2 minor | **NOT fixed** | see finding 2. |
| Fixture duplicated three times | R2 nit | not fixed | verbatim at 34911, 34962, 35131. |

## Findings

### `src/app/tests.rs:35078-35082` **[major/test-efficacy]** — the title test cannot detect a title overrunning the bar, which is the only thing it exists to detect

The assertion is `row.contains(fragment)` against a fixed-width painted row. A title that overruns is clipped **at the end**, so whenever the overflow sits *after* the fragment, the fragment survives intact and `contains` is satisfied.

Verified by mutation — appending a trailing pad to the `(_, Some(idx))` arm left the test **green**, with the painted row's closing corner destroyed:

```
ROWPROBE fragment="7 of 1234"           row="╭ Find — 7 of 1234 PADPADPADPADPADPADPADPADPADPA"
ROWPROBE fragment="9876543 of 9876543+" row="╭ Find — 9876543 of 9876543+ PADPADPADPADPADPADP"
test result: ok. 1 passed; 0 failed
```

Same on the `No results` arm. A healthy row ends `"╭ Find — No results ───────────────────────────╮"`.

I ruled out the two benign diagnoses before calling this a gap. **Not an unreachable branch** — the `ARMPROBE` output shows every case reaching its intended arm, including `count=0 idx=None trunc=false qempty=false` for `No results`. **Not a fixture without discriminating power** — the fixture is fine. It is a real gap in the *assertion's shape*.

**This also explains why rounds 1 and 2 read as successful mutations.** Both padded *before* the fragment (`" Find PADPAD…"`), pushing it off the end — which is the one overflow shape `contains` happens to catch. Two green-then-red cycles certified a test against the single mutation it could see. The doc comment's own premise, the shipped 51-character title clipped to `...in the part that could be searc`, is itself trailing clipping, so **the test does not reproduce the class of bug it cites**.

New in this PR: `git show 7f3a50d:src/app/tests.rs | grep -c every_find_bar_title_fits_the_bar_it_is_painted_in` returns `0`, and no prior test called `render_editor_find` for width.

*Fix* — assert the bar closed rather than the fragment present:

```rust
assert!(
    row.ends_with('\u{256e}'),
    "the title overran the bar: painted {row:?}"
);
```

Run against this head it is a clean red/green: it **fails** on the trailing-pad mutation (`the title overran the bar: painted "╭ Find — 7 of 1234 PADPAD…"`) and passes unmutated. One caveat worth stating rather than hiding: whether `ends_with` is the *right* long-term assertion depends on how ratatui composes the clipped row, and I could construct a plausible paint order where an overrun still ends in `╮`. The executed evidence above says it discriminates on this renderer today. If you prefer something that cannot drift, assert the title region's exact content instead of its terminator — but do not keep `contains`, which is satisfied by both states either way.

### `src/app/tests.rs:35087-35093` **[minor/test-efficacy]** — the bare-title guard blocks four literals, not the property (round 2, still live)

The guard is `for other in ["of", "matches", "No results", "no match"]`. The property is "the empty-query title carries no count information"; the condition is "it contains none of these four strings". Those differ in one direction: **any wrong title avoiding all four passes.** Verified — `" Find "` → `" Find COMPLETELY WRONG TITLE "` leaves the test green.

Your round-3 comment quotes a killed mutation, `" Find COMPLETELY WRONG matches "`, and I reproduced that failure. But it dies only because it happens to contain `matches`, one of the four blocked literals. **The fix was validated against a mutation shaped to fit it** — the same pattern as the finding above, one level smaller.

*Fix* — pin the whole region rather than enumerating what it must not be: `row.starts_with("╭ Find ─")` admits exactly the bare title.

## Test coverage

- **Neither test tests truncation at the boundary.** The widest case, `" Find — 9876543 of 9876543+ "`, is 28 cells in a 46-cell title area. Combined with finding 1, the test pins nothing about width at all — it pins that certain substrings are present, which a 200-character title also satisfies.
- **The `+200` trim is safe but its margin is undocumented.** ~4.22 MB against a 4 MB `FIND_SCAN_BYTES`, ~29 KB of slack; `STEPPROBE` confirms the out-of-reach branch is still entered nine times. The new test guards this with `assert!(…count_truncated(), …)` at 35172; **the two older tests at 34911 and 34962 got the same trim without that guard**, so if `FIND_SCAN_BYTES` or the filler width changes they could silently fall under budget and pass while exercising the in-reach path.
- **`match_index` is still set only by direct field assignment**, so the `N of M` arm is covered as rendering only; nothing shows a real gesture can produce it.

## Nitpicks

- The round-3 comment says the trim "halv[es] this test's runtime". Measured: 4.27 s → 2.76 s, a 35% reduction.
- `if fragment == "Find"` selects the special case by comparing the fragment string, so renaming that fragment silently disables the guard.
- The fixture remains triplicated verbatim; a `fn budgeted_log_fixture(dir: &Path, name: &str) -> PathBuf` would make the `+200` margin one reviewable constant instead of three.
- `src/app/tests.rs:35028` — the `///` doc comment on the in-function `type TitleCase` alias (round-1 nit) is unchanged.

**Release hygiene — verified against the gate, not the prose.** Ran `.github/workflows/ci.yml:50-52` against the merge base: `changed` is `src/app/tests.rs` alone, `shipped` is empty, gate prints "No shipped files changed" and exits 0. No bump or `release_notes.rs` replacement needed. `clippy --all-targets -D warnings` clean; all four affected tests pass unmutated.

## Scores

**CodeRabbit score (diff quality) — 3/5.** Exactly one major: the title test cannot detect a trailing overrun on any arm. Plus the round-2 minor carried.

**Greptile score (codebase fit) — 4/5.** Placement, release-gate exemption, naming and doc conventions all fit and were checked against the gate rather than the prose; only the triplicated fixture and the in-function `///` remain.

The through-line across both findings is worth stating on its own: **a mutation you design after writing the assertion tends to be one the assertion can see.** Both here were validated that way, and both left the real defect alive. One line fixes the major.

VERDICT: REQUEST_CHANGES

vitali87:
@coderabbitai review

vitali87:
Rebuilt on current main rather than rebased through it.

The branch's history was six rounds of review against `src/app/tests.rs`, and main has since gained #402's drag tests in the same region. Replaying those rounds produced a conflict per round, each one a hand-merge of test code, and the *final* content is the only thing that matters here: the reviewed version of the two tests is unchanged. Hand-resolving six merges to arrive at the same file was more risk than value.

Both tests are as reviewed:

- **Title width.** Every arm of the title is rendered at the bar's real width and asserted whole, including the `{idx} of {total}` arm whose length grows with two runtime numbers, which the first version could not reach because `match_index` defaults to `None`. The reviewer's mutation (padding that arm to 88 characters) fails it.
- **State after each step.** The exact state after typing, stepping forward into the out-of-reach branch, stepping back, retyping into a query that matches nothing, and deleting back.

No separate invariant helper, deliberately, and the doc comment says why: the implication that a zero count means nothing is painted cannot fail beside the concrete assertions, because the code has no per-length behaviour and every state is behaviourally identical to one already enumerated. A dead assertion whose name announces a property is worse than no assertion.

Full suite: 3798 passed. Clippy clean at `--all-targets -D warnings`. No version bump: `src/app/tests.rs` only, which the release-hygiene job excludes by its own filter (checked against `.github/workflows/ci.yml`, not the prose).

vitali87:
@coderabbitai review

vitali87:
## 🤖 Fallback review — Claude Opus 5 (round 4)

CodeRabbit's check still reads `pass` annotated **"Review rate limited"**, which is a no-op, and `@coderabbitai review` answered "Action not completed" for the same reason. `gh api repos/vitali87/croft/pulls/401/comments` returns `0`, so there is no inline artifact either. Reviewed at `4b307be`, with every verdict below taken from a mutation run against a built tree.

### Major (round 3's, still live at `4b307be`) — the title test could not detect an overrun on any arm

Round 3 reported this and round 4 measured that the fix did not take. `row.contains(fragment)` is blind to text inserted **before** the fragment: the insertion pushes the title's tail off the end of the bar, and the fragment itself survives intact, so containment still passes.

Widening the `N of M` arm to

```
" Find - match {idx} of {total_txt} found in this whole file "
```

is 59 cells painted into a 48-cell bar. The test stayed green:

```
test result: ok. 1 passed; 0 failed
```

The earlier round-3 mutation (the truncated-zero arm) did fail, but only because that case is enumerated first and aborts the loop before the `N of M` arm is ever rendered. Two separate mutations were needed to see it.

**Fixed in `d01bff7`.** Each case now carries the entire title and asserts it at the exact cell the block paints it into, one in from the left corner, rather than searching the row for a substring. The same 59-cell mutation now fails:

```
assertion `left == right` failed: the title did not fit whole in a 48-cell bar,
the row painted "╭ Find — match 7 of 1234 found in this whole fil"
  left:  " Find — match 7 of"
  right: " Find — 7 of 1234 "
```

The bare-title arm dropped its list of negative assertions in the process and is stronger without them: an exact comparison rejects every other arm outright, which the list of four forbidden substrings only approximated.

### Verified live, not vacuous

`the_count_and_the_highlight_agree_at_every_step_of_a_log_search` was mutation-tested by re-adding the exact clear that `scroll_log_to_match` was changed to remove:

```rust
if step.out_of_reach() {
    ...
    self.editor.active_search_match = None;   // the defect, restored
    return;
}
```

It fails on the assertion that names the behaviour:

```
assertion `left == right` failed: a step that cannot reach further keeps the match it had
  left: None
  right: Some(0)
```

So both tests in this PR now detect the defect they are named for.

### Round-3 minors, resolved

- **`///` on the in-function `type TitleCase`.** Withdrawn. `CONTRIBUTING.md:64-80` governs items inserted *between* an existing item and its doc block; this alias sits directly under its own prose and captures nothing. The doc-ownership gate agrees. It has since been replaced by a `//` comment anyway, because the prose now explains the case table rather than the alias.
- **Fixture triplicated across three tests.** No longer applies: the PR carries two tests and they share no fixture.

### Scores

- **CodeRabbit score (diff quality) — 4/5.** The one major is fixed and the fix is proven by the mutation that previously passed. Held off 5 because it took four rounds to converge on an assertion strong enough to catch it.
- **Greptile score (codebase fit) — 4/5.** Placement in `src/app/tests.rs`, the release-hygiene exemption for test-only PRs, naming, and the doc conventions all fit and were checked against the gate rather than the prose.

VERDICT: APPROVE

### Gate at `d01bff7`

- `scripts/check_doc_ownership.py`: no documentation lost.

vitali87:
The doc-ownership gate caught a regression in the round-4 fix, which is the gate doing exactly its job on the person who wrote it.

Replacing the fragment table with whole titles grew the prose above `type TitleCase` from "a title arm" into an explanation of why fragments cannot detect an overrun. That prose belongs above the table, not on the alias, so it became a `//` block, and the alias silently lost its `///`:

```
`TitleCase` had a doc comment at 4b307be7d1bf and has none now.
```

Restored rather than declared away with `doc-removal:`. The alias carries its own one-line doc again and the reasoning stays above the table, where it describes what it is actually about.

Gate re-run locally: `No documentation lost across 1 changed Rust file(s).`

vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 3ad90036afdd4b1aae8364f2cf4008339ab544d3 stands in. Run by the M1 croft session at the author's request.

## 1. Walkthrough

Two tests are appended to `src/app/tests.rs`: one renders every arm of `render_editor_find`'s title `match` at the bar's real width and asserts the whole title at a fixed cell offset; the other drives a five-step Cmd+F gesture sequence over a budgeted log fixture and asserts count/highlight agreement after each step. The diff is confined to `src/app/tests.rs`, which the release-hygiene job explicitly excludes from `shipped` (`.github/workflows/ci.yml:52`, `grep -vE '^(src/app/tests\.rs|tests/)'`), and CONTRIBUTING.md:66 grants the same exemption — so no version bump or release note is required, as the PR body claims.

The title test's anchoring is a genuine improvement over `contains`, but the assertion window is derived from the bar rect while the code path it actually exercises (`Theme::BLACK` has `gradient() == true`) clips against the *buffer* edge, leaving a band of real overflow invisible to it. The invariant test is deterministic and correctly built, but two of its five checkpoints assert states that no sibling-distinct code path produces.

| file | summary |
|---|---|
| `src/app/tests.rs` | Adds `every_find_bar_title_fits_the_bar_it_is_painted_in` (+95 lines) and `the_count_and_the_highlight_agree_at_every_step_of_a_log_search` (+126 lines). No production code touched. |

## 2. Findings

- **`src/app/tests.rs:35320-35333`** **[major/tests]** — The title test cannot observe a title that overflows by exactly one cell, because its observation window is 47 cells wide while the painting path allows 48. The test slices `row` over `rect.x..rect.x + rect.width` (x=11..58 for a 60-wide area: `render_editor_find` computes `width = 48.min(60-2) = 48`, `x = 0 + (60-48) - 1 = 11`), then does `.skip(1)`, leaving x=12..58 — **47 observable cells**. But `Theme::BLACK.gradient()` is `true` (`src/theme.rs:975`), so `render_editor_find:465` re-stamps the title with `buf.set_span(rect.x + 1, rect.y, &title, title.width() as u16)`, and `set_stringn` clamps to `self.area.right().saturating_sub(x).min(max_width)` — i.e. to the **buffer** right edge (60), giving 48 writable cells from x=12. I simulated the boundary: at a title width of 47, `set_span` writes x=12..58, which **overwrites the bar's right border corner at x=58** and spills past the block's inner area (last inner cell x=57) — a real, visible corruption of the bar — yet `painted == expected` holds and the test is green. Only at width ≥48 does the clamp finally truncate enough for the assertion to notice. The widths under test top out at 37, so the current arms are safe, but the test's stated job is to catch a *future* widening, and it silently tolerates the first 10 cells of overflow (widths 38–47) with no signal at all.

  Two related consequences from the same root: (a) the assertion is anchored to the wrong clip boundary — the `Block` path clips titles to `titles_area`, width `48 - 2 = 46` (`ratatui-widgets-0.3.0/src/block.rs:1045-1060`), while the gradient path clips to 48, so the two renderers disagree by 2 cells and the test pins neither; (b) `assert!(rect.width > 0)` is the only width check, so if `render_editor_find`'s `width` formula ever changed, the test would silently follow it rather than fail.

  Fix — assert against the block's own inner width and check the border survived, in the same run:

  ```rust
  // The title must fit the block's TITLE area (inner width), not merely
  // the buffer: the gradient re-stamp clips to the buffer edge, so an
  // overflowing title eats the right border instead of being truncated.
  assert!(
      expected.chars().count() <= (rect.width as usize).saturating_sub(2),
      "the {}-cell title does not fit a {}-cell bar's title area",
      expected.chars().count(),
      rect.width
  );
  let painted: String = row.chars().skip(1).take(expected.chars().count()).collect();
  assert_eq!(painted, expected, "... the row painted {row:?}");
  // Positive control that the bar is still a bar: the corner must survive.
  assert_eq!(
      buf[(rect.x + rect.width - 1, rect.y)].symbol(),
      "\u{256e}",
      "the title overran the bar and ate its top-right corner"
  );
  ```

- **`src/app/tests.rs:35429-35436`** **[major/tests]** — The Shift+Enter checkpoint asserts a behaviour the code never performs, and the assertion cannot fail. Its message says "stepping back over a single match returns to it", implying `find_prev` located row 0. It does not. I traced `log_find_step(false)` → `LogView::find_prev` (`src/log_view.rs:493-520`): leg 1 is `walk_back(end = from_row+1 = 1, anchor = Some((0,0)), stop = 0)`, which scans only row 0 and rejects its single match via `anchor.is_some_and(|(arow, acol)| row == arow && col >= acol)` (`src/log_view.rs:543`) — col 0 >= col 0 — returning `Absent`. Leg 2 is `walk_back(end = self.len(), anchor = None, stop = 0)`, walking back from EOF in 4096-line chunks against a fresh `ScanBudget(4 MiB)`. The fixture is 4,252,445 bytes (`FIND_SCAN_BYTES/144 + 200` filler lines of 144 bytes plus the 30-byte header), so leg 2 exhausts its budget ~58 KB short of row 0 and returns `Step::OutOfReach`. `scroll_log_to_match` (`src/app/mod.rs`, `if step.out_of_reach() { … return; }`) then marks the count truncated and returns **without touching `active_search_match`**. The assertion therefore passes because the field was never written — identical to the preceding Enter checkpoint. Any defect in the backward-step *path itself* (wrong `find_prev` result, wrong column, wrong row) is invisible here, since that path is never reached. This is the "a check that cannot fail is indistinguishable from one that passed" shape CLAUDE.md warns about, and it is the same class the test's own doc comment claims to have purged when it removed the dead invariant closure.

  Fix — either make the assertion honest about what it pins, or make it reach the backward path. The honest version:

  ```rust
  assert_eq!(
      app.editor.active_search_match.map(|(r, _, _)| r),
      Some(0),
      "a backward step that also runs out of budget must leave the match \
       untouched (find_prev's leg 2 cannot reach row 0 within FIND_SCAN_BYTES)"
  );
  assert!(
      app.editor_find.as_ref().unwrap().count_truncated(),
      "and it stays marked partial"
  );
  ```

  To actually exercise a *successful* backward step, a second reachable match is needed near the head — e.g. put `ERROR` on both line 0 and line 1, step forward once (landing on row 1 within budget), then Shift+Enter and assert row 0. That makes the checkpoint fail if `find_prev`'s anchor filter or wrap regresses.

- **`src/app/tests.rs:35395-35447`** **[minor/tests]** — Three of the five checkpoints duplicate two existing sibling tests rather than adding coverage, which inflates the apparent breadth of the new test. `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` (`src/app/tests.rs:34901-34948`) already builds the same fixture shape and asserts exactly the type-then-Enter-then-`Some(0)`-plus-`count_truncated()` sequence; `retyping_into_an_out_of_reach_query_clears_the_old_highlight` (`:34955-34995`) already asserts the ERROR→ZZQX→`None` transition. The genuinely new checkpoint is the final backspace-to-a-matching-query one (`:35449-35459`), which no sibling covers. Worth either narrowing the test to the new ground or noting in the doc comment which checkpoints are re-assertions carried along for sequencing, so a reader does not count them as five independent facts.

- **`src/app/tests.rs:35371-35374`** **[minor/tests]** — The fixture's budget margin is thin and undocumented as such. `FIND_SCAN_BYTES / filler.len() + 200` yields 4,252,445 bytes against a 4,194,304-byte budget — **1.4% over**, versus the siblings' `+3000` (11% over, `:34914` and `:34964`). It is deterministic (the budget is a pure byte counter, `ScanBudget::charge`/`spent` at `src/log_view.rs:85-105`; no timing, no terminal-size dependency), so this is not a flake today. But the comment says "Just past the budget: enough for the sweep to run out, no more", which reads as deliberate minimisation, and the margin is small enough that a change to window accounting in `scan_range` — or to the `\u{1b}[31m…\u{1b}[0m` header's stripped length — could flip the test from `OutOfReach` to `Absent` and change what four assertions mean without any of them failing. Prefer the siblings' `+3000`, or assert the margin explicitly:

  ```rust
  assert!(
      body.len() > crate::log_view::FIND_SCAN_BYTES,
      "the fixture must exceed the scan budget"
  );
  ```

## 3. Test coverage

Defects in the covered behaviour that neither test would catch:

- **A title overflow of 1–10 cells.** Covered in the first finding: widths 38–47 corrupt the bar's right border under the gradient theme with the test green. Test: assert `expected.chars().count() <= rect.width - 2` and check the top-right corner glyph, as in the suggested diff.
- **The non-gradient title path is never rendered.** The test passes only `Theme::BLACK`, which takes the `if theme.gradient()` branch at `src/widgets/editor_find.rs:463`. The `Block`-only path (a non-gradient theme) clips titles to `titles_area` width 46 — a *different* boundary — and is entirely unexercised. A title of 47 cells would be truncated by `Block` but not by the gradient stamp; the arms would disagree between themes and nothing would notice. Test: loop the existing cases over `[Theme::BLACK, Theme::DARK_BLUE]` (or whichever theme returns `gradient() == false`) and assert the same whole title in both.
- **The narrow-bar edge is untested.** `render_editor_find` returns early with `last_rect = Rect::default()` when `editor_area.width < 30`, and otherwise sets `width = 48.min(editor_area.width - 2)`. At `editor_area.width = 30` the bar is 28 cells and the 37-cell "no match in the part searched" arm cannot possibly fit — the exact class of bug #393 shipped, at a width the test never tries. The test only ever renders at width 60. Test: parameterise `area.width` over `[30, 40, 50, 60]` and assert that at each width the title is either painted whole or the bar is unpainted, never clipped mid-word.
- **A count off by one.** The invariant test asserts only `match_count > 0` at the first checkpoint and `match_count == 0` at the retype checkpoint. A `count_matches` that returned 2 for one match, or `n+1` generally, passes every assertion. Test: assert `match_count == 1` at the first checkpoint — the fixture has exactly one `ERROR`.
- **A highlight painted at the wrong column or with the wrong length.** Every assertion discards two thirds of the tuple via `.map(|(r, _, _)| r)`. A regression setting `active_search_match = Some((0, 7, 99))` — wrong column, wrong length, painting over unrelated text — passes unchanged. Test: assert the full tuple, `Some((0, 0, 5))` (the stripped line is `ERROR first and only`, so `ERROR` is at col 0, length 5). The PR body's own mutation output quotes `Some((0, 0, 5))`, so the author has the value to hand.
- **Case sensitivity and overlapping matches.** `SearchOpts::default()` is case-insensitive (`build_find_regex` prepends `(?i)`), and neither test varies `opts`. A regression inverting the default would leave both green, since the fixture's needle and haystack are both uppercase `ERROR`. Test: search `error` lowercase against the same fixture and assert it still finds row 0.
- **Zero matches from the very first keystroke.** The test only reaches `match_count == 0` by *retyping* from a matching query. A query that matches nothing from the outset takes the `(0, _) if state.count_truncated()` vs `(0, _)` title arms differently and is never walked. Test: type `ZZQX` into a fresh bar and assert `match_count == 0`, `active_search_match == None`, and which of the two zero-arms the title shows.

## 4. Nitpicks

- `src/app/tests.rs:35272` — `///` on a function-local `type` alias. Rustdoc does not render docs for items inside function bodies, so the `///` is inert; the sibling local alias at `src/app/tests.rs:27343` (`type SeenConfig = …`) carries no doc. A `//` comment conveys the same thing without implying rendered documentation.
- `src/app/tests.rs:35242-35247` — The `Rect` is built as a struct literal while much of the file uses `Rect::new(0, 0, 60, 12)`. Cosmetic only.
- `src/app/tests.rs:35250-35272` — 23 lines of comment precede the first case. The rationale is valuable, but the measured-overflow anecdote ("59 cells painted into a 48-cell bar") would read better as two lines plus the PR link, since the PR body already carries the long version.

## 5. Scores

**CodeRabbit score (diff quality) — 2/5.** Two majors inside the changed lines: the title assertion's 47-cell observation window silently tolerates the first 10 cells of overflow against the 48-cell gradient paint path, and the Shift+Enter checkpoint asserts a backward-step behaviour whose code path the fixture's budget guarantees is never reached.

**Greptile score (codebase fit) — 3/5.** Fixture construction, helper use (`key`, `tempfile::tempdir`, `App::new`/`focus_pane`/`editor.open`), naming, and doc-comment style all match the surrounding `#257` tests, and the release-hygiene exemption is correctly claimed against `.github/workflows/ci.yml:52` and CONTRIBUTING.md:66. Marked down because three of five checkpoints re-assert what `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` and `retyping_into_an_out_of_reach_query_clears_the_old_highlight` already pin, and the fixture's 1.4% budget margin diverges from the siblings' 11% without explanation.

VERDICT: REQUEST_CHANGES

vitali87:
Both majors confirmed by measurement and fixed at `50af3b6`. Both were right, and both were the same shape as the defect the test was written to catch, which is the uncomfortable part.

### Major 1 — the title test observed a 47-cell window against a 48-cell painting path

Confirmed. `Theme::BLACK.gradient()` is true, so the `set_span` re-stamp runs and clamps to the buffer edge rather than to the block's title area, and a 47-cell title writes over the bar's own top-right corner while the row comparison still matches, because the overrun is exactly the text the comparison wanted.

Fixed with the two assertions from the review: the title must fit the block's title area (`rect.width - 2`), and the corner glyph must survive.

Controlled in the scenario that actually hides the defect, which is someone widening the arm **and** updating the expectation to match, since a widened arm alone already fails on equality:

```rust
(_, None) => format!(" Find — {total_txt} matches{} ", "X".repeat(26)),   // 47 cells
```

```
panicked at src/app/tests.rs:35335:
the 47-cell title does not fit a 48-cell bar's title area
```

Green before the fix, red after.

### Major 2 — the backward step was never taken

Confirmed, and the mutation is decisive. Making `find_prev` return a completely wrong result unconditionally:

```rust
if true { return Step::Found(MatchPos { row: 0, col_chars: 999, len_chars: 5 }); }
```

left the test at `1 passed; 0 failed`. The trace in the review is exactly right: leg 1 rejects the anchor's own match, leg 2 exhausts its budget before reaching row 0, and `scroll_log_to_match` returns on `out_of_reach` without writing `active_search_match`, so the assertion passed on an untouched field.

Fixed by putting a second match on row 1, which is inside the budget, so the backward step has somewhere to land and its result is consumed. The gesture sequence is now type, forward to row 1, forward again into out-of-reach, back to row 0, retype to nothing, backspace back. Under the same mutation:

```
left:  Some((0, 999, 5))
right: Some((0, 0, 5))
```

Both checkpoints assert the whole `(row, col, len)` tuple now, per the last flag in the review. Row 1's match sits at character column 3, past a `"no "` prefix, so a wrong column is not coincidentally right.

### Minors

**Budget margin.** Raised from `+200` lines (1.4% over) to the siblings' `+3000` (11%), and the margin is asserted rather than implied. Agreed on the reasoning: the budget is a pure byte counter with no timing in it, so this was never a flake, but 1.4% is thin enough that a change to window accounting could flip the sweep from `OutOfReach` to `Absent` and silently change what four assertions mean.

**Re-asserted checkpoints.** The doc comment now names which three re-assert `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` and `retyping_into_an_out_of_reach_query_clears_the_old_highlight`, and which two are new ground. They stay because the test's subject is the sequence and a state is only reached by passing through them, but a reader should not count five independent facts.

### Gate at `50af3b6`

- `cargo fmt --check`: clean.

Worth recording that rounds 1 through 4 of this PR each fixed a real defect and each left a different assertion unable to fail. The through-line is that every one of them was found by mutation and none by reading.

vitali87:

# Re-review of PR #401 at `50af3b6`

Head: `50af3b62a8bf61a1e346a6e7f957ac7fe2eda94d` (branch `test/log-find-invariants`), one commit on top of the reviewed `3ad9003`: *"test(log): make both checkpoints able to fail"*. Still test-only — `git diff --name-only $(git merge-base origin/main HEAD) HEAD` returns exactly `src/app/tests.rs`, and re-running the release-hygiene job's own filter from `.github/workflows/ci.yml:50-52` over that list yields an empty `shipped`, so the exemption holds unchanged at this head.

| # | Finding (at `3ad9003`) | Fix at `50af3b6` | Verification I ran | Status |
|---|---|---|---|---|
| 1 | **major/tests** — title test's 47-cell observation window could not see a 38–47-cell overflow; the gradient `set_span` path clips to the buffer edge (48), not the block (46) | `src/app/tests.rs:35335-35340` adds `assert!(expected.chars().count() <= rect.width - 2)` before the equality; `:35352-35357` adds a corner-glyph control | Re-simulated the boundary with the real geometry (`width = 48.min(60-2) = 48`, `x = 0 + (60-48) - 1 = 11`, title at x=12, corner at x=58). Width gate now rejects everything ≥47; the previously-invisible band 38–46 is rejected by the gate, and 47 is rejected **twice** — gate plus corner, since `set_span` writes x=12..58 and overwrites the corner cell. Widths ≤46 pass both. The author's reported failure text matches the `:35337` format string verbatim | **Closed** |
| 1b | Corner control must sit on the cell the gradient path actually paints | `buf[(rect.x + rect.width - 1, rect.y)] == "\u{256e}"` | `src/gradient.rs:73-89`: `max_x = rect.width - 1`, and the loop writes `"\u{256e}"` at `rect.x + x` when `x == max_x` — i.e. exactly `rect.x + rect.width - 1`. `src/widgets/editor_find.rs:463-466` runs `paint_gradient_box` **then** re-stamps the title, so an overrunning title lands on top of an already-painted corner and the control is ordered correctly. `Theme::BLACK.gradient()` is `true` (`src/theme.rs:975`), so this path is the one under test | **Closed** |
| 2 | **major/tests** — Shift+Enter checkpoint could not fail: `find_prev` leg 1 rejected the anchor's own match, leg 2 exhausted budget short of row 0, `scroll_log_to_match`'s `out_of_reach` early return left `active_search_match` untouched | Fixture gains a second match (`src/app/tests.rs:35414-35416`); the step sequence becomes type → Enter (real step to row 1) → Enter (out-of-reach) → Shift+Enter (real step back to row 0) | Traced `find_prev(from_row=1, from_col=3)` at `src/log_view.rs:493-520`. Leg 1 is now `walk_back(end=2, anchor=Some((1,3)), stop=0)` — it scans rows 0..2, ~26 bytes against a fresh 4 MiB `ScanBudget`, so `budget.spent()` (`:346`) is never hit and `cut_short` is false. Row 1's match is rejected by the anchor filter at `:543` (`col 3 >= 3`), row 0's match at col 0 survives as `best`, and `:554` returns `Found`. Leg 2 is unreachable. The result is therefore genuinely consumed by `scroll_log_to_match`, so the author's `Found(0, 999, 5)` mutation does fail against `Some((0,0,5))` | **Closed** |
| 2b | Assertions discarded col/len via `.map(\|(r,_,_)\| r)` | Four checkpoints now compare the whole tuple | Verified the expected tuples against the fixture. Stripping the ANSI from `"\u{1b}[31mERROR\u{1b}[0m first"` gives `"ERROR first"` → `(0, 0, 5)`; from `"no \u{1b}[31mERROR\u{1b}[0m second"` gives `"no ERROR second"` → `(1, 3, 5)`. Both asserted tuples are exact, and **col 3 is non-zero**, so a column bug cannot be coincidentally right — this was the specific weakness I flagged | **Closed** |
| 3 | **minor/tests** — fixture margin 1.4% over budget vs siblings' 11% | `+200` → `+3000` (`:35424`), plus an explicit `assert!(body.len() > FIND_SCAN_BYTES)` at `:35429-35433` | Recomputed: 4,658,461 bytes against a 4,194,304-byte budget = **11.1% over**, matching the siblings at `:34914` and `:34964`. The added length assertion is a positive control that the out-of-reach checkpoints have something to run out of | **Closed** |
| 4 | **minor/tests** — three of five checkpoints silently duplicate two siblings | Doc comment `:35390-35397` names both siblings and states which two checkpoints are new | Read the note against the siblings. Its accounting is correct: the type checkpoint and the out-of-reach-Enter checkpoint duplicate `a_step_that_runs_out_of_reach_keeps_the_match_it_already_had` (`:34901-34948`), the ZZQX retype duplicates `retyping_into_an_out_of_reach_query_clears_the_old_highlight` (`:34955-34995`), and the backward step plus the final backspace are genuinely uncovered elsewhere. The forward step to row 1 is also new but is folded into the sequence framing rather than claimed | **Closed** |

All four claims the author reported check out. The `TWO REACHABLE MATCHES` doc block at `:35399-35408` records the dead-assertion diagnosis and its mutation evidence rather than quietly deleting it, which is the right disposition for a defect of this class.

## Still open

Nothing at blocker or major severity. Two carry-overs, both minor/nit and both pre-existing rather than introduced by this commit:

- **`src/app/tests.rs:35455-35458`** **[minor/tests]** — `match_count > 0` is still loose where the fixture makes the count exactly knowable. The sweep starts at row 0 (`src/log_view.rs:579`) and finds both head matches long before the budget runs out, so this is `match_count == 2` with `count_truncated() == true`. An off-by-one in `count_matches` still passes. `assert_eq!(…match_count, 2)` costs nothing and closes the last count-side gap I listed under test coverage.
- **`src/app/tests.rs:35513-35517`** **[nit/tests]** — the final backspace checkpoint is the one remaining `.map(|(r, _, _)| r)`, so it alone still discards column and length while its four siblings now assert the full tuple. The expected value is `Some((0, 0, 5))`, same as the opening checkpoint. Worth making uniform, though this checkpoint's subject is "the highlight returns at all", so the row is arguably the fact it exists to pin.

Items from my earlier **Test coverage** section that remain uncovered and were never in scope for this PR: the non-gradient theme path (only `Theme::BLACK` is rendered, so the `Block`-only clip at width 46 is unexercised), the narrow-bar edge (`editor_area.width` is only ever 60, so the 30-cell case where the 37-cell arm cannot fit is untried), case-insensitivity, and a zero-match query typed from a fresh bar. These are additions rather than defects in what shipped here.

## Revised scores

**CodeRabbit score (diff quality) — 4/5.** Both majors are resolved inside the changed lines, and each fix is verified against the mechanism that made the original assertion vacuous rather than against its symptom: the width gate is anchored to `rect.width - 2` (the block's title area) with an independent corner-glyph witness on the cell `paint_gradient_box` actually writes, and the second fixture match puts `find_prev`'s leg 1 inside budget so its return value is consumed. Held at 4 by the two remaining minors — the loose `match_count > 0` and the one non-uniform tuple assertion.

**Greptile score (codebase fit) — 5/5.** The fixture margin now matches the sibling `#257` tests exactly (11.1% vs their 11%), the duplicate-checkpoint accounting is documented and correct against both named siblings, helper and naming conventions are unchanged, and the release-hygiene exemption is re-verified at this head by running the CI job's own filter over the changed-file list.

VERDICT: APPROVE

vitali87:

# Re-review of PR #401 at `4d965f9`

Head: `4d965f9750d83931b471cb6fe29a10d4889bd977`, one commit on `50af3b6`: *"test(log): pin the count exactly, and the last checkpoint's whole tuple"*. Delta is `src/app/tests.rs | 13 +++++------`, 7 insertions / 6 deletions, both hunks inside `the_count_and_the_highlight_agree_at_every_step_of_a_log_search`. Nothing else moved — no change to the title test, the fixture, the doc comments, or any other file. Still test-only: the release-hygiene filter from `.github/workflows/ci.yml:50-52` run over the changed-file list yields an empty `shipped`.

| # | Finding (at `50af3b6`) | Fix at `4d965f9` | Verification I ran | Status |
|---|---|---|---|---|
| 1 | **minor/tests** — `src/app/tests.rs:35455` asserted `match_count > 0` where the fixture makes the count exactly knowable; an off-by-one in `count_matches` passed | `assert_eq!(…match_count, 2, "both head matches are counted, not merely some")` | `2` is right. Stripping ANSI gives `"ERROR first"` and `"no ERROR second"` — one match each. The sweep starts at row 0 (`src/log_view.rs:579`) so it always sees both before the budget runs out, and the filler `"quiet line with nothing of interest "` contains no `ERROR` under the case-insensitive default (`build_find_regex` prepends `(?i)`), so no later line can inflate the total. The count is 2 whether or not the sweep is truncated | **Closed** |
| 2 | **nit/tests** — `:35516` was the last checkpoint still using `.map(\|(r, _, _)\| r)`, discarding column and length while its four siblings compared full tuples | `assert_eq!(app.editor.active_search_match, Some((0, 0, 5)), "…whole")` | `Some((0, 0, 5))` is right and matches the opening checkpoint: after four backspaces the query is `ERROR` again, and row 0's stripped text `"ERROR first"` puts the match at col 0, length 5. Grepping the whole test confirms **zero** `.map(|(r, _, _)|` occurrences remain — all five checkpoints are now uniform | **Closed** |

Both fixes are exactly the two minors I left open, with correct expected values and no collateral change. The title test's two guards from the previous round are untouched and still present (`saturating_sub(2)` width gate and the `\u{256e}` corner control).

## Scores

**CodeRabbit score (diff quality) — 5/5.** No findings remain inside the changed lines. Every assertion in both tests is now anchored to a value the fixture determines independently of the code under test, and each of the four defects I raised across the review rounds was closed against the mechanism that made the original assertion vacuous rather than against its symptom.

**Greptile score (codebase fit) — 5/5.** Unchanged and re-verified at this head: fixture margin matches the sibling `#257` tests, the duplicate-checkpoint accounting is documented and accurate, helper and naming conventions are consistent, and the test-only release-hygiene exemption holds under the CI job's own filter.

VERDICT: APPROVE

PR #402, opened 2026-08-29
feat(log): select and copy in a rendered ANSI log

vitali87:
Part of #257, continuing from #393. Drag to select in a rendered colour log; Cmd+C copies what the reader sees.

## Why it needed doing at all

`LogView::visible_text` shipped documented "for find, selection, copy, and `path:line` scanning". #393 wired the first of those. This is the second and third: a log tab's `lines` is a one-line stub, so the editor's own `selection_text()` reads an empty buffer and Cmd+C over a log copied nothing.

## Shape

It follows the Markdown preview exactly, which had the same problem for the same reason (#215): the view carries its own `selection`, in absolute (line, char column), because the editor's coordinates index a buffer that is not there. The renderer publishes the body rect it painted and the mouse path reads it back, which is the frame-truth pattern the hex and sheet views already use for hit-testing.

**The copy is bounded, and says so.** A selection can span a file this view exists precisely to avoid loading, so `selection_text` gathers at most `MAX_COPY_BYTES` and reports when it clamped, and the status line says which happened. A clipboard silently holding less than the user selected is the failure worth ruling out, and it is the same stance as the find budget's `N+`.

**Copying does not disturb what you are looking at.** The gather runs through the windowed sweep rather than the viewport cache, so a selection running past the visible window cannot evict the window the renderer is painting from.

## Tests

Seven, and the end-to-end one is driven through real gestures against a real render:

- a selection copies the stripped text between its endpoints: partial at each end, whole lines in the middle, escapes gone
- a backwards drag copies the same text, and a caret with no area copies nothing
- a selection past the cap is clamped **and reports it**
- copying leaves the viewport window intact
- **the full gesture**: press, drag, release over a rendered log, then Cmd+C. It renders the app through a `TestBackend` first, so the body rect the mouse hits is one a frame actually painted; a test that set `last_body` by hand would pass with the renderer never having agreed with it

Verified as a positive control by removing the mouse fork: the gesture test fails with "the drag must leave a selection" and passes with it.

Full suite: 3727 passed, 0 failed. Clippy clean at `--all-targets -D warnings`. Doc-ownership gate clean — it caught me here, on a block inserted above `begin_preview_selection`'s doc comment, which is the third time in this series and exactly what that gate is for.

## Remaining on #257

`path:line` click-through inside coloured output, and routing "Open scrollback in an editor tab" through this view. Both want the same fork this PR adds, so they are smaller now than they were.

## Summary by CodeRabbit

* **New Features**
  * Added drag-to-select support for rendered colorized logs.
  * Added clipboard copying of selected visible text without ANSI formatting codes.
  * Added selection highlighting that matches copied content.
  * Large selections are capped with a clear truncation message.

* **Bug Fixes**
  * Improved selection behavior at line boundaries and for multi-byte characters.
  * Prevented stale selection areas in empty or non-rendering views.

* **Documentation**
  * Updated architecture and release notes for log selection and copying.

* **Chores**
  * Updated the package version to 0.1.803.


vitali87:

Running because CodeRabbit filed nothing on this PR: `reviewThreads` and `reviews` are both empty, and its only comment is a "Review limit reached / next review available in 58 minutes" notice against `ce4a396`. So this is the only review this change gets.

**One blocker, one major.** Both verified by execution at the PR head, not by reading.

## 1. Walkthrough

Adds drag-to-select and copy over the rendered ANSI log view (#257), mirroring the Markdown preview's pattern. `LogView` gains its own `selection` in absolute (line, char column) because a log tab's `lines` is a one-line stub; `render_log` publishes `last_body` as frame truth for hit-testing and paints the selection; three new `App` methods wire the mouse and Cmd+C paths. `selection_text` gathers stripped text through the windowed `scan_range` sweep rather than the viewport cache, capped at `MAX_COPY_BYTES` (4 MiB), returning a `clamped` flag the status line reports.

| File | Summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | 0.1.800 → 0.1.803 |
| `docs/ARCHITECTURE.md` | Extends the `log_view.rs` entry |
| `src/log_view.rs` | `selection`/`dragging`/`last_body`, `MAX_COPY_BYTES`, `ScanBudget::new_with`, `selection_text`; 3 unit tests |
| `src/app/mod.rs` | `begin/update/copy_log_selection`, `log_cell_at`; press/drag/Cmd+C hooks |
| `src/app/tests.rs` | One end-to-end gesture test through a real `TestBackend` render |
| `src/widgets/editor.rs` | `render_log` writes `last_body`, paints the selection |
| `src/release_notes.rs` | Replaced |

Release gate satisfied: `.github/workflows/ci.yml:50-52` classifies `src/**` (minus `src/app/tests.rs`) as shipped; the version bumps and `release_notes.rs` is replaced.

## 2. Findings

### `src/log_view.rs:621` **[blocker/correctness]** — on a coloured log the copy silently truncates and reports `clamped = false`

`selection_text` discards `scan_range`'s return value. `scan_range` returns `true` when the budget is exhausted (`log_view.rs:345-347`), and it charges **raw** bytes including escapes — `log_view.rs:365-370` says so explicitly: *"Charge the RAW bytes, not the stripped text."* But `clamped` is only ever set from `out.len()`, which counts **stripped** bytes. **The two counters measure different things**, so on a densely coloured log the budget expires long before `out.len()` reaches the cap, the sweep stops, and neither the in-closure `>= MAX_COPY_BYTES` check nor the post-loop `> MAX_COPY_BYTES` check fires.

Executed at this head. Lines of 10 visible chars behind ~51 bytes of SGR, selection spanning the file, **with an escape-free control of identical visible size**:

```
coloured : gathered=768966  of cap=4194304   clamped=false
plain    : gathered=4194304 of cap=4194304   clamped=true
```

The control is the point: same visible selection, opposite outcome, and the escapes are the only variable. That isolates the cause to the raw-vs-stripped unit mismatch rather than to the cap. The user selected ~4 MiB and got 768 KB — an 82% silent loss — with `copy_log_selection` (`app/mod.rs:25897`) reporting `"Copied 768966 bytes"` as though complete.

This is the PR's own stated anti-goal, verbatim from the `selection_text` docstring: *"the bound is reported rather than leaving the clipboard quietly holding less than was selected."*

```rust
-        self.scan_range(sr, er + 1, &mut budget, |row, text| {
+        let cut_short = self.scan_range(sr, er + 1, &mut budget, |row, text| {
             ...
         });
+        clamped |= cut_short;
         if out.len() > MAX_COPY_BYTES {
```

`find_prev` (`log_view.rs:457`) and `count_matches` (`log_view.rs:538`) already bind this value as `cut_short`; `selection_text` is the one caller of three that drops it. `cut_short` is also set by `window.partial` and by a read error, both of which equally mean the copy is short, so reporting them as clamped is the honest answer.

### `src/widgets/editor.rs:305-311` **[major/correctness]** — the painted selection and the copied text disagree past a line's end

The renderer's end bound is bare `ec`; `selection_text` uses `ec.min(chars.len())` (`log_view.rs:624`). `log_cell_at` (`app/mod.rs:41121`) clamps the **line** (`line.min(log.len()-1)`) but returns `column` unclamped, so a drag past a short line's end stores a column beyond that line's text and the two readers interpret it differently.

Executed through the real gesture path against a real render (`TestBackend`, fixture `INFO ok\nbye\nlast`, press at body origin, drag to `body.x + 60` on row 1):

```
selection = Some(((0, 0), (1, 60)))
copied    = "INFO ok\nbye"
cells painted with selection bg on line 1 = 60   (the line is 3 chars)
```

60 cells highlighted, 3 characters copied. Dragging rightward off the end of a short line is how anyone selects "to end of line", so this is a routine gesture, not an edge case.

Preferred fix is to clamp at the source so both readers see one value:

```rust
-    (line.min(log.len().saturating_sub(1)), column)
+    let line = line.min(log.len().saturating_sub(1));
+    (line, column.min(log.line_chars(line)))
```

If a per-line length is awkward there, clamp in the renderer instead (`let to = if idx == er { ec.min(len) } else { len };` plus `let from = from.min(len);`), accepting that the stored state stays out of range.

### `src/log_view.rs:592-596` **[minor/codebase fit]** — `ordered_selection_public` is a naked wrapper

It exists solely to call the private `ordered_selection`, whose only other caller is `selection_text` in the same module. Making that method `pub` and deleting the wrapper is the same visibility with one fewer name. One call site (`widgets/editor.rs:255`).

### `src/app/mod.rs:25987` **[minor/correctness]** — the Cmd+C log guard falls through to the stub buffer

`if self.editor.log.is_some() && self.copy_log_selection()` returns early only when the copy *succeeded*. With a log open and no selection, control falls into the `diff` branch and then reads `self.editor.selection_text()` — the one-line stub this whole design exists to avoid. The Markdown branch just above (`25977-25985`) returns unconditionally once `markdown_preview.is_some()`, which is the right shape. Benign today because the stub yields nothing, but the guard claims to handle logs and then doesn't.

## 3. Test coverage

**The rendering half of the feature has zero coverage.** I deleted the entire selection-painting block (`widgets/editor.rs:300-320`, replaced with `let _ = &selection;`) and ran the suite: **3738 passed, no related failures.** The three failures were `double_click_in_terminal_word_selects_in_split_and_maximised_layout`, `terminal_session_restores_pane_layout_names_and_focus_across_restarts`, and `a_client_arriving_mid_swap_is_told_to_reconnect_not_seated` — all environmental (real shells / PTY / sockets), baselined on the unmutated head where the first two pass in isolation and the third fails identically. **No test observes a highlighted cell.** The end-to-end gesture test renders a frame but only reads `last_body` back out; it never inspects the buffer it drew.

**`a_selection_larger_than_the_copy_cap_is_clamped_and_reported` (`log_view.rs:766`) cannot test the property its name claims.** Its fixture is `"x".repeat(120)` — escape-free — so `budget.charge(raw.len())` and `out.len()` advance in lockstep and the cap is always reached before the budget. The divergence is structurally invisible to it. That is exactly the control row above: the plain fixture reports `clamped=true` and looks like a pass. **The fixture is not merely weak, it is the one shape that hides the bug**, on a view whose entire subject is colour-bearing text.

Its second assertion is loose in a way worth naming separately: `text.len() <= MAX_COPY_BYTES + 200` is satisfied by every value from 0 upward, including the 768 KB truncation. An upper bound alone cannot distinguish "stopped at the cap" from "stopped far short of it" — it needs a lower bound.

Tests I would write:

1. **The blocker's regression test**, coloured fixture plus the escape-free control, asserting `clamped` *and* a lower bound on length. Fails on this head (`clamped=false`, 768966), passes with the `cut_short` fix.
2. **A render assertion tying paint to copy** — assert the count of cells carrying `theme.selection()` on each row equals the character count `selection_text` contributed for that row. That one test kills both the deletion mutation and the paint/copy divergence, and it is one step past where the current gesture test stops.
3. **Drag past a short line's end** — assert painted width on the end line does not exceed the line length.
4. **`begin_log_selection`'s dismissal branch** (`app/mod.rs:25829-25834`) — a press outside `last_body` with a live selection must clear it and return `false`. Uncovered, and it decides whether a click elsewhere leaves a stale highlight.

## 4. Nitpicks

- `widgets/editor.rs:303-320`: the inner bare `{ ... }` block inside the `if let` is redundant.
- `app/mod.rs:25890`: the status line reports **bytes** here while every other copy path in this file reports **chars** (`25817`, `25984`, `25998`, `26009`). A CJK log line reports ~3× the characters copied.
- `log_view.rs:94-97`: `ScanBudget::new_with` has one caller; a wash either way.
- `app/mod.rs:41100-41109`: the doc comment on `log_cell_at` naming the cell-vs-character assumption is genuinely good — it states the drift and why matching the renderer is deliberate. Keep as-is. (Worth noting the column clamp above is a different axis from the one it documents.)

## 5. Scores

**CodeRabbit score (diff quality) — 1/5** — one blocker: `selection_text` drops the sweep's truncation signal, so a coloured log copies a fraction of the selection and reports it complete (768966 of 4194304, `clamped=false`, against a plain control at the full cap).

**Greptile score (codebase fit) — 2/5** — the Markdown-preview pattern is mirrored correctly, but the renderer and copy path disagree on the end-line column (60 painted, 3 copied), the entire painting path survives deletion with a green suite, and the Cmd+C guard falls through to the stub buffer.

The design is right and the hard parts — sweeping rather than caching, publishing `last_body` as frame truth, absolute coordinates against a stub `lines` — are all correct. Both defects are in reporting the result of that work rather than in doing it.

VERDICT: REQUEST_CHANGES

vitali87:
@coderabbitai review

vitali87:
Merging on CodeRabbit's re-review of the current head: all four of its findings are fixed and its threads resolved, every check is green, and the branch is rebased onto main.

The four, since one was a crash:

- **A panic.** `out.truncate(MAX_COPY_BYTES)` split a multi-byte character whenever the cap landed inside one, so a large selection of accented or non-Latin text crashed croft. Every earlier test here used ASCII, where every byte offset is a boundary, so none could reach it.
- Cmd+C fell through to the editor's copy path when the log copy had nothing to do, and a log tab's buffer is a one-line stub, so an empty selection could put placeholder text on the clipboard.
- The status line named the 4 MiB cap whenever the copy was clamped, but a sweep can also stop on the raw-byte budget at a fraction of that, so it reported four megabytes for a copy of a few hundred kilobytes.
- A frame that painted no log body left its previous rect published for the mouse to hit-test against.

That last fix is not where the finding placed it, and the test is why: with a zero-sized area the editor returns before `render_log` runs at all, so a reset inside `render_log` never executes. Every rect the frame publishes is now cleared at the top of the editor's render, before any early return.

The earlier fallback review's blocker and major are fixed too, with both mutations reproduced as controls: dropping the truncation signal gives 440412 bytes reported as a complete copy, and removing the column clamp gives 60 painted cells against 3 copied characters.

The session running those fallback reviews went offline before it could re-review, so the last round's fixes carry CodeRabbit's verdict rather than a second opinion. Saying so rather than merging quietly on a green tick.

Full suite: 3751 passed. Clippy clean. Doc-ownership gate clean.

PR #406, opened 2026-08-29
fix(ci): a doc captured and then restored within a branch is not a loss

vitali87:
A false positive in the pairwise pass shipped an hour ago in #403.

## What happens

That pass exists because a file the branch **adds** has no base version, so a doc capture inside it is invisible to a merge-base comparison. It walks every adjacent pair of branch commits instead.

Walking every pair means reporting every intermediate state the branch passed through. A doc captured in one commit and **restored in a later one** is reported as a loss, even though the branch tip is correct.

## Found immediately, on a real PR

#400 had a genuine capture: a helper inserted between `apply` and its doc block, in a file that branch adds, which is precisely the case #403 extended the gate to see. The gate caught it, named the right commit, I fixed it, and the gate kept reporting it, because the broken state still exists in an earlier commit on that branch.

So the gate was simultaneously right (there was a capture) and wrong (the branch was fine by then).

## Why this is the failure mode that matters

For a gate, missing a defect is bad and accusing a correct branch is worse. The first thing a maintainer does with a gate that cries wolf is stop reading it, and then it catches nothing at all. The merge-base pass never had this problem because it compares two endpoints; the pairwise pass walks every pair and needed the same endpoint check.

Losses from the pairwise pass are now filtered against the item's state at head. The merge-base pass is unchanged.

## Test

One, and it is a positive control: removing the filter fails it with

```
AssertionError: 1 != 0 : a doc restored before the branch tip is not a loss
```

The true positive is untouched, still covered by `test_a_capture_inside_a_file_the_branch_added_is_reported`, so the fix narrows the report rather than disabling it.

44 script tests pass. No version bump: scripts and their tests only, which the release-hygiene job's own filter excludes.

## Note

#403's own PR body argued that a structural guard is code, and can be wrong in exactly the ways it exists to catch. This is that, one hour later, in my own guard, found by the guard blocking a branch I had already fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Documentation checks no longer report false positives when documentation is removed temporarily but restored before the final branch state.

* **Tests**
  * Added regression coverage for restored documentation across multiple commits.

### DIFF
--- a/scripts/check_doc_ownership.py
+++ b/scripts/check_doc_ownership.py
@@ -162,6 +162,13 @@ def main():
     # commit-by-commit instead, which is the only place their history exists.
     added = [f for f in changed if not git("cat-file", "-e", f"{base}:{f}", allow_missing_path=True, allow_fail=True)]
     if added:
+        # What matters is the state at HEAD. A doc captured in one commit and
+        # restored in a later one is not a loss: the branch is fine, and
+        # reporting it blocks a PR whose head is correct. Without this the
+        # pairwise pass reports every intermediate state a branch passed
+        # through, which is exactly the false accusation that makes a gate
+        # untrustworthy.
+        head_state = {f: documented(git("show", f"{head}:{f}", allow_missing_path=True)) for f in added}
         commits = git("log", f"{base}..{head}", "--format=%H", "--reverse").split()
         for older, newer in zip(commits, commits[1:]):
             for f in added:
@@ -173,6 +180,7 @@ def main():
                         and name in after
                         and not after[name]
                         and (f, name) not in exempt
+                        and not head_state[f].get(name, False)
                         and not any(l[0] == f and l[1] == name for l in losses)
                     ):
                         # Name the commit the doc was last seen at. Saying
--- a/scripts/check_doc_ownership.py
+++ b/scripts/check_doc_ownership.py
@@ -103,6 +103,71 @@ def is_doc(line):
     return s.startswith("///") and not s.startswith("////")
 
 
+# What a line is, for the backward scan. Computed in one FORWARD pass,
+# because both multi-line constructs (an attribute spanning lines, a `/** */`
+# doc block) can only be recognised by reading downward: from below, `))]` and
+# `*/` are indistinguishable from code.
+DOC, ATTR, SKIP, CODE = "doc", "attr", "skip", "code"
+
+
+def classify(lines):
+    """Label every line DOC, ATTR, SKIP (blank or ordinary comment) or CODE.
+
+    Both multi-line shapes were false-ACCUSATION bugs rather than misses: an
+    attribute spanning lines, or a `/** */` block doc, made a documented item
+    read as undocumented and the gate reported a loss that had not happened.
+    For a gate that is the worse direction, since a gate that cries wolf stops
+    being read.
+    """
+    kinds = []
+    attr_depth = 0
+    in_block_doc = False
+    in_block_comment = False
+    for raw in lines:
+        s = raw.strip()
+        if in_block_doc:
+            kinds.append(DOC)
+            if "*/" in s:
+                in_block_doc = False
+            continue
+        if in_block_comment:
+            kinds.append(SKIP)
+            if "*/" in s:
+                in_block_comment = False
+            continue
+        if attr_depth > 0:
+            kinds.append(ATTR)
+            attr_depth += s.count("[") - s.count("]")
+            continue
+        # `/** ... */` is an outer doc comment lowering to the same `#[doc]`
+        # attribute as `///`. `/*! */` is INNER (it documents the enclosing
+        # item, not the next one) and `/***` is an ordinary comment, so
+        # neither counts.
+        if s.startswith("/**") and not s.startswith("/***"):
+            kinds.append(DOC)
+            if "*/" not in s[2:]:
+                in_block_doc = True
+            continue
+        if s.startswith("/*"):
+            kinds.append(SKIP)
+            if "*/" not in s[2:]:
+                in_block_comment = True
+            continue
+        if s.startswith("#["):
+            kinds.append(ATTR)
+            depth = s.count("[") - s.count("]")
+            attr_depth = max(depth, 0)
+            continue
+        if is_doc(raw):
+            kinds.append(DOC)
+            continue
+        if s == "" or s.startswith("//"):
+            kinds.append(SKIP)
+            continue
+        kinds.append(CODE)
+    return kinds
+
+
 def documented(text):
     """Map item name -> True when ANY definition of it carries a doc comment.
 
@@ -120,6 +185,7 @@ def documented(text):
     reported documentation as missing when rustc could see it perfectly well.
     """
     lines = text.splitlines()
+    kinds = classify(lines)
     state = {}
     for i, line in enumerate(lines):
         m = ITEM.match(line)
@@ -128,13 +194,9 @@ def documented(text):
         # Exactly one alternative captures per match.
         name = next(g for g in m.groups() if g)
         j = i - 1
-        while j >= 0:
-            s = lines[j].lstrip()
-            if s.startswith("#[") or s == "" or (s.startswith("//") and not is_doc(lines[j])):
-                j -= 1
-                continue
-            break
-        has_doc = j >= 0 and is_doc(lines[j])
+        while j >= 0 and kinds[j] in (ATTR, SKIP):
+            j -= 1
+        has_doc = j >= 0 and kinds[j] == DOC
         state[name] = state.get(name, False) or has_doc
     return state
 
--- a/scripts/tests/test_doc_ownership.py
+++ b/scripts/tests/test_doc_ownership.py
@@ -121,6 +121,41 @@ class ItemKinds(unittest.TestCase):
                 f"to the gate rather than merely undocumented",
             )
 
+    def test_multi_line_attributes_and_block_docs_are_read_correctly(self):
+        """Both shapes made a DOCUMENTED item read as undocumented.
+
+        That is a false accusation, not a miss: the gate would report a loss
+        that never happened, on a branch that had merely reformatted an
+        attribute or used `/** */`. For a gate that is the worse direction,
+        because one that cries wolf stops being read.
+
+        The inner form `/*!` and the ordinary `/*` and `/***` comments must
+        NOT count, or the gate invents documentation instead.
+        """
+        documented_shapes = [
+            ("multi-line attribute", "/// doc\n#[cfg(all(\n    feature = \"x\",\n    unix\n))]\nconst A: u8 = 1;\n"),
+            ("one-line block doc", "/** doc */\nconst A: u8 = 1;\n"),
+            ("multi-line block doc", "/**\n * doc\n */\nconst A: u8 = 1;\n"),
+            ("block doc then attribute", "/** doc */\n#[cfg(test)]\nconst A: u8 = 1;\n"),
+        ]
+        for label, text in documented_shapes:
+            self.assertEqual(
+                gate.documented(text).get("A"), True, f"{label} documents A"
+            )
+
+        undocumented_shapes = [
+            ("inner block doc documents the MODULE", "/*! module */\nconst A: u8 = 1;\n"),
+            ("ordinary block comment", "/* a note */\nconst A: u8 = 1;\n"),
+            ("triple-star rule", "/*** rule ***/\nconst A: u8 = 1;\n"),
+            ("four slashes is a rule", "//// rule\nconst A: u8 = 1;\n"),
+        ]
+        for label, text in undocumented_shapes:
+            self.assertEqual(
+                gate.documented(text).get("A"),
+                False,
+                f"{label} must not count as documentation",
+            )
+
     def test_a_declaration_inside_a_string_is_not_an_item(self):
         """The regex is line-anchored, so a declaration quoted mid-line is
         not mistaken for one."""
--- a/scripts/tests/test_doc_ownership.py
+++ b/scripts/tests/test_doc_ownership.py
@@ -249,6 +249,38 @@ class GitShapes(unittest.TestCase):
                 repo.exit_code(), 1, "the gate must fail on a captured doc"
             )
 
+    def test_a_doc_captured_and_then_restored_within_a_branch_passes(self):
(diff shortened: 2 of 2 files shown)


vitali87:

Running because CodeRabbit filed no inline threads on this PR (`reviewThreads` is empty at `b704158`), so this is the only review it gets. Prioritised ahead of four older croft PRs because it sits on main's CI path for every PR.

**The fix is correct on every axis I probed. Both findings are pre-existing gaps in `documented()`, on the merge-base pass this PR does not touch.**

---

## 1. Walkthrough

#406 adds an eight-line filter to the pairwise pass that #403 shipped an hour earlier. The pairwise pass walks every adjacent commit pair so that a doc capture inside a branch-*added* file is representable at all; the cost was that it reported every intermediate state, including captures the author had already repaired. The fix computes `head_state` once before the walk and drops any pairwise loss whose item is documented at head. The merge-base pass is untouched, correctly, since it already compares two endpoints.

| Path | Change | Assessment |
|---|---|---|
| `scripts/check_doc_ownership.py` | +8: `head_state` dict plus `and not head_state[f].get(name, False)` | Correct. Right property, no KeyError path, no true positive lost. |
| `scripts/tests/test_doc_ownership.py` | +32: capture-then-restore in a branch-added file | Genuine red/green; mutation-verified in both directions. |

## 2. The fix, verified in both failure directions

The "positive-control verified" claim holds, and understates the coverage. Mutated at the PR head (44 tests pass unmutated):

| Mutation | Result | Reading |
|---|---|---|
| Filter removed (`and True`) | FAIL: `a doc restored before the branch tip is not a loss` | Positive control confirmed. |
| Filter always drops (`and False`) | FAIL ×2, incl. `the gate must fail on a captured doc` | **Over-aggressive direction is covered** by the pre-existing true-positive test. |
| Wrong property (`name not in head_state[f]`) | FAIL ×2 | The suite distinguishes "name exists at head" from "documented at head". |

That third row is the one I most expected to fail, since it is the substitution a fix-then-verify cycle is least likely to catch — you write the test for the bug you just saw. It holds.

Every false-accusation scenario also behaves correctly at this head: rename within a branch-added file, deliberate deletion, file added then removed, capture then `git revert`, capture-restore-capture, single-commit branch (empty `zip`), and `git mv` leaving the path absent at head (no KeyError). None of these is covered by a test; see §4.

## 3. Findings

### `scripts/check_doc_ownership.py:31-38` **[major/correctness]** — a multi-line attribute makes a documented item read as undocumented

`ITEM`'s prefix `(?:#\[[^\]]*\]\s*)*` requires the attribute to open and close on one line, and `documented()`'s backward scan only steps over lines starting with `#[`. An attribute spanning lines matches neither, so the scan stops on the continuation line and the item reads as undocumented.

Verified by execution against `b704158` — no capture occurred, yet the gate fails the branch:

```
/// doc + #[cfg(unix)]                       -> {'a': True}
/// doc + #[cfg(all(\n unix,\n feature))]    -> {'a': False}

::error file=a.rs::`a` had a doc comment at main and has none now.   (exit 1)
```

**The trigger is a formatting change, not a code change.** `rustfmt` splits a `#[cfg]` across lines on its own once the attribute exceeds the line width, so this surfaces without anyone touching a doc comment, which is the worst way for a gate to start crying wolf.

**Fix:** join physical lines into logical ones by `[`/`(` balance before matching, or minimally extend the backward scan to step over a line ending in an unbalanced bracket.

### `scripts/check_doc_ownership.py:97-103` **[major/correctness]** — `/** */` block docs are invisible, so a comment-style change reports a false loss

`is_doc()` recognises only `///`. Rust's `/** ... */` is an equally valid outer doc comment lowering to the same `#[doc]` attribute, and the backward scan treats `*/` as an ordinary line and stops there.

```
/// doc\npub fn a() {}     -> {'a': True}
/** doc */\npub fn a() {}  -> {'a': False}    (exit 1)
```

The docstring at line 96 records the `////` narrowing but not this one, while the module docstring's premise — "a function that had a doc comment and now has none is the exact fingerprint" — is false for this shape.

**Fix:** recognise `/** */` in `is_doc()` and step over `*/`…`/**` in the scan; or, if block docs are deliberately out of scope, say so in the docstring and pin the known false positive in a test, so the decision is recorded rather than latent.

### `scripts/check_doc_ownership.py:118` **[minor/correctness]** — raw identifiers collapse to the key `r`

`([A-Za-z_]\w*)` stops at the `#`, so every raw-ident item shares one key:

```
documented("/// a\nfn r#match() {}\n\nfn r#type() {}\n")  ->  {'r': True}
```

Under "any documented wins" this under-reports, matching the deliberate behaviour already pinned in `test_same_named_items_are_merged_which_under_reports`, so it fails safe rather than loud. **Fix:** `((?:r#)?[A-Za-z_]\w*)`.

**Why these are major and not blocker.** I checked whether they fire today, and the check itself is worth reporting because my first attempt was wrong. I swept all 159 tracked `.rs` files comparing `documented()` against the source and got **0 disagreements** — then control-tested the probe by planting a known-broken multi-line attribute, and it *also* returned empty. The probe had reused the gate's own `#[`-skip rule, so it walked onto `))]`, failed `ITEM.match`, and never examined the case: it had inherited the exact blind spot it was built to find, and its silence meant nothing. Rebuilt to join logical lines by bracket balance first, it now fires on the planted defect and stays silent on clean input, and the real-tree sweep is still 0. So the gaps are genuinely latent — but that is a measurement, not the grep it started as.

## 4. Test coverage

The gaps that matter most, given this PR's own thesis:

- **Capture-restore-capture is absent**, and it is the sharpest test of the filter. The new test proves a restore is forgiven; nothing proves a *later* re-capture is still caught. It passes today, unpinned.
- **The `doc-removal:` exemption is untested on the pairwise path.** Both the exempt test and the rename test use files that exist at base, so only the merge-base pass. The exemption works on branch-added files (verified), but that is unpinned behaviour on the newer, less-exercised path.
- **`head_state[f]` is an unguarded subscript.** Safe because both loops iterate the same `added` list, but an edit that filters one and not the other turns a false accusation into a crash. Worth pinning via the add-then-remove shape.
- **No test covers a multi-line attribute or a block doc**, which is how both findings stayed invisible. `test_declaration_shapes_the_regex_must_not_miss` reads as an exhaustive sweep — its docstring says "Each was probed rather than assumed" — but every case in it is single-line, so it promises more than it delivers.

## 5. Nitpicks

- `documented(git("show", f"{older}:{f}"))` inside the double loop re-parses each commit's version twice (once as `newer`, once as `older` next iteration): O(2·commits·files) `git show` calls. Memoising on `(rev, path)` halves it if a branch ever gets long.
- The module docstring still says "Flag functions that lost their doc comment" and the CI job is named "doc comments stay with their function", but #403 widened scope to nine item kinds. The error message is what a maintainer reads first.
- `test_doc_ownership.py:271` builds the restored body as an inline concatenation; a `RESTORED_CONST` beside `DOCUMENTED_CONST`/`CAPTURED_CONST` would match the file's convention.

**Release hygiene:** verified against the job, not CONTRIBUTING. `.github/workflows/ci.yml:50-52` filters shipped files to `^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)`; both changed files are under `scripts/`, so `shipped` is empty and the job exits 0 at line 55. No version bump or release-notes replacement required — the PR body is correct.

## 6. Scores

**CodeRabbit score (diff quality) — 5/5.** The eight added lines are correct on every axis probed: right property, no KeyError path, no true positive lost, over-aggressive mutation caught by an existing test, and every false-accusation scenario behaving correctly. The findings live in code this diff does not touch.

**Greptile score (codebase fit) — 3/5.** One major, counted once: `documented()` reports a loss for two shapes where no capture occurred. Both are the exact false-accusation class this PR argues is the one that matters, and both are invisible to a shape test presenting itself as exhaustive. Neither occurs in `src/` today, which is why this is major rather than blocker.

Nothing here blocks the merge: the fix is sound and closes a live false positive. The regex gaps deserve a follow-up issue, **multi-line attribute first**, since `rustfmt` can trigger it with no one editing a doc comment.

VERDICT: APPROVE

vitali87:
Pre-empting the obvious question about this fix, since a positive control alone cannot answer it: **does the filter now swallow a real capture?**

It cannot, and the reason is the keying. The filter tests whether the item is **documented** at head, not whether it still exists there:

```python
and not head_state[f].get(name, False)
```

`head_state[f]` is `documented(...)`, a map of name to has-doc. Keying on mere existence would be the dangerous version: a genuine capture where the item survives and its doc does not would then be skipped, which is the exact failure a fix-then-verify cycle is least likely to catch, because the test you write is the one for the bug you just saw.

Measured on the two shapes:

```
capture persisting at head -> A documented? False   (loss reported)
restored at head           -> A documented? True    (loss skipped)
```

Both directions are covered by tests, not just the new one:

- `test_a_capture_inside_a_file_the_branch_added_is_reported` is the true positive through the pairwise path, and it still passes. If this fix had over-reached, that is the test that would have gone green-but-wrong, so it is the one that matters most here.
- `test_a_doc_captured_and_then_restored_within_a_branch_passes` is the new false positive, verified as a positive control: removing the filter fails it with `1 != 0 : a doc restored before the branch tip is not a loss`.

Other false-accusation shapes I checked while here, all already correct and covered: a **rename** within a branch is not a loss (`test_a_rename_is_not_a_lost_doc` — the check requires the name to still exist, so a rename drops out), a **declared removal** is exempt (`doc-removal:` trailer), and a declaration **inside a string** is not an item.

The one shape I have NOT covered is a commit that reverts an earlier one wholesale. It should behave identically to the restore case, since only the head state is consulted, but I have not written that test and would rather say so than imply it.

vitali87:
## 🤖 Fallback review — Claude Opus 5 (round 2, `abad571`)

Re-reviewing because `abad571` rewrote `documented()` rather than deferring my two majors to follow-ups. CodeRabbit still has zero inline threads, so this remains the only review.

**Both majors are fixed, all four near-misses are correctly rejected, and the rewrite changes nothing else.** Verdict flips to a clean approve.

## Prior findings

| Finding (round 1) | Fixed? | Evidence executed at `abad571` |
|---|---|---|
| Multi-line attribute reads as undocumented | **yes** | `/// doc` + `#[cfg(all(\n unix,\n feature="x"\n))]` → `{'a': True}` (was `False`) |
| `/** */` block doc invisible | **yes** | `/** doc */\npub fn a() {}` → `{'a': True}` (was `False`) |
| Raw identifiers collapse to key `r` | not addressed | Still `{'r': True}`; minor, fails safe, fine to leave |

## The rewrite does exactly what it claims and no more

The risk in replacing a backward scan with a forward classification pass is that it changes behaviour you did not intend. I diffed the two implementations over a shape matrix:

| case | `b704158` | `abad571` | delta |
|---|---|---|---|
| multi-line attribute | False | **True** | CHANGED (intended) |
| block `/** */` doc | False | **True** | CHANGED (intended) |
| doc, blank line, item | True | True | same |
| doc, `//` note, item | True | True | same |
| `/** doc */ pub fn a()` same line | False | False | same |

**Exactly the two intended shapes moved.** That is the strongest evidence available that a rewrite of this size is behaviour-preserving, and it is the check I would have wanted before merging it.

Your reasoning for fixing rather than filing is right, and stronger than the one I gave: both gaps are the *same failure direction the PR exists to fix*. Shipping "false accusations are the worse failure" while leaving two of them in the modified function would have been incoherent. The forward-pass justification also holds up — from below, `))]` and `*/` are genuinely indistinguishable from code, so no backward scan can classify them.

**The three near-misses are the part I would have missed.** A fix aimed at "recognise more things as docs" swallows all of them, and each fails in the opposite direction, making the gate *invent* documentation. Verified independently:

```
/*! inner doc */  -> False   (documents the enclosing item, not the next one)
/* comment */     -> False
/*** rule */      -> False
//// rule         -> False
```

I also probed shapes neither of us named, all correct: multi-line `/** ... */`, block doc combined with a multi-line attribute, attribute preceding the doc block, `*/` inside a string literal, `]` inside `#[doc = "]"]`, and a nested `/* /* */ */`. The string-literal cases are the ones a bracket-balance approach typically gets wrong; this one does not.

## Findings

**None against this diff.**

Two shapes deviate from what a reader might expect, and I checked both against `b704158`: **unchanged by this PR, so they are pre-existing and not findings here.** Recording them so they are not rediscovered as regressions:

- `/// doc` + blank line + item → `True`. Rust attaches nothing across a blank line, so this is the gate inventing documentation.
- `/// doc` + ordinary `//` line + item → `True`. Same direction.

I checked whether these fire, with an independently-written scanner (not reusing the gate's logic, after being burned by exactly that in round 1) and a planted-positive control confirming it fires. One real hit: `src/highlight.rs:356`, where a `///` block is separated from `static SYNTAX` by a `// ponytail:` note. **That instance is benign** — the doc block genuinely documents `SYNTAX`, so reading it as documented is correct. So the lenient behaviour is currently doing the right thing on the only place it applies, and tightening it would need care rather than being an obvious win. Worth a note in `is_doc`'s docstring, not a change.

## Test coverage

`test_doc_ownership.py` gains the shape sweep and the near-misses, which is the right pairing: the near-misses are what stop a future "recognise more docs" edit from over-reaching. Remaining gaps, unchanged from round 1:

- **Capture-restore-capture** — still absent, still the sharpest test of the head filter. The restore test proves forgiveness; nothing pins that a *later* re-capture is still caught.
- **Wholesale revert** — you flagged this yourself rather than implying coverage. I ran it at this head: exit 0, correct. It reduces to the restore case since only head state is consulted, so the test is cheap and worth having as a pin.
- **`doc-removal:` exemption on the pairwise path** — both existing tests use files present at base, so only the merge-base pass.

Your answer on the filter keying is confirmed by mutation at this head: `name not in head_state[f]` (existence) fails two tests, so the suite does distinguish it from `documented at head`. That was the failure I most expected to slip through.

## Scores

**CodeRabbit score (diff quality) — 5/5.** Both majors fixed at the mechanism rather than patched at the symptom; the forward pass is the right shape; near-misses pinned in both directions; behaviour-preserving elsewhere by direct comparison.

**Greptile score (codebase fit) — 5/5.** The round-1 majors are resolved in the function they lived in. The two remaining lenient shapes predate this PR, and their only real occurrence is benign.

Release hygiene: both files are under `scripts/`, so `.github/workflows/ci.yml:50-52` computes an empty `shipped` set and exits 0 at line 55. No version bump or release-notes replacement required.

VERDICT: APPROVE

PR #408, opened 2026-08-29
perf(log): index a log's newlines with memchr

### LINKED ISSUE #394: Rendered log view: opening indexes byte-by-byte, so a large log is not instant
## What happens

`LogView::open` (`src/log_view.rs`) builds its line-offset index by reading the file in 1 MiB chunks and testing **every byte** for a newline:

```rust
for (i, &b) in buf[..n].iter().enumerate() {
    if b == b'\n' { ... }
}
```

Measured on a 147 MiB, 2,000,000-line colour log, release build: **922ms** to open. That scales linearly, so the 2 GB case the module and #257 both name lands around 13 seconds before the first frame.

## Why it matters

The module's own header states the opposite:

> Logs are the worst case for size, so the file is NEVER loaded whole. This keeps only a line-offset index plus a small byte window around the viewport, exactly the posture `crate::hex` uses: opening a multi-gigabyte log costs one index pass, and scrolling costs one bounded read per refill.

Both halves of that are true except the implied cost: the file is not held in memory (2M lines of `u64` offsets is 16 MB) and scrolling is genuinely bounded, but "one index pass" is a full read of every byte at roughly 160 MB/s. #257's acceptance criterion "a 2 GB log opens without loading it fully into memory; scrolling stays smooth" is met on memory and missed on latency: the open blocks.

## Suggested direction

`memchr` is already in the dependency tree (pulled in via regex, among others). Replacing the per-byte loop with `memchr::memchr_iter` over each chunk is a contained change to one loop and typically runs several times faster on the same read, since it vectorises the newline scan rather than branching per byte.

Worth deciding alongside it: whether a file past some size should index lazily (index the head, open immediately, extend in the background) rather than paying the whole pass up front. That is a larger change and probably a separate issue, but it is the only route to a genuinely instant 2 GB open, since even a perfect scan still has to read 2 GB.

## How it was found

Measuring the find sweeps added in #393 (search in a rendered log). The find path itself is budgeted and stays in the tens of milliseconds; the open pass was the number that stood out beside it. Not a regression from that PR, and not touched by it.

vitali87:
Part of #394. Deliberately NOT `Closes`: this makes the index scan about three times faster, and #394 also asks for the open to stop reading the whole file, which stays open for a background-extension change.

## The one loop that reads every byte

`LogView::open` builds a line-offset index, and that scan is the entire cost of opening a large log: the view exists precisely so the file is never read whole, so nothing else in the open touches every byte. It tested one byte at a time, which cannot use the vector instructions that make a newline search memory-bound.

## Measured in one binary, not across two builds

Both implementations timed against the same buffer, alternately, in a single release binary:

| | |
|---|---|
| byte loop | **188ms** |
| `memchr` | **53ms** |
| full `open` (read + index) | 107ms |

122 MiB, 2,000,000 lines.

Timing two separate builds minutes apart on a shared machine would have measured the machine as much as the change. The probe also **asserts the two scans produce identical offsets before timing either**, so this compares two implementations of the same result rather than a fast wrong answer against a slow right one.

## The test the rewrite actually needed

Not "does it still find newlines" — every existing test covers that, and all of them use files far smaller than one `INDEX_CHUNK`. The failure a rewrite of this loop introduces is at the **seam**: an off-by-one at a chunk boundary loses or duplicates a line, and only for files past the first megabyte, which is exactly the size where nobody looks.

Both alignments are covered: a newline as the final byte of a chunk, and one as the first byte of the next.

## Dependency

`memchr` was already in the tree via `regex`; this makes it direct. That is a `Cargo.toml` change, so the version bump applies.

## What this does not fix

#394 also notes that a 2 GB log still costs a full read at open. This makes that read about three times faster, and it does not make it lazy. Genuinely instant would need indexing the head and extending in the background, which is a different change and a larger one; the issue says so and stays open for it.

Full suite: 3724 passed, 0 failed. Clippy clean at `--all-targets -D warnings`.


vitali87:
@coderabbitai review

vitali87:

CodeRabbit's check reads `pass` annotated **"Review rate limited"**, which is a no-op, and `@coderabbitai review` answered "Action not completed". `gh api repos/vitali87/croft/pulls/408/comments` returns `0`, so there is no inline artifact. This PR had no review at all. Reviewed at `bd4de74`, against a built tree.

### The rewrite itself is exactly equivalent

The old loop pushed on `b == b'\n'` after the cap check and broke out of the block on the cap. `memchr_iter` yields the same indices in the same order, and the cap check sits in the same place relative to the push, so `truncated` is set on the same newline and `line_starts` ends identical. Nothing about `pos`, the chunk stride, or the trailing-line handling moved.

### The seam test is live, not a passenger

Verified by mutation. Dropping a newline that lands at offset 0 of a chunk, which is precisely the second alignment the test constructs:

```rust
for i in memchr::memchr_iter(b'\n', &buf[..n]) {
    if i == 0 { continue; }   // MUTATION
```

fails on the second case:

```
  left: 1
  right: 2
```

The existing tests all use bodies far smaller than one `INDEX_CHUNK` and stay green under that mutation, so this test covers ground none of them reach. The reasoning in the PR body about which failure a rewrite of this loop actually introduces is correct.

### Minor: `docs/ARCHITECTURE.md` has an unbalanced parenthesis

The `memchr` aside opened a parenthesis of its own and closed it early, which orphaned the original closing paren:

> a `u64` line-offset index built with `memchr` (a per-byte comparison ... 122 MiB)**,** capped at MAX_INDEXED_LINES, past which the view reports truncation rather than showing a prefix as if it were the whole file**)** plus a 256 KiB parsed window

Counting delimiters over that line gives a final depth of `-1`. The cap clause used to live inside the parentheses and now sits in running text with a stray `)` after it.

Suggested shape, which keeps both asides and balances:

> a `u64` line-offset index built with `memchr` (a per-byte comparison cannot use the vector instructions that make a newline search memory-bound, and this scan IS the cost of opening a large log: 188ms against 53ms over 122 MiB), capped at MAX_INDEXED_LINES, past which the view reports truncation rather than showing a prefix as if it were the whole file, plus a 256 KiB parsed window

### Minor: `Closes #394` contradicts the body

The body's closing section says the issue "stays open" for the lazy-indexing half of #394. `Closes #394` in the first line will auto-close it on merge, so the two disagree and the keyword wins. Either drop the keyword and reference the issue plainly, or accept the close and open the follow-up for background extension before merging.

### Dependency

`memchr = "2.8"` matches what the lockfile already resolved through `regex` (`memchr 2.8.0`), so this promotes an existing transitive dependency to a direct one without moving any version. The `Cargo.lock` diff confirms it: the only change is the new entry under `croft-software`.

### Scores

- **CodeRabbit score (diff quality) — 4/5.** The change is small, equivalent, measured in one binary with the two scans asserted to agree before timing, and the one new test is proven live. Held off 5 for the two minors above, both in the surrounding prose rather than the code.
- **Greptile score (codebase fit) — 5/5.** The `Cargo.toml` comment follows the existing convention for promoting a transitive dependency (the `color_quant` entry directly below it does the same thing), the version bump and release-notes replacement match the per-change discipline, and the architecture entry is updated in the same commit.

VERDICT: REQUEST_CHANGES on the two minors. Both are one-line edits; re-review is not needed once they land.

vitali87:
Both minors fixed in `84ff74d`.

**Parenthesis.** The cap clause reads as running text now, which is what it became when the `memchr` aside took the parentheses. Delimiter depth over that line is back to `0`.

**`Closes #394`.** Replaced with `Part of #394`, and the body now says why: the scan is about three times faster and the open still reads the whole file, which is the half of #394 that stays open. The keyword would have closed the issue while the body said it should not.

No code changed, so the existing version bump still describes this branch accurately.

vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 84ff74d3ab8f014efdae4475b958aa4becaa7e20 stands in. Run by the M1 croft session at the author's request.

## Walkthrough

Replaces the per-byte newline loop in `LogView::open`'s index pass with `memchr::memchr_iter`, promotes `memchr` from a transitive dependency to a direct one, and adds one test covering a newline at an `INDEX_CHUNK` seam. The scan rewrite is semantically exact — I modelled both implementations over CRLF, no-final-newline, empty, newline-only, oversized-line, UTF-8-straddling-a-seam, and cap-boundary inputs, plus 6,000 randomised short-read/small-cap fuzz trials, with zero divergence in either `line_starts` or `truncated`. Release hygiene (version bump, release notes, lockfile) is complete. The findings below are about the test's advertised scope, not the production change.

| file | summary |
|---|---|
| `src/log_view.rs` | Scan loop swapped to `memchr_iter`; one new seam test. Offsets and truncation semantics verified identical. |
| `Cargo.toml` | `memchr = "2.8"` added as a direct dep with a rationale comment; version 0.1.807 → 0.1.808. |
| `Cargo.lock` | `memchr` added to `croft-software`'s dep list; resolves to 2.8.0, already the version in main's lock. No new package, no new transitive edge. |
| `src/release_notes.rs` | Highlight replaced, `Feature` → `Fix`. |
| `docs/ARCHITECTURE.md` | `log_view.rs` entry now names `memchr` and the 188ms/53ms measurement. |

## Findings

- **`src/log_view.rs:1136-1145`** **[minor/tests]** The test's contract line claims coverage the fixture does not provide — the review request and the PR body both describe "a multi-byte character straddling a boundary", but neither fixture contains a non-ASCII byte. I extracted both fixture bodies and checked: `vec![b'a'; INDEX_CHUNK - 1] + b'\n' + b"second\nthird\n"` (1,048,589 bytes) and `vec![b'b'; INDEX_CHUNK] + b'\n' + b"tail\n"` (1,048,582 bytes) are `all-ascii: True`. The seam behaviour that a multi-byte char would exercise is real and lives outside the indexer: `read_window`/`ensure` call `String::from_utf8_lossy` on a window whose end is a byte offset, so a code point split across that boundary becomes U+FFFD. That path is untouched by this diff, so this is a doc/scope mismatch rather than a bug — but the claim should either be dropped or backed by a fixture:

  ```rust
  // A code point that straddles the seam: its lead byte ends chunk one,
  // its continuation bytes begin chunk two.
  let mut body = vec![b'x'; INDEX_CHUNK - 1];
  body.extend_from_slice("é".as_bytes());
  body.extend_from_slice(b"\ntail\n");
  let (_d3, p3) = write_tmp(&body);
  let mut v3 = LogView::open(&p3).unwrap();
  assert_eq!(v3.len(), 2);
  v3.ensure(0, 1).unwrap();
  assert!(v3.visible_text(0).unwrap().ends_with('é'), "no replacement char at the seam");
  ```

- **`src/log_view.rs:1148-1149`** **[nit/tests]** The comment "Fill the first chunk exactly, so its final byte is a newline" is accurate about the fixture but rests on an unstated premise: that `reader.read(&mut buf)` with `buf.len() == INDEX_CHUNK` returns exactly `INDEX_CHUNK`. `BufReader::read` bypasses its internal buffer when the target is at least as large as its capacity, so the chunking is the OS's, not the constant's. I confirmed with a raw `read(2)` against a 1,048,589-byte file on this platform: sizes were `[1048576, 13]`, so the seam does land where the test intends *here*. It is not guaranteed by the API — on a short read the fixture still passes, but it stops testing the seam. Worth a word in the comment so a future reader does not treat the alignment as contractual.

### Verified, not findings

Recording these because they were the specific things I was asked to attack, and each came back clean:

- **The seam test is a real control.** I built four broken indexers against both fixtures. Chunk-relative offsets (`pos` dropped): `len` 3→4 and 2→3. Off-by-one (`pos+i` instead of `pos+i+1`): 3→4 and 2→3, with wrong starts. Duplicate newline at chunk end: 3→5 and 2→3. Skipping the first byte of each later chunk: fixture 1 passes but fixture 2 goes 2→1. Every bug is caught by at least one fixture, and the two fixtures together catch all four — the second alignment is load-bearing, not decoration. The test reads `INDEX_CHUNK` itself rather than a hardcoded number.
- **Cap equivalence is exact.** Both loops evaluate `line_starts.len() >= MAX_INDEXED_LINES` before every push and `break` out of the inner scan on the same iteration, leaving `truncated = true` and the identical `line_starts` prefix; the outer `if truncated { break }` is unchanged. Modelled at caps 1–6 over multi-chunk inputs and fuzzed 3,000 times with randomised short reads: zero divergence in offsets or flag. The `break` correctly exits only the inner loop in both versions, so the trailing-newline `pop()` sees the same vector.
- **Callers hold.** The diff touches exactly one loop plus one test — `Window`, `partial`, `ensure()`, `line()`, `visible_text()`, `len()`, `line_range()`, and `read_window()` are byte-identical to `origin/main`. Since `line_starts` and `truncated` are provably unchanged for every input, `render_log` (reads `view.truncated`, `view.len()`, `ensure`, `line`), `log_cell_at`, and the selection/drag path in `src/app/mod.rs` see no behaviour change.
- **Dependency hygiene.** `memchr` resolves to 2.8.0 in both main's lock and this branch's — `git show origin/main:Cargo.lock` already carried the identical `version`/`checksum` pair, so the only lock delta is the `croft-software` dep-list entry. No new package, no version bump elsewhere, no new build surface. Default features are on, which is what gives the SIMD path; that is the right choice here.
- **Release gate.** `Cargo.toml` 0.1.807 → 0.1.808 and `RELEASE_NOTES` fully replaced — both required by CONTRIBUTING.md:57-65 for any PR touching `src/`. Present. (Per instructions I am not scoring the 0.1.808 contention with sibling PRs.)
- **ARCHITECTURE.md matches the code.** The entry says the index is "built with `memchr`", "capped at MAX_INDEXED_LINES, past which the view reports truncation", and "a 256 KiB parsed window". All three check out against `src/log_view.rs:22`, `:27`, `:161-162`, `:237`. The 188ms/53ms figures are the PR's measurement, which I cannot re-run under the no-cargo constraint; the parenthesis balances after 84ff74d.

## Test coverage

- **The truncation path has no test at all, in this PR or before it.** `MAX_INDEXED_LINES` is 10,000,000, so a fixture that reaches it would be a ~20 MB file — which is why it has never been covered. This diff rewrites the exact loop that sets `truncated`, making it the moment to close the gap. The clean fix is to make the cap injectable for tests (`fn open_capped(path: &Path, cap: usize)`, with `open` delegating at `MAX_INDEXED_LINES`), then assert the three cases my model distinguishes: a file with exactly `cap` lines is *not* truncated and keeps all of them; `cap + 1` lines *is* truncated and reports exactly `cap`; and the truncated `line_starts` still holds correct absolute offsets when the cap is hit in a later chunk rather than the first. Without injection this stays untestable and a future rewrite of the same loop gets no signal — the `break`-out-of-inner-loop structure is precisely the kind of thing that silently changes meaning.
- **No test covers a UTF-8 code point split across a `WINDOW_BYTES` refill boundary** in `ensure`/`read_window`. Not introduced here, but it is the failure the new test's prose gestures at, and the suggested fixture above is a two-line addition.

## Nitpicks

- **`src/log_view.rs:1155`** The second fixture reuses the binding name `body` by shadowing, and the temp-dir guards are `_d` / `_d2`. Fine as written, but the two halves are independent scenarios; splitting into two `#[test]` functions would make a failure report say which alignment broke without reading the assert message.
- **`Cargo.toml:99-101`** The comment says memchr is "already in the tree via regex". True, and it is also pulled by ~19 other packages per the lock. Not wrong, just narrower than reality.

## Scores

**CodeRabbit score (diff quality) — 4/5** — The production change is exactly equivalent and well-motivated; only a minor test-scope overclaim and two nits.

**Greptile score (codebase fit) — 4/5** — Matches the module's existing posture and comment style, dependency and release hygiene are complete, docs track the code; the untested truncation path is the one gap the rewrite should have closed.

VERDICT: APPROVE

vitali87:
Both items addressed at `9ac9a6c`.

**The comment claimed more than the fixture did.** Correct, and the wording was mine. "Straddling the seam by a byte" described a newline sitting at offset 0 of the second chunk, which is an *alignment*, not a straddle: every fixture in the test is ASCII, so nothing in it is split across a boundary at all. That case is now described as what it is, and a genuine straddle is added: a 3-byte CJK sequence with its first byte as the last of one chunk and its other two as the first of the next.

Worth saying why it is worth having even though the index cannot fail it. `memchr` matches bytes, and no byte of a multi-byte UTF-8 sequence can be `0x0A`, so the index is indifferent. The windowed reader downstream is not: a window boundary falling mid-sequence hands `visible_text` invalid UTF-8. The new case pins the pair.

The first version of that fixture was wrong in an instructive way. It built a single line the size of a chunk, which exceeds the 256 KiB parsed window, so `visible_text` read back `""` and the assertion failed for a reason that had nothing to do with the seam. The fixture keeps its lines short now and asserts on the last one.

**The cap had never been tested.** Agreed, and it is the one branch in the index loop the rewrite moved.

Tested against a real file rather than by making the cap injectable. The injectable version is tempting and I think it is the wrong trade here: a cap that exists so a test can lower it becomes a second definition of the limit, and the two can disagree. The file is 10 MB of newlines and `memchr` walks it in one pass, so the test costs about 3 seconds.

It is a control, not a passenger. Removing the report while keeping the stop:

```rust
if line_starts.len() >= MAX_INDEXED_LINES {
    break;                    // MUTATION: stop, but do not report it
}
```

```
panicked: the view must say it did not index the whole file
```

The test also pins that the index stops **at** the cap rather than one past it, which nothing else did.

### Gate at `9ac9a6c`

- `cargo fmt --check`: clean.

One run before this reported a single failure, and the re-run above was clean. It is the PTY-timing load-flake class in #397, which I reproduced independently today: eight runs of the `app` and `terminal` tests under artificial load, one of which failed two tests, neither of them touching this diff.

vitali87:
Rebased onto main and renumbered to **0.1.809**, since #411 landed and took 0.1.808.

The rebase carried one conflict, in `src/release_notes.rs`, and it is the conflict this whole file causes: two versions describing two different changes, colliding because they share a file rather than because they disagree. That is exactly what #410 removes, and it is why #410 goes last in the merge order rather than first.

Resolved to this version's own note (the memchr one), since a release describes itself. No code changed in the rebase.

### Gate at the rebased head

- `cargo test --bin croft`: `3800 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.

vitali87:

# Re-review of PR #408 at `974d8db`

Verified against `974d8dbb1d22b844b5a51187e7dbe8a069cb6371` (branch `perf/log-index-memchr`), true PR scope taken from `git diff origin/main...HEAD` — five files, unchanged in shape from what I approved plus the new tests.

| # | Finding (at `84ff74d`) | Fix at `974d8db` | Verification |
|---|---|---|---|
| 1 | **[minor/tests]** Seam test's prose claimed a multi-byte straddle; both fixtures were pure ASCII (`all-ascii: True` on 1,048,589 and 1,048,582 bytes). | `src/log_view.rs:1169-1192` adds a third fixture: `unit = b"0123456789\n"` repeated `(INDEX_CHUNK-1)/11` times, padded to exactly `INDEX_CHUNK - 1`, then `"中文 tail\n"`. | **Resolved, and the straddle is genuine.** I computed the offsets: U+4E2D encodes to `E4 B8 AD`; byte `0xE4` lands at offset **1048575** (last byte of chunk 1) and `0xB8 0xAD` at **1048576–1048577** (first bytes of chunk 2). The sequence is split across the `INDEX_CHUNK` boundary, not merely near it. The in-test `assert_eq!(body.len(), INDEX_CHUNK - 1, "the next byte starts the CJK")` makes the fixture self-checking, so the alignment cannot silently drift if `unit` changes. Modelling `ensure(last, 1)` → `line_range` → `from_utf8_lossy` → `split('\n')` gives `visible_text(last) == "中文 tail"` with **no U+FFFD**, satisfying the `ends_with` assert. The author's note about the earlier attempt is borne out: the last line's span is **12 bytes**, well inside `WINDOW_BYTES` (262144), so the window genuinely holds the line being read — a chunk-sized single line would have exceeded it and read back empty, proving nothing. |
| 2 | **[nit/tests]** Comment implied `read` always returns exactly `INDEX_CHUNK`; that is the OS's choice, not the constant's. | `src/log_view.rs:1158-1159` reworded to "The other alignment: the newline is the FIRST byte of the second chunk rather than the last of the first." | **Partially addressed.** The misleading "straddling the seam by a byte" phrasing is gone and the two alignments are now named plainly. The underlying `BufReader::read` premise still is not called out, but I re-confirmed empirically that raw `read(2)` on a 1,048,589-byte local file returns `[1048576, 13]` on this platform, so the fixtures land where intended. Remains a nit; not worth another round. |
| 3 | **[test coverage gap]** No test exercised `MAX_INDEXED_LINES` / `truncated` at all — the one branch the rewrite moved. | `src/log_view.rs:1195-1218` adds `a_file_past_the_line_cap_reports_truncation_rather_than_a_prefix`: a real `MAX_INDEXED_LINES + 1`-byte all-newline file, asserting `v.truncated` and `v.len() == MAX_INDEXED_LINES`. | **Resolved, and the test is not vacuous.** I modelled the exact mutant the author names — delete `truncated = true`, keep the `break`. Correct impl yields `(len 10000000, truncated True)`; the mutant yields `(len 10000000, truncated False)`, so the first assert fails. The second assert (`len == cap`, not `cap + 1`) independently catches an off-by-one in the `>=` comparison. I withdraw my suggestion of an injectable cap: the author's reasoning is right — a cap a test can lower is a second definition of the limit, and the whole point of this branch is what happens at the *real* boundary. **CI-viable:** 9.54 MB tempfile, one `memchr` pass, and a 76.3 MB `Vec<u64>` peak. Python's byte-loop equivalent ran in 2.83s as a loose upper bound; the Rust path is a vectorised scan plus 10M pushes, comfortably sub-second. The 76 MB transient is the documented cost of the cap itself and is freed at test end. |
| 4 | *(not a finding — instructed not to score it)* Version contention with sibling PRs. | Rebased to **0.1.809** after #411 landed 0.1.808. | **Consistent.** `origin/main` is at 0.1.808; `Cargo.toml` = 0.1.809; `Cargo.lock`'s `croft-software` = 0.1.809; `memchr = "2.8"` still direct in `Cargo.toml` and still resolving to the already-present 2.8.0. `RELEASE_NOTES` remains the single replaced `NoteKind::Fix` entry describing this change. Both CONTRIBUTING.md:57-65 requirements met. |

## Production loop: unchanged

`git diff 84ff74d..HEAD -- src/log_view.rs` touches **only** the `mod tests` block. Against `origin/main`, the entire non-test delta is still the single hunk at `src/log_view.rs:153-168` I verified last round — same `memchr_iter` loop, same cap check before every push, same inner `break` leaving the outer `if truncated { break }` intact, same trailing-newline `pop()`. My prior equivalence evidence stands unmodified: zero divergence in `line_starts` or `truncated` across CRLF, no-final-newline, empty, newline-only, oversized-line, UTF-8-straddle, and cap-boundary probes, plus 6,000 randomised short-read and small-cap fuzz trials.

## One observation, not a finding

The branch is two commits behind `origin/main` (`b295743`, `7dedd3b` — both `scripts/check_doc_ownership.py` and CI doc-gate work). My first pass at the scope used a two-dot diff and those showed up as apparent reversals; the three-dot diff confirms the PR proposes only the five intended files. No content overlap, so no conflict — worth a `git merge origin/main` before the merge purely to keep the head current for the checks, per the repo's merge-not-rebase convention for a single branch.

## Scores

**CodeRabbit score (diff quality) — 5/5** *(was 4/5)* — Both prior findings addressed with fixtures I verified by computing the byte offsets rather than trusting the comments; production code unchanged and provably equivalent.

**Greptile score (codebase fit) — 5/5** *(was 4/5)* — The truncation gap that held the previous score is closed, and closed the way this codebase argues for elsewhere: a real file at the real limit rather than a test-only seam in the definition of the limit.

VERDICT: APPROVE

PR #410, opened 2026-08-30
build: one release-notes file per version

### LINKED ISSUE #399: build: one release-notes file per version so parallel PRs stop colliding on src/release_notes.rs
## Problem

Every shipped PR must bump `Cargo.toml` and REPLACE the single entry in `src/release_notes.rs`. With several PRs open at once, each one carries a different version and a different note in the same three files (`Cargo.toml`, `Cargo.lock`, `src/release_notes.rs`), so every merge conflicts every other open PR on exactly those files and each one needs a rebase with `git checkout --theirs` on the three. Today that was five rebases across #390/#391/#392/#393/#395, plus a version-slot ladder (.797–.801) agreed by hand between two sessions so the bumps would not collide.

The content never actually conflicts — each PR's note is about its own change — only the file does.

## Proposal

- `src/release_notes/<version>.md` (e.g. `src/release_notes/0.1.798.md`), one file per version, containing that release's highlights in the same one-sentence-per-line shape the panel renders now.
- `build.rs` reads `CARGO_PKG_VERSION`, picks the matching file, and emits it into `OUT_DIR` for `include_str!`; a missing file for the current version is a build error (the same guarantee as today: the binary always describes itself).
- The CI job "version bump + release notes" changes from "`src/release_notes.rs` modified" to "a file named after the new version exists and is non-empty", and additionally fails if a note for a version that is not the current one was modified.
- The welcome panel code is unchanged apart from where the text comes from.

Two PRs then touch disjoint files; only `Cargo.toml`/`Cargo.lock` still collide, and those resolve mechanically. The slot ladder can stay as a convention for the version number itself.

## Acceptance

- Bumping to a version with no notes file fails the build with a message naming the expected path.
- The welcome panel shows the current version's notes byte-for-byte.
- Two branches adding `0.1.X.md` and `0.1.Y.md` merge without a conflict outside `Cargo.toml`/`Cargo.lock`.
- CONTRIBUTING's release-notes rule is rewritten for the new layout.

vitali87:
Closes #399.

## The problem, measured

Every shipped PR must bump `Cargo.toml` and **replace** the single entry in `src/release_notes.rs`. With several PRs open, each carries a different version and a different note in the same file, so every merge conflicts every other open PR and each needs a rebase. That happened **five times in one day** across #390/#391/#392/#393/#395, and two sessions had to agree a version ladder (.797 to .801) by hand so the bumps would not collide.

The notes' *content* never conflicts. Each PR's note is about its own change. Only the file did.

## What changed

- `src/release_notes/<version>.md`, one file per version. One highlight per line, prefixed `feature:` or `fix:`, which selects the card's glyph and tint. Blank lines and `#` headings are ignored so a file can carry a title.
- `build.rs` bakes the file matching `CARGO_PKG_VERSION` into `OUT_DIR` for `include_str!`. **A missing or empty file is a build error**, which preserves exactly what the shared file guaranteed: a binary always describes itself, rather than shipping an empty panel.
- The CI job changes from "`src/release_notes.rs` was modified" to "this version's file exists and is non-empty", and **additionally fails if a PR edits another version's notes**. A release's notes describe that release; rewriting an older one changes what a binary already in the world claims to be.
- The welcome panel is unchanged apart from where its text comes from.

Two PRs adding different versions' notes now touch disjoint files. `Cargo.toml` and `Cargo.lock` still collide, and those resolve mechanically.

## Acceptance, checked rather than asserted

Removing this version's notes file and rebuilding:

```
/root/croft/src/release_notes/0.1.808.md could not be read (No such file or
directory (os error 2)). Every version needs a notes file, so the welcome
panel always describes the binary it is in. Create it with one highlight per
line, each prefixed `feature:` or `fix:`.
```

The panel's content is asserted non-empty and text-bearing at runtime, which is the "shows the current version's notes" criterion in the form a test can hold.

## Tests

Four, and two exist because a naive parser gets them wrong:

- a summary that itself contains a colon (`croft plot: numbers in, a chart out`) keeps its whole text; splitting on the first colon would cut it in half
- an unrecognised prefix reads as a **fix**, not a feature, and the test says why: overselling a release is the worse error
- headings and blank lines are not highlights
- the baked notes are non-empty and every summary has text, since a blank one would paint an empty row

Full suite: 3800 passed, 0 failed. Clippy clean at `--all-targets -D warnings`.

## Note for the next PR after this merges

CONTRIBUTING's release rule is rewritten for the new layout, so the next contributor writes `src/release_notes/<their version>.md` instead of editing a shared file.

## Summary by CodeRabbit

- **New Features**
  - The welcome screen now displays curated release highlights from version-specific release notes.
  - Release highlights support feature and fix categories with corresponding visual styling.

- **Documentation**
  - Updated contribution guidance to describe the per-version release notes format and required entry prefixes.

- **Chores**
  - Updated the application version to 0.1.808.
  - Added validation to ensure release notes are present and correctly maintained for each version.


vitali87:

CodeRabbit's check reads `pass` annotated **"Review rate limited"**, which is a no-op. `gh api repos/vitali87/croft/pulls/410/comments` returns `0`, so there is no inline artifact and this PR had no review at all. Reviewed at `33de8fb` against a built tree; the one finding below was confirmed by mutation, not by reading.

### Major — `the_baked_notes_are_this_versions_and_are_not_empty` checked only the second half of its name

The test asserted the notes are non-empty and that no summary is blank. Neither can see the version. **Any** version's file satisfies both, so a `build.rs` that baked the wrong one passed silently, which is precisely the failure this layout introduces: the file is now selected by a computed path rather than being the one file that exists.

Measured. Pointing the bake at a fixed file:

```rust
let path = dir.join("0.1.999.md");   // MUTATION
```

with `src/release_notes/0.1.999.md` reading `fix: A different release entirely.`:

```
test release_notes::tests::a_kind_prefix_selects_the_glyph_and_the_rest_is_the_summary ... ok
test release_notes::tests::a_colon_inside_the_summary_is_kept ... ok
test release_notes::tests::headings_and_blank_lines_are_not_highlights ... ok
test release_notes::tests::the_baked_notes_are_this_versions_and_are_not_empty ... ok
```

Four for four green while the binary carried another release's story: exactly the outcome the whole PR exists to make impossible, and the panel would have lied with nothing failing.

**Fixed in `eec82a6`.** The test now reads the current version's file directly, through a path built from `CARGO_PKG_VERSION` at compile time, and requires the baked text to equal it:

```rust
const FROM_SOURCE: &str = include_str!(concat!(
    env!("CARGO_MANIFEST_DIR"),
    "/src/release_notes/",
    env!("CARGO_PKG_VERSION"),
    ".md"
));
```

The same mutation now fails that one test and leaves the other three green:

```
assertion `left == right` failed: the binary carries notes that are not v0.1.808's
test result: FAILED. 3 passed; 1 failed
```

### Checked and sound

- **`parse` and the shared table.** `split_once(':')` splits only on a recognised kind prefix, so `feature: croft plot: numbers in, a chart out.` keeps its inner colon, and an unrecognised prefix keeps the whole line. The three parser tests each fail under a targeted change, and the unknown-prefix arm defaults to `Fix`, which is the right direction: overselling a release is the worse error.
- **`rerun-if-changed` on the directory** rather than the file. Correct for this shape: cargo scans a watched directory, so both editing this version's file and adding the next one rebuild. Watching only the file would miss the bump.
- **The CI check moved from "was this file touched" to "does this version's file exist and say something".** That is the change that removes the collision, and the version-monotonicity check above it still guarantees `$new` differs from `$old`, so the file cannot be one a previous PR already wrote.
- **The stale guard.** Filters `$changed` (which includes deletions) rather than `$shipped`, so removing an older release's notes is caught as well as editing them.
- **`Cargo.lock`.** Version line only; no dependency moved.

### Minor, fixed in the same commit

The `release-hygiene` job's header comment still said "src/release_notes.rs is replaced", naming the file this branch removes. It now describes the per-version file and cites #399.

### Scores

- **CodeRabbit score (diff quality) — 4/5.** The design is right and the parser is well covered. Held off 5 because the one test guarding the new failure mode did not guard it, and no amount of reading the diff would have shown that; it took a mutation.
- **Greptile score (codebase fit) — 5/5.** `CONTRIBUTING.md`, the module docs, the architecture comment on the CI job and the welcome-panel doc link were all updated in the same change, and the build-error-not-empty-panel choice keeps the guarantee the single file gave.

VERDICT: APPROVE at `eec82a6`.

### Gate at `eec82a6`

- `cargo test --bin croft`: `3800 passed; 0 failed; 10 ignored`.
- Build-error path verified by deleting the notes file: the build stops with the path it wanted, rather than shipping an empty panel.

vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head eec82a69cb03137adf7ba1cf3824c1171121e81f stands in. Run by the M1 croft session at the author's request.

## 1. Walkthrough

Replaces the single hand-edited `RELEASE_NOTES` const with one Markdown file per version (`src/release_notes/<version>.md`), baked into `OUT_DIR` by a new `bake_release_notes()` in `build.rs` and read back through `include_str!`. A missing or empty file is a hard build error, preserving the old guarantee that a binary always describes itself. The CI `release-hygiene` job switches from "the shared file was modified" to "this version's file exists and is non-empty", plus a new guard rejecting edits to other versions' notes. The migration is correct for this PR (version bumps to 0.1.808, so `0.1.808.md` is the only file needed) and the fixed test is now a genuine control.

| File | Summary |
|---|---|
| `.github/workflows/ci.yml` | `release-hygiene` now checks existence + non-emptiness of `src/release_notes/$new.md` and rejects changes to other versions' notes. Two existence-vs-authorship holes (see findings). |
| `build.rs` | New `bake_release_notes()`: reads `src/release_notes/$CARGO_PKG_VERSION.md`, panics on missing/empty, writes to `OUT_DIR`. Correct; `rerun-if-changed` semantics unchanged from main. |
| `src/release_notes.rs` | Const replaced by `include_str!(OUT_DIR)` + `parse()` + `OnceLock`-backed `release_notes()`. Four new unit tests. Parser is total — no panic path. |
| `src/release_notes/0.1.808.md` | New; this version's single `fix:` highlight. Tracked, not gitignored. |
| `src/app/welcome.rs` | `RELEASE_NOTES` → `release_notes()`. Behaviour-neutral. |
| `src/app/tests.rs` | Doc comment on `long_test_note()` updated for the new layout. |
| `CONTRIBUTING.md` | Release rule rewritten for one-file-per-version, with format and rationale. |
| `Cargo.toml` / `Cargo.lock` | 0.1.807 → 0.1.808. |

## 2. Findings

- **`.github/workflows/ci.yml:74`** **[major/ci]** — **The gate checks the file EXISTS, not that this PR wrote it, so a bump onto a pre-existing notes file passes with notes describing a different change.** The check is `git cat-file -e "HEAD:$notes"`, which asks only whether the blob is reachable at HEAD — it never consults `$changed`. Verified by simulation against real commits: I built a main carrying `0.1.809.md` from a bump that was later reverted (the notes file survived the revert, since only `Cargo.toml` and the code were backed out), then branched a PR that bumps 0.1.808 → 0.1.809 and touches `src/app/welcome.rs` while authoring **no** notes. The job exits **0**, and the welcome panel then advertises `feature: shipped then reverted.` — a feature that is not in the binary. This is exactly the "existence vs. authorship" hole suspected in the PR description, and a revert is a realistic route into that state. Fix — require the file to be in this PR's diff, which is also what the old gate meant:
  ```diff
  -          if ! git cat-file -e "HEAD:$notes" 2>/dev/null; then
  +          if ! printf '%s\n' "$changed" | grep -qx "$notes"; then
  +            echo "::error file=$notes::This PR changes shipped files but does not write $notes. Every release describes itself: write this version's highlights there, one per line, each prefixed 'feature:' or 'fix:'."
  +            fail=1
  +          elif ! git cat-file -e "HEAD:$notes" 2>/dev/null; then
  ```

- **`.github/workflows/ci.yml:84`** **[major/ci]** — **A `git mv` of the previous release's notes onto the new version passes every check, and the stale-notes guard cannot see the deletion.** `changed` comes from `git diff --name-only` (line 48), which collapses a rename to its **destination path only**. Verified: on a branch that bumps 0.1.808 → 0.1.809, runs `git mv src/release_notes/0.1.808.md src/release_notes/0.1.809.md`, and touches `src/app/welcome.rs`, `git diff --name-status` reports `R100 src/release_notes/0.1.808.md → src/release_notes/0.1.809.md`, but `--name-only` prints only `src/release_notes/0.1.809.md`. The job exits **0** on the reasoning that the new file exists, is non-empty, and no *other* notes file was "changed" — while 0.1.808's shipped notes have in fact been destroyed and v0.1.809 now ships v0.1.808's text verbatim. Note the control: an outright `git rm` of the old file **is** caught (I confirmed it exits 1), so the guard works and it is specifically rename-detection that defeats it. Fix — disable rename detection so a rename surfaces as add+delete:
  ```diff
  -          changed="$(git diff --name-only "$base" HEAD)"
  +          # --no-renames so a `git mv` of an old release's notes onto the new
  +          # version surfaces as a deletion the stale-notes guard below can see.
  +          changed="$(git diff --no-renames --name-only "$base" HEAD)"
  ```

- **`src/release_notes.rs:82`** **[minor/correctness]** — **A notes file containing only a heading passes CI and `build.rs` but parses to zero highlights, and the resulting build failure names the wrong cause.** `build.rs:50` guards on `text.trim().is_empty()`, and CI line 77 on `tr -d '[:space:]'` — both see `# Release 0.1.809` as non-empty content — but `parse()` filters `#` lines, yielding an empty `Vec`. `has_recent_panel()` (`src/app/welcome.rs:24`) then returns false and the "IN THIS RELEASE" panel vanishes. Verified by simulating the job (exit 0) and modelling `parse()` (0 notes). It is not shipped, because `the_baked_notes_are_this_versions_and_are_not_empty` asserts `!notes.is_empty()` — but the author sees a failing unit test rather than build.rs's targeted message. Fix: make `build.rs` reject content with no highlight lines, by applying the same heading/blank filter before the emptiness check.

- **`src/release_notes.rs:83`** **[minor/correctness]** — **A Markdown bullet (`- feature: …`) silently degrades to a `Fix` with the bullet in the summary.** `split_once(':')` puts `- feature` on the left, which fails `eq_ignore_ascii_case("feature")`, so the fallback arm at line 92 keeps the whole line: modelled output is `("Fix", "- feature: did a thing")`. The panel then paints a bug glyph plus a literal `- ` on a genuine feature. In a `.md` file bullets are the natural way to write a list, and nothing — CI, build.rs, or any test — flags it. Fix: strip a leading bullet marker before matching, `let line = line.strip_prefix("- ").or_else(|| line.strip_prefix("* ")).unwrap_or(line);`.

- **`docs/ARCHITECTURE.md:111`** **[minor/docs]** — **Stale reference to the deleted `RELEASE_NOTES` const.** The line reads "read synchronously from `release_notes::RELEASE_NOTES`". That symbol no longer exists anywhere in the tree — a `git grep RELEASE_NOTES` returns this line as its *only* hit (positive control: `git grep release_notes` returns 39 hits, so the corpus and pattern are both live). Fix: `release_notes::release_notes()`.

- **`docs/ARCHITECTURE.md:54`** **[minor/docs]** — **The file-tree entry still describes the old layout.** It documents `release_notes.rs` as the "baked-in data, updated on every version bump", which is now the job of `src/release_notes/<version>.md`; the `.rs` file holds only the parser and types. `CONTRIBUTING.md` was updated thoroughly, so this is the one doc left behind. Fix: amend the entry and add a `release_notes/` line naming the per-version files.

## 3. Test coverage

- **The CI job's own logic is untested**, and it is where both majors live. The repo already has precedent for pinning a CI filter with a fixture test (commit d7b6a2e, "ci: pin the sweep's jq filter with a fixture test"). I would add a fixture test that runs the release-hygiene script against synthetic git revisions and asserts the verdict for each shape: bump+new file (pass), bump+renamed old file (must fail), bump onto a pre-existing file with no authorship (must fail), file without bump (fail), no bump at all (fail), whitespace-only file (fail), and an old version's notes edited (fail). Every one of those is cheap to construct — I built all seven in a scratch clone for this review.
- **`parse()` has no test for a heading-only file or for bulleted input.** Both are inputs a contributor will plausibly write and both currently misbehave. I would add `assert!(parse("# 0.1.809\n").is_empty())` as an explicit statement of the gap (paired with the build.rs fix above), and `assert_eq!(parse("- feature: x\n")[0].kind, NoteKind::Feature)`.
- **No test renders the real baked notes end-to-end.** Every welcome-panel render test injects fixtures via `set_test_notes` (`src/app/tests.rs:21711`, `:21751`, `:5493`), so the glyph and summary shape is well covered but the live `release_notes()` → panel path is not. `the_baked_notes_are_this_versions_and_are_not_empty` checks the data, not the paint. A single render test using `WelcomeState::new()` asserting the panel is non-empty would close the seam where a parse regression ships a blank card.

Positive note on the fixed test: `the_baked_notes_are_this_versions_and_are_not_empty` (`src/release_notes.rs:155-167`) is now a real control. It compares `NOTES_MD` against a second `include_str!` built from `CARGO_PKG_VERSION` at compile time, so baking any other version's file fails on the `assert_eq!` — the vacuity the author found is genuinely gone, not merely relabelled.

## 4. Nitpicks

- `.github/workflows/ci.yml:87` — `printf '  %s\n' $stale` is unquoted, relying on word-splitting. Harmless here (paths match `[0-9.]+\.md` and the `[ -n ]` decision on line 85 uses the quoted variable), and it matches the existing style on line 60, but `printf '  %s\n' "$stale"` is not equivalent for multi-line input; a `while read` loop says it precisely.
- `.github/workflows/ci.yml:84` — `grep -v "^$notes$"` interpolates `$notes` into a regex, leaving its `.` characters unescaped. Confirmed cross-match: `src/release_notes/0X1Y809.md` is treated as the current version's file. Unreachable through the `ver()` sed, so cosmetic only; `grep -vx -F "$notes"` removes the class.
- `build.rs:48` — the `rerun-if-changed` on the notes *directory* is right for adds and deletes, but cargo watches directory mtime, so an in-place edit to `0.1.808.md` that does not change the directory entry may not retrigger. Adding `println!("cargo:rerun-if-changed={}", path.display());` for the file itself makes the "editing this one must too" comment true by construction rather than by filesystem behaviour.
- `src/release_notes.rs:101` — `release_notes()` parses under a `OnceLock`, but the input is a compile-time constant, so the lock buys nothing over parsing on each of the handful of calls. Not worth changing; noting only that the `'static` lifetime here comes from leaking through `get_or_init`, which is fine.

Checked and clean, for the record: `cargo package` will include the new directory (no `include`/`exclude` in `Cargo.toml`, file is tracked and not gitignored); `set -euo pipefail` does not abort on the `git cat-file -e` guard inside `if !`; `exit $fail` propagates correctly; the `rerun-if-changed` emission adds no regression since both branches of `main()` on `origin/main` already emitted one; the render path (`src/app/mod.rs:11555-11578`) is total, so a malformed line cannot panic at startup; and the version bump to 0.1.808 with its matching notes file is present.

## 5. Scores

**CodeRabbit score (diff quality) — 2/5** — Two majors: the gate accepts a pre-existing notes file as if this PR wrote it (verified exit 0 on a revert-derived scenario), and a `git mv` of the previous release's notes passes because `--name-only` hides the rename's source.

**Greptile score (codebase fit) — 3/5** — The mechanism, docs, and tests fit the repo's conventions well and the vacuous test is genuinely fixed, but `docs/ARCHITECTURE.md` still names the deleted `RELEASE_NOTES` const, and the repo's own precedent of pinning a CI filter with a fixture test (d7b6a2e) was not followed for a job this change substantially rewrites.

VERDICT: REQUEST_CHANGES

vitali87:
Both majors and all three minors are fixed at `a2538c4`. Both majors were real, and both were in the gate rather than in the Rust, which is the half of this PR that had no test at all.

**Authorship.** `git cat-file -e HEAD:$notes` proves the file exists, not that this PR wrote it. The revert scenario is the one that matters: a revert that drops the code but leaves the notes file behind puts notes for an unreleased version on main, and a later bump onto that version then passed every check while the panel advertised a feature the binary does not contain. The check is now that `$notes` appears in the changed set. The existence test is kept **behind** it rather than replaced, so a PR that changes this version's notes by *deleting* them is still caught.

**Renames.** Confirmed: `git diff --name-only` collapses a rename to its destination alone, so `git mv 0.1.808.md 0.1.809.md` reads as "wrote this version's notes" while destroying the previous release's. The detail that makes this convincing is the one in the review: a plain `git rm` **is** caught, which is the tell that rename detection was doing the hiding rather than the stale guard being wrong. `--no-renames` splits it back into a delete plus an add and the stale guard sees the delete.

**Heading-only file.** `build.rs` checked `!text.trim().is_empty()` while `parse` filters `#` lines, so a file holding only `# 0.1.808` passed the build guard and produced zero highlights: an empty panel arriving past the guard that exists to prevent one. `build.rs` now applies the same filter, and its message says "blank, or nothing but headings" rather than "empty", which was going to be a confusing thing to read while looking at a file with text in it.

**`- feature:` bullet.** A markdown list is the obvious way to write a file of one-liners, and it degraded to a `Fix` whose summary still carried the dash. Leading `- ` and `* ` are stripped before the prefix match, with `a_bullet_does_not_hide_the_kind_prefix` covering all three spellings.

**`docs/ARCHITECTURE.md`** named the deleted const on lines 54 and 111. Line 54 now describes the per-version layout and why it exists; line 111 points at `release_notes()`.

### On the suggested CI fixture test

Not in this PR, and the reason is that it wants to be bigger than this change. The job is 40 lines of inline shell inside `ci.yml`, so a fixture test has to either extract it to `scripts/` first (a change with its own review surface, touching the one workflow every PR depends on) or re-implement it in the test, which tests the copy rather than the gate. Both majors here came from simulating the script against real commits, which is what a fixture test would automate. Worth doing, and it should carry the extraction with it.

### Gate at `a2538c4`

- `cargo test --bin croft`: `3801 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.

vitali87:

## Re-review of PR #410 at `a2538c4`

Re-verified against head `a2538c4f7e0d266e4a8aae00fae2f0b1cdc3a593` (branch `build/release-notes-per-version`). All nine CI scenarios were re-simulated from scratch git repos built against the a2538c4 workflow text, with negative controls confirming the harness can still emit a pass.

| # | Finding (from eec82a6) | Fix at a2538c4 | Verification | Status |
|---|---|---|---|---|
| 1 | **major/ci** `ci.yml:74` — existence, not authorship: a bump onto a notes file left on main by a revert passed with notes describing a different change | `ci.yml:83` — authorship check `grep -qx "$notes"` against `$changed`, with the existence check demoted to an `elif` at :86 | Rebuilt the revert route: main at 0.1.808 carrying an orphaned `0.1.809.md`, PR bumps onto it and writes no notes. Now **exits 1** — `does not write src/release_notes/0.1.809.md`. Was exit 0. | **Fixed** |
| 2 | **major/ci** `ci.yml:84` — `git mv` of the old release's notes onto the new version passed; `--name-only` collapses a rename to its destination, hiding the delete | `ci.yml:48` — `git diff --no-renames --name-only` | `git mv 0.1.808.md 0.1.809.md` + bump now surfaces both paths; stale guard fires: `Only src/release_notes/0.1.809.md may change.` **Exit 1**, was exit 0. | **Fixed** |
| 3 | **minor/correctness** `release_notes.rs:82` — heading-only file passed CI and build.rs but parsed to zero highlights; failure surfaced as a confusing unit-test error | `build.rs:53-57` — `has_highlight` closure applying parse's own filter | Modelled `has_highlight` against `parse` over 12 inputs (heading-only, heading+blanks, blank, whitespace, bullets, plain): **zero disagreements**. Heading-only now panics in build.rs with a message naming the cause. | **Fixed** |
| 4 | **minor/correctness** `release_notes.rs:83` — `- feature: x` degraded to a Fix carrying the dash | `release_notes.rs:86` — `strip_prefix("- ")`/`("* ")` + `trim()`, and new test `a_bullet_does_not_hide_the_kind_prefix` (:125-133) covering `-`, `*`, and bare | Modelled: `- feature: One.` → `(Feature, "One.")`, `* feature: One.` → `(Feature, "One.")`, `- fix: Two.` → `(Fix, "Two.")`. Test asserts kind **and** summary for all three forms. | **Fixed** |
| 5 | **minor/docs** `ARCHITECTURE.md:111` — referenced deleted `RELEASE_NOTES` const | Now `release_notes::release_notes()` | `git grep RELEASE_NOTES` → **no hits** (control: `git grep release_notes` hits 10 files, so corpus and pattern are live) | **Fixed** |
| 6 | **minor/docs** `ARCHITECTURE.md:54` — tree entry described the old layout | Rewritten: "the text is DATA, one file per version in `src/release_notes/<version>.md`, baked in by `build.rs`", with the #399 rationale and the build-error guarantee | Read at :54; accurate to the implementation | **Fixed** |

### Scenario matrix at a2538c4

| Scenario | Expected | Result |
|---|---|---|
| PR #410 itself vs merge base | pass | **exit 0** |
| bump + adds its own `0.1.809.md` | pass | **exit 0** |
| bump + `git mv` old notes onto new version | fail | **exit 1** (stale guard) |
| bump onto pre-existing file, no authorship (revert route) | fail | **exit 1** (authorship) |
| adds notes file but forgets the bump | fail | **exit 1** (version + authorship + stale) |
| touches `src/`, no bump, no notes | fail | **exit 1** (version + authorship) |
| bump + whitespace-only notes | fail | **exit 1** (empty) |
| bump + edits another version's notes | fail | **exit 1** (stale) |
| PR deletes this version's notes (new `elif`) | fail | **exit 1** (changed but absent at HEAD) |
| *control:* docs-only PR | pass | **exit 0** — harness can still print 0 |

The `elif` ordering is correct: authorship first, then existence, then emptiness. Because the authorship check gates the chain, the deletion case reaches the existence check only when the file is genuinely in `$changed`, which is what makes the new error message ("cannot be the file the PR deleted") accurate rather than misleading.

### Still open

- **`build.rs:4` [minor/docs]** — the file's header comment still reads "the panel now renders hand-curated release highlights from `src/release_notes.rs` (compiled-in data…)". That is the old layout, and it sits directly above the function this PR added to replace it. The two `ARCHITECTURE.md` lines were fixed but this one was missed. Fix: `…highlights from src/release_notes/<version>.md, baked into OUT_DIR by bake_release_notes below`.
- **`src/release_notes.rs:86` [nit/style]** — possible `cargo fmt --check` failure, which CI enforces at `ci.yml:162` with a pinned rustfmt. The line is 91 chars (under `max_width = 100` from `rustfmt.toml`), but the chained expression inside the closure is 73 chars, over rustfmt's default `chain_width` of 60. Comparable `.map(|l| …)` bodies elsewhere in the tree are broken across lines by rustfmt once they exceed that (`src/app/mod.rs:10303-10305`). I could not run `cargo fmt` under the no-cargo constraint, so this is a flag, not a confirmed failure — worth a local `cargo fmt --check` before merge.
- **`src/release_notes.rs:87` [nit/correctness, carried over]** — `feature:` with nothing after it still parses to a note with an empty summary, which would paint a bare glyph. Caught by `the_baked_notes_are_this_versions_and_are_not_empty` (:189-194, asserts `!n.summary.trim().is_empty()`), so it cannot ship — but as with the old heading-only case, the diagnosis arrives as a test failure rather than a targeted build error. Low value; noting for completeness.
- **CI job logic remains untested.** The author explicitly declined extracting the inline shell to `scripts/` for a fixture test, and that is a reasonable scope call for this PR — the repo has precedent both ways (d7b6a2e added a fixture test for the sweep's jq filter). The nine scenarios above are now the only record that this gate behaves correctly; worth a follow-up issue so the next rewrite of this job has a net under it.

### Revised scores

**CodeRabbit score (diff quality) — 4/5** — Both majors are fixed at the root cause rather than patched around, and each fix is accompanied by a comment explaining the failure mode it prevents. Remaining: one stale comment in `build.rs`, plus a formatting risk on the new chain line.

**Greptile score (codebase fit) — 4/5** — Docs now match the implementation, the bullet fix ships with a test covering all three forms, and `has_highlight` deliberately mirrors `parse`'s filter so the build guard and the parser cannot drift. Held back from 5 by the untested CI job and the missed `build.rs:4` comment.

Under the rubric (4 = only minors/nits), no blockers or majors remain.

VERDICT: APPROVE

vitali87:
Both items handled.

**`build.rs`'s header comment** still said the panel renders highlights "from `src/release_notes.rs`", which is the file this branch removes. It now names `src/release_notes/<version>.md` and the function that bakes it.

**The `cargo fmt --check` concern:** unfounded, and I checked rather than took it on trust. The chain at `src/release_notes.rs:86` is 89 columns as written and rustfmt leaves it alone; `cargo fmt --check` is clean at this head with the pinned toolchain (rustfmt 1.9.0-stable). Reasonable thing to flag from a reviewer that could not run it.

Re-simulating all nine CI scenarios against the new workflow text was the right way to check this half of the PR, and it is what neither of us did the first time round: both majors here were in the shell, which is the part of this change that has no test.

vitali87:
Merged with main and renumbered to **0.1.810**, since #408 landed and main now owns 0.1.809.

**A confirmation worth recording, because this PR is the thing being tested.** Renumbering meant:

```
git mv src/release_notes/0.1.808.md src/release_notes/0.1.810.md
```

which is the exact gesture the new `--no-renames` guard exists to catch. I simulated the gate against my own commit before pushing:

```
--- changed notes files ---
src/release_notes/0.1.810.md
--- authorship check ---
PASS: this PR writes src/release_notes/0.1.810.md
--- stale guard ---
PASS: no other version's notes touched
```

It correctly passes. `0.1.808.md` never reached main, so the diff against the merge base shows an add and no delete. The guard catches destroying a **shipped** release's notes and permits a branch renumbering its own, which is the distinction it needs to draw and the one I would not have trusted without running it.

The merge conflict was, again, `src/release_notes.rs`: main carrying #411's note against this branch's parser. That file has now conflicted on three of my branches today, which is the case for this change stated better than the PR body states it.

### Gate at this head

- `cargo test --bin croft`: `3803 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.

One earlier run on this tree reported a single failure and the run above was clean; it is the PTY load-flake class, which #423 addresses.

vitali87:
Merged with main again after #401 landed, still at **0.1.810**.

This time the manifest conflict was resolved by taking main's `Cargo.toml` whole and changing only the version line, rather than picking a side. On the previous pass, on a different branch, `git checkout --theirs Cargo.toml` silently dropped the `memchr` dependency #408 had just added, and only `git diff origin/main --stat` caught it. A blunt side-pick on a manifest is not a resolution: the two sides differ in the version *and* in the dependency set, and taking one wholesale discards the other's changes.

`src/release_notes.rs` conflicted for the fourth time today, which continues to be this PR's own argument.

### Gate at this head

- `cargo test --bin croft`: `3807 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.
- Diff against `origin/main` is the ten intended files and nothing else.

vitali87:

## Re-review of PR #410 at `4fbe4a2`

Head `4fbe4a2e0d42bfc9fea1f1bf914f1b3116dd29fd`. **The workflow text is unchanged since a2538c4** (`git diff a2538c4..HEAD -- .github/workflows/ci.yml` is empty), so the nine-scenario matrix I ran at a2538c4 carries over unmodified. I re-ran the gate once against the current base as a live control: the real PR vs `origin/main` (now 0.1.809) exits **0**.

Critically, the two merges of main did not disturb reviewed logic — `ci.yml`, `src/release_notes.rs`, and `src/app/welcome.rs` are **byte-identical to what I approved** (0 diff lines each since a2538c4). The apparent movement in `tests.rs`/`CONTRIBUTING.md`/`ARCHITECTURE.md` is incoming main; the PR's own footprint in those files is unchanged (`tests.rs` is still just the one doc-comment edit).

| # | Open item at a2538c4 | Fix at 4fbe4a2 | Verification | Status |
|---|---|---|---|---|
| 1 | **minor/docs** `build.rs:4` — header still said the panel renders highlights "from `src/release_notes.rs`" | Rewritten to `src/release_notes/<version>.md`, "baked in by `bake_release_notes` below" | Read `build.rs:1-6`. `git grep "release_notes\.rs"` now returns exactly one hit, `ARCHITECTURE.md:54`, which is the file-tree entry legitimately naming the `.rs` file and whose prose already describes the new layout. Control: `git grep release_notes` hits 10 files, so pattern and corpus are live. | **Fixed** |
| 2 | **nit/style** `release_notes.rs:86` — possible `cargo fmt --check` failure (chain over default `chain_width`) | Author ran the pinned rustfmt; reports clean | Cannot run cargo, so this rests on the author's run — which is the right evidence, and it is the check CI enforces at `ci.yml:162`. I can confirm the line is **91 chars / 91 bytes, no non-ASCII**, under `max_width = 100` from `rustfmt.toml`. The 89-vs-91 gap is a counting convention, not a disagreement about the limit. | **Resolved** (author-verified) |

### Renumber and manifest trap

| Claim | Verified |
|---|---|
| Version 0.1.808 → 0.1.810 (main owns 0.1.809) | `Cargo.toml` = **0.1.810**; main = 0.1.809 ✓ |
| `src/release_notes/0.1.810.md` present | Present and sole file in the dir; carries one `fix:` highlight ✓ |
| `memchr = "2.8"` survived the manifest re-take | `Cargo.toml:101` — present ✓ (this is the #408 dep the trap would have dropped) |
| Lock's `croft-software` version | **0.1.810** ✓, and `memchr` present in the lock ✓ |
| No stale notes files | Neither 0.1.808.md nor 0.1.809.md exists on main or HEAD ✓ |

Post-merge build risk checked and clear: main does not yet contain `bake_release_notes`, so main builds today without a notes dir; the baking machinery and `0.1.810.md` arrive together in this PR, and both `build.rs`'s read path and the test's `include_str!(CARGO_PKG_VERSION)` resolve to the present file.

### Still open

Nothing blocking. The one item carried forward is the follow-up noted before: **the CI job's logic has no fixture test** — the author declined extracting the inline shell to `scripts/` for this PR, a reasonable scope call, but my nine scenarios remain the only record that this gate behaves correctly. Worth an issue so the next rewrite has a net under it.

### Scores

**CodeRabbit score (diff quality) — 5/5** — Both majors and both minors from the original review are fixed at the root cause; the two open items from a2538c4 are closed, and the renumber avoided the manifest trap cleanly.

**Greptile score (codebase fit) — 5/5** — Docs, build guard, and parser now agree with each other and with the implementation; the merge of main preserved reviewed logic byte-for-byte and kept #408's `memchr` dependency intact.

No findings remain (rubric: 5 = no findings). The untested CI job is logged as a follow-up rather than a finding against this diff.

VERDICT: APPROVE

PR #411, opened 2026-08-30
fix(workspace): keep non-ASCII folder paths in a .code-workspace

### LINKED ISSUE #396: workspace: the .code-workspace JSONC stripper walks bytes, mangling non-ASCII

vitali87:
Closes #396.

## The defect

`workspace::strip_jsonc` walked the file as bytes and rebuilt it with `bytes[i] as char`. That maps each byte to the Latin-1 code point of the same value, so every multi-byte UTF-8 sequence came back out as one char per byte.

A `.code-workspace` listing a folder named `café` reached serde as `cafÃ©`. `中文` became three mojibake chars. Non-ASCII in a *key* made the document fail to parse outright (`key must be a string at line 2 column 3`).

## The fix

`tasks::strip_jsonc` already does the same job over `chars()`. The single caller now uses it, and the byte-walking copy is gone.

## Why the swap is safe

An equivalence probe ran both implementations over 14 JSONC shapes: line comments, block comments, trailing commas in objects and in arrays, comment markers inside string literals, escaped quotes, CRLF, empty input, and the three non-ASCII cases. Eleven agreed byte for byte. The three that differed are exactly the defect:

```
DIFF non-ascii value   workspace: "{ \"name\": \"CafÃ© Noir\" }"   tasks: "{ \"name\": \"Café Noir\" }"
DIFF non-ascii key     workspace: "{ \"ä¸­æ\u{96}\u{87}\": 1 }"      tasks: "{ \"中文\": 1 }"
DIFF emoji             workspace: "{ \"a\": \"ð\u{9f}\u{98}\u{80}\" }"  tasks: "{ \"a\": \"😀\" }"

PROBE 3 of 14 cases differ
```

So the consolidation changes behaviour on the broken cases and on nothing else.

## The test is a control

`a_non_ascii_folder_path_survives_the_jsonc_strip` opens a commented `.code-workspace` whose folders are `café`, `中文` and `файл`. Grafting it onto the previous implementation (`git show origin/main:src/workspace.rs`) turns it red with the defect spelled out:

```
assertion `left == right` failed: the folder names must arrive intact rather than as mojibake
  left: ["cafÃ©", "ä¸\u{ad}æ\u{96}\u{87}", "Ñ\u{84}Ð°Ð¹Ð»"]
 right: ["café", "中文", "файл"]
```

## Gate

- `cargo test --bin croft`: 3796 passed, 1 failed. The failure is `cmd_clicking_a_non_web_hyperlink_refuses_instead_of_opening`, the PTY-timing load-flake class tracked in #397; it passes in isolation and touches nothing in this diff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.807"
+version = "0.1.808"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.807"
+version = "0.1.808"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/import_vscode.rs
+++ b/src/import_vscode.rs
@@ -325,9 +325,10 @@ fn read_jsonc(path: &Path) -> Result<Option<Value>> {
     }
     let raw =
         std::fs::read_to_string(path).with_context(|| format!("reading {}", path.display()))?;
-    // `crate::tasks::strip_jsonc` walks CHARS. The `workspace.rs` one walks
-    // bytes and mangles any non-ASCII value (#396), which a settings file
-    // carries routinely in paths and snippet bodies.
+    // `crate::tasks::strip_jsonc` walks CHARS. The copy that used to live in
+    // `workspace.rs` walked bytes and mangled any non-ASCII value (#396),
+    // which a settings file carries routinely in paths and snippet bodies;
+    // it was deleted in favour of this one rather than repaired.
     let stripped = crate::tasks::strip_jsonc(&raw);
     if stripped.trim().is_empty() {
         return Ok(None);
--- a/src/release_notes.rs
+++ b/src/release_notes.rs
@@ -60,6 +60,6 @@ pub struct ReleaseNote {
 
 /// What shipped in the current release. Replace on every version bump.
 pub const RELEASE_NOTES: &[ReleaseNote] = &[ReleaseNote {
-    kind: NoteKind::Feature,
-    summary: "croft theme-import turns a VS Code colour theme (.json) into a croft theme: workbench chrome, the 16 terminal colours, and the code palette, with a note for every colour croft had to derive because the theme did not name it.",
+    kind: NoteKind::Fix,
+    summary: "A .code-workspace file whose folder paths carry non-ASCII characters (accents, CJK, Cyrillic) now opens with those names intact instead of mojibake.",
 }];
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -734,6 +734,31 @@ mod tests {
         assert_eq!(v["b"], 1);
     }
 
+    /// The non-ASCII guarantee belongs on the function that carries it.
+    ///
+    /// Seven modules route their JSONC through here, and #396 was a second
+    /// copy in `workspace.rs` that walked BYTES: it rebuilt every multi-byte
+    /// character one byte at a time, so `caf\u{e9}` came back as `caf\u{c3}\u{a9}`
+    /// and the JSON still parsed. That copy is gone, but the only tests that
+    /// covered this were in the CALLERS, and each of those reaches this
+    /// function through file IO and a JSON parse. A future rewrite of this
+    /// stripper could reintroduce the defect for every module whose own
+    /// tests happen to be ASCII, which was all of them.
+    ///
+    /// Non-ASCII in a key as well as a value: a mangled key does not merely
+    /// arrive wrong, it makes the document fail to parse outright.
+    #[test]
+    fn strip_jsonc_preserves_non_ascii() {
+        let cleaned = strip_jsonc(
+            "{\n  // a comment, so the comment path actually runs\n  \"caf\u{e9}\": \"\u{4e2d}\u{6587} \u{444}\u{430}\u{439}\u{43b}\",\n}",
+        );
+        let v: serde_json::Value = serde_json::from_str(&cleaned).expect("valid strict JSON");
+        assert_eq!(
+            v["caf\u{e9}"],
+            "\u{4e2d}\u{6587} \u{444}\u{430}\u{439}\u{43b}"
+        );
+    }
+
     #[test]
     fn empty_workspace_has_no_tasks() {
         let tmp = tempfile::tempdir().unwrap();
--- a/src/vscode_theme.rs
+++ b/src/vscode_theme.rs
@@ -230,9 +230,10 @@ fn load_chain(path: &Path, depth: usize, out: &mut Vec<RawTheme>) -> Result<()> 
 
 /// Parse one theme document, tolerating VS Code's JSONC extras.
 fn parse_theme(raw: &str) -> Result<RawTheme> {
-    // `crate::tasks::strip_jsonc` walks CHARS. The `workspace.rs` one walks
-    // bytes and turns a non-ASCII theme name ("Caf\u{e9} Noir") into mojibake,
-    // which is exactly the sort of theme most likely to carry one.
+    // `crate::tasks::strip_jsonc` walks CHARS. The copy that used to live in
+    // `workspace.rs` walked bytes and turned a non-ASCII theme name
+    // ("Caf\u{e9} Noir") into mojibake (#396); it was deleted in favour of
+    // this one, which is why there is only ever one of these.
     let stripped = crate::tasks::strip_jsonc(raw);
     let value: serde_json::Value = serde_json::from_str(&stripped)?;
     // `tokenColors` is camelCase in the file; take it by hand rather than
@@ -1163,10 +1164,11 @@ mod tests {
         );
     }
 
-    /// Review finding: the byte-walking `strip_jsonc` in `workspace.rs`
-    /// turned every non-ASCII character into mojibake, so a theme called
-    /// "Caf\u{e9} Noir" imported under a mangled name. The chars-based
-    /// stripper in `tasks.rs` is the right one.
+    /// Review finding: the byte-walking `strip_jsonc` that used to live in
+    /// `workspace.rs` turned every non-ASCII character into mojibake, so a
+    /// theme called "Caf\u{e9} Noir" imported under a mangled name. It was
+    /// deleted for the chars-based stripper in `tasks.rs` (#396); this pins
+    /// that the theme path keeps using the right one.
     #[test]
     fn a_non_ascii_theme_name_survives_the_jsonc_strip() {
         let src = r##"{
--- a/src/workspace.rs
+++ b/src/workspace.rs
@@ -247,10 +247,8 @@ pub fn parse_workspace_file(path: &Path) -> Result<Vec<PathBuf>, String> {
     // so a workspace folder named "Caf\u{e9} Noir" arrived as "CafÃ© Noir"
     // with no error anywhere: the JSON still parsed (#396).
     //
-    // The two agreed on all eleven JSONC shapes probed (line and block
-    // comments, trailing commas in objects and arrays, comment markers inside
-    // strings, escaped quotes, CRLF, empty input) and differed on exactly the
-    // three non-ASCII ones, so this is the same behaviour minus the defect.
+    // The two treat comments, trailing commas and string literals alike, so
+    // the swap changes what happens to non-ASCII input and nothing else.
     let stripped = crate::tasks::strip_jsonc(&raw);
     let v: serde_json::Value =
         serde_json::from_str(&stripped).map_err(|e| format!("parse {}: {e}", path.display()))?;
@@ -542,7 +540,7 @@ mod tests {
     /// under a mangled name.
     #[test]
     fn a_non_ascii_folder_path_survives_the_jsonc_strip() {
-        let tmp = tempfile::TempDir::new().unwrap();
+        let tmp = tempfile::tempdir().unwrap();
         // A directory whose name needs more than one byte per character, in
         // three scripts, since the corruption is per byte.
         for name in [
--- a/src/workspace.rs
+++ b/src/workspace.rs
@@ -242,7 +242,16 @@ where
 /// workspace names at least one folder.
 pub fn parse_workspace_file(path: &Path) -> Result<Vec<PathBuf>, String> {
     let raw = std::fs::read_to_string(path).map_err(|e| format!("read {}: {e}", path.display()))?;
-    let stripped = strip_jsonc(&raw);
+    // `crate::tasks::strip_jsonc` walks CHARS. The copy that used to live
+    // here walked BYTES and rebuilt every multi-byte character as mojibake,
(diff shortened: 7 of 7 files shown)


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 031b349d8e10421ab659ccfbf664d156e3d1c0aa stands in. Run by the M1 croft session at the author's request.

## Walkthrough

`workspace::parse_workspace_file` no longer uses a module-local, byte-walking `strip_jsonc`; it calls the existing char-based `crate::tasks::strip_jsonc` and the byte-walking copy (~87 lines) is deleted. The old implementation rebuilt input via `bytes[i] as char`, which maps each byte to its Latin-1 code point, so every multi-byte UTF-8 sequence emerged as one char per byte (`café` → `cafÃ©`); the JSON still parsed, so the folder silently opened under a mangled name. A regression test covering Latin-1-supplement, CJK, and Cyrillic folder names is added, plus the required version bump and release-note replacement.

| File | Summary |
|---|---|
| `src/workspace.rs` | Swaps the local `strip_jsonc` call for `crate::tasks::strip_jsonc` with a 10-line rationale comment; deletes the 87-line byte-walking implementation; adds `a_non_ascii_folder_path_survives_the_jsonc_strip`. |
| `src/release_notes.rs` | Replaces the single `RELEASE_NOTES` entry: `NoteKind::Feature` (theme-import) → `NoteKind::Fix` (non-ASCII workspace paths). |
| `Cargo.toml` / `Cargo.lock` | Version `0.1.807` → `0.1.808`. |

**Equivalence verified, not assumed.** I modeled both algorithms in Python from the sources (`git show origin/main:src/workspace.rs` for the old one, `src/tasks.rs:296-370` for the new) and ran every shape you named — nested block comments, `//` inside a block comment, `/*` inside a line comment, a lone `/`, a trailing `/`, unterminated string/block comment at EOF, `\\` before the closing quote, `*/` inside a string, `'`, comment markers inside keys — plus 300,000 random ASCII inputs over a JSONC-weighted alphabet. **Zero ASCII divergences.** The positive control (a non-ASCII input) diverged as expected, so the harness could detect a difference and did not. The two structural differences I checked by hand are both benign: the old code re-skipped comments during comma lookahead while the new code's second pass sees a comment-free string, and both preserve the `\n` that ends a line comment.

The BOM case diverges (old: `ï»¿{…}`; new: `\u{feff}{…}`) but that is the non-ASCII fix, and both fail to parse — `serde_json` rejects a leading BOM either way, so no behavior regresses.

## Findings

None. Every claim I probed held up: the consolidation is equivalent on ASCII, the test is a real control, both `CONTRIBUTING.md` release requirements are met (`Cargo.toml` `0.1.808` vs `origin/main`'s `0.1.807`, `src/release_notes.rs:66-69` rewritten to describe this fix), and no caller depended on the deleted function — `grep -rn strip_jsonc src/` shows it was `fn` (private, never `pub`), with a single in-module caller, and every other consumer (`config_layers.rs:265,381`, `dap/configs.rs:109,182`, `import_vscode.rs:331,555`, `problem_matchers.rs:408`, `vscode_theme.rs:236`) already routed through `crate::tasks::strip_jsonc`. Nothing in `parse_workspace_file` depends on byte semantics: the stripped string goes straight to `serde_json::from_str`.

## Test coverage

The new test is a genuine control, not a tautology. Grafted onto the old implementation it fails for the right reason: the paths become `cafÃ©`/`ä¸­æ`/`ÑÐ°Ð¹Ð»`, those directories do not exist on disk, so `abs.canonicalize().unwrap_or(abs)` (`src/workspace.rs:277`) falls back to the mojibake path and the `file_name()` assertion fails with the mangled names spelled out. The fixture is not over-rich — the leading `//` comment is load-bearing (without it the stripper's comment path never runs), the three scripts cover 2-byte, 3-byte, and 2-byte-Cyrillic sequences, and the trailing commas after both the array and the object exercise the comma path in each container. I also confirmed APFS on this machine preserves NFC for these names, so the assertion is not normalization-fragile here.

One genuine gap, though it is the pre-existing one this PR inherits rather than creates:

- **`crate::tasks::strip_jsonc` has one test, `strip_jsonc_leaves_strings_intact` (`src/tasks.rs:726-736`), and it is entirely ASCII.** The non-ASCII guarantee that six modules now depend on is asserted only indirectly, from `workspace.rs` and `vscode_theme.rs` (`a_non_ascii_theme_name_survives_the_jsonc_strip`, `src/vscode_theme.rs:1171`), both of which reach it through file I/O and JSON parsing. I would add a direct unit test next to the existing one:

```rust
#[test]
fn strip_jsonc_preserves_non_ascii() {
    let cleaned = strip_jsonc("{\n  // c\n  \"k\": \"caf\u{e9} \u{4e2d}\u{6587}\",\n}");
    let v: serde_json::Value = serde_json::from_str(&cleaned).expect("valid strict JSON");
    assert_eq!(v["k"], "caf\u{e9} \u{4e2d}\u{6587}");
}
```

That puts the assertion on the function that actually carries the invariant, so a future refactor of `tasks::strip_jsonc` cannot silently reintroduce the bug in a module whose own tests happen to be ASCII.

## Nitpicks

- `src/workspace.rs:245-253` — the comment says "The two agreed on all eleven JSONC shapes probed". A count of shapes probed in a since-deleted throwaway script is not something a future reader can verify or maintain; it will read as stale the moment anyone re-probes. The surrounding comments in this file explain *why* (see `:265-267`, the `#164 review` note) rather than citing evidence counts. Consider trimming to the mechanism and the issue reference, dropping the "eleven"/"three" tallies.

- `src/workspace.rs:545` — `tempfile::TempDir::new()` where the other seven tests in this module all use `tempfile::tempdir()` (`:428, 453, 472, 508, 580, 667, 693`). Same thing; match the file.

- `src/import_vscode.rs:328-330`, `src/vscode_theme.rs:233-235`, `src/vscode_theme.rs:1166-1169` — three comments in two other files describe the deleted function in the **present tense** ("The `workspace.rs` one walks bytes and mangles…", "the byte-walking `strip_jsonc` in `workspace.rs`"). After this PR there is no such function, so a reader who greps for it finds nothing and cannot tell whether the comment is stale or the code is missing. Not a blocker — the advice they give (use the chars-based one) stays correct — but this PR is the moment to move them to past tense, e.g. `// The copy that used to live in workspace.rs walked bytes and mangled any non-ASCII value (#396); this one walks CHARS.`

## Out of scope, worth an issue

`src/app/mod.rs:40947-40961`, `normalise_dropped_token` — **the identical bug class, unfixed, in a path context.** The `file://` branch does `path.push((h * 16 + l) as char)` for `%xx` escapes and `path.push(bytes[i] as char)` for literal bytes. Modeling it: dropping a file at `/tmp/café` sends `file:///tmp/caf%C3%A9`, and this decodes to `/tmp/cafÃ©` — the same mojibake, so the drop silently opens the wrong (nonexistent) path. Both branches are wrong; the `%xx` one needs to accumulate bytes into a `Vec<u8>` and `String::from_utf8` at the end. It has no test (`grep -rn normalise_dropped_token src/` shows two callers at `:40834` and `:40885`, no test). Not this PR's job, but it is the same defect and it is live.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: no findings against the changed lines; the consolidation is equivalent on ASCII under 300k differential inputs with a passing positive control, and the regression test is a verified control with a right-reason failure mode. Only nits (comment tallies, `TempDir::new` vs `tempdir`).
- **Greptile score (codebase fit) — 5/5**: no findings outside the diff; the deleted function was private with one caller, every other consumer already used the chars-based stripper, and both documented release requirements (version bump, `release_notes.rs` replacement) are met. The stale cross-file comments and the `tasks.rs` unit-test gap are nits and a coverage suggestion, not defects.

VERDICT: APPROVE

vitali87:
All four items from the review are addressed, plus the coverage suggestion.

**The coverage gap was the real one.** `tasks::strip_jsonc` carries the non-ASCII guarantee for seven modules and had one ASCII-only test; every assertion about the behaviour lived in a *caller* and reached the function through file IO and a JSON parse. `strip_jsonc_preserves_non_ascii` now asserts it directly, next to the existing test.

It puts non-ASCII in the **key** as well as the value, which is the louder half of the defect: a mangled value arrives wrong and parses, while a mangled key makes the whole document fail to parse (`key must be a string at line 2 column 3`). The fixture keeps a leading `//` so the comment path runs rather than being skipped as a plain string copy.

**Three present-tense comments** in `import_vscode.rs` and `vscode_theme.rs` described the deleted function as though it still existed. Now past tense, each naming #396, so a reader who greps and finds nothing knows why.

**The shape tally** is gone from the rationale comment in `workspace.rs`. A count from a throwaway probe script is not something a later reader can re-derive or maintain; the comment says what the two implementations agree on instead. The measurement stays in this PR's description, where it is dated and attached to the change that made it.

**`TempDir::new()` to `tempdir()`**, matching the other seven tests in the module.

### Out of scope, and now filed

The `normalise_dropped_token` finding is the same defect class, live, in a path context. It is #414, and PR #416 addresses it.

### Gate at `87ed3b7`

- `cargo fmt --check`: clean. Worth noting because the first draft of the new test was **not**: the raw fixture string pushed the following assertion past the width, and only `cargo fmt --check` caught it.
- `scripts/check_doc_ownership.py origin/main HEAD`: no documentation lost across 5 changed files.

One earlier run of the suite on this tree reported a single failure. Three subsequent full runs were clean, and the failure was in the PTY-timing load-flake class tracked in #397, which is unrelated to anything in this diff; the run above is the one these numbers come from.

vitali87:

# Re-review: PR #411 @ `87ed3b7`

Head confirmed: `87ed3b7b09632e53c66f817e7c7497e16d6022db`, one commit on top of the `031b349` I reviewed (`test(tasks): pin the non-ASCII guarantee on the function that carries it`).

**Production change is unchanged.** Filtering the incremental diff to non-comment lines leaves exactly two hunks, both test-only: the new `strip_jsonc_preserves_non_ascii` function and the `TempDir::new` → `tempdir` swap. The line I approved — `let stripped = crate::tasks::strip_jsonc(&raw);` at `src/workspace.rs:251` — is untouched, and no `src/` file outside comments and `#[cfg(test)]` moved.

| # | Original finding | Fix at `87ed3b7` | Verification |
|---|---|---|---|
| 1 | Nit: `workspace.rs` rationale cited an unverifiable "eleven shapes / three non-ASCII" tally from a deleted script | `src/workspace.rs:250-251` now reads "The two treat comments, trailing commas and string literals alike, so the swap changes what happens to non-ASCII input and nothing else" | Read the hunk; tallies gone, claim is now a maintainable statement of mechanism. Re-ran the differential fuzz (200k random ASCII inputs, fresh models) — **0 divergences**, positive control diverges, so the comment's claim is true as written. |
| 2 | Nit: `TempDir::new()` against the file's 7-of-8 `tempfile::tempdir()` convention | `src/workspace.rs:543` → `tempfile::tempdir()` | Read the hunk; all 8 tempdir sites in `workspace.rs` now use one spelling. |
| 3 | Nit: three present-tense comments describing the deleted function | `import_vscode.rs:328-331`, `vscode_theme.rs:233-236`, `vscode_theme.rs:1167-1171` all moved to "The copy that used to live in `workspace.rs` walked bytes…", each citing #396 | `grep -rn "workspace.rs\` one walks\|byte-walking \`strip_jsonc\` in" src/` returns **no matches** — the absence-proof has a live control, since the same grep matched all three sites at `031b349`. |
| 4 | Coverage: the non-ASCII guarantee had no direct test on `tasks::strip_jsonc` | `src/tasks.rs:737-761` adds `strip_jsonc_preserves_non_ascii`, going beyond my suggestion by putting non-ASCII in the **key** as well as the value | Modeled the exact fixture through both implementations (below). |

## The new test is a real control

Fed the literal fixture from `src/tasks.rs:750` through both algorithms, with the old one re-extracted from `origin/main:src/workspace.rs` rather than reused from my earlier session:

```
OLD (byte-walking):  {\n  \n  "cafÃ©": "ä¸­æ file..." }   → parses, key is "cafÃ©"
                     → v["café"] is Null → assert_eq! FAILS
NEW (tasks::strip):  {\n  \n  "café": "中文 файл"\n}        → assertion holds
```

It fails on the old code for the right reason, and the author's choice to mangle a **key** is the stronger test: the doc comment claims a mangled key "makes the document fail to parse outright", which is true for the `workspace.rs` caller (an unknown key means no `folders` array) though here it surfaces as a wrong-key lookup rather than a parse error. The comment slightly overstates the mechanism for this specific fixture; not worth a change, since the test itself is correct and the distinction does not affect what it pins.

Release gate still clean: `Cargo.toml` and `Cargo.lock` at `0.1.808` vs `origin/main`'s `0.1.807`, and `RELEASE_NOTES` describes this fix.

## Note on method

My first re-run produced output about `file://` percent-decoding — unrelated to this PR. The shared scratchpad `probe.py` had been overwritten at 13:27 by another session working the `normalise_dropped_token` issue I flagged as out-of-scope, so the import failed and I read a stale artifact. I rebuilt both models from source under `rv411_87ed3b7/` and re-derived every result above; nothing in this table rests on the clobbered file. Worth knowing that the scratchpad is contended between sessions.

## Scores

Unchanged, now with the nits cleared:

- **CodeRabbit score (diff quality) — 5/5**: no findings against the changed lines. Equivalence holds under 200k differential ASCII inputs with a passing positive control, and both tests are verified red-on-old / green-on-new controls.
- **Greptile score (codebase fit) — 5/5**: no findings outside the diff. Deleted function was private with one caller, every other consumer already routed through `tasks::strip_jsonc`, both documented release requirements met, and the cross-file comments no longer point at a function that does not exist.

Still open and unrelated to this PR: `src/app/mod.rs:40947-40961` (`normalise_dropped_token`) carries the identical byte-as-char defect for dropped `file://` paths — another session appears to be on it.

VERDICT: APPROVE

PR #412, opened 2026-08-30
ci: key the doc-ownership gate's methods by their enclosing impl

### LINKED ISSUE #405: ci: the doc-ownership gate merges same-named items, so a capture on one is invisible
## What happens

`scripts/check_doc_ownership.py` keys its before/after comparison by item **name**, and where a file defines the same name more than once it takes "any definition documented" as documented:

```rust
/// One
fn new() {}
/// Two
fn new() {}
```

Insert an item above the first `new`'s doc and it captures that prose, but `state["new"]` stays `True` because the second `new` is still documented. The gate reports no loss.

`new` across several `impl` blocks is the common shape, so this is not exotic.

## Why it is keyed that way today

Two revisions cannot be lined up by position: any edit above an item moves it. Name is the only key that survives arbitrary edits, and "any documented" is the conservative reading of a duplicate. The failure is therefore an under-report (a real capture goes unnoticed), never a false accusation, which is the safer direction for a job that blocks merges.

Documented in `documented()`'s docstring and pinned by `test_same_named_items_are_merged_which_under_reports`, so a change here has to state which way it went.

## What a fix needs

An occurrence-unique key that still aligns across revisions. Candidates and their problems:

- **Positional index** (`new` #1, #2): reorders break it, and a mis-aligned key turns a silent miss into a false accusation on the wrong item.
- **Enclosing `impl` type** (`X::new`): correct for the common case and stable under edits within the block, but needs the script to track `impl` headers, which currently it deliberately does not parse (the block's name is not a unique key either, given `impl Trait for X`).
- **Byte offset or a hash of the item's own body**: robust against renames of neighbours, defeated by any edit to the item itself, which is exactly when its doc is most likely to move.

The `impl`-qualified key is probably the right trade, with a fallback to the current name-only behaviour where the enclosing block cannot be identified. It is a heavier change than it looks, which is why it is filed rather than folded into #403.

## How it was found

CodeRabbit's review of #403, which extends the same gate to consts and to branch-added files. Not introduced there: the keying dates from #314.

vitali87:
The gate keyed its before/after comparison by bare item name, and where
a file defined the same name more than once took "any definition
documented" as documented. `new` across impl blocks is the common
shape, so a capture on one `new` hid behind the other's surviving doc
and the gate reported no loss (#405).

Methods are now keyed `<impl header>::<name>` (`Foo::new`,
`Display for Foo::fmt`), with generics and lifetimes stripped so a
reformatted header does not move the key. The header is the whole
`Trait for Type` rather than the type alone, since two trait impls on
one type both define `fmt`. Block structure is tracked by brace depth
with string and char literals removed first, so a format string's
braces do not leave the scanner inside a block that has closed. Items
outside any impl, and items in a `mod` or `fn` body, keep the bare
name, so every existing key for those survives; where one key still
covers several definitions the conservative "any documented" reading
stays.

`doc-removal:` accepts the qualified key (`a.rs::A::new`).

Control: the original and the new gate report the same two items over
origin/main~40..origin/main, so the new keys add no false accusation on
real history.

Closes #405

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,7 +88,13 @@ so in a commit message on the branch:
 
 ```text
 doc-removal: src/path/to/file.rs::some_function_name
+doc-removal: src/path/to/file.rs::SomeType::method_name
 ```
+
+The key after the path is the one the gate's own error names: a bare name
+for a free item, or the enclosing `impl` header for a method (`Foo::new`,
+`Display for Foo::fmt`), so a declared removal of one `new` cannot excuse
+another.
 
 The file qualifier matters: an exemption keyed on the bare name would excuse
 every function of that name in every changed file, so a deliberate removal of
--- a/scripts/check_doc_ownership.py
+++ b/scripts/check_doc_ownership.py
@@ -12,8 +12,11 @@ comment and now has none is the exact fingerprint the insertion leaves behind,
 and it is what both known instances did.
 
 Deliberate removals are rare and are declared: put
-`doc-removal: <path>::<fn name>` in a commit message on the branch. The path
-qualifier keeps one declared removal from excusing a same-named function
+`doc-removal: <path>::<item key>` in a commit message on the branch, where
+the key is the one the gate's own error names: a bare name for a free item
+(`src/foo.rs::bar`), or the enclosing impl header for a method
+(`src/foo.rs::Foo::new`, `src/foo.rs::Display for Foo::fmt`). The path
+qualifier keeps one declared removal from excusing a same-named item
 elsewhere.
 
 Usage: check_doc_ownership.py <base-rev> <head-rev>
@@ -64,20 +67,39 @@ MISSING_PATH = re.compile(
 )
 
 
-# The head of an `impl` block. Its methods are keyed `<header>::<name>` so
-# that `new` in two impl blocks is two items, not one merged key whose "any
-# documented" reading hides a capture on either (#405). The whole header is
-# the key rather than the type alone: `impl Display for X` and `impl Debug
-# for X` both define `fmt`. Generics and lifetimes are stripped so a branch
-# that reformats them does not move every method to a new key.
-IMPL_HEADER = re.compile(
-    r"^\s*(?:pub(?:\([^)]*\))?\s+)?(?:unsafe\s+)?impl\b(?P<rest>.*)$"
+# The head of an `impl` or `trait` block. Its methods are keyed
+# `<header>::<name>` so that `new` in two impl blocks is two items, not one
+# merged key whose "any documented" reading hides a capture on either
+# (#405). The whole header is the key rather than the type alone: `impl
+# Display for X` and `impl Debug for X` both define `fmt`. A trait's own
+# method declarations get the same treatment, keyed `trait Foo::id`, since
+# two traits declaring `id` are the same shape. Generics and lifetimes are
+# stripped so a branch that reformats them does not move every method to a
+# new key.
+BLOCK_HEADER = re.compile(
+    r"^\s*(?:(?:pub(?:\([^)]*\))?\s+)?trait|(?:unsafe\s+)?impl)\b(?P<rest>.*)$"
 )
 
-# String and char literals, so the braces inside them are not counted as
-# block structure. `format!("{x}")` is on most lines that matter, and one
+# Everything on a CODE line that is not block structure, in the order a
+# left-to-right scan meets it: a complete raw string (`r#"{"#` is not a
+# `"..."` and its braces are content), a complete ordinary string, a char
+# literal (exactly one, possibly escaped, character between quotes, which
+# is what keeps a lifetime `'a` from opening one), a raw string that runs
+# past its line, an ordinary one that does (a `"...\` continuation), and a
+# `//` comment. `format!("{x}")` is on most lines that matter, and one
 # miscounted brace keys every item after it to the wrong block.
-LITERAL = re.compile(r'"(?:\\.|[^"\\])*"' + r"|'(?:\\.|[^'\\])'")
+TOKEN = re.compile(
+    r'(?P<raw>b?r(?P<hashes>#*)"(?:(?!"(?P=hashes)).)*"(?P=hashes))'
+    r'|(?P<plain>b?"(?:\\.|[^"\\])*")'
+    r"|(?P<char>b?'(?:\\.|[^'\\])')"
+    r'|(?P<open_raw>b?r(?P<open_hashes>#*)")'
+    r'|(?P<open_plain>")'
+    r"|(?P<comment>//)"
+)
+
+# The first `"` not escaped by a backslash: where an ordinary string that
+# ran past its line ends.
+UNESCAPED_QUOTE = re.compile(r'(?<!\\)(?:\\\\)*"')
 
 
 def strip_generics(text):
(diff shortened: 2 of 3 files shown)


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 609365bfcc1e2b71edb84be7fd38d661cc460fcf (branch ci/doc-ownership-impl-keys) stands in.

## Walkthrough

`scripts/check_doc_ownership.py` compared two revisions by bare item name and merged same-named items under an "any documented" reading, so a doc captured on `A::new` stayed invisible behind `B::new`'s surviving prose (#405). This PR tracks `impl` block structure by brace depth (string and char literals stripped first) and keys methods `<impl header>::<name>` — generics and lifetimes removed — while items outside any impl keep their bare name; the `doc-removal:` escape hatch was widened to accept the qualified form. The core fix is correct and well-tested, but block tracking is a line-oriented brace counter with three verified holes.

| File | Summary |
|---|---|
| `scripts/check_doc_ownership.py` | Adds `IMPL_HEADER`, `LITERAL`, `strip_generics()`, `impl_key()`, `brace_delta()`; `documented()` maintains a `blocks` stack + `pending` header and qualifies names by the innermost impl; `doc-removal:` regex widened for qualified keys. |
| `scripts/tests/test_doc_ownership.py` | Replaces `test_same_named_items_are_merged_which_under_reports` with 7 tests: impl-keying, trait-impl distinction, header shapes, brace tracking in literals, bare names outside impls, same-scope merging pinned, and an end-to-end #405 case. |

## Findings

- **`scripts/check_doc_ownership.py:252,257,260-261`** **[major/correctness]** — `pending` is set by an `impl` header but only cleared when a block actually opens, so a **body-less impl leaks its key onto the next unrelated block**, and that silently masks a real #314 capture the original scanner caught.

  `impl Eq for X {}` has `brace_delta == 0` (one `{`, one `}`), so line 259's `if delta > 0` never fires, `pending` is never reset at line 261, and it survives to qualify whatever block opens next. This is not hypothetical — `src/widgets/file_finder.rs:287` is exactly `impl Eq for RankedSlot {}`.

  Probe (before → after, a marker impl added *and* an item inserted above a doc block — the classic #314 capture):
  ```rust
  // before                          // after
  struct X;                          struct X;
                                     impl Eq for X {}
  mod util {                         mod util {
      /// Utility prose.                 /// Utility prose.
      pub fn go() {}                     pub fn newcomer() {}   // captures the doc
  }                                      pub fn go() {}         // lost its prose
                                     }
  ```
  ```
  [NEW]  reports_loss=False   after={'X': False, 'Eq for X::newcomer': True, 'Eq for X::go': False}
  [ORIG] reports_loss=True  ['go']
  ```
  The key moved from `go` to `Eq for X::go`, so `name in after` is false and the loss is dropped. **The new scanner misses a capture the old one reported** — a regression in the gate's core purpose, hidden behind a key rename.

  Fix — clear `pending` on any line that resolves it, not only one that opens a block:
  ```python
             delta = brace_delta(line)
             if delta > 0:
                 blocks.append((depth, pending))
                 pending = None
  +          elif pending is not None and "{" in LITERAL.sub("", line).split("//", 1)[0]:
  +              # A body-less `impl Eq for X {}` opens and closes on one line.
  +              # Its key must not survive onto the next block.
  +              pending = None
  ```

- **`scripts/check_doc_ownership.py:80,110`** **[major/correctness]** — `LITERAL` cannot match Rust **raw strings** (`r"…"`, `r#"…"#`), so braces inside them are counted as block structure and the depth counter desyncs on **7 of 161 real files under `src/`**.

  `LITERAL` is `"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])'` — the `r#"` prefix and the `"#` terminator are outside its grammar, and it has no multi-line string handling at all. Verified against a correct reference tokenizer (raw strings, byte strings, char-vs-lifetime, nested block comments), comparing the gate's per-line `brace_delta` sum at EOF:
  ```
  src/app/tests.rs         depth=-18   src/notebook.rs          depth=+4
  src/dap/configs.rs       depth=-12   src/pair.rs              depth=+5
  src/keymap.rs            depth= -4   src/problem_matchers.rs  depth=-2
  src/tasks.rs             depth= -6   (reference tokenizer: all balanced)
  ```
  Concrete culprits: `src/keymap.rs:830` `r#"[{"key": "ctrl+click", …` (delta −1, should be 0); `src/tasks.rs:491` `r#"{ "tasks": [` (delta +1); `src/pair.rs:3135` (delta +2). Line-by-line trace on the verbatim `keymap.rs:830` shape shows the stack corrupting:
  ```
  1 d= 1 blk=[(0,'Foo')]   :: impl Foo {
  3 d= 1 blk=[(0,'Foo')]   :: let j = r#"[{"key": "ctrl+click", …   <- swallowed a }
  6 d=-1 blk=[]            :: }                                     <- depth NEGATIVE
  7 d= 0 blk=[(-1,'Bar')]  :: impl Bar {                            <- block opens at -1
  ```
  Mitigating: I checked every mis-scoped item against the reference tokenizer and **today zero items are actually mis-keyed**, because in all 7 files the desync begins inside `#[cfg(test)] mod tests`, below every `impl`. So this is a live invariant break with no live symptom — but it makes the `blocks` stack unsound for any file whose raw strings sit above an impl, and the gate has no assertion that would notice.

  Fix — extend `LITERAL` to cover raw strings (single-line, which is what the counter can reason about) and skip lines it cannot balance:
  ```python
  -LITERAL = re.compile(r'"(?:\\.|[^"\\])*"' + r"|'(?:\\.|[^'\\])'")
  +# Raw strings first: `r#"…"#` is not a `"…"` and its braces are content.
  +LITERAL = re.compile(
  +    r'r(#*)"(?:(?!"\1).)*"\1' + r'|b?"(?:\\.|[^"\\])*"' + r"|b?'(?:\\.|[^'\\])'"
  +)
  ```
  and, since a multi-line raw string still cannot be balanced line-wise, consider tracking an "inside raw string" flag in `classify()` alongside the existing `in_block_comment` machinery (lines 172-188), which already solves exactly this shape for `/* */`.

- **`scripts/check_doc_ownership.py:263`** **[minor/correctness]** — depth is allowed to go negative and is never clamped, which turns a transient miscount into permanent corruption. Once `depth == -1` (trace above), `blocks[-1][0] >= depth` stops popping correctly and every subsequent block is keyed at a negative opening depth. Given the counter is known-fallible (finding above), a floor is cheap insurance:
  ```python
             depth += delta
  +          # A miscount must not corrupt every later block; a file's depth
  +          # can never be negative.
  +          depth = max(depth, 0)
             while blocks and blocks[-1][0] >= depth:
  ```

- **`scripts/check_doc_ownership.py:246-247,272`** **[minor/correctness]** — a `trait` body pushes `None` (it is not an `impl`), so **trait methods keep a bare name and same-named methods across two traits still merge** — the exact #405 bug, unfixed for traits. Probe:
  ```rust
  trait A { /// doc a
            fn id(&self); }
  trait B { /// doc b
            fn id(&self); }
  ```
  → `{'A': False, 'id': True, 'B': False}` — one `id` key for two declarations; inserting an item above `A::id`'s doc is not reported (`reports_loss=False`, same as ORIG). Latent only: I confirmed the repo currently defines **zero** traits (`grep -rn 'trait ' src/` yields 6 hits, all inside string literals; positive control: the same flags find 294 `impl` lines). Worth either handling `trait` in `IMPL_HEADER` or noting the carve-out in the docstring so the next reader knows it was a decision.

- **`scripts/check_doc_ownership.py:255,257` / `:73-75`** **[minor/correctness]** — a **multi-line `impl` header keyed only by its first line** loses the qualifier entirely. `IMPL_HEADER` matches `impl<T>` and `impl_key` strips the generics to the empty string, so `pending` becomes `""` — falsy — and line 272's `next((k for _, k in reversed(blocks) if k), None)` skips it:
  ```rust
  impl<T>
      Trait for X<T>
  {   /// doc
      fn m() {}   }
  ```
  → `{'m': True}` (bare), where the single-line form gives `{'Trait for X::m': True}`. A rustfmt rewrap therefore silently moves an item between keys — `before={'Trait for X::m': True}`, `after={'m': True}` — which is a missed detection on that item for that PR. `test_impl_header_shapes` covers the `where`-clause wrap but not this one. Latent: only one multi-line `impl` header exists in `src/` today (`file_finder.rs:287`, body-less).

- **`CONTRIBUTING.md:88-95` and `scripts/check_doc_ownership.py:14-17`** **[minor/docs]** — the `doc-removal:` regex now accepts `a.rs::A::new` and `a.rs::Display for Foo::fmt` (I verified all six realistic forms parse), and the gate's own error message at line 349 *emits* the qualified key — but both the module docstring (`doc-removal: <path>::<fn name>`) and CONTRIBUTING's fenced example (`doc-removal: src/path/to/file.rs::some_function_name`) still show only the bare form. A contributor following the docs will paste a key the gate no longer matches. Update both to show the qualified shape.

## Test coverage

- **The `pending` leak from a body-less impl** (finding 1) — the regression above. Add:
  ```python
  def test_a_bodyless_impl_does_not_qualify_the_next_block(self):
      text = ("impl Eq for X {}\n"
              "mod util {\n    /// doc\n    pub fn go() {}\n}\n")
      self.assertEqual(gate.documented(text), {"go": True})
  ```
  This fails today with `{'Eq for X::go': True}`.

- **Raw strings in brace tracking** (finding 2) — `test_braces_in_strings_and_chars_do_not_break_block_tracking` covers `format!("{{ {x} }}")` and `'{'` but no `r#"…"#`. Add the verbatim `src/keymap.rs:830` shape:
  ```python
  def test_raw_strings_do_not_break_block_tracking(self):
      text = ('impl Foo {\n    /// doc\n    fn m() {\n'
              '        let j = r#"[{"key": "click"}]"#;\n    }\n}\n'
              '/// top\nfn top() {}\n')
      self.assertEqual(gate.documented(text), {"Foo::m": True, "top": True})
  ```

- **A multi-line `impl` header whose first line is only `impl<T>`** (finding 5) — `test_impl_header_shapes` should gain `("header wrapped by rustfmt", "impl<T>\n    Trait for X<T>\n{")` asserting `Trait for X::m`.

- **An EOF invariant over real files.** No test asserts the counter is self-consistent. A cheap guard that would have caught finding 2 at authoring time: scan every `src/**/*.rs` and assert `depth == 0` and `blocks == []` at EOF.

**Verified sound, no action needed:** the author's parity claim reproduces exactly — running both scanners over the 45 changed `.rs` files in `origin/main~40..origin/main` gives identical output (`src/session_host.rs::broadcast`, `src/widgets/terminal.rs::TrigRowPaint`) under both. No documented-true bare name disappears under the new scanner across all 161 `src/` files (10182 → 10278 keys, zero losses). 6 of the 7 new tests fail against `origin/main`'s scanner, so they are genuine controls; the 7th (`test_same_named_items_in_the_same_scope_are_still_merged`) passes on both by design as a pinning test, and it does fully preserve the replaced test's assertions (both `documented()` calls return identical values under old and new). The `::error file=…` annotation is safe — the qualified key lands in the message body, after the property terminator, so spaces and `::` cannot corrupt GitHub's parse. This PR touches only `scripts/`, which is outside the `release-hygiene` job's `shipped` glob (`.github/workflows/ci.yml:52`), so the version-bump / release-notes rule correctly does not apply.

## Nitpicks

- `scripts/check_doc_ownership.py:74` — `IMPL_HEADER` accepts `pub impl`, which is not legal Rust (probe: `pub impl Foo {…}` → `{'Foo::m': True}`). Harmless, but the `pub` alternative is dead syntax copied from `ITEM`; dropping it would make the regex say what it means.
- `scripts/check_doc_ownership.py:80` — the `r'…' + r"…"` string concatenation splits one regex across two literals for no reason now that neither half is commented separately; a single raw literal reads better.

## Scores

- **CodeRabbit score (diff quality) — 2/5**: two majors whose evidence is inside the changed code — the `pending` leak at lines 252-261 that regresses a capture the old scanner caught, and `LITERAL` at line 80 being unable to match raw strings.
- **Greptile score (codebase fit) — 2/5**: both majors reproduce against real repo content — `src/widgets/file_finder.rs:287` is the body-less impl, and 7 of 161 real `src/` files desync — plus a docs drift between the widened regex and CONTRIBUTING's documented form.

VERDICT: REQUEST_CHANGES

vitali87:
## 🤖 Fallback review — follow-up at head eab8d0b

The reviewer re-verified the tree with the fixes for all six findings, by behaviour rather than by reading the diff:

| Finding | Fix | Verification |
|---|---|---|
| 1. `pending` leak from a body-less `impl Eq for X {}` (major) | a header is spent on any line that opens a brace | the masked #314 capture is reported again (`['go']`, was missed) |
| 2. raw strings and `"...\` continuations miscounted (major) | left-to-right `TOKEN` tokenizer with cross-line `open_string` state | EOF depth invariant over all 161 `src/` files: 0 desynced, down from 7 |
| 3. negative depth unclamped (minor) | `depth = max(depth + delta, 0)` | present in `feed()` |
| 4. trait methods merged (minor) | `trait Foo::id` keys | two traits declaring `id` now separate, capture reported |
| 5. rustfmt-wrapped `impl<T>` header lost its key (minor) | header accumulated until its brace | `Trait for X::m`, identical to the single-line form |
| 6. docs drift (minor) | docstring + CONTRIBUTING show the qualified form | regex accepts every prior form plus `trait Foo::id` |

Tests 19 → 24; each new one fails against the previous head and passes now, including the EOF-invariant test over every real file. CI's exact invocation passes 56/56. Parity on `origin/main~40..origin/main` is byte-identical to the original gate (the same two pre-existing items, each printed twice by a pre-existing double-report between the added-file and changed-file loops that both scanners share). The `pub impl` nit is also taken.

**Revised verdict: APPROVE** at eab8d0b (CodeRabbit 4/5 → no majors remain; Greptile 4/5).

PR #413, opened 2026-08-30
fix(theme-import): one id for canonically equivalent theme names

### LINKED ISSUE #407: theme-import: canonically-equivalent theme names mint two different ids
## What happens

`slug` in `src/vscode_theme.rs` derives a theme id from the theme's name without normalising Unicode, so two names that are **canonically equivalent** produce different ids and install as two separate themes:

```
"Café Dark"            (NFC, precomposed é)  ->  "café-dark"
"Cafe\u{301} Dark"     (NFD, e + acute)      ->  "cafe\u{301}-dark"
```

Those two strings render identically, compare equal under NFC normalisation, and are the same theme to any user. Importing the same theme from two sources, or re-importing after an upstream re-encode, yields two picker entries and two directories.

`İstanbul` behaves the same way, and worse: its lowercase under Rust's `to_lowercase` is `i̇stanbul` (i + combining dot above), so the id carries a mark that the precomposed form does not.

## Not a regression from #395

Worth stating clearly, because it looks like one. The `\p{M}` change in that PR was suspected of introducing this, on the reasoning that marks never previously reached an id. That reasoning is wrong, and running the old implementation shows why: `char::is_alphanumeric()` is true for a precomposed `é` (category `Ll`, not a mark), so the old code already minted `café-dark` for one form and `cafe-dark` for the other. Different ids either way; the marks change what the second one looks like, not whether they diverge.

Verified by running the previous `slug` rather than reasoning about it, which is the only thing that distinguishes "new" from "pre-existing" here.

## Suggested fix

Normalise to NFC before slugging. That collapses the two forms to one id and leaves every existing single-form id unchanged, since a name already in NFC normalises to itself.

`unicode-normalization` is the usual crate and is not currently a dependency; it is small and has no transitive weight. Worth checking whether anything else in croft would want it (the editor's grapheme handling may already imply it) before adding it for one call site.

An existing installed theme whose id was derived from an NFD name would change id under this fix, so the change wants the same collision handling the importer already has: a re-import recognises its own manifest by header, so it would overwrite in place rather than mint a suffix.

## How it was found

Review of #395, where the reviewer flagged it as introduced by that PR and then corrected itself after running the parent revision. The correction is the useful part: a model of the old code (a reimplementation of `is_alphanumeric` in another language) gave the right verdict for the wrong reason, and only executing the real thing settled it.

vitali87:
`slug` derived a theme id from the name without normalising Unicode, so
two names that are canonically equivalent minted two ids and installed
as two separate themes: "Café Dark" (precomposed é) became `café-dark`
while "Cafe\u{301} Dark" (e + combining acute) became `cafe\u{301}-dark`.
Both render identically and are the same theme to anyone; importing it
from two sources, or after an upstream re-encode, gave two picker
entries and two directories (#407).

The name is now NFC-normalised before slugging, and again after
lowercasing, since lowercasing can itself decompose (`İ` becomes `i` +
a combining dot above). A name already in NFC normalises to itself, so
no existing single-form id changes. `unicode-normalization` is the one
new dependency; it has no transitive weight.

Two fixture expectations in the non-Latin-script test pinned the NFD
spelling of the id (`cafe\u{301}`, `u\u{308}ber`) and now pin the NFC
one (`café`, `über`); the point of that test, that a combining mark is
not a separator, is unchanged and still exercised by the Devanagari case,
which carries marks in NFC.

Closes #407

## Gate
- `cargo test --bin croft vscode_theme::`: 28 passed (the new test was red on the parent: `left: "cafe\u{301}-dark" right: "café-dark"`).
- `cargo clippy --all-targets -- -D warnings`: clean. `cargo fmt --check`: clean.
- Version 0.1.809 (main is 0.1.807; #408/#410/#411 each claim 0.1.808). Release notes replaced.


vitali87:

Requested by the author. CodeRabbit is rate limited repo-wide (its check reads `pass` annotated "Review rate limited", which is no review), and `gh api repos/vitali87/croft/pulls/413/comments` returns `0`. Reviewed at head, against a built tree; every claim below is a measurement.

### The fix is right, and one half of it is unexercised

The trailing `.nfc()` after `to_lowercase()` is **load-bearing, not defensive**, and nothing in the PR would notice if it were deleted. Removing it:

```rust
let lowered: String = s.nfc().collect::<String>().to_lowercase(); // MUTATION
```

leaves the entire module green:

```
test result: ok. 28 passed; 0 failed
```

Including `canonically_equivalent_names_share_an_id`, which is the test written for this. So the step that the comment justifies at length is the one step no test defends.

**Why the existing cases cannot reach it.** The comment cites `İ` (U+0130), whose lowercase expands to `i` + U+0307. That sequence has no precomposed form, so the second NFC is a no-op there, and `slug("İstanbul") == slug("i\u{307}stanbul")` holds with or without it. A sweep over every code point from U+0020 to U+30000 finds **zero** single characters where `NFC(lower(NFC(c)))` differs from `lower(NFC(c))`.

**An input that does reach it.** Extending the sweep to base + combining mark pairs finds 81, the first being `H` + U+0331 (combining macron below). There is no precomposed uppercase, so the first NFC leaves it alone; lowercasing gives `h` + U+0331, which **does** compose, to U+1E96 `ẖ`. So without the trailing NFC, `H̱ana` and `ẖana` mint two different ids: exactly the bug #407 is about, surviving the fix for it.

Verified in Rust, not just in the sweep. This assertion passes as written and fails under the mutation:

```rust
// The trailing NFC is load-bearing: lowercasing can expose a composition
// the uppercase spelling had no precomposed character for. "H" + U+0331
// has no precomposed uppercase, so the first NFC leaves it; its lowercase
// "h" + U+0331 composes to U+1E96.
assert_eq!(slug("H\u{331}ana"), slug("\u{1e96}ana"));
```

```
  left: "h\u{331}ana"
  right: "ẖana"
```

Suggest adding it to `canonically_equivalent_names_share_an_id`. One line, and it turns the comment's argument into something the suite checks.

### Blocker — `cargo fmt --check` fails on this branch

Two hunks in the diff are unrelated to the change and are formatting *regressions*, which suggests an older rustfmt ran over the file:

```
Diff in /root/croft/src/vscode_theme.rs:24:
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result, anyhow};

Diff in /root/croft/src/vscode_theme.rs:1398:
-        assert!(std::fs::read_to_string(&first.path)
-            .unwrap()
-            .starts_with(GENERATED_HEADER));
+        assert!(
+            std::fs::read_to_string(&first.path)
+                .unwrap()
+                .starts_with(GENERATED_HEADER)
+        );
```

The import ordering is the edition-2024 style (lowercase items sort after uppercase), so the branch reverted it to the 2021 ordering. CI runs `cargo fmt --check` at `.github/workflows/ci.yml:130-131`, so this fails the `clippy + tests` job. Running `cargo fmt` on the branch with the pinned toolchain (rustfmt 1.9.0-stable) reverts both hunks and shrinks the diff to the actual change.

### Checked and sound

- **The NFC/NFD fixtures are genuinely different.** The two `parse_theme` literals in `canonically_equivalent_names_share_an_id` render identically in a diff, which is the classic way a self-comparison hides. Dumping the code points: `43 61 66 e9` against `43 61 66 65 301`. Real pair, not a tautology.
- **Existing ids are unchanged.** NFC is idempotent on already-composed text, so `one-dark-pro` and `ñandú` are untouched. The two moved fixture expectations (`cafe\u{301}` to `caf\u{e9}`, `u\u{308}ber` to `\u{fc}ber`) are the intended consequence and the PR states them as such rather than quietly editing them.
- **`unicode-normalization 0.1.25`** pulls only `tinyvec`, which the lockfile already carried. The `Cargo.toml` comment follows the convention of the entries around it.
- **The release note** describes this version's change and replaces rather than appends.

### Scores

- **CodeRabbit score (diff quality) — 3/5.** The fix is correct and the reasoning in the comment is right, but the half of it that the comment works hardest to justify has no test, and I only found that by mutation rather than by reading. Plus the fmt failure, which is mechanical but red.
- **Greptile score (codebase fit) — 4/5.** Placement, the dependency comment convention, the version bump and the notes replacement all match the repo. Docked one for the unrelated reformat, which puts two hunks in the diff that have nothing to do with #407.

VERDICT: REQUEST_CHANGES

Two items: run `cargo fmt` with the pinned toolchain, and add the `H\u{331}` assertion. Happy to re-check both once they land.

### One thing worth knowing about the merge order

#410 replaces `src/release_notes.rs`'s const with one file per version (`src/release_notes/<version>.md`). If it lands first, this branch's `RELEASE_NOTES` edit becomes a conflict and the note moves to `src/release_notes/0.1.809.md`. I am holding #410 until the branches that still use the old layout are in.

vitali87:
Both addressed, head is now 694f6fb1f9728fe93731652a01fa9fdd129a5a39.

- **fmt blocker**: fixed in 2a972a5 (my hand-run rustfmt used edition 2021; the pinned check is edition 2024). The two unrelated hunks are reverted by that commit.
- **Unexercised trailing NFC**: added your assertion to `canonically_equivalent_names_share_an_id` (`slug("H\u{331}ana") == slug("\u{1e96}ana")`, and `== "ẖana"`), and replaced the comment's İ example, which you showed never reaches that branch, with the H + U+0331 one. Mutation control re-run here: with the trailing `.nfc()` removed the module fails on exactly that line (`left: "h\u{331}ana", right: "ẖana"`); restored, 28 passed.
- Merge order noted: this stays on the old release-notes layout, and moves to `src/release_notes/0.1.809.md` if #410 lands first.

vitali87:
## 🤖 Re-score at `694f6fb` — Claude Opus 5

Both round-1 items are fixed and verified, and the fix surfaced its mirror image.

**fmt.** `cargo fmt --check` is clean at this head, and both unrelated hunks are gone from the diff.

**The trailing NFC is now defended.** Under the same mutation as before:

```rust
let lowered: String = s.nfc().collect::<String>().to_lowercase(); // MUT-TRAIL
```

```
test vscode_theme::tests::canonically_equivalent_names_share_an_id ... FAILED
test result: FAILED. 27 passed; 1 failed
```

Round 1 had this at `28 passed; 0 failed`, so the assertion does the work it was added for.

### New finding — the LEADING `.nfc()` is now the unexercised one, and it looks redundant

The same mutation on the other side:

```rust
let lowered: String = s.to_lowercase().nfc().collect(); // MUT-LEAD
```

```
test result: ok. 28 passed; 0 failed
```

That is not just missing coverage. Sweeping for an input where the leading pass changes the answer, comparing `NFC(lower(s))` against `NFC(lower(NFC(s)))`:

- every code point from U+0020 to U+30000, and
- every base + combining mark pair over U+0020..U+0600, U+1E00..U+1F00 and U+2100..U+2180 against U+0300..U+0370,

gives **zero** differing inputs.

There is a reason it comes out zero rather than merely rare: Unicode requires case conversion to preserve canonical equivalence, so `lower(s)` and `lower(NFC(s))` are canonically equivalent for every `s`, and normalising afterwards collapses them to the same string. The trailing NFC therefore already does everything the leading one was there for.

The sweep is not a proof over all strings, and the stability guarantee is the part that generalises it, so I would not call this a defect. It is one extra full pass and one extra `String` allocation per id derivation, in code whose comment argues at length for both passes. Two honest options:

1. **Drop it**, and let the comment say why one pass suffices: case conversion preserves canonical equivalence, so normalising after lowercasing is enough.
2. **Keep it**, and let the comment say it is belt-and-braces against a case mapping that violates that guarantee, rather than implying an input exists that needs it.

Either is fine. What is not fine is the current state, where the comment reads as though both passes are load-bearing and only one of them is.

### Scores at `694f6fb`

- **CodeRabbit score (diff quality) — 4/5.** The fix is correct, the fmt noise is gone, and the load-bearing half is now pinned by a test that fails without it. Held off 5 for the leading pass: unexercised, and on the evidence unnecessary.
- **Greptile score (codebase fit) — 5/5.** Dependency comment convention, version bump, notes replacement and test placement all match the repo, and `unicode-normalization` adds only `tinyvec`, which the lockfile already carried.

VERDICT: APPROVE

The leading-NFC item is a comment or a one-line deletion, not a blocker. It does not need another round from me either way.

vitali87:
Leading `.nfc()` dropped at 2f7a344e1460ccb8000e86806c0ab1868fd107e2, comment reworded to say the trailing pass is the load-bearing one and why the leading one added nothing (case mapping preserves canonical equivalence; your sweep). Mutation control re-run: with the remaining pass removed the module fails on the NFD fixture (`left: "cafe\u{301}-dark", right: "café-dark"`); restored, 28 passed. fmt clean.

vitali87:
## 🤖 Re-score at `2f7a344` — Claude Opus 5

The leading pass is gone and the remaining one is fully defended.

Removing the normalisation entirely:

```rust
let lowered: String = s.to_lowercase();   // MUTATION
```

```
test vscode_theme::tests::a_name_with_no_letter_or_digit_is_refused ... FAILED
test vscode_theme::tests::a_theme_named_in_a_non_latin_script_can_be_imported ... FAILED
test vscode_theme::tests::canonically_equivalent_names_share_an_id ... FAILED
test result: FAILED. 25 passed; 3 failed
```

Three tests, not one, which is the right shape: the normalisation is load-bearing for the id, for the non-Latin path, and for the empty-slug guard. Compare with where this PR started, where dropping half of it changed nothing at all.

The `H` + U+0331 assertion is the one that makes the single pass sufficient rather than merely shorter, since it is the case where lowercasing exposes a composition the uppercase spelling had no precomposed character for.

### Scores at `2f7a344`

- **CodeRabbit score (diff quality) — 5/5.** Every line of the fix is now covered by a test that fails without it, and the code is one pass rather than two with a comment arguing for both.
- **Greptile score (codebase fit) — 5/5.** Unchanged from the last round: dependency comment convention, version bump, notes replacement and test placement all match, and `unicode-normalization` adds only `tinyvec`, which the lockfile already carried.

VERDICT: APPROVE

Nothing outstanding from my side. Note for whoever merges: this branch still carries a `src/release_notes.rs` edit, so if #410 lands first the note moves to `src/release_notes/0.1.809.md`.

PR #415, opened 2026-08-30
fix(log): paint and hit-test the rendered log by display width

### LINKED ISSUE #404: Rendered log view: double-width characters break column alignment
## What happens

`render_log` (`src/widgets/editor.rs`) advances the paint cursor by character count rather than display width:

```rust
buf.set_stringn(x, y, text, room, style);
x = x.saturating_add(text.chars().count().min(room) as u16);
```

A double-width character (CJK, most emoji) occupies two terminal cells but counts as one here, so every span after it on that line is painted one cell to the left of where it belongs, and the drift accumulates with each wide character. A log line of Chinese text with a colour change mid-line paints the colour boundary in the wrong place.

## Why it matters beyond cosmetics

Anything that maps screen cells back to text inherits the same arithmetic. Mouse selection in the log view (#402) maps a clicked cell to a character column by straight subtraction, deliberately matching the renderer: a mapping that measured width correctly would disagree with what is actually on screen, which is worse than agreeing with it wrongly. So the renderer is the place to fix this, and the consumers follow.

`path:line` click-through, still unwired on #257, will inherit it too.

## Suggested direction

The editor's text path already has `visible_text_width` and friends for this class of problem; the log renderer was written as a simpler paint loop and did not pick them up. The fix is to advance by display width and to have `log_cell_at` map cells back through the same measure, changed together so the two never disagree.

Worth checking whether the same assumption exists in the other windowed viewers (`src/hex.rs` is fixed-width by construction, so probably not; `src/archive.rs` lists member names, which can be CJK).

## How it was found

Reviewing my own #402 before its review round, by asking what the column arithmetic assumes rather than whether it works on the fixtures. It is not a regression from that PR: the assumption predates it, in #257's original renderer.

vitali87:
Closes #404.

## The defect

`render_log` advanced its paint cursor by `chars().count()`, so after a double-width character (CJK, most emoji) every later span landed one cell to the left of where the terminal had put the text, over the wide character's second half, and the drift grew with each one. The selection band and the find highlight painted by character index the same way, and `log_cell_at` matched the drift on purpose so a click agreed with the wrong picture.

## The fix

A `CellMap` (`src/cell_map.rs`) gives each character's first cell and width, computed the way ratatui's `Buffer::set_stringn` paints: by grapheme, dropping any grapheme with a control character and any of zero width, advancing by `unicode-width`. The painter places each span at its start byte's cell, the selection and find bands restyle exactly the cells a character occupies, and `log_cell_at` maps a cell back through the same map, so either half of a wide character is a click on it. Changed together so the two never disagree.

The find-highlight styles move into `search_highlight_styles`, shared by the editor's highlighter and the log's.

The archive viewer paints each row with one `set_stringn` from the left edge, so it has no per-span advance to drift; the hex view is fixed-width by construction.

## The test is a control

`double_width_characters_keep_log_columns_aligned` opens a log whose first line is a red `中文`, a space, and a green `X`, clicks cell 5, and reads the painted buffer. On the parent it fails for the right reason: cell 5 is empty because the X was painted on cell 3, over `文`:

```
assertion `left == right` failed: cell 5 is the X, two cells past the wide pair
  left: ""
 right: "X"
```

`cell_map::tests` pins the map itself: ASCII, a wide pair (both halves map back to it), a combining mark (a second character with no cell of its own), and a tab (ratatui drops it, so the next character lands one cell earlier).

## Gate

- Test run over `cell_map::`, the new test, and the existing log tests: 53 passed.
- `scripts/check_doc_ownership.py origin/main HEAD`: no documentation lost across 6 changed Rust files.
- Version 0.1.810 (main is 0.1.807; #408/#410/#411 claim 0.1.808, #413 claims 0.1.809). Release notes replaced.
- Does not touch `src/log_view.rs`, so it is independent of #408.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed rendered log alignment for CJK, emoji, and other wide characters.
  - Improved mouse hit-testing, text selection, color boundaries, and search highlights for variable-width characters.
  - Prevented clipped wide characters at the screen edge from receiving incorrect selection highlighting.
- **Release**
  - Updated the package to version 0.1.813.


vitali87:

Requested by the author. CodeRabbit is rate limited repo-wide, and `gh api repos/vitali87/croft/pulls/415/comments` returns `0`. Reviewed at `1d067b5` against a built tree.

The design is the right one: one `CellMap` per line, built from the same rules `Buffer::set_stringn` uses, shared by the painter, the selection band, the find highlight and `log_cell_at`, so the four cannot drift. Reading `set_stringn` in `ratatui-core-0.1.0/src/buffer/buffer.rs:335-368` against `CellMap::new` confirms the rules match: walk graphemes with `true`, drop any containing a control character, drop zero-width ones, advance by `unicode-width`.

The test is live. Reverting the geometry to the old behaviour by making every grapheme one cell wide:

```rust
u16::try_from(grapheme.chars().count()).unwrap_or(u16::MAX)   // MUTATION
```

```
assertion `left == right` failed: cell 5 is the X, two cells past the wide pair
  left: ""
  right: "X"
```

Three findings.

### Major — `paint_log_cells` and `set_stringn` disagree at the right edge, which is the exact class this PR exists to close

`set_stringn` drops a grapheme that does not fit **whole**:

```rust
.map_while(|(symbol, width)| {
    remaining_width = remaining_width.checked_sub(width)?;
    Some((symbol, width))
})
```

A wide character starting on the body's last cell is therefore not painted at all, and the cell keeps whatever was there.

`paint_log_cells` does something different:

```rust
for x in first..first.saturating_add(cells.width_of_char(col)) {
    if x >= right { return false; }
    let cell = &mut buf[(x, y)];
    cell.set_style(restyle(cell.style()));
}
```

It styles the first half **before** discovering the second does not fit. So for a line whose wide character begins on the final cell of the body, a selection or a find highlight lands a coloured cell where the text painter deliberately put nothing. Narrow, but it is a painter/annotator disagreement in the one direction the PR is closing everywhere else, and it will read as a stray band rather than as a clipped character.

The fix is to check the whole character up front:

```rust
let width = cells.width_of_char(col);
let first = left.saturating_add(cells.cell_of_char(col));
if first >= right { return false; }
if first.saturating_add(width) > right { return false; }   // clipped, as set_stringn drops it
```

Worth a test at the boundary: a body whose width puts a CJK character's first half on the last cell.

### Minor — a `CellMap` per line per frame, allocating by BYTE length

`CellMap::new` runs for every visible row on every frame, and:

```rust
let mut chars = Vec::with_capacity(text.len());
```

is `text.len()` in **bytes**, holding 16-byte tuples. A 500-character ASCII line reserves 8 KB; the CJK lines this PR is about reserve three times what they need, since every character is three bytes. Across a 50-row viewport that is on the order of hundreds of kilobytes of allocation per frame, on the redraw path.

This module already solved exactly this problem next door: `ansi_text::parse_into` exists so "a bulk scan reuses one allocation instead of building a String and a span Vec per line" (`src/ansi_text.rs:79`, and the architecture entry says so explicitly). A `CellMap::build_into(&mut self, text)` that `clear()`s and refills a map hoisted out of the row loop would follow that precedent and cost nothing in clarity. At minimum, `with_capacity(text.chars().count())` rather than `text.len()`.

`log_cell_at` building a map per mouse event is fine and needs no change.

### Minor — the find-highlight segment walk is now implemented twice

`render_log` grew its own copy of the loop that `paint_search_highlight` already runs: split on the needle, count columns, compare against `active_on_line` to pick the active style. Only the two `Style` values were extracted into `search_highlight_styles`. The active-match test (`c == abs_col && l == chunk_cols`) is now in two places and has to stay in step by hand.

`paint_search_highlight` could take an optional `&CellMap` and paint through `paint_log_cells` when given one, which would leave one implementation of the walk. The repo already argues this case for the escape-stripping rules: "There is ONE implementation of the stripping rules on purpose - a second 'just strip the escapes' copy would drift from this one" (`docs/ARCHITECTURE.md`, the `ansi_text.rs` entry). The same reasoning applies to the thing that decides which match is the active one.

### Checked and sound

- **No new dependency weight.** `unicode-width` and `unicode-segmentation` were already resolved through ratatui; the `Cargo.lock` diff adds them only to `croft-software`'s list, with no new package entries.
- **`cell_of_byte`** is a `partition_point` over a Vec sorted by byte offset. Correct, and the right shape for span boundaries.
- **Combining marks.** A non-first character in a cluster gets width 0 and is skipped by `char_at_cell`'s `width > 0` filter, so it can never be the answer to a click while still being counted by the selection arithmetic. That is the distinction the module docs promise and the test pins it.
- **The doc comment on `log_cell_at`** was rewritten rather than left describing the old deliberate drift, which is the failure mode I have been filing against this repo all day.

### Scores

- **CodeRabbit score (diff quality) — 3/5.** The core is right and well tested, but the right-edge case reintroduces a painter/annotator disagreement, and the per-frame allocation goes against a pattern this exact module established for this exact reason.
- **Greptile score (codebase fit) — 4/5.** New module, `main.rs` registration, dependency comments, version bump and notes replacement all match the repo. Docked one for the duplicated segment walk, which the architecture doc argues against by name.

VERDICT: REQUEST_CHANGES

The right-edge case is the one I would not merge without: it is the same class of defect the PR fixes, at a boundary. The other two are worth doing here rather than later, but I would not block on them alone.

vitali87:
All three addressed at f29292a7a61ff7209398028537b1bcb8c04503bf.

- **Right edge (major)**: `paint_log_cells` now checks the whole character fits before styling any cell and returns false when it is clipped, matching `set_stringn`. New test `a_wide_character_clipped_at_the_right_edge_gets_no_selection_band` learns the body width from a probe render, puts a CJK character on the last cell, and asserts the cell is blank and unselected while the one before it is selected.
- **Per-frame allocation (minor)**: `CellMap::build_into` refills one map hoisted out of the row loop; `new` delegates to it. No per-row `Vec` any more.
- **Duplicated walk (minor)**: `paint_search_highlight` takes `Option<&CellMap>`; the log passes `Some` and paints through `paint_log_cells`, the editor callers pass `None`. The inline copy in `render_log` is gone.

Gate at this head: 105 tests over `cell_map::`, the log tests, and the find/search-highlight tests pass; clippy `-D warnings` and fmt clean; doc-ownership gate clean.

vitali87:
## 🤖 Re-score at `f29292a` — Claude Opus 5

All three items landed, and the major's fix is controlled.

**The right-edge case.** `paint_log_cells` now refuses a character that does not fit whole, which is what `set_stringn`'s `map_while` does. Reverting just that half of the condition:

```rust
if first >= right {   // MUTATION: back to the half-cell band
```

```
assertion `left != right` failed: so the band must not colour the cell it left blank
  left:  Some(Rgb(38, 79, 74))
  right: Some(Rgb(38, 79, 74))
```

`a_wide_character_clipped_at_the_right_edge_gets_no_selection_band` fails on exactly the disagreement it is named for. Good test: it asserts the cell is NOT the selection colour, which is the only assertion that can distinguish "the painter left it blank" from "the band coloured it anyway".

**The allocation.** `CellMap::build_into` refills in place and the map is hoisted out of the row loop in `render_log`, so a redraw costs one allocation after the first frame rather than one per visible row. The doc comment cites `ansi_text::parse_into` and the byte-versus-character sizing, which is the reasoning I would have wanted a reader to find there.

**The duplicated walk.** `paint_search_highlight` takes `Option<&CellMap>` and the log path calls it, so the segment split and the active-match test exist once. That was the item I was least sure would come out clean, and it did.

### Scores at `f29292a`

- **CodeRabbit score (diff quality) — 5/5.** The geometry matches the painter's rules including at the boundary, every rule is pinned by a test that fails without it, and the hot path allocates once per frame rather than once per row.
- **Greptile score (codebase fit) — 5/5.** The new module, its registration, the dependency comments, the version bump and the notes replacement all match the repo, and the single-implementation rule the architecture doc argues for is now honoured rather than argued against.

VERDICT: APPROVE

Nothing outstanding. Same note as on #413: this branch still carries a `src/release_notes.rs` edit, so the note moves to `src/release_notes/0.1.810.md` if #410 lands first.

PR #416, opened 2026-08-30
fix(drop): decode file:// URLs as UTF-8 bytes, not Latin-1 chars

### LINKED ISSUE #414: Drag-and-drop: file:// URLs with non-ASCII paths decode to mojibake
## What happens

`normalise_dropped_token` (`src/app/mod.rs`) decodes a dropped `file://` URL by pushing each byte as a char:

```rust
path.push((h * 16 + l) as char);   // %xx escape
path.push(bytes[i] as char);       // literal byte
```

`u8 as char` maps a byte to the Latin-1 code point of the same value, so every multi-byte UTF-8 sequence comes out one char per byte. Dropping `/tmp/café` sends `file:///tmp/caf%C3%A9`, which decodes to `/tmp/cafÃ©`; the drop then silently opens a path that does not exist. A literal non-ASCII byte in the URL (some terminals do not percent-encode) takes the second branch and mangles the same way.

This is the #396 defect class (`workspace::strip_jsonc` walked bytes and rebuilt with `as char`), in the drop path.

## Fix

Accumulate decoded bytes into a `Vec<u8>` and `String::from_utf8` (or `from_utf8_lossy`) once at the end, so a percent-encoded multi-byte sequence reassembles. There is no test for `normalise_dropped_token` today; the fix wants one with a `%C3%A9` path and a CJK one.

## How it was found

The fallback review of #411 (https://github.com/vitali87/croft/pull/411#issuecomment-5468480124), which modelled the byte-walking decode after fixing the same shape in `workspace.rs`.

vitali87:
Closes #414.

## The defect

`normalise_dropped_token` decoded a dropped `file://` URL by pushing each `%xx` byte, and each literal byte, `as char`. That maps a byte to the Latin-1 code point of the same value, so every multi-byte UTF-8 sequence came out one char per byte: dropping `/tmp/café` sent `file:///tmp/caf%C3%A9`, which decoded to `/tmp/cafÃ©`, and the drop silently opened a path that does not exist. The #396 defect class, in the drop path.

## The fix

Decode into a `Vec<u8>` and build the string once at the end with `from_utf8_lossy`, so a percent-encoded multi-byte sequence reassembles. A sequence that is not valid UTF-8 (a Latin-1 filename percent-encoded raw) has no faithful spelling either way; the replacement character keeps the rest of the path rather than dropping it.

## The test is a control

`a_dropped_file_url_with_a_non_ascii_path_decodes_to_utf8` covers a percent-encoded path with 2- and 3-byte sequences, a literal non-ASCII character in the URL, and the two ASCII shapes the decoder always handled (`%20`, single quotes). On the parent it fails with the mojibake spelled out:

```
  left: Some("/tmp/cafÃ©/ä¸\u{ad}æ\u{96}\u{87}.txt")
 right: Some("/tmp/café/中文.txt")
```

## Gate

- The new test: green after the fix, red on the parent as above.
- Version 0.1.811 (main is 0.1.807; #408/#410/#411 claim 0.1.808, #413 0.1.809, #415 0.1.810). Release notes replaced.

Found by the fallback review of #411.


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 65a76c619586daa0d54f73bc4367ff65180f714b (branch fix/drop-path-utf8) stands in.

## 1. Walkthrough

`normalise_dropped_token` decoded a dropped `file://` URL by pushing each `%xx` byte and each literal byte `as char`, which maps a byte to the Latin-1 code point of the same value, so every multi-byte UTF-8 sequence became mojibake and the drop opened a path that does not exist (#414). The fix accumulates decoded bytes into a `Vec<u8>` and builds the string once with `String::from_utf8_lossy`, so a percent-encoded sequence reassembles. A test is added, and the version/release-notes release gate is satisfied.

| File | Summary |
|---|---|
| `src/app/mod.rs` | `normalise_dropped_token`: decode accumulator changed `String` → `Vec<u8>`; the two `as char` pushes become raw byte pushes; the result is materialised via `String::from_utf8_lossy(&path).into_owned()`. Comments explain the Latin-1 trap and the lossy choice. |
| `src/app/tests.rs` | New `a_dropped_file_url_with_a_non_ascii_path_decodes_to_utf8`: `%C3%A9` + CJK, a literal non-ASCII char, plus two ASCII regression pins. |
| `Cargo.toml` / `Cargo.lock` | Version `0.1.807` → `0.1.811`. |
| `src/release_notes.rs` | Single note replaced: `Feature` (theme-import) → `Fix` describing this decode fix. |

I modelled both decoders over the full probe set. **No ASCII behaviour changed** — `%20`, `%zz`, `%2`, trailing `%`, `%` as last/second-last byte, upper/lowercase hex, `+`, `?`/`#`, `file://localhost/…`, `file:///C:/x`, `\ ` escapes, and quote stripping are byte-identical between old and new. The only diffs are the intended ones (`%C3%A9`, 3-/4-byte, literal non-ASCII) plus the invalid-sequence cases below. I also checked the `i + 2 < bytes.len()` bound: for an escape at `i` we need `i+2 <= len-1`, which is exactly `i+2 < len` — correct, not an off-by-one.

## 2. Findings

- **`src/workspace.rs:285`** **[major/correctness]** — The identical byte-as-char bug is still live here, and this PR's own test doc-comment calls #414 "the #396 defect class" as though #396 were fixed. It is not fixed at source. `strip_jsonc` does `let c = bytes[i] as char;` (line 285) and `out.push(bytes[i + 1] as char);` (line 289) over `src.as_bytes()`. Verified by modelling: input `{"folders":[{"path":"/tmp/café"}]}` → output `{"folders":[{"path":"/tmp/cafÃ©"}]}`. This is reached from live code — `parse_workspace_file` (`src/workspace.rs:245`) calls it on every `.code-workspace` file, so a workspace naming a non-ASCII folder resolves to a path that does not exist, the exact user-visible symptom of #414. The repo already *knows*: `src/import_vscode.rs:328` and `src/vscode_theme.rs:233` both carry comments routing around it ("The `workspace.rs` one walks bytes and mangles any non-ASCII value (#396)"). Two call sites were given a workaround; the defective function was left in place with a third caller still on it. Fix — make it walk chars, matching the sibling that the workarounds already prefer:
  ```rust
  -    let bytes = src.as_bytes();
  ...
  -        let c = bytes[i] as char;
  ```
  Simplest correct form: delete `workspace.rs::strip_jsonc` and have `parse_workspace_file` call `crate::tasks::strip_jsonc(&raw)`, then drop the now-stale warning comments at `import_vscode.rs:328` and `vscode_theme.rs:233`. Out of scope for this PR is defensible, but shipping a fix whose test doc asserts the class was already handled, while the original instance is still reachable, is not — at minimum file it and correct the doc-comment wording.

- **`src/app/mod.rs:40966`** **[minor/correctness]** — The inline decoder duplicates `src/shell_integration.rs:160`, `pub(crate) fn percent_decode(bytes: &[u8]) -> Vec<u8>`, which after this change is byte-for-byte the same algorithm — same `i + 2 < bytes.len()` bound, same hex pair, same fallthrough push — and already returns `Vec<u8>` correctly. `src/file_ref.rs:174` consumes it as `String::from_utf8(percent_decode(...)).ok()?`. This PR reimplements it rather than reusing it, so the next percent-decoding bug has to be fixed in two places. Fix: replace the loop body with `let decoded = crate::shell_integration::percent_decode(rest.as_bytes()); s = String::from_utf8_lossy(&decoded).into_owned();`. Note `shell_integration` is declared in `src/main.rs:58` and `percent_decode` is `pub(crate)`, so it is already reachable from `src/app/mod.rs`; the local `hex_digit` (`src/app/mod.rs:41013`) would then have no caller in this function.

- **`src/app/mod.rs:40971`** **[minor/correctness]** — `from_utf8_lossy` is the wrong call for an *invalid* sequence, and the function's own doc (`src/app/mod.rs:40936-40939`, "Anything that isn't a plausible filesystem path returns None and is dropped by the caller") says so. Verified: `file:///tmp/%FF` → `/tmp/\u{FFFD}` and `file:///tmp/%C3` → `/tmp/\u{FFFD}`, both `is_absolute() == true`, so both survive the `is_pure_path_payload` gate at `src/app/mod.rs:40834` and the `parse_foreign_dropped_paths` filter at `src/app/mod.rs:40885`. U+FFFD is a legal filename character, so the synthesised path is not merely wrong, it is *plausible* — it defeats the only screen either caller applies. The consequences differ per call site and both are bad: `is_pure_path_payload` (used at `src/app/mod.rs:26205` to decide whether a paste is a drag) returns `true`, hijacking what may be ordinary pasted text into a drop; `parse_foreign_dropped_paths` hands the replacement-char path to the relay to `scp`. The inline comment's rationale — "a Latin-1 filename percent-encoded raw ... the replacement character keeps the rest of the path rather than dropping it" — does not hold up: the replacement is lossy and irreversible, so the "kept" path can never open the intended file; it can only silently open or fetch the wrong one, or nothing. A drag source on macOS/Linux (Finder, iTerm2, GTK) percent-encodes the raw bytes of the OS path, and on macOS those are always UTF-8, so an invalid sequence means the token is not a path this process can name. Refusing is both the documented contract and the safer default. Fix:
  ```rust
  -        s = String::from_utf8_lossy(&path).into_owned();
  +        // A sequence that is not valid UTF-8 names no path this process can
  +        // open; per this function's contract, drop the token rather than
  +        // inventing a plausible-looking path out of replacement characters.
  +        s = String::from_utf8(path).ok()?;
  ```

## 3. Test coverage

Behaviour this diff changed or touches that no test exercises:

- **Invalid UTF-8 in a `file://` URL.** The single most behaviour-changing case (`%FF`: `/tmp/ÿ` before, `/tmp/\u{FFFD}` now) has no assertion, so whichever policy the finding above settles on is unpinned. Write `a_dropped_file_url_with_invalid_utf8_is_dropped`: assert `normalise_dropped_token("file:///tmp/%FF")` is `None` (or, if lossy is kept deliberately, assert the replacement char explicitly so the choice is recorded), plus `"file:///tmp/%C3"` for a truncated multi-byte lead.
- **The `%`-escape edge shapes.** `%zz`, `%2`, a trailing `%`, and `%` as the second-last byte all pass through literally, and all four run through the rewritten bounds/hex logic. None is asserted. Add them to the existing test as a group — they are the cases a refactor of the `i + 2 < bytes.len()` bound would silently break.
- **`+` must not become a space.** `file:///tmp/a+b` → `/tmp/a+b` is correct (form-encoding rules do not apply to file URLs) but unpinned; a future "use a URL crate" change could regress it invisibly.
- **`workspace.rs::strip_jsonc` non-ASCII.** No test covers it (`grep` for `strip_jsonc` finds tests only for the `tasks.rs` sibling, at `src/tasks.rs:727`). Write `a_workspace_file_with_a_non_ascii_folder_resolves_to_that_folder`, parsing a `.code-workspace` whose `folders[].path` is `/tmp/café`; it fails today.

On whether the new test is a real control: **yes for the two non-ASCII assertions.** Modelled against the old decoder, `"file:///tmp/caf%C3%A9/%E4%B8%AD%E6%96%87.txt"` produced `/tmp/cafÃ©/ä¸\u{ad}æ\u{96}\u{87}.txt` and `"file:///tmp/café"` produced `/tmp/cafÃ©` — both differ from the asserted value, confirming the PR's `cafÃ©` claim. The two ASCII assertions pass on the old code and are regression pins, not controls, which is the right role for them; the comment labels them accurately. The fixture is well-specified for what it covers (`\u{e9}` escapes rather than literals keep it robust against source-encoding surprises) but under-specified on the edges listed above.

## 4. Nitpicks

- `src/app/mod.rs:40946-40951` — the comment explains the old bug in the past tense at length. Once merged, the "why it is a `Vec<u8>`" is the durable half; the `cafÃ©` archaeology is what the test doc-comment and #414 are for.
- `src/app/tests.rs:35238` — "The #396 defect class, in the drop path" reads as though #396 closed the class. Per the first finding it did not; "the same defect class as #396" avoids asserting a fix that has not landed.

## 5. Scores

- **CodeRabbit score (diff quality) — 4/5**: no defect inside the changed lines — the decode is correct, the bound is right, and no ASCII behaviour moved (verified across ~26 probes); the two minors (duplicated decoder, `from_utf8_lossy` vs the documented `None` contract) are shape issues, not breakage.
- **Greptile score (codebase fit) — 3/5**: exactly one major, all of whose evidence sits outside the diff — `workspace.rs:285` still carries the same bug on a live path, with two other files already commented as routing around it, while this PR's test doc treats the class as handled.

VERDICT: REQUEST_CHANGES

vitali87:
Addressed at c81e0cd96e1ebe9d78a36b832fb4b3476bb99804.

- **Duplicated decoder (minor)**: the loop is gone; `normalise_dropped_token` now calls `shell_integration::percent_decode` and the local `hex_digit` is removed.
- **`from_utf8_lossy` (minor)**: replaced with `String::from_utf8(decoded).ok()?`, per the function's documented contract. New test `a_dropped_file_url_with_invalid_utf8_is_dropped` pins `%FF` and a truncated `%C3` to `None`; the existing test gains the `+`, `%zz`, `%2`, and trailing-`%` pass-through shapes.
- **`workspace.rs::strip_jsonc` (major)**: that is exactly what PR #411 fixes (it deletes the byte-walking copy and routes `parse_workspace_file` through `tasks::strip_jsonc`, with a non-ASCII control test); it is reviewed and ahead of this PR in the merge order, so the fix is not duplicated here. The test doc-comment now says "the same defect class as #396" rather than implying #396 had already landed.

Gate at this head: the two drop tests pass, clippy `-D warnings` clean, fmt clean.

vitali87:

# Re-review: PR #416 @ `c81e0cd`

Head confirmed: `c81e0cd fix(drop): reuse percent_decode and drop a token that is not UTF-8`, on top of `65a76c6`.

Both minors from my previous review are fixed as recommended, and the deferred major is genuinely covered elsewhere.

| # | Prior finding | Fix at `c81e0cd` | Verification |
|---|---|---|---|
| 1 | **major** `workspace.rs:285` — same byte-as-char bug live in `strip_jsonc` | Deliberately not in this diff; owned by PR #411 | Confirmed real, not a deferral-in-name: `gh pr diff 411` **deletes** `workspace.rs::strip_jsonc` (removing `bytes[i] as char` at old lines 285/289/322) and routes `parse_workspace_file` to the char-walking `crate::tasks::strip_jsonc` — exactly the fix I proposed. It also adds `strip_jsonc_preserves_non_ascii` and rewrites the stale warning comments in `import_vscode.rs` and `vscode_theme.rs`. PR #411 is OPEN on `fix/workspace-jsonc-chars`. **Does not block #416** — see below. |
| 2 | **minor** `mod.rs:40966` — decoder duplicated `shell_integration::percent_decode` | `let decoded = crate::shell_integration::percent_decode(rest.as_bytes());`, local `hex_digit` deleted | **Behaviourally identical over the whole probe set: 0 divergences / 28 inputs.** The one substitution that could have bitten is the hex predicate — `percent_decode` uses `(b as char).to_digit(16)` where the removed `hex_digit` used explicit `b'0'..=b'9' \| b'a'..=b'f' \| b'A'..=b'F'` ranges. I enumerated **all 256 byte values**: the two agree on every one (`to_digit(16)` rejects `g`–`z`/`G`–`Z` via its `< 16` check and rejects `0x80..=0xFF` because Latin-1 chars are not ASCII alphanumerics). The `i + 2 < bytes.len()` bound is character-for-character the same in both. `hex_digit` has no callers left in `src/` (the surviving `type_hex_digit` hits are the unrelated hex editor in `src/hex.rs`). `crate::shell_integration::percent_decode` resolves: `mod shell_integration` at `src/main.rs:58`, `pub(crate)`, and `src/file_ref.rs:174` already calls it by that exact path. |
| 3 | **minor** `mod.rs:40971` — `from_utf8_lossy` contradicted the doc contract and defeated both callers' path screens | `s = String::from_utf8(decoded).ok()?;` | `file:///tmp/%FF` → `None`, `file:///tmp/%C3` → `None`, `file:///tmp/%C3%A9%FF` → `None`. The token no longer reaches `is_pure_path_payload` (`mod.rs:40834`) or `parse_foreign_dropped_paths` (`mod.rs:40885`), so it can neither hijack a paste into a drop nor be handed to the relay to `scp`. Now matches the doc at `mod.rs:40936-40939`. The `?` sits after quote-stripping and before backslash-unescaping, so no other branch is skipped. |
| 4 | **nit** comment over-long on past-tense archaeology | Rewritten; now leads with the invariant and the reason for dropping | Read at `mod.rs:40946-40955`. |
| 5 | **nit** test doc asserted #396 had closed the class | Now "The same defect class as #396, in the drop path" | Read at `tests.rs:35232-35235`. |

**No ASCII behaviour changed** across both heads. Re-ran the full probe set against a model of the new `percent_decode` path: `%20`, `%zz`, `%2`, trailing `%`, `%` as last/second-last byte, upper/lowercase hex, `+`, `?`/`#`, `file://localhost/…` (host component still not stripped — unchanged from `origin/main`, and it fails the `is_absolute()` gate as it always did), `file:///C:/x`, `\ ` escapes, quote stripping, and bare `file://` are all byte-identical to `origin/main`.

## Test coverage

My previous gaps are closed. All ten assertions across both tests verified to hold against the new decoder:

- `a_dropped_file_url_with_invalid_utf8_is_dropped` is a **real red control**: under `origin/main` `%FF` → `/tmp/ÿ` and `%C3` → `/tmp/Ã`, both `Some(..)`, so both assertions fail on the old code for the right reason.
- The `+`, `%zz`, `%2`, trailing-`%` shapes are now pinned in a table-driven loop with `"{url}"` as the failure message — these are regression pins (green on old code), correctly labelled as such by the comment.
- The two non-ASCII assertions remain red controls (`origin/main` yields `cafÃ©`).

One residual gap, a nit rather than a hole: no test pins that `crate::shell_integration::percent_decode` and this call site stay in agreement — if someone later "fixes" `percent_decode`'s bound for OSC 7, the drop path silently inherits it. The existing tests would catch a behaviour change, so this is covered in effect.

## Does the deferred major block this PR?

**No.** Scoping it out is correct rather than convenient: the two changes touch disjoint files (#416 → `src/app/`, #411 → `src/workspace.rs`, `src/tasks.rs`, `src/import_vscode.rs`, `src/vscode_theme.rs`), neither depends on the other, and #411 is reviewed and ahead in the merge order. My original objection was narrower than the fix — I objected to shipping a test doc-comment that *asserted the class was already handled* while an instance was still reachable. That wording is now corrected, so the objection is answered within this diff.

One merge-order note, not a finding against the diff: **#411 bumps to `0.1.808` and #416 bumps to `0.1.811`, both from `origin/main`'s `0.1.807`.** They do not collide, and #416 merging second stays above main. But per CLAUDE.md the release gate cannot catch a version collision itself — re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks` before merging, since main will have moved by then.

## Release gate

- `Cargo.toml` / `Cargo.lock`: `0.1.807` → `0.1.811`; `origin/main` re-fetched and still at `0.1.807`. Above main. Clean.
- `src/release_notes.rs`: single note replaced, `NoteKind::Feature` → `NoteKind::Fix`, summary describes this change. Clean.
- Diff touches `src/` beyond `src/app/tests.rs`, so the gate applies and both halves are satisfied.

## Findings

None. No blocker, no major, no minor, no nit against this diff.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: both prior minors fixed inside the changed lines; the `percent_decode` substitution is behaviour-preserving over the full probe set and over all 256 bytes for the hex predicate; the invalid-UTF-8 policy now matches the function's documented contract.
- **Greptile score (codebase fit) — 5/5**: the change now reuses the crate's existing decoder instead of a third copy, deletes the dead local helper, and the one out-of-diff major is fully covered by PR #411, which lands first and applies the fix at source.

VERDICT: APPROVE

PR #417, opened 2026-08-30
feat(import-vscode): bring the colour theme across

### LINKED ISSUE #351: croft import-vscode: bring settings, keybindings, and snippets across

vitali87:
Closes #351.

## What was left of #351

`croft import-vscode` already carries settings, keybindings, and snippets across (#400). The design's third step, the colour theme, was not done: `workbench.colorTheme` came out under "croft has no equivalent for", which is the wrong answer for the one setting the user sees all day when the theme is sitting in their extensions directory.

## What this does

When `workbench.colorTheme` names a theme an installed extension contributes (`<extensions>/<publisher.name-version>/package.json` → `contributes.themes[]`, matched on `id` or `label`), the import converts that theme's JSON through the same path as `croft theme-import`, installs the manifest under croft's extensions directory when applied, and sets croft's `theme` to the converted id alongside the other settings. The theme is installed before `config.json` is written, so a config never names a theme whose manifest is missing. Conversion notes (colours croft had to derive) are printed with the report, as `theme-import` prints them.

Each product keeps its extensions beside its own dot-directory (`~/.vscode`, `~/.vscode-insiders`, `~/.vscode-oss`, `~/.cursor`, `~/.windsurf`), so the directory is chosen from the product that owns the user directory being imported; a `--from` pointing anywhere probes all five, most standard first.

VS Code's built-in themes (Dark+, Light Modern, …) live inside the app bundle, not the extensions directory, so they are not found: the key stays listed as unmapped and a warning names the theme and the by-hand route (`croft theme-import <theme.json>`).

## Not done, on purpose

The issue's example settings list (tab size, word wrap, minimap, `files.exclude`, …) has no croft-side keys to land on: croft's user config is the fixed set in `prefs.rs` / `WORKSPACE_ALLOWED_KEYS`. The table can only grow when croft grows the setting, so those stay honestly reported as unmapped rather than mapped to something that sounds similar.

## Tests

Three new tests in `import_vscode::tests`, all red on the parent (the API did not exist):

- `an_installed_vscode_theme_is_converted_and_selected` — a fixture extensions directory with one theme; the scan resolves it, `settings["theme"]` follows the converted id, `workbench.colorTheme` leaves the unmapped list; `apply_into_dirs` writes `theme-fixture-dark/extension.toml` and `config.json` with `theme`.
- `a_built_in_or_absent_theme_stays_unmapped_with_the_reason` — `Dark+` with no matching extension: no theme, no `theme` key, still unmapped, and the warning names both the theme and `theme-import`.
- `extension_dirs_follow_the_product_of_the_user_directory` — the five product mappings and the probe-all fallback.

## Gate

- `import_vscode::` tests: 22 passed.
- Version 0.1.812 (main is 0.1.807; #408/#410/#411 claim 0.1.808, #413 0.1.809, #415 0.1.810, #416 0.1.811). Release notes replaced.


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 16a49c6 stands in.

## 1. Walkthrough

This change closes the last step of #351 by resolving `workbench.colorTheme` into a real croft theme instead of reporting it as unmapped. It locates the product's extensions directory from the folder holding `User`, scans `<dir>/*/package.json` `contributes.themes[]` for a matching `id` or `label`, converts the theme JSON through the existing `vscode_theme::convert_file`, and installs the manifest before `config.json` is written so a config never names a missing theme. The core design is sound and the ordering decision is right; the problems are in a hermeticity leak into the test suite, an unreported conflict case, and version-ordering when one extension has several versions installed.

| File | Change |
|---|---|
| `src/import_vscode.rs` | `EXTENSION_DIRS`, `theme_extension_dirs_under`/`theme_extension_dirs`, `find_installed_theme`, `ThemeSource`, `Report.theme`, `resolve_color_theme`, `scan_profile_with_extensions`, `apply_into_dirs`; `apply_into` becomes `#[cfg(test)]`. Three new tests. |
| `src/cli.rs` | Prints the resolved theme line and the conversion notes with the report. |
| `src/release_notes.rs` | Release note replaced for this feature. |
| `Cargo.toml` / `Cargo.lock` | 0.1.807 → 0.1.812. |

## 2. Findings

**`src/vscode_theme.rs:552` [major/test-isolation] — the new scan path drags the developer's real `~/.config/croft` into the test suite, and this PR is what makes it reachable from `import_vscode` tests.**

Evidence: `resolve_color_theme` calls `convert_file` → `convert` → `existing_ids()` (`src/vscode_theme.rs:531`). `Theme::all()` is *deliberately* hermetic under test — `src/theme.rs:181` reads `let disabled = if cfg!(test) { BTreeSet::new() } else { ... read_extension_sources(user_extensions_dir()) ... }`, with a comment stating the registry "must be hermetic" because "reading the developer's real ~/.config/croft ... made this process-wide cache depend on machine state". But `existing_ids()` then filters through `was_generated_by_import(id)` (`src/vscode_theme.rs:552`), which has **no** `cfg!(test)` guard:

```rust
fn was_generated_by_import(id: &str) -> bool {
    let path = crate::lsp::manifest::user_extensions_dir()
        .join(format!("theme-{id}"))
        .join("extension.toml");
    std::fs::read_to_string(path)...
```

`user_extensions_dir()` is `prefs::config_dir().join("extensions")`, and `config_dir()` (`src/prefs.rs:542`) resolves `XDG_CONFIG_HOME` or `$HOME/.config/croft` — the real one. So `an_installed_vscode_theme_is_converted_and_selected` reads the developer's home directory, contradicting the module's own stated posture and the docstring on `apply_into` ("Nothing here touches the environment"). It is a read, not a write, so it will not corrupt anything, and today it is benign because the bundled ids cannot collide with `fixture-dark`. But it is exactly the machine-state dependency the sibling guard exists to prevent, and it makes a future collision test non-deterministic.

Fix: guard `was_generated_by_import` the same way `Theme::all()` is guarded (`if cfg!(test) { return false; }`), or thread the extensions dir through `convert_file` so the scan reads the same directory `install_into` writes to.

**`src/import_vscode.rs:775` [major/correctness] — when the user's `config.json` already has `theme`, the import installs the manifest but silently keeps the old theme; the conflict is listed generically and the report claims a theme was applied.**

Evidence: `apply_into_dirs` installs unconditionally when `report.theme.is_some()`, before the merge loop. The merge loop (`for (key, value) in &report.settings`) is never-overwrite: if `existing` already has `theme`, it pushes `config.json theme: kept "x", VS Code had "fixture-dark"` and `continue`s. Two consequences, both user-visible:

1. The manifest is installed and reported in `written`, so the CLI prints `Wrote: .../theme-fixture-dark/extension.toml` — for a theme the user is not using. Installing it is arguably fine (it becomes selectable in the picker), but nothing says so.
2. `src/cli.rs:974` prints `workbench.colorTheme "Fixture Dark" -> croft theme "fixture-dark", converted from ...` **before** `apply` runs, and the conflict list is printed after. The theme line is unconditional and reads as an accomplished fact. The user sees "your theme is now fixture-dark" and a separate generic conflict line further down, when the truth is the theme did not change.

This is the one case where the report actively misleads. Fix: after `apply_into_dirs`, when `report.theme` is set and a `theme` conflict was recorded, say so explicitly — e.g. "the converted theme was installed and is selectable, but config.json already sets `theme`, so it was left alone; run `croft theme-import` or set `theme = "fixture-dark"` to switch." Moving the theme line after `apply` (or making it conditional on no conflict) would also work.

**`src/import_vscode.rs:355` [minor/correctness] — with two versions of one extension installed, lexicographic sort picks an arbitrary version, not the newest, and the docstring overstates what the sort buys.**

Evidence: `extensions.sort()` on `Vec<PathBuf>` is bytewise. Modelled:

```
sorted: ['acme.theme-1.10.0', 'acme.theme-1.2.0', 'acme.theme-1.9.0', 'acme.theme-2.0.0']
first (wins): acme.theme-1.10.0
```

VS Code leaves old versions in place on update (it prunes lazily), so having `acme.theme-1.9.0` and `acme.theme-1.10.0` side by side is the ordinary state, not an edge case. The first match wins, so `1.10.0` beats `2.0.0` here purely by string order. The docstring says the sort makes "two versions of one extension resolve the same way on every run" — that claim is true (it is deterministic) but the comment invites the reader to believe the right one is chosen. Determinism is not correctness here.

Fix: iterate newest-last (or reverse the scan and keep the last match), parsing the trailing `-<semver>` when present; at minimum, amend the docstring to say the choice is deterministic but *arbitrary* among versions.

**`src/import_vscode.rs:328` [minor/portability] — `theme_extension_dirs` uses `HOME` only, while `discover_profiles` in the same file already handles `APPDATA`.**

Evidence: `discover_profiles` (`src/import_vscode.rs:51-60`) probes `$HOME/Library/Application Support`, `$HOME/.config`, **and** `$APPDATA`. `theme_extension_dirs` (line 328) is `match std::env::var_os("HOME") { Some(home) => ..., None => Vec::new() }`. On Windows the profile is found via `APPDATA` but the extensions dir is `%USERPROFILE%\.vscode\extensions`, and `HOME` is typically unset — so profile discovery succeeds and theme resolution returns an empty vec, silently taking the "not an installed extension theme" branch. The warning then blames a built-in theme for what is a lookup that never ran.

This is consistent with the rest of the file being HOME-centric and croft may not target Windows, but the asymmetry is new and within one file. Fix: fall back to `USERPROFILE` when `HOME` is absent, or state in the docstring that theme resolution is HOME-only by design.

**`src/import_vscode.rs:360` [minor/robustness] — `theme["path"]` is joined to the extension dir with no containment check.**

Evidence: `return Some(ext.join(rel))`. Rust's `Path::join` replaces the base when `rel` is absolute, and does not normalise `..`. Modelled:

```
'/etc/passwd'            -> /etc/passwd
'../../../../etc/passwd' -> /home/t/.vscode/extensions/acme.x-1.0.0/../../../../etc/passwd
```

I am rating this minor, not a blocker, and deliberately: the input is the user's own extensions directory on their own machine, the operation is a read that is then parsed as a theme, and a user who has a malicious VS Code extension installed has already lost. The realistic failure is a badly packaged extension, not an attack. But the tool reports the path it read, and reading outside the extensions directory on the strength of a `package.json` field is worth one cheap guard.

Fix: reject a `path` that is absolute or that escapes `ext` after normalisation, and skip to the next theme entry.

**`src/import_vscode.rs:337` [minor/correctness] — matching on both `id` and `label` is right, but a label collision across two extensions resolves silently to whichever sorts first.**

Evidence: `theme["id"].as_str() == Some(label) || theme["label"].as_str() == Some(label)`. This is correct behaviour for VS Code: `workbench.colorTheme` stores the `label` for themes with no `id`, and the `id` when the extension declares one, so matching both is the right call and the docstring says so accurately. The gap is what happens on a tie — `label` is not unique across extensions (several popular themes ship a `Dark` or `Monokai` label), and the first sorted match wins with no note in the report. Fix: when more than one extension matches, take the first but push a warning naming the alternates, so the user can tell why they got the wrong one.

## 3. Test coverage

The three new tests are real controls, not shape assertions — the PR states they were red on the parent, and the assertions are specific enough to back that up. `an_installed_vscode_theme_is_converted_and_selected` drives the genuine write path (`apply_into_dirs`) and asserts on the manifest existing on disk, its presence in `written`, and `config["theme"]` read back from the written file, so unwiring the install would fail it. `a_built_in_or_absent_theme_stays_unmapped_with_the_reason` correctly uses a *populated* fixture with a non-matching label rather than an empty directory — that is a positive control on the search, and it means "found nothing" is distinguishable from "could not search". `extension_dirs_follow_the_product_of_the_user_directory` covers all five products plus the fallback, and asserts both `all[0]` and `all.len()`.

On writes outside tempdirs: the test writes only under two `TempDir`s. But as noted above it *reads* the real `~/.config/croft/extensions` through `was_generated_by_import`, so the claim in `apply_into`'s docstring that "nothing here touches the environment" is no longer strictly true on the read side.

Untested edge cases, in rough order of how likely they are to bite:

- **The `Err` branch of `convert_file`** (line 629) — a `workbench.colorTheme` matching a theme whose JSON is malformed. This is the branch that decides whether a broken theme aborts the import or degrades to a warning, and nothing exercises it.
- **The conflict case** — an existing `config.json` with `theme` already set. This is the major finding above; a test here would have surfaced the misleading report.
- A `package.json` that is not valid JSON, and an extension directory with no `contributes.themes` — both are `continue` arms in the scan loop and both are plausible in a real extensions directory (many extensions contribute no themes at all).
- A `contributes.themes[]` entry whose `path` does not exist on disk.
- Two versions of one extension (the ordering finding).
- `apply_into_dirs` when `install_into` fails.

On that last point, to answer the ordering question directly: the `?` on `install_into` is correct. It propagates before the `config.json` block is reached, so a failed install cannot leave a config naming a missing theme. That is the right direction, it matches the stated intent in the comment, and it is the part of this change I would least want altered.

## 4. Nitpicks

- `src/import_vscode.rs:750` — `apply_into` keeps its full doc comment ("Split out so a test can drive the REAL write path...") but is now `#[cfg(test)]`, and `apply_into_dirs` below it is documented only as "[`apply_into`] with the extensions directory ... named explicitly". The substantive prose now describes a test-only wrapper while the real production entry point gets the thinner comment. Worth moving the rationale down to `apply_into_dirs`, which is where a future reader will land. Rustdoc on a `#[cfg(test)]` item is also not built for the public docs, so that paragraph effectively disappears.
- `docs/ARCHITECTURE.md:79` describes `import_vscode.rs` as converting "`settings.json`, `keybindings.json` and `snippets/*.json`" — the colour theme is now a fourth conversion and a notable one. `docs/SETTINGS.md:98` also describes the import without mentioning the theme. Neither is touched by this PR. Given how carefully that architecture line is written, a clause about the theme belongs in it.
- The `.vscode-oss` mapping is keyed on the product folder `VSCodium`; some VSCodium builds use a `VSCodium - Insiders` folder, which will fall through to the probe-all path. Harmless, and arguably the right default, but it is a sixth case the five-entry table quietly does not cover.

## 5. Scores

Release hygiene verified: `origin/main` is at 0.1.807, the branch is at 0.1.812, `Cargo.lock` is bumped in step, and `src/release_notes.rs` is replaced rather than appended — matching the "replace on every version bump" contract in the file. Per CLAUDE.md, re-read main's version in the same breath as the final `gh pr checks` before merging, since 0.1.808–0.1.811 are claimed by open PRs.

Two majors — the test-hermeticity leak through `was_generated_by_import`, and the unreported theme conflict where the CLI announces a theme change that did not happen.

**CodeRabbit score (diff quality) — 3/5**
**Greptile score (codebase fit) — 3/5**

The design is right and the install-before-config ordering is the correct call, carefully reasoned and correctly implemented. What holds it back is that both majors are places where this change steps outside conventions the surrounding code states explicitly in its own comments: `theme.rs` documents why the theme registry must be hermetic under test and this path routes around that guard, and `import_vscode.rs` documents "merge, never overwrite" with conflicts reported, while the theme's conflict is reported in a way that contradicts the line the CLI already printed.

VERDICT: REQUEST_CHANGES

vitali87:
All findings addressed at 8ffac394a646530e129b5441d17082ef6de28389.

| Finding | Fix | Pinned by |
|---|---|---|
| test reads the real `~/.config/croft` via `was_generated_by_import` (major) | `cfg!(test)` guard, same reason as `Theme::all()` | the existing install test now runs hermetic |
| theme conflict reported misleadingly (major) | conflict line says the converted theme is installed and selectable and how to switch; CLI line reads "converts to" | `an_existing_theme_setting_is_kept_and_the_conflict_says_how_to_switch` |
| byte-order version pick (minor) | `extension_order`: name asc, version desc | `the_newest_installed_version_wins_and_an_escaping_path_is_skipped` (1.10.0 > 1.9.0 > 1.2.0) |
| HOME-only (minor) | falls back to `USERPROFILE` | — |
| path containment (minor) | absolute or `..` paths skipped | same test, `Evil` rows |
| label collision (minor) | warning names the alternates | — |
| `convert_file` Err branch untested | | `a_theme_that_does_not_convert_is_a_warning_not_a_failure` |
| docs | rationale moved to `apply_into_dirs`; ARCHITECTURE.md names the theme conversion | |

Gate at this head: 25 `import_vscode::` tests pass; clippy `-D warnings`, fmt, and the doc-ownership gate clean.

vitali87:

# Re-review of PR #417 — head `8ffac39`

Verified in `/private/tmp/…/scratchpad/wt351`, confirmed at `8ffac394a646530e129b5441d17082ef6de28389` ("fix(import-vscode): keep the theme import hermetic, honest, and newest-first"), sitting on `16a49c6`. Every finding from the previous round is addressed. Reviewed by reading the incremental diff and modelling the ordering and path logic; no cargo/rustc run.

## 1. Finding → fix → verification

| # | Finding (prev. severity) | Fix at this head | Verification |
|---|---|---|---|
| 1 | **major** — `was_generated_by_import` read the developer's real `~/.config/croft` during tests, routing around the hermeticity guard `Theme::all()` states explicitly | `src/vscode_theme.rs:553` now opens with `if cfg!(test) { return false; }`, with a comment tying it to the `Theme::all()` guard | Read the diff. The guard is the first statement, before `user_extensions_dir()` is called, so no filesystem access can precede it. It returns `false`, matching the pre-existing behaviour on a missing file (`.unwrap_or(false)`), so non-test behaviour is unchanged and no bundled id is now wrongly treated as import-generated. **Fixed.** |
| 2 | **major** — an existing `theme` in `config.json` conflicted silently while the CLI announced the theme had changed | `src/import_vscode.rs:851` special-cases `key == "theme" && report.theme.is_some()`: *"kept {current}; the converted theme {value} is installed and selectable under the gear menu, set theme = {value} to switch"*. `src/cli.rs:976` changed from `-> croft theme {:?}, converted from` to `converts to croft theme {:?} (from {})` | Both halves of the complaint are answered: the conflict line now states the manifest *was* installed and how to switch, and the CLI line is future-tense ("converts to"), so it no longer asserts a change that the merge may decline. The `report.theme.is_some()` conjunct is right — a `theme` conflict arising from a config-only source keeps the generic message. **Fixed.** |
| 3 | **minor** — lexicographic sort made an arbitrary version win (`1.10.0` before `1.9.0` and `2.0.0`) | New `extension_order` at `src/import_vscode.rs:342` returns `(String, Reverse<Vec<u64>>)`; `find_installed_theme` → `find_installed_themes` returns *all* matches newest-first, and `resolve_color_theme` takes `.next()` | Modelled with correct `Reverse<Vec<u64>>` semantics (whole-vector reversed comparison, not per-element negation). `['1.9.0','1.10.0','1.2.0','2.0.0']` orders `2.0.0, 1.10.0, 1.9.0, 1.2.0` — newest first, the exact case that was wrong. Also modelled the no-version-suffix case the coordinator asked about: `acme.theme` yields `("acme.theme", Reverse([0]))` and sorts *after* every versioned sibling of the same stem, which is the safe direction (a real version beats an unparseable one). Hyphenated extension names (`pub.my-cool-theme-1.10.0`) split correctly because `rsplit_once` takes the last hyphen. **Fixed.** |
| 4 | **minor** — `HOME`-only lookup broke Windows, where `discover_profiles` already used `APPDATA` | `src/import_vscode.rs:330`: `std::env::var_os("HOME").or_else(|| std::env::var_os("USERPROFILE"))`, commented with the `discover_profiles` asymmetry | `USERPROFILE` is the correct variable — the Windows extensions dir is `%USERPROFILE%\.vscode\extensions`, not `%APPDATA%`, so this composes properly with `EXTENSION_DIRS` joining `.vscode/extensions`. `HOME` still wins where both are set, so Unix is unaffected. **Fixed.** |
| 5 | **minor** — `theme["path"]` joined unchecked; absolute paths replace the base and `..` was not normalised | `src/import_vscode.rs:387-394` computes `escapes = rel.is_absolute() \|\| rel.components().any(\|c\| matches!(c, Component::ParentDir))` and pushes only when `!escapes` | Both vectors I modelled are closed: `is_absolute()` catches `/etc/passwd` (which `Path::join` would otherwise let replace the base entirely), and the `ParentDir` component scan catches `../../../etc/passwd`. Checking components rather than substring-matching `".."` is the right call — it will not false-positive on a directory legitimately named `..foo`. Note the check is purely lexical, so a *symlink* inside the extension could still point out; that is a deliberate and reasonable line for a local-trust tool and I am not raising it. **Fixed.** |
| 6 | **minor** — a label matching several extensions resolved silently | `src/import_vscode.rs:649-658`: after taking the first match, remaining matches become `alternates` and a warning names both what was taken and what was ignored | The iterator is consumed correctly — `matches.next()` takes the winner, the rest map into `alternates`, so the taken path is never listed among the ignored. The warning fires only when `!alternates.is_empty()`, so the common single-match case stays quiet. **Fixed.** |
| nit | `apply_into`'s substantive rationale sat above a `#[cfg(test)]` item | The "Split out so a test can drive the REAL write path…" paragraph moved down to `apply_into_dirs` (now *"[`apply`] into an explicit config directory and extensions directory"*); `apply_into` keeps a two-line pointer | The prose now documents the production entry point, so it is built into rustdoc rather than dropped with the test cfg. **Fixed.** |
| nit | `docs/ARCHITECTURE.md` described only three conversions | The `import_vscode.rs` line now reads "…`snippets/*.json`, and the colour theme named by `workbench.colorTheme` (resolved from the product's extensions directory and converted through `vscode_theme.rs`)" | Matches the file's house style and cross-references the sibling module. **Fixed.** |

I also confirmed no stale references to the renamed `find_installed_theme` survive: `grep -rn "find_installed_theme\b" src/` exits 1 (no matches) while `find_installed_themes` resolves at four sites (definition, caller, two test uses) — the paired result is the positive control that the pattern was capable of matching.

## 2. Test coverage at this head

The three new tests are real controls, and each is aimed at a branch that was previously untested:

- **`a_theme_that_does_not_convert_is_a_warning_not_a_failure`** — covers the `Err` branch of `convert_file` I flagged as the highest-value gap. It overwrites the fixture theme with `{ not json` and asserts three separate things: `report.theme.is_none()`, that the *other* setting (`format_on_save`) still landed, and that the warning says "could not be converted". That middle assertion is what makes it a degradation test rather than a rejection test.
- **`an_existing_theme_setting_is_kept_and_the_conflict_says_how_to_switch`** — seeds `config.json` with `{"theme": "nord"}` and asserts the manifest was installed, `config.json` is *not* in `written`, the on-disk value is still `nord`, and the conflict contains both "installed and selectable" and "fixture-dark". It pins all four facts that the previous head got wrong, and reading the written file back means the assertion cannot pass on report state alone.
- **`the_newest_installed_version_wins_and_an_escaping_path_is_skipped`** — asserts the full ordering `["new.json", "old.json", "older.json"]` from directories `1.9.0`/`1.10.0`/`1.2.0`, which is exactly the sequence byte order would get wrong, so it would fail against the old `extensions.sort()`. The escaping half is not vacuous: without the guard, `find_installed_themes("Evil", …)` would return two joined paths and `is_empty()` would fail. The guard runs before any existence check (`find_installed_themes` has no `is_file`/`exists` gate), so the test does not depend on `/etc/passwd` being present — it passes or fails on the logic, not the machine.

Remaining untested, all minor and none regressions: a `package.json` that is not valid JSON, an extension with no `contributes.themes`, a theme whose `path` names a file that does not exist, and `apply_into_dirs` when `install_into` itself fails. The first two are `continue` arms that the escaping-path fixture partially exercises incidentally. The `?` on `install_into` still propagates before the `config.json` block, so a failed install cannot leave a config naming a missing manifest — that ordering is unchanged and remains correct.

## 3. Still open

Nothing blocking. Two observations I am recording rather than raising as findings, since neither is reachable from a real VS Code layout:

- `extension_order` parses a prerelease directory such as `acme.theme-1.0.0-rc1` as stem `acme.theme-1.0.0` with version `rc1` → `[0]`, because `rsplit_once('-')` takes the last hyphen and `"rc1".parse::<u64>()` fails to `0`. It would sort after a plain `1.0.0`, which is the safe direction anyway. VS Code's extension directories are `publisher.name-<semver>` without prerelease suffixes, so this is theoretical.
- With differing part counts (`1.0` vs `1.0.0`), `Reverse<Vec<u64>>` puts the *longer* vector first, so `1.0.0` beats `1.0`. Harmless, and again not a shape VS Code produces.

Release hygiene re-verified at this head: `origin/main` is still `0.1.807`, the branch and `Cargo.lock` are both `0.1.812`, and `src/release_notes.rs` is replaced rather than appended. Per CLAUDE.md, re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks` before merging — 0.1.808–0.1.811 are claimed by open PRs and main can move under you.

## 4. Scores

Both majors are fixed at the source rather than papered over, and each fix landed with a test that fails against the previous head. The hermeticity fix in particular is placed where the guard belongs and cites the sibling precedent, so the codebase's stated convention is now consistent instead of bypassed.

**CodeRabbit score (diff quality) — 5/5**
**Greptile score (codebase fit) — 5/5**

VERDICT: APPROVE

Issue #378, opened 2026-08-29
Epic: croft web, a browser client for the session host


vitali87:
The session host already fans one croft out to N clients; a browser is one more client. This epic tracks the RFC, the transport, the page, and the trust boundary. Suggested order: RFC first, nothing else lands before its trust-boundary section is agreed.

## Issues

- [ ] #340 croft web: design RFC for a browser client of the session host
- [ ] #341 croft web: WebSocket client transport for the session host
- [ ] #342 croft web: embedded browser page that renders the session
- [ ] #343 croft web: token auth, TLS, and listener lifecycle

vitali87:
Surveyed against `origin/main`: none of the four children has started, and they are a strict chain.

| # | state | absence proof (with control) |
| --- | --- | --- |
| 340 RFC | not started | no `docs/WEB.md`; `git ls-tree origin/main docs` = 11 files, none web-related. The only mention anywhere is `src/vscode_extensions.rs:499` "croft web is tracked in #378" |
| 341 WebSocket transport | not started | `git grep -i websocket -- src` = 0 files; control: `session_host` appears 11× in `mod.rs` and `Frame` 131× in `session_host.rs`, so the transport layer it would attach to is present and greppable. No `Web` variant in `cli.rs`'s `Command` enum |
| 342 browser page | not started | depends on 341; no HTML/JS assets in tree, no `include_bytes!` page |
| 343 auth / TLS / lifecycle | not started | depends on 341; no listener exists to secure |

Worth noting the groundwork that does exist, since it changes the size of 341: `src/session_host.rs` already broadcasts frames verbatim to multiple clients over its own transport. A WebSocket client is a second transport onto an existing mux rather than a new server.

No action taken. Flagging only that this epic is entirely unstarted, in case its position in the roadmap was unclear from the outside.

Issue #371, opened 2026-08-29
Commit scrubber: drag through history and watch the workspace re-render


vitali87:
## Motivation

TIMELINE lists a file's history; the COMMITS graph lists the repo's. Neither lets you *move* through time. A scrubber that re-renders the open editor, its git gutter, and the Explorer decorations as you drag across commits is a way of understanding a codebase that no editor offers, and a striking demo.

## Proposed design

1. **Source Control: Scrub History** opens a slider along the bottom of the editor listing commits on the current branch (first-parent), with the graph's colour rails; `←`/`→` step, drag scrubs, `Home` returns to the working tree.
2. At each position the active editor shows the file at that commit (read-only, via `git show`, cached), the git gutter shows the delta from the previous commit, the breadcrumbs and Outline follow, and the Explorer dims files that did not exist yet.
3. **Open here** turns the read-only view into a normal tab at that commit; **Diff to working tree** opens the diff widget.
4. Large repos: commits load lazily around the cursor; the slider never blocks the UI (git worker thread, as in the graph).

## Acceptance criteria

- [ ] Scrubbing across 200 commits of `src/app/mod.rs` stays interactive (under 16 ms per step after warm-up).
- [ ] The gutter at each step shows that commit's changes for the file, matching `git show --stat`.
- [ ] Leaving the scrubber restores the live buffer with dirty state untouched.

## Related

- Tracking epic: #384

vitali87:
## The cursor has shipped in #448 — the surfaces have not

**Shipped:** `src/scrubber.rs`, behind "Source Control: Scrub History". Arrows step, `Home` returns to the working tree, `Esc` closes, and `git::branch_history` (`--first-parent HEAD`) supplies the commits.

**Still open, which is what this issue is now for:** the slider widget with the graph's colour rails, showing the file at each commit (read-only, cached `git show`), the gutter delta per step, breadcrumbs and Outline following, Explorer dimming for files that did not exist yet, "Open here" / "Diff to working tree", and lazy loading around the cursor.

**Criteria:** ❌ 200 commits stays interactive (nothing renders yet); ❌ the gutter matches `git show --stat`; ⚠️ leaving restores the live buffer — true, but vacuously, since nothing is replaced.

**Two things from the implementation worth not re-deriving.**

[...]

Issue #370, opened 2026-08-29
.http files: run HTTP requests from the editor and show the response in a tab


vitali87:
## Motivation

REST Client is one of the most-installed VS Code extensions and the whole feature is a parser for `.http` files plus an HTTP call plus a pretty-printer. croft already has JSON/XML pretty views and a tab model; this is a few hundred lines that many users check for on day one.

## Proposed design

1. Recognise `.http` / `.rest` files: tree-sitter-style highlighting for the request line, headers, body, `###` separators, and `{{variables}}` (from a `.http.env.json` next to the file, plus `$env` references).
2. A ▷ glyph above each request (same affordance as runnable docs); `Cmd+Enter` runs the request under the caret via a minimal HTTP/1.1 client (no TLS dependency beyond what the crate tree already has; if none, shell out to `curl` with a clear provenance line, decided in the PR).
3. The response opens in a `Response · GET /users` tab: status line, headers, timing, and the body pretty-printed by content type (JSON/XML through the existing formatters, images through the preview pipeline), with **Save Response** and **Copy as curl**.
4. Request history per file in the same store the terminal command history uses.

## Acceptance criteria

- [ ] A three-request file runs each independently with variables substituted from the env file.
- [ ] A JSON response is pretty-printed and foldable; a PNG response renders inline.
- [ ] Secrets in the env file are never written to the response tab or history.

## Related

- Tracking epic: #384

Issue #365, opened 2026-08-29
croft pr <n>: PR review mode with a Files Changed section and branch checkout


vitali87:
## Motivation

Reviewing a pull request means reading a diff with real navigation: go to definition, hover types, run the tests. GitHub's web diff has none of that; lazygit shows the diff but is not an editor. croft has both the LSP editor and `gh` already in its workflow (Source Control shells out to git; the vetting workflow uses `gh`), so a PR review mode is mostly assembly.

## Proposed design

1. `croft pr 123` (and **Source Control: Review Pull Request** with a picker over `gh pr list`) fetches the PR head, checks it out into a worktree (reusing the worktree-lane plumbing) or the current tree with a confirm, and switches the sidebar to a **PULL REQUEST #123** view: title, author, checks status, and a **Files Changed** tree with per-file `+n −m` and a viewed checkbox (GitHub's model).
2. Clicking a file opens the diff against the base; the editor side is a real buffer with LSP, so go-to-definition and hover work in the changed code.
3. Checks and CI status refresh on a timer; a failing check row opens its log in OUTPUT through `gh run view --log`.
4. Review comments and submission are the sibling issue; this one is read-only review.

## Acceptance criteria

- [ ] `croft pr <n>` on this repo opens the file list within a few seconds and marks files viewed persistently (GitHub's viewed state via the API when available, local otherwise).
- [ ] Go-to-definition works from a changed line into unchanged code.
- [ ] Leaving review mode restores the previous branch/worktree with nothing dirty left behind.

## Related

- Tracking epic: #384

PR #418, opened 2026-08-30
feat(extensions): map VS Code extensions to what croft has

### LINKED ISSUE #352: Map VS Code extension recommendations to croft equivalents
## Motivation

`.vscode/extensions.json` and a user's installed-extensions list are the last part of a VS Code profile. croft cannot run VS Code extensions, but a lot of them correspond to something croft already has (language servers, debug adapters, themes, REST client, GitLens-style blame), and telling the user "you have this, you don't need that, and here's what has no equivalent" is a better answer than silence.

## Proposed design

1. A curated table `extension_id → { croft: <feature or manifest>, status: builtin|installable|none, note }`, shipped as data next to the language manifests.
2. `croft import-vscode` (and a palette command **Extensions: Compare with VS Code**) reads `.vscode/extensions.json` recommendations and `code --list-extensions` when available, and renders the result in the Extensions panel as a third group: **FROM VS CODE**, each row showing the mapping and, for `installable`, a one-click install of the croft manifest.
3. The `none` rows link to a tracking discussion so the table doubles as a demand signal for what to build next.

## Acceptance criteria

- [ ] The top 50 marketplace extensions have a row in the table.
- [ ] A workspace recommending `rust-analyzer`, `ms-python.python`, and `esbenp.prettier-vscode` shows builtin/builtin/installable respectively.
- [ ] Nothing is downloaded or executed without the user clicking install.

## Related

- Tracking epic: #380
- Depends on: #351

vitali87:
Closes #352.

## What this does

croft cannot run VS Code extensions, but most of the popular ones name a thing croft already has: a language server, a debug adapter, a test runner, a theme, a viewer. This ships the comparison as data and shows it in three places.

1. **`src/vscode_extensions.rs`** — a curated table of 92 marketplace extensions (`publisher.name` → croft extension id or feature, `builtin` / `installable` / `none`, one-line note), with `lookup`, `compare` (dedup, ordered built-in → installable → none), and two read-only readers: `workspace_recommendations(root)` parses `.vscode/extensions.json` (JSONC), and `installed_via_code()` runs `code --list-extensions` when `code` is on PATH.
2. **Extensions panel: FROM VS CODE.** A fourth group after AVAILABLE. Each row is the marketplace id with the blurb "built in: lsp-rust — …", "install mcp-…: …", or "no equivalent — …". Only an installable row carries the `+Add` control, and clicking it installs the *croft* catalog id (`selected_install_target()`), not the VS Code id; a built-in or no-equivalent row has no control and its toggle just restates the answer in the status bar. Workspace recommendations appear on every refresh (a file read); the installed list only after the user asks.
3. **`Extensions: Compare with VS Code`** palette command (`compare_extensions_with_vscode`): runs `code --list-extensions` once, merges with the recommendations, opens the panel, and says how many rows it compared (and that `code` was not found, when it was not).
4. **`croft import-vscode`** prints the same comparison as a section of its report, read-only, pointing at the panel for installs.

Nothing is downloaded or executed without a click: the readers only read a file and list extensions; install goes through the existing `mcp::catalog::install`.

## The acceptance triple, and one honest deviation

`rust-lang.rust-analyzer` → built in (`lsp-rust`); `ms-python.python` → built in (`lsp-python`). The issue expected `esbenp.prettier-vscode` → *installable*; there is no croft manifest to install for it, so it maps to **built in: `format_on_save`** (each language server's formatter runs on save). The installable path is implemented and tested against a synthetic row; the shipped table has no installable rows because croft's catalog holds only MCP servers today, and no popular VS Code extension is one. Rows for `none` carry the tracking issue in their note where one exists (#264, #355, #365/#366, #370, #371, #378) rather than a discussion link.

## Tests

- `vscode_extensions::tests` — the table has ≥ 50 unique lowercase `publisher.name` rows, every `none` row has an empty target and every other row a non-empty one; the acceptance triple and case-insensitive lookup; every built-in target that looks like an extension id (`lsp-`/`dap-`/`test-`) is an id croft actually bundles, and every installable target is a catalog entry; JSONC recommendations with comments and trailing commas, and a missing file; `compare` dedups case-insensitively and orders by status.
- `widgets::extensions` — FROM VS CODE rows sort last under their header, the built-in and unknown rows say so, exactly one `+Add` appears, and only the installable row reports an install target.
- `app::tests` — a workspace with `.vscode/extensions.json` shows its recommendations under FROM VS CODE after `refresh_extensions`.

The panel and app tests were mutation-checked (unconditional `available: true` fails the panel test; disabling the built-in-row guard fails the app test); the module tests pin the table's shape and the acceptance triple.

## Gate

- Test run over `vscode_extensions::`, `widgets::extensions::`, and the app test: 22 passed.
- Version 0.1.813 (main is 0.1.807; open PRs claim 0.1.808–0.1.812). Release notes replaced.


vitali87:

Requested by the author. CodeRabbit is rate limited repo-wide, and `gh api repos/vitali87/croft/pulls/418/comments` returns `0`. Reviewed at `0c13147` against a built tree.

The install path is the part I attacked hardest, since a table of marketplace ids feeding an installer is where a wrong id would do real work. It is correctly defended, in two layers, and one of the two layers has no test.

### The install path is right

`available: install.is_some()` means a built-in or no-equivalent row is never `available`, so `selected_available()` is false and the install branch is unreachable for it. Then `selected_install_target()` supplies the **croft** catalog id, and the `unwrap_or(id)` fallback is reached only for a genuine AVAILABLE catalog row, where `id` already is the catalog id. A VS Code marketplace id cannot arrive at `mcp::catalog::install`.

The outer layer is pinned. `available: true` unconditionally:

```
test widgets::extensions::tests::from_vscode_rows_sort_last_and_only_installable_ones_offer_add ... FAILED
```

### Minor — the inner guard has no test

The early return that makes a built-in or no-equivalent row *say what croft has* rather than do nothing:

```rust
if let Some(item) = self.extensions.selected_item()
    && item.vscode
    && item.install.is_none()
{
    self.status = format!("{}: {}", item.id, item.description);
    return;
}
```

Disabling it entirely (`if false && let Some(item) = …`) leaves all 48 tests in `vscode_extensions::`, `widgets::extensions::` and the app extension tests green.

It is defence in depth rather than the only thing standing between a user and a wrong install, so this is a minor. But it is also the row's entire user-visible behaviour: with it gone, toggling `rust-lang.rust-analyzer` under FROM VS CODE falls through to whatever the non-available path does, and the answer the group exists to give never appears. One app test that toggles a built-in row and asserts the status names the croft equivalent would cover it.

### Minor — `installed_via_code` has no timeout, and it blocks the frame

```rust
let Ok(out) = std::process::Command::new("code")
    .arg("--list-extensions")
    .stdin(std::process::Stdio::null())
    .output()
```

`stdin` nulled and a non-zero exit treated as empty are both right. What is missing is a bound. `.output()` blocks until the child exits, and `compare_extensions_with_vscode` calls it on the render thread, so a `code` that hangs (a stale lock, a slow network share, an unresponsive Electron start) freezes croft with no way out.

This repo already has the pattern and the scar. `gui_path.rs` runs its login-shell probe "bounded by a 3s timeout and reaped off-thread so a wedged rc cannot wedge startup" (`docs/ARCHITECTURE.md`), which is the same shape of problem: an external program that is usually instant and occasionally is not. The palette command is explicit and rare, which argues for keeping it simple, but "rare" describes how often it runs, not how bad it is when it wedges.

### Minor — "red on the parent" is not a control for a new module

The PR says the tests "were all red on the parent (the API did not exist)". For a new module that is a compile error, not evidence that any test detects the defect it names. Two of the ones I mutated do detect real changes, so the tests are good; the claim is just not the one that establishes it. Worth stating as "mutation-checked" or not claiming a control at all, since this repo has had several tests that passed against their own bug.

### Checked and sound

- **`compare`** lowercases and dedups through a `BTreeSet` before lookup, so `Publisher.Name` and `publisher.name` collapse, and the ordering is by `Status` then id, which matches the panel's grouping.
- **`workspace_recommendations`** goes through `crate::tasks::strip_jsonc` (the chars-based one), so a `.vscode/extensions.json` with a non-ASCII comment does not mangle. This matters more than it looks: the byte-walking copy was the subject of #396 and was deleted today.
- **The table's own consistency tests** are the right ones: every `lsp-`/`dap-`/`test-` target must be an id croft actually bundles, and every installable target must be a catalog entry, so the table cannot drift from what ships.
- **The prettier deviation** is stated in the PR body rather than quietly resolved, and mapping it to `format_on_save` is the honest answer given no croft manifest exists for it.
- **Nothing is downloaded or executed without a click.** The readers read a file and list extensions; install goes through the existing catalog path.

### Scores

- **CodeRabbit score (diff quality) — 4/5.** Large, well-structured, and the one genuinely dangerous path is defended twice. Held off 5 for the unbounded subprocess on the render thread and the untested inner guard.
- **Greptile score (codebase fit) — 5/5.** New module registered, panel group added in the existing group-rank shape, palette command wired through `Cmd`, JSONC read through the shared stripper, version bump and notes replacement all match the repo.

VERDICT: REQUEST_CHANGES on the timeout. The two test items are worth doing here but I would not block on them alone.

vitali87:
Addressed at 1c5f41196d6cb19849c80eefa13ff6576a811b3b.

- **Unbounded `code --list-extensions` on the render thread (minor, the blocking one)**: `installed_via(cmd, timeout)` now spawns with piped stdout, reads on a helper thread, `recv_timeout(LIST_TIMEOUT = 3 s)`, kills, and reaps off-thread — the `gui_path::run_probe` shape. `listing_installed_extensions_gives_up_on_a_hung_child` drives it with `sleep 30` under a 300 ms bound (asserts return well under 5 s), a normal listing (non-ids filtered), and a missing program.
- **Untested inner guard (minor)**: `workspace_extension_recommendations_show_under_from_vscode` now selects the `rust-lang.rust-analyzer` row, toggles it, and asserts the status starts with `rust-lang.rust-analyzer: built in: lsp-rust` and `disabled_extensions` is unchanged. Disabling the guard makes it fall through to the toggle path and fail.
- **"red on the parent" wording (minor)**: taken — the PR body will say mutation-checked rather than claiming a control for a new module.

Also since your pass: two commits tightening table notes to what croft ships today (no agent-pane badging claim — that is #344; runnable fences are #353; Tokyo Night and GitHub Dark are bundled; tag auto-close is `none`).

Gate at this head: 23 tests over `vscode_extensions::`, `widgets::extensions::`, and the app test pass; clippy `-D warnings`, fmt, and the doc gate clean.

vitali87:
## 🤖 Re-score at `1c5f411` — Claude Opus 5

All three items landed.

**The timeout.** `installed_via` spawns, reads on its own thread, takes `recv_timeout`, then kills and reaps **off-thread**. That last part is the one that is easy to get wrong and is right here: a synchronous `wait()` after `kill()` would wedge the frame anyway for a child in uninterruptible sleep, which is the failure the bound exists to prevent. Same shape as `gui_path::run_probe`, and the comment says why rather than just what.

`listing_installed_extensions_gives_up_on_a_hung_child` passes in 0.31s against a `sleep 30`, and it asserts the elapsed time rather than only the empty result, so it fails if the bound stops being a bound rather than merely if the result changes. Splitting `installed_via(cmd, timeout)` out so the test can inject both is the right seam: the production entry point still owns the 3s constant.

**The guard.** Now pinned. Disabling it:

```rust
if false && let Some(item) = self.extensions.selected_item()
    && item.vscode
    && item.install.is_none()
```

```
test app::tests::workspace_extension_recommendations_show_under_from_vscode ... FAILED
test result: FAILED. 48 passed; 1 failed
```

Green before, red now.

**The wording.** The body says mutation-checked rather than red-on-the-parent, which is the claim that actually holds for a new module.

### Scores at `1c5f411`

- **CodeRabbit score (diff quality) — 5/5.** The one dangerous path is defended twice with both layers pinned, the subprocess is bounded in the shape this repo already learned the hard way, and the table's consistency tests keep it from drifting from what ships.
- **Greptile score (codebase fit) — 5/5.** Unchanged from the last round.

VERDICT: APPROVE

Nothing outstanding. Same note as the others: this branch carries a `src/release_notes.rs` edit and will need converting once #410 lands.

PR #419, opened 2026-08-30
ci: probe a file's existence at the base by exit status, not stdout

vitali87:
`git cat-file -e` prints nothing on success, so the added-file detection
that read its stdout took every changed file for a branch-added one:
each went through the pairwise commit walk (over origin/main~40..main
that was more than ten minutes of git show calls, for nothing) and a
capture in an existing file was reported twice, once by that walk and
once by the base comparison.

`exists_at` reads the exit status, `added_files` uses it, and `git()`
loses the `allow_fail` mode that only that probe used. The same 40
commits now take under a minute and report each of the two real
losses once. Two tests pin it: a file present at the base is not
"added", and a capture in an existing file yields exactly one
annotation.

## Control
- `scripts/check_doc_ownership.py origin/main~40 origin/main`: 2 annotations in 54 s (before: 4 annotations, the run did not finish inside 10 minutes).
- `python3 -m unittest scripts.tests.test_doc_ownership`: 26 passed; the two new tests were red on the parent (`AttributeError: exists_at`, and `2 != 1` annotations).

CI/scripts only: no version bump or release note.

### DIFF
--- a/scripts/check_doc_ownership.py
+++ b/scripts/check_doc_ownership.py
@@ -213,33 +213,48 @@ class BlockTracker:
             self.blocks.pop()
 
 
-def git(*args, allow_missing_path=False, allow_fail=False):
+def git(*args, allow_missing_path=False):
     """Run git, failing closed.
 
     A silent failure here is worse than a crash: an errored `git diff` yields
     an empty file list, the checker reports "no documentation lost" and exits
     zero, and the gate has passed by not running.
 
-    `allow_missing_path` narrows that tolerance to exactly the expected case,
-    and `allow_fail` is separate and narrower still: it is for a probe whose
-    exit status IS the answer, and its empty return is never read as content.
+    `allow_missing_path` narrows that tolerance to exactly the expected case.
     Tolerating every non-zero exit would let an invalid revision or a
     malformed object read as empty content, which is the same fail-open bug
-    one door further in.
+    one door further in. A probe whose exit status IS the answer does not go
+    through here at all: see `exists_at`.
     """
     proc = subprocess.run(["git", *args], capture_output=True, text=True)
     if proc.returncode != 0:
-        # `allow_fail` is for a PROBE whose non-zero exit is the answer
-        # (`cat-file -e` asking whether a path exists), never for a command
-        # whose output is then read as content.
-        if allow_fail:
-            return ""
         if allow_missing_path and MISSING_PATH.search(proc.stderr):
             return ""
         raise SystemExit(
             f"git {' '.join(args)} failed ({proc.returncode}): {proc.stderr.strip()}"
         )
     return proc.stdout
+
+
+def exists_at(rev, path):
+    """Whether `path` is present at `rev`.
+
+    A probe: `git cat-file -e` prints nothing either way and answers with
+    its exit status. Reading its stdout, as the added-file detection once
+    did, made every changed file look branch-added: each took the pairwise
+    commit walk (minutes on a real branch, for nothing) and a capture in an
+    existing file was reported twice, once per loop.
+    """
+    proc = subprocess.run(
+        ["git", "cat-file", "-e", f"{rev}:{path}"], capture_output=True, text=True
+    )
+    return proc.returncode == 0
+
+
+def added_files(base, changed):
+    """The changed files that do not exist at `base`: the branch added them,
+    so their history lives only in its own commits."""
+    return [f for f in changed if not exists_at(base, f)]
 
 
 def is_doc(line):
@@ -389,7 +404,7 @@ def main():
     # empty and no loss can ever be reported in it: a doc captured within the
     # branch is invisible to a merge-base comparison. Those files are checked
     # commit-by-commit instead, which is the only place their history exists.
-    added = [f for f in changed if not git("cat-file", "-e", f"{base}:{f}", allow_missing_path=True, allow_fail=True)]
+    added = added_files(base, changed)
     if added:
         # What matters is the state at HEAD. A doc captured in one commit and
         # restored in a later one is not a loss: the branch is fine, and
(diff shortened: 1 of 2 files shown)


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head b295743 stands in.

## 1. Walkthrough

The added-file detection called `git("cat-file", "-e", ...)` and tested its **stdout**, but `cat-file -e` is a pure exit-status probe that prints nothing either way — so `not ""` was always `True` and every changed `.rs` file was classified as branch-added. This PR extracts the probe into `exists_at(rev, path)`, which reads `returncode`, adds `added_files(base, changed)` on top of it, and drops the now-unused `allow_fail` mode from `git()`. I reproduced the bug and the fix against scratch repos: the parent emits **2** annotations for a single capture, the head emits **1**.

|---|---|---|
| `scripts/check_doc_ownership.py` | `exists_at` / `added_files` added; `allow_fail` removed from `git()`; docstrings and the `main()` comment updated | Correct. Diagnosis verified; gate remains fail-closed on a bad base |
| `scripts/tests/test_doc_ownership.py` | Two tests in `GitShapes`, plus `contextlib`/`io`/`os` imports | Both are genuine red controls on the parent |

**Verification performed.** `git cat-file -e main:a.rs` → rc=0, stdout 0 bytes (confirms the premise). Old script on a 2-commit branch with one capture → 2 annotations; new script → 1. Full suite: 26 passed. `grep -rn allow_fail` → no hits, with `allow_missing_path` (9 hits) as the positive control that the corpus and pattern are live.

## 2. Findings

**`scripts/check_doc_ownership.py:246` [nit/robustness]** — `exists_at` returns `False` for an *invalid revision*, not just a missing path, so at that layer alone an unreadable base would silently reclassify every file as branch-added. I checked whether this is exploitable and it is **not**: `main()` calls `git("log", f"{base}..{head}")` at line 387 before `added_files` is ever reached, and that raises `SystemExit` on a bad rev (`fatal: ambiguous argument …` does not match `MISSING_PATH`). I confirmed directly — `python3 scripts/check_doc_ownership.py nosuchrev HEAD` exits non-zero with `git log … failed (128)`, and `git("show", "nosuchrev:a.rs", allow_missing_path=True)` also raises rather than returning `""`. So the gate stays fail-closed today via two independent doors. The residual risk is ordering-coupled: `exists_at`'s own safety depends on a call several lines earlier in a different function. Not worth blocking on; worth a line in the docstring noting that the fail-closed guarantee comes from the earlier `git log`, so a future refactor that reorders `main()` does not quietly remove it.

No blockers, majors, or minors. Specifically checked and found clean:

- **Paths with spaces and non-ASCII.** `exists_at("main", "sp ace.rs")` → `True`, `"ünï.rs"` → `True`. No shell is involved (`subprocess.run` with a list), so no quoting hazard. The gate correctly reported the capture in `sp ace.rs`.
- **A file the branch deletes.** `exists_at` → `True` at base, `False` at head; it is not in `added`, and the `for f in changed` loop's `git show {head}:{f}` is tolerated by `allow_missing_path`. `after` is empty, so `name in after` is False and no loss is invented. Correct.
- **Submodules.** A gitlink returns `False` from `exists_at` (`cat-file -e` will not resolve a commit entry as a blob), but this is unreachable: `git diff --name-only` emits the submodule path itself (`subm`), never `subm/s.rs`, and the `.rs` suffix filter drops it. No behaviour change.
- **Empty path.** `exists_at(rev, "")` returns `True` because `git cat-file -e main:` resolves the root tree — unreachable, since `"".endswith(".rs")` is `False`.
- **Fork PRs / PRs that add a file.** The workflow passes a real merge-base and `HEAD` with `fetch-depth: 0` and `persist-credentials: false`; the change is confined to classification and cannot alter fork behaviour. The added-file path is *better* served now: only genuinely added files take the pairwise walk.
- **Version-bump exemption.** Both changed paths fail the `^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)` filter, so `shipped` is empty and `release-hygiene` exits 0. I ran the pipeline with `src/main.rs` as a positive control to prove the filter can match — it does. Exempt, as the PR body states.

## 3. Test coverage

Both new tests are real controls; I reconstructed the parent script alongside the new tests and ran them:

- `test_a_file_present_at_the_base_is_not_treated_as_added` → `AttributeError: module 'check_doc_ownership' has no attribute 'exists_at'`
- `test_a_capture_in_an_existing_file_is_reported_exactly_once` → `AssertionError: 2 != 1`

Both match the PR body exactly, and an existing test (`test_a_capture_in_an_existing_file_is_reported`) passed on the same parent as a positive control, proving the harness itself was working rather than uniformly broken.

**One real gap, and it answers the question directly.** `test_..._exactly_once` is **not** meaningful as a guard on the pairwise loop's own dedupe. I deleted the `and not any(l[0] == f and l[1] == name for l in losses)` line in place in the worktree and ran the full suite: **26 tests, OK**. Nothing catches it. (My first attempt at this in a scratch copy produced a misleading failure in `test_the_block_tracker_is_balanced_at_eof_on_every_real_file` — that was only the missing `src/` corpus, which the test's own `assertGreater(len(files), 50, "the control corpus is missing")` guard correctly flagged; I re-ran in place to get the true answer.)

That is not a regression this PR introduces — the dedupe was already uncovered on the parent — and after this fix an existing file no longer enters the pairwise loop at all, so the two mechanisms guard different things. But the test's docstring ("with every file mistaken for branch-added, the capture was reported by that walk AND by the base comparison") describes what it pins, and a reader could reasonably take it as covering the dedupe too. Worth a follow-up test for a genuinely branch-added file whose doc is captured across two intermediate commits, which is the case the dedupe actually exists for.

The `os.chdir` / `sys.argv` save-restore in both tests uses `try/finally` correctly. `Repo.exit_code` already encapsulates this pattern; the second test does not reuse it only because it needs stdout, which is fair.

## 4. Nitpicks

- `scripts/tests/test_doc_ownership.py:9` — `import os` is now at module scope, but `Repo.exit_code` (line ~50) still has a function-local `import os`. Now redundant; drop the local one.
- `scripts/check_doc_ownership.py:249` — `added_files` is a one-line comprehension used exactly once. Fine as-is (it names the concept and gives the test a seam), but the docstring's opening line runs into the summary; PEP 257 prefers a one-line summary then a blank line.
- Line length: the `added_files` call site is now well within limits, but `head_state = {...}` (line 415) remains ~100 chars. Pre-existing, and there is no Python linter in CI, so no action needed.

## 5. Scores

**CodeRabbit score (diff quality) — 4/5**
The diagnosis is correct and I reproduced it end to end; the fix is minimal and removes a mode rather than adding one. Held off 5 by the nit-level ordering coupling in `exists_at`'s fail-closed guarantee and the redundant local `import os`.

**Greptile score (codebase fit) — 4/5**
Matches the file's established conventions: fail-closed by default, a narrowly-scoped tolerance, and comments that explain the failure rather than the code. Both new tests are honest red controls, and the comment that referenced `allow_fail` was correctly updated to `allow_missing_path` rather than left stale. Held off 5 because the second test's docstring implies coverage of the pairwise dedupe that it demonstrably does not provide.

VERDICT: APPROVE

Issue #358, opened 2026-08-29
Notification sinks: forward finished commands, red tests, and waiting agents to ntfy, termux-notification, or a webhook


vitali87:
## Motivation

croft already knows when a long command finishes in an unfocused pane, when the test suite goes red, and (with the agent-lane series) when an agent is waiting for input. Today all of that ends in the status bar. Forwarding it to a phone or a chat is what turns a remote session on a server into something you can walk away from.

## Current state

- OSC 9 notifications and "long command finished" notices surface in the status bar (`docs/LAYOUT.md`, Terminal section).
- `src/termux.rs` already shells out to `termux-api` for other features.

## Proposed design

1. A `notifications` block in `config.json`: a list of sinks (`{kind: "ntfy", topic, server?}`, `{kind: "termux"}`, `{kind: "webhook", url, headers?}`, `{kind: "command", argv}`), each with an event filter (`command_finished` with a min-duration, `tests_failed`, `agent_waiting`, `osc9`, `trigger:<name>`).
2. Delivery on a background thread with a short retry, never on the render path; failures logged to a **Notifications** OUTPUT channel.
3. Payload: title, body, workspace, host, and a deep link (`croft://attach?host=…&path=…`) the phone-side issue consumes.
4. Secrets: webhook headers live in `config.local.json` (machine-local layer) and are never synced.

## Acceptance criteria

- [ ] `cargo build` taking longer than the threshold in an unfocused pane posts to a configured ntfy topic with the exit code.
- [ ] A failing Test Explorer run posts once, not once per test.
- [ ] Sink failure never blocks input or rendering (measured with the perf HUD).

## Related

- Tracking epic: #383
- Depends on: #344

Issue #347, opened 2026-08-29
Approval popup: show an agent's pending edit as a live diff with LSP diagnostics, approve or deny in place


vitali87:
## Motivation

Once an agent's edit proposal reaches croft (hook issue), the editor can do what a terminal prompt cannot: show the change as a real diff, run it through the language server *before* it lands, and let you approve, deny, or edit the proposal.

## Proposed design

1. A modal diff popup (reusing `src/widgets/diff.rs` and the merge-editor plumbing) titled `claude wants to edit src/foo.rs`, side by side old/new, with the proposal applied to a shadow buffer that is sent to the LSP as a `didChange` so new diagnostics show in the gutter of the *proposed* side.
2. Actions: **Approve** (`Enter`), **Deny** (`Esc`, with an optional reason typed back to the agent), **Edit then approve** (open the proposal in a scratch editor, the edited text is what gets approved), **Approve all from this agent for 10 min** (a visible countdown chip in the status bar; `Cmd+K X` revokes).
3. Multiple pending proposals queue in arrival order with a `2 pending` badge; the agent-lane rows show the pending one with a pulsing glyph.
4. When the file is open and dirty in croft, warn that the agent's proposal is based on disk content and show a three-way view (disk, yours, agent's) using the merge editor.

## Acceptance criteria

- [ ] A proposal that introduces a type error shows the diagnostic on the proposed side before approval.
- [ ] Deny with a reason reaches the agent's stdout payload.
- [ ] "Approve all" auto-answers subsequent proposals only from that agent and only until revoked or expired.
- [ ] Dirty-buffer case surfaces the three-way view rather than silently approving.

## Related

- Tracking epic: #379
- Depends on: #346

vitali87:
claimed by croft-34 — building on #346's hook (PR #606)

Issue #345, opened 2026-08-29
Agent lane: a per-agent list of files it changed since you last reviewed them


vitali87:
## Motivation

The daily pain with a coding agent is losing track of what it touched while you were looking elsewhere. croft owns the file watcher and git, so it can keep a per-agent ledger and present a review queue: each file it changed, a diff, a "reviewed" tick, and an unread count that only goes down when you actually looked.

## Current state

- `src/app/fs_watch.rs` delivers workspace change events; `src/git.rs` computes per-file diffs against HEAD; `src/widgets/diff.rs` renders them.
- The agent-detection issue provides `AgentEvent`s and the pane identity.

## Proposed design

1. While an agent pane is `working`, attribute every fs-watch write in the workspace to that agent (ties between two working agents attribute to both and are flagged).
2. A **Agent Lane** section in the sidebar (Explorer ⋯ menu toggle, like Outline/Timeline): one collapsible group per agent, one row per touched file with `+n −m`, sorted by last write. Click opens the diff of that file versus the content at the moment you last reviewed it (a snapshot taken through the existing local-history store, `src/history.rs`), not just versus HEAD.
3. "Mark reviewed" (`Enter` on the row / `Cmd+K V` on all) snapshots the current content as the new baseline; the count badge on the row and the lane header drop.
4. The Explorer file tree gets a small dot decoration on unreviewed files, in the agent's colour.

## Acceptance criteria

- [ ] With a script that writes three files while "working", the lane lists exactly those three, and a fourth written after the agent goes idle is not attributed.
- [ ] Diff for a row compares against the last-reviewed snapshot, verified after two rounds of edits.
- [ ] Badge counts and tree decorations clear on Mark reviewed and persist across restart within the same session.

## Related

- Tracking epic: #379
- Depends on: #344

vitali87:
## Where this stands after #444

Deliberately left open — three of the four design points have shipped, one has not, and the split is worth recording so the next session does not have to re-derive it.

**Shipped:**

1. **Attribution while working** (#421, #429). A write lands against every agent whose pane is `Working`; two working agents share the row and it is flagged rather than guessed; a write with no agent working belongs to the user and never enters a lane.
2. **Mark reviewed** (#429), per file and per lane, baselining the content on disk so a row returns only on a later write.
3. **Tree decoration** (#444). Unreviewed files carry a trailing dot in the Explorer; the status chip counts the same set.

**Not shipped — what this issue is still for:**

[...]

PR #421, opened 2026-08-30
feat(agents): agent lanes: badge a pane running a coding agent

### LINKED ISSUE #344: Agent lanes: detect a coding agent running in a terminal pane and badge it
## Motivation

Claude Code, Codex CLI, aider, and gemini-cli all run in a terminal pane, which means croft is already where the agents live. Everything in the agent cockpit series starts from one primitive: knowing that pane 2 is an agent, which one, and whether it is idle, working, or waiting for the user.

## Current state

- croft samples the foreground process group of each pane for shell-integration and re-root purposes (`src/terminal_session.rs`, `src/shell_integration.rs`) and already renders name pills, progress gauges, and `host_accents` decorations on panes.
- OSC 133 marks give command start/end; agents that stay in the foreground never produce those.

## Proposed design

1. Identify the foreground process by name against a small table (`claude`, `codex`, `aider`, `gemini`, extensible from config as `agents.json`), re-sampled at the existing cadence.
2. Show an agent badge in the pane's name pill (`◆ claude`), and a status: **working** (output flowing), **waiting** (no output for N seconds and the last screen rows match a prompt pattern from the table), **idle**.
3. Emit an internal event stream (`AgentEvent::{Seated, Working, Waiting, Gone}`) that the follow-up issues (agent lane, notifications, approval) consume; nothing else in this issue.
4. Status bar chip `2 agents · 1 waiting`, click to focus the waiting pane.

## Acceptance criteria

- [ ] Running `claude` in a pane badges it within one sample interval; exiting clears it.
- [ ] A `waiting` transition fires exactly once per prompt, verified with a fake agent script in tests.
- [ ] Unknown agents can be added through `agents.json` without a rebuild.

## Related

- Tracking epic: #379

vitali87:
Closes #344.

## What this does

Claude Code, Codex CLI, aider and gemini-cli run in a terminal pane, so croft is already where the agents live. This adds the one primitive the agent-cockpit series starts from: knowing that pane 2 is an agent, which one, and whether it is working, waiting for you, or idle.

**Recognition.** The pane's foreground process is already sampled every 1.5 s for the name pill (`refresh_terminal_labels`). The same `(shell_pid, process_name)` samples now go through `App::apply_agent_samples`, which classifies the name (basename, any case) against an `AgentTable`: built-in rows for `claude`, `codex`, `aider`, `gemini`, extended by `~/.config/croft/agents.json` (seeded from a commented template by **Preferences: Open Agent Lanes (JSON)**, reloaded on save, no rebuild). A row named like a built-in replaces it.

**Status.** An agent that stays in the foreground never emits OSC 133 marks, so status is read from two things the pane already has: how long since the PTY last produced a byte (a new `last_output_ms` stamp on the reader, `quiet_for()`), and its last six non-empty screen rows (`tail_rows`). Under 3 s of quiet is **working**; quiet with a row matching one of the agent's prompt patterns (a permission question, a `>` line, a Y/n) is **waiting**; quiet otherwise is **idle**.

**Surfaces.** The pane pill wears `◆ claude ◐` (badge first, label clipped to the room left). The status bar gains a `◆ 2 agents · 1 waiting` chip while any agent is seated; clicking it focuses the first waiting pane, else the first agent pane. A `Waiting` transition puts "claude in Terminal 2 is waiting for your input" in the status bar.

**Events.** `AgentEvent::{Seated, Working, Waiting, Gone}` are computed by a pure `transition(prev, next)` and queued on the app for the tick to drain. `Waiting` fires only on the way *into* waiting, so a prompt that sits on screen across many samples fires once. Nothing else consumes them yet (#345, #347, #358's `agent_waiting` wire up next).

## Tests

- `agents::tests` — the built-in table classifies the four agents by name or path and rejects lookalikes; `agents.json` extends the table and can replace a built-in, a broken or missing file is the built-in table, and the template parses; `judge` is working while output flows, waiting on a prompt row, idle otherwise, and a question buried ten rows up does not count; `transition` fires Waiting once per prompt, Seated/Gone bracket the lane, and a different agent taking the pane is Gone then Seated.
- `app::tests::an_agent_pane_is_badged_and_a_prompt_fires_waiting_once` — the acceptance test with a fake agent: a `/bin/sh` that prints `Do you want to proceed? (y/n)` and sits, sampled as `claude`; asserts the lane is seated straight into Waiting, the announcement is made once and not on the next identical sample, the pill shows `◆ claude ◐`, the chip shows `1 agent · 1 waiting` with a hit rect, and a sample naming `zsh` unseats it.
- The existing pill-budget test (`a_long_pane_label_never_overpaints_the_maximize_button`) and host-accent test still pass; the badge takes room inside the pill's existing budget rather than widening it.

## Gate

- Test run over `agents::`, the new app test, the pill-budget test, and the host-accent test: 7 passed; clippy `-D warnings` and fmt clean.
- Version 0.1.815 (main is 0.1.808; open PRs claim 0.1.809–0.1.814). Release notes replaced.


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head 25e27a5 stands in.

## 1. Walkthrough

PR #421 adds the agent-lanes primitive from #344: a new `src/agents.rs` (an `AgentTable` of built-in + `agents.json` rows, a pure `judge()` reading quiet-time plus the last six non-empty screen rows, and a total `transition()` producing `AgentEvent::{Seated,Working,Waiting,Gone}`), a `last_output_ms` stamp and `quiet_for()`/`tail_rows()`/`agent()`/`set_agent()` on `PtyTerminal`, and app wiring that classifies the existing 1.5 s foreground-process samples, badges the pane pill, adds a status-bar chip with a click target, and drains events on the tick. The pure core is well factored and its unit tests are genuine controls. The wiring around it has three verified problems: a lane that can never be cleared when a sample goes missing, a `quiet_for()` that reads ~56 years before the pane's first byte, and a chip click that focuses a pane the user may not be able to see.

| File | What changed | Notes |
| --- | --- | --- |
| `src/agents.rs` | New: `AgentKind`/`AgentStatus`/`AgentLane`/`AgentEvent`, `AgentTable::{builtin,load,from_json,classify,names}`, `judge`, `transition`, `TEMPLATE`, 4 unit tests | Pure and well tested; `transition` is total and correct across all 20 prev/next pairs |
| `src/widgets/terminal.rs` | `last_output_ms: Arc<AtomicU64>` stamped in the PTY reader; `quiet_for`, `agent`/`set_agent`, `tail_rows(n)` | `quiet_for` has an unhandled zero-initial state; `tail_rows` calls `grid_lines()` (whole grid + scrollback under the term mutex) to keep 6 rows |
| `src/app/mod.rs` | `apply_agent_samples`, `drain_agent_events`, `agent_counts`, `focus_agent_pane`, pill badge, `seg_texts` → `Vec` + 6th chip segment + `status_agents_rect`, seed/reload/palette wiring | Config wiring is symmetric with triggers at all three sites; chip indexing and rect reset are correct; focus idiom diverges from every peer |
| `src/app/tests.rs` | `an_agent_pane_is_badged_and_a_prompt_fires_waiting_once` | Real render assertions, exercises Seated→Waiting, once-per-prompt, and Gone |
| `src/widgets/command_palette.rs`, `src/main.rs` | `OpenAgentsJson` command in all five arms; `mod agents` | Complete |
| `Cargo.toml`, `src/release_notes.rs` | 0.1.808 → 0.1.815, notes replaced | Verified unique against all open PRs |

## 2. Findings

**`src/app/mod.rs:27139` [major/correctness] — a pane whose foreground process fails to resolve keeps its agent badge forever; `Gone` never fires.**

Evidence: `refresh_terminal_labels` builds the sample vector in the background thread as

```rust
let resolved: Vec<(i32, String)> = targets
    .into_iter()
    .filter_map(|(sp, fg)| {
        crate::widgets::terminal::process_name(fg).map(|n| (sp, n))
    })
    .collect();
```

`process_name` (`src/widgets/terminal.rs:1008`) returns `None` for `pid <= 0`, for a pid `sysinfo` cannot see, and for an empty name — and `targets` itself already drops any pane whose `foreground_pid()` (`tcgetpgrp`) returns `None`. `apply_agent_samples` then iterates **only** the samples that survived:

```rust
for (shell_pid, name) in samples {
    let Some(t) = ... else { continue };
```

so a pane with a seated lane that produces no sample this round is simply not visited. Nothing anywhere else calls `set_agent(None)`. The unseat path in the PR — and in the acceptance test — depends entirely on a *positive* sample naming a non-agent (`zsh`). That is the lucky case: the agent exits, the shell reclaims the foreground group, `process_name` resolves `zsh`, `classify` returns `None`, `Gone` fires. The unlucky cases are the ones with no coverage: the agent exits and the pane's shell dies with it (pane held open, `foreground_pid` gone), the agent is killed and `tcgetpgrp` races the reap, or `sysinfo` transiently misses the pid. In each, the pill keeps `◆ claude ◐`, `agent_counts()` keeps counting it, the chip keeps claiming "1 waiting", and clicking the chip focuses a pane with no agent in it. The acceptance criterion in #344 is "exiting clears it" — it clears only on the resolvable path.

Fix: drive the loop over panes, not over samples. Build a `HashMap<i32, &str>` from the samples, then iterate `self.terminals` and pass `map.get(&shell_pid)` — `None` classifies to `None`, which `transition` already turns into `Gone` correctly. That also removes the `find()` per sample (quadratic in panes today).

**`src/widgets/terminal.rs:2167` [major/correctness] — `quiet_for()` reports ~56 years before the pane's first byte, so a freshly spawned agent is judged Idle or Waiting immediately.**

Evidence: `last_output_ms` is initialised `AtomicU64::new(0)` and only ever stamped in the reader's `Ok(n)` arm. `quiet_for` computes `now_ms.saturating_sub(0)`, i.e. the full Unix epoch — roughly 56 years — which is unconditionally `>= QUIET_AFTER`. The doc comment ("a pane that never has is quiet since the epoch") states the behaviour but not that it is wrong for the intended semantics: an agent that has been *seated but has not printed yet* is, by any reading of #344, working — it is starting up. Instead the first sample after spawn skips the `quiet_for < quiet_after` guard in `judge` and goes straight to the screen scan, and whatever the shell's startup left on the grid decides the badge. Any residual row ending in `?` (trivially common — see below) badges a just-launched agent `◐ waiting for you` and fires a spurious `Waiting` announcement, once, before the agent has done anything.

This is not hypothetical for the intended flow: the user runs `claude`, the sampler fires within 1.5 s, and the agent's own first paint may not have landed yet.

Fix: seed `last_output_ms` with the spawn time rather than 0, so a pane that has never output is quiet *since it started* — the honest reading, and it makes `Working` the correct initial verdict for the first `QUIET_AFTER` of a pane's life.

**`src/app/mod.rs:27209` [major/correctness] — clicking the agents chip focuses a pane the user may not be able to see.**

Evidence: `focus_agent_pane` does:

```rust
self.active_terminal = idx;
self.focus_pane(Pane::Terminal);
```

The status bar renders unconditionally in `render()` (it is not inside any `show_terminal` guard), so the chip is visible with the bottom panel hidden and with the panel showing PROBLEMS, OUTPUT, CAPTURES, or PORTS. Every peer that reaches a terminal pane from elsewhere in this file goes through the panel-revealing idiom: `captures_open_selected` (`:25643`) does `self.active_terminal = idx; self.set_bottom_panel_tab(BottomPanelTab::Terminal); self.focus_pane(Pane::Terminal);`, `undo_close_terminal` (`:12635`) sets `show_terminal = true` before focusing, and `set_bottom_panel_tab` (`:13568`) itself is the helper that both un-hides the panel and switches the tab. `focus_agent_pane` does neither, so the advertised affordance — "click the chip to reach the waiting agent" — silently does nothing visible in exactly the situation where it is most useful (you are in the editor, the panel is collapsed, and the chip is telling you an agent is blocked on you).

Fix: `self.active_terminal = idx; self.set_bottom_panel_tab(BottomPanelTab::Terminal);` — that helper already calls `focus_pane(Pane::Terminal)`, matching `captures_open_selected` exactly.

**`src/agents.rs:132,162` [minor/correctness] — `\?\s*$` makes "waiting" noisy for gemini and aider.**

Evidence, modelled against the actual patterns:

| agent | row | matches |
| --- | --- | --- |
| gemini | `Does that make sense?` | `\?\s*$` |
| aider | `Should we refactor this?` | `\?\s*$` |
| aider | `if (x == 1) { return y; } // ?` | `\?\s*$` |
| codex | `Allowlist updated` | `Allow` |
| codex | `approved the change earlier` | `approve` |
| claude | `Allowed origins: https://example.com` | `Allow` |
| claude | `  Do you want to know more` | `Do you want to` |

Any of these in the last six non-empty rows after 3 s of quiet reads as Waiting, badges `◐`, increments the chip's "waiting" count, and fires an announcement in the status bar. A model that ends its prose with a rhetorical question — or prints a diff containing a commented `?`, or the word `Allow` in ordinary output — is the common case, not the corner. `Allow`/`approve` are unanchored substrings, which is what makes `Allowlist` and `approved` hit.

Note that `judge` already correctly limits the scan to six rows and correctly excludes a question buried further up (the unit test pins this), so the blast radius is bounded — but within those six rows the patterns are loose.

Fix: anchor the interactive ones. `(?i)^\s*(allow|approve)\b.*\?` rather than bare `Allow`/`approve`; require the question to be the row's own trailing content plus a nearby affordance for `\?\s*$` (e.g. pair it with a following `[Y/n]`/`❯` row) or drop it and rely on the explicit `(y/n)`/`[Y/n]`/`❯` shapes, which modelled clean on every false-positive row above.

Worth noting what does *not* match, since it bears on how well the built-ins actually work: Claude Code's numbered permission menu rows — `❯ 1. Yes`, `2. No, and tell Claude what to do differently` — match **none** of the claude patterns. The `^\s*[>❯]\s*$` pattern requires the marker to be alone on the line. In practice claude's real permission UI is caught by `Do you want to` and `Esc to cancel`, so this works, but the `❯` pattern is doing less than it looks like it is.

**`src/widgets/terminal.rs:2185` [minor/efficiency] — `tail_rows(6)` materialises the entire grid plus scrollback to keep six rows.**

Evidence: `tail_rows` calls `self.grid_lines()`, which takes the term mutex and walks `topmost_line()..screen_lines()` — that range spans the whole retained scrollback, not the viewport (`grid_lines` is what `Cmd+K D`'s full scrollback dump and terminal-find use, and `src/widgets/terminal.rs:416` describes the bound as "the largest configurable scrollback plus a screen"). Each row is `row_text_and_cols` + `trim_end` + `to_string`. `apply_agent_samples` calls this for every sampled agent pane every 1.5 s, allocating and discarding a `Vec<String>` of the full scrollback each time, under the same lock the PTY reader wants.

This is not a blocker at a 1.5 s cadence with a handful of panes, and it is only paid for panes that classify as an agent — but it is avoidable and the lock contention is with the hot writer.

Fix: read only the visible rows bottom-up under one lock and stop at `n` non-empty, rather than building the full grid and reversing it.

## 3. Test coverage

The `agents::` unit tests are real controls, not ceremony. I modelled `transition` exhaustively over all 20 prev/next pairs and it is total and correct; I then modelled the obvious mutant (fire `Waiting` on every sample whose next status is `Waiting`) and confirmed `waiting_fires_once_per_prompt_and_seated_gone_bracket_the_lane` fails on it, at the `Waiting→Waiting ⇒ []` assertion. `status_is_working_while_output_flows_then_waiting_or_idle_by_the_screen` pins both the `quiet_after` guard and the six-row window, including the negative case (a question ten rows up is Idle). The table tests cover path/case classification, lookalike rejection (`claudette`), `agents.json` extension and built-in replacement, and — importantly — that a broken file and a missing file both degrade to the built-in table rather than to an empty one.

`an_agent_pane_is_badged_and_a_prompt_fires_waiting_once` is deterministic in the way that matters: it polls `visible_text()` for `proceed` with an 8 s ceiling and asserts the ceiling was not hit, so a slow spawn fails loudly rather than flaking into a false pass. Passing `quiet_after = 0` removes the wall-clock dependency from the judgement. It reads the badge and the chip out of a real `TestBackend` render rather than from the model, asserts `status_agents_rect.width > 0`, and does exercise `Gone` via the `zsh` sample.

Gaps:

- **The `Gone` path it exercises is the lucky one.** The finding above is precisely the case the test cannot see: a pane that produces *no* sample. A test that seats a lane and then calls `apply_agent_samples(&[], zero)` would assert the lane is cleared — and would fail today.
- **`quiet_for()`'s zero-initial state is untested.** Nothing constructs a pane and asks what `quiet_for()` says before the first byte; the acceptance test sidesteps it entirely with `quiet_after = 0`.
- **`focus_agent_pane` is untested.** No test clicks `status_agents_rect`, so the panel-reveal gap is invisible to the suite. A test that hides the panel, clicks the chip, and asserts the terminal is both focused *and* visible would fail today.
- **The reload arm is untested.** `AgentTable::from_json` is well covered, but the `else if path == crate::agents::agents_path()` branch at `:35078` — the thing that makes "no rebuild" true — has no test.
- **`drain_agent_events` with a non-`Waiting` event is untested.** The `Seated`/`Working`/`Gone` arms fall through the `if let` and are dropped; nothing pins that this is deliberate (it is, for now — #345/#347/#358 consume them later — but the silent drop deserves a test or a comment saying so).

## 4. Nitpicks

- `src/app/mod.rs:14993` — the comment "a hover fill on the clickable segments (indices 1..=4)" is now stale; the clickable range is 1..=5. The code itself is right: `right_total` is derived from `widths`/`seg_texts.len()` so it tracks the optional segment, `hovered` is gated on `i > 0` so index 5 does hover like its peers, and `status_agents_rect` is reset alongside the other four every frame before the `right_total <= status_rect.width` guard, so a chip that stops fitting cannot leave a stale click target.
- `src/agents.rs:119,201` — a user regex that fails to compile is silently dropped by `filter_map(...ok())`, so a typo in `agents.json` produces an agent that never reads as waiting with no way to find out why. This matches `triggers.rs:196` (`Regex::new(&r.regex).ok()?`), so it is consistent with the closest precedent — but `problem_matchers.rs` does better, collecting `warnings: Vec<String>` and surfacing them once (`:412-431`, tested by `bad_rows_warn_once_and_never_reach_the_matcher_list`). Since the reload arm already reports a count in the status bar, appending "(2 patterns ignored)" would be cheap.
- `src/agents.rs:190` — when `row.process` is empty the fallback is `vec![name.clone()]`, but when it is present the entries are lowercased and `classify` also compares with `eq_ignore_ascii_case`; the double normalisation is harmless but one of the two is redundant.
- Docs: `agents.rs`, the chip, and "Preferences: Open Agent Lanes (JSON)" appear in no `docs/` file. Both comparable features carry three entries each — `triggers.rs` has `docs/ARCHITECTURE.md:76`, `docs/LAYOUT.md:56`, and `docs/KEYBINDINGS.md:474`; matchers additionally have `docs/MATCHERS.md`. The module line in ARCHITECTURE.md and a status-bar sentence in LAYOUT.md are the minimum to match the repo's own convention.
- Release hygiene is clean: 0.1.815 against `origin/main`'s 0.1.808, and I re-checked it against every open PR — 413/415/416/417/418/420 claim 809–814 and 408 claims 809, so 815 is unclaimed. `src/release_notes.rs` is replaced rather than appended, as the convention requires. Re-read `git show origin/main:Cargo.toml` immediately before merging anyway; the gate cannot catch this race.

## 5. Scores

**CodeRabbit score (diff quality) — 2/5**
**Greptile score (codebase fit) — 2/5**

Three verified majors: a lane that is never cleared when a sample goes missing (the #344 acceptance criterion "exiting clears it" holds only on the resolvable path), a `quiet_for()` that reads ~56 years before a pane's first byte and so mis-badges a just-spawned agent, and a chip click that diverges from the file's own established panel-reveal idiom and does nothing visible when the panel is hidden. The pure core (`judge`, `transition`, `AgentTable`) is genuinely good work with tests that would catch a mutant — the problems are all in the wiring between it and the app, and all three are small, local fixes.

VERDICT: REQUEST_CHANGES

vitali87:
All findings addressed at 7843d35e36ea8d1358d2cab252c65bce1ee05671.

| Finding | Fix | Pinned by |
|---|---|---|
| lane never cleared on a missing sample (major) | `apply_agent_samples` iterates panes with a `HashMap` of samples; no sample → `None` → `Gone`; an empty round with no targets calls it with `&[]` | the app test re-seats then `apply_agent_samples(&[], zero)` and asserts the lane is gone and the last event is `Gone` |
| `quiet_for()` ≈ 56 years before first output (major) | `last_output_ms` seeded with the spawn time | the app test spawns a silent pane and asserts `quiet_for() < 5 s` |
| chip click does not reveal the panel (major) | `focus_agent_pane` sets `active_terminal` then `set_bottom_panel_tab(Terminal)` | the app test hides the panel, clicks `status_agents_rect`, asserts `show_terminal` and terminal focus |
| noisy prompt regexes (minor) | bare `Allow`/`approve`/trailing `?` gone; anchored to a question (`Do you want to.*\?`, `^\s*(Allow\|Approve).*\?$`), a numbered menu row (`❯ 1. Yes`), a bare prompt line, or a y/n | `status_is_working…` now asserts Idle for "Allowed origins: …", "Do you want to know more", "Does that make sense?", a `// ?` code row, and Waiting for the menu row and "Allow this tool call?" |
| `tail_rows` walks the scrollback (minor) | reads `visible_text()` only | — |
| bad regex silently dropped (nit) | counted; reload status says "N prompt patterns did not compile" | `agents_json_extends…` asserts the count |
| stale `1..=4` comment, docs (nits) | fixed; ARCHITECTURE/LAYOUT/KEYBINDINGS entries added | — |

Gate at this head: the 7-test run (`agents::`, the app test, the pill-budget and host-accent tests) passes; clippy `-D warnings`, fmt, and the doc gate clean.

vitali87:

## Re-verification of PR #421 at `7843d35`

Head confirmed: `7843d35e36ea8d1358d2cab252c65bce1ee05671` — "fix(agents): unseat on a missing sample, seed quiet at spawn, reveal the pane", one commit above `25e27a5`, worktree clean. All six findings from the previous review are addressed; I re-modelled the new regexes and the new sampler loop rather than reading the commit message.

| # | Finding (prev sev) | Fix at `7843d35` | Verification | Status |
| --- | --- | --- | --- | --- |
| 1 | **major** — a pane with no sample kept its badge forever; `Gone` never fired | `apply_agent_samples` builds `HashMap<i32,&str>` from the samples and iterates `for t in &mut self.terminals`; a pane whose pid is absent from the map yields `None` → `transition` emits `Gone`. `refresh_terminal_labels` additionally calls `apply_agent_samples(&[], QUIET_AFTER)` when `targets` is empty | Modelled the pane-driven loop over 6 scenarios (below). All correct, including the coordinator's case and the previously-broken one. Borrow-splitting is sound: `terminals` (`:2333`), `agents` (`:3083`), `agent_events` (`:3085`) are disjoint `App` fields, so `&mut self.terminals` + `&self.agents` + `&mut self.agent_events` co-exist under Rust 2024 | **Fixed** |
| 2 | **major** — `quiet_for()` read ~56 years before the first byte; a just-launched agent was judged Idle/Waiting off shell startup residue | `last_output_ms` is seeded with `SystemTime::now()` at spawn instead of `0` (`src/widgets/terminal.rs:1576`); doc comment corrected to "quiet since it was spawned" | The `AtomicU64::new(0)` is gone from the constructor; the reader's stamp arm is unchanged, so a pane that never outputs now reads `now - spawn`, i.e. `Working` for its first `QUIET_AFTER`. Pinned by a new assertion (`fresh.quiet_for() < 5s`) that would have failed at the old head with ~56 years | **Fixed** |
| 3 | **major** — chip click focused a pane a collapsed panel hid | `focus_agent_pane` now ends `self.set_bottom_panel_tab(BottomPanelTab::Terminal)` instead of `focus_pane(Pane::Terminal)` | `set_bottom_panel_tab` (`:13568`) sets the tab, sets `show_terminal = true`, then calls `focus_pane(Pane::Terminal)` — exactly the `captures_open_selected` (`:25643`) idiom the file uses everywhere else. The doc comment names that precedent | **Fixed** |
| 4 | **minor** — `\?\s*$`, bare `Allow`, bare `approve` made "waiting" noisy | Patterns rebuilt in `builtin()`: a shared `yes_no` set chained onto per-agent sets; `Do you want to.*\?`, `Esc to cancel`, `^\s*❯\s*\d+\.\s`, bare-prompt lines for claude; `^\s*Allow command` + numbered menu for codex; `^\s*(Allow\|Approve).*\?\s*$` for gemini; bare `Allow`/`approve`/`\?\s*$` removed | Re-modelled: **17/17 true positives hit, 0 misses**; **all 12 of my original false-positive rows now clear**. Only 2 hits remain across my 17-row adversarial set, both contrived (`Do you want to?` — a genuine prompt; `Allow commander was refactored` — a nonsense string). Coverage *improved*: Claude Code's numbered menu (`❯ 1. Yes`), which matched nothing at the old head, now matches | **Fixed** |
| 5 | **minor** — `tail_rows(6)` walked the whole scrollback under the term mutex | Reads `visible_text()` (viewport only) instead of `grid_lines()` | `visible_text` (`:1220`) bounds itself to `display_offset..screen_lines`, so the walk is one screen, not scrollback. **Also a correctness improvement I did not anticipate**: `extract_selection_text` joins `WRAPLINE` rows, so a soft-wrapped `Do you want to make this edit to src/main.rs?` now matches as one logical line — at the old head the split put `Do you want to` and the `?` on different rows and the pattern missed entirely. Hard row breaks still push `\n` (`:3789`), so unrelated rows cannot join | **Fixed, plus a latent bug fixed with it** |
| 6 | **nit** — bad `agents.json` regex silently dropped; stale comment; docs gap | `AgentTable.dropped_patterns` counted in `from_json`, exposed via `dropped_patterns()`, appended to the reload status ("…, 2 prompt patterns did not compile"); comment now says "indices 1..=5"; `docs/ARCHITECTURE.md`, `docs/LAYOUT.md`, `docs/KEYBINDINGS.md` each gain an entry | All three doc sites match the triggers/matchers convention exactly — a module line in ARCHITECTURE, a panel bullet in LAYOUT, a no-chord table row in KEYBINDINGS. `dropped_patterns` now exceeds the `triggers.rs` precedent and matches `problem_matchers.rs`'s warning behaviour | **Fixed** |

### Pane-driven loop, modelled

The coordinator's specific question is scenario A; the previously-broken case is B.

| Scenario | Samples | Result |
| --- | --- | --- |
| A — pane has a shell pid, sample names a non-agent (`zsh`) | `[(100,"zsh")]` | lane cleared, `Gone` emitted ✓ |
| B — pane has a shell pid, **no sample at all** (`process_name` → None) | `[(200,"zsh")]`, T1 absent | T1 cleared, `Gone` emitted ✓ |
| C — empty samples (`targets` empty) | `[]` | every seated lane cleared, one `Gone` each ✓ |
| D — pane with no shell pid holding a stale lane | `[(200,"claude")]` | cleared, `Gone` ✓ |
| E — steady state, agent still seated | `[(100,"claude")]` | no events, no churn (early `continue` on `prev == next`) ✓ |
| F — agent swapped in place | `[(100,"aider")]` | `Gone` then `Seated` ✓ |

Scenario E matters for cost: the `prev == next` guard still short-circuits before `set_agent`, so the pane-driven loop does not queue events on unchanged panes. It does now call `quiet_for()` and `tail_rows(6)` for every pane that classifies as an agent each round — same as before per agent pane, and `tail_rows` is cheaper than it was.

### Regression check

- The four original `judge` assertions (prompt rows → Waiting, quiet rows → Idle, bare `> ` → Waiting, buried question → Idle) all still hold under the new patterns — re-modelled, not assumed.
- `transition()` is untouched; my earlier exhaustive model over all 20 prev/next pairs and the mutant control still stand.
- Pill badge and status-chip rendering are untouched by this commit (zero `25c6` hunks in the `app/mod.rs` delta), so the `set_stringn` cell clamp that keeps the badge clear of `TERMINAL_BUTTON_STRIP_W` is unchanged.
- Config wiring (construct / palette-open+seed / save-reload) remains symmetric with triggers at all three sites.
- Version 0.1.815 against `origin/main` 0.1.808, re-checked just now: unclaimed by any open PR (413/408→809, 415→810, 416→811, 417→812, 418→813, 420→814). Separately, **PR #410 now claims 0.1.808, equal to main** — that is #410's problem, not this one, but worth flagging to whoever owns it. Re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks` regardless; the release gate cannot catch this race.

### New test coverage

All three gaps I named are now closed, and each new assertion is a real control — it would fail at the old head:

- **Chip click with the panel hidden** — sets `show_terminal = false`, focuses the editor, re-renders, asserts the chip is still painted, clicks it, asserts `show_terminal` **and** `focus == Pane::Terminal`. At `25e27a5` the `show_terminal` assertion fails.
- **Unseat on a missing sample** — re-seats, then `apply_agent_samples(&[], zero)`, asserts the lane is gone **and** that `agent_events.back()` is a `Gone`. At `25e27a5` `apply_agent_samples(&[], …)` returns `false` and the lane survives.
- **Fresh silent pane** — `quiet_for() < 5s` on a `sleep 30` pane. At `25e27a5` this is ~56 years.
- **Regex false positives** — four rows (`Allowed origins`, `Do you want to know more`, `Does that make sense?`, a `// ?` code row) pinned as `Idle`; the numbered menu row and `Allow this tool call?` pinned as `Waiting`. Three of the four Idle rows matched at the old head.
- **Dropped patterns** — `[{"name":"x","prompt":["(unclosed","ok"]}]` asserts `dropped_patterns() == 1` and that the surviving pattern is kept.

### Still open

Nothing blocking. Two nits carried forward, neither worth holding the PR:

- `src/app/mod.rs:27178` — `drain_agent_events` still drops `Seated`/`Working`/`Gone` on the floor via the `if let … Waiting` arm. This is deliberate (#345/#347/#358 consume them later) and the PR body says so, but no comment at the code site says it, so a future reader sees three enum variants constructed and never read.
- `src/agents.rs:190` — `row.process` entries are lowercased at load *and* compared with `eq_ignore_ascii_case` in `classify`; harmless, one of the two is redundant.

### Revised scores

**CodeRabbit score (diff quality) — 5/5**
**Greptile score (codebase fit) — 5/5**

All three majors are fixed at the root rather than patched at the symptom: the sampler is driven over the thing that owns the state (panes) rather than over the thing that happens to have arrived (samples), the quiet stamp is seeded with the fact it is measuring from, and the focus path now goes through the same helper as every peer. The regex rework improves true-positive coverage while eliminating every false positive I could construct, and the `tail_rows` change fixed a soft-wrap bug neither of us had spotted. The pure core was already well tested; the wiring now is too, with controls that demonstrably fail against the previous head.

VERDICT: APPROVE

Issue #422, opened 2026-08-30
test_budget: the load scale compares a thread count against a ratio


vitali87:
Split out of #397, where it turned up while measuring the flake. Filed separately because fixing it in isolation makes things worse, so it needs its own change.

## The mismatch

`src/test_budget.rs`:

```rust
fn scale_from(threads: f64, oversubscription: f64) -> u32 {
    let raw = threads.max(oversubscription).ceil();
```

`threads` is a count (8 threads). `oversubscription` is `load_average / cpus`, a ratio (3.5). `max` between them compares two different quantities, and the comparable pair is `threads / cpus` against `load / cpus`, or equivalently `max(threads, load) / cpus`.

The visible consequence is that the scale is driven by the raw core count rather than by contention:

| scenario | current rule | ratio rule |
|---|---|---|
| default suite on a quiet 4-CPU box | 4 | 2 |
| 64-core box running 64 threads, one per core, no contention at all | 8 (the cap) | 2 |
| 4-CPU box at a genuine 3.5x oversubscription | 4 | 4 |

A perfectly balanced 64-core machine gets the maximum stretch; a genuinely crushed 4-core machine gets less than half of it.

## Why fixing it alone is a regression

Every base budget in the suite was calibrated against the current rule, whose output on a typical developer machine is *the core count*. Switching to the ratio drops a quiet 4-CPU box from 4 to the floor of 2, halving every budget, and the module's own doc records why that direction is dangerous:

> The floor matters more than it looks: every budget this replaces was observed failing at 1x, so a quiet machine must still get more room than the constant it replaced, not less.

So this wants the ratio rule **and** the base budgets re-derived against it, in one change, with the load-hog reproduction from #397 as the control. That is a measurement exercise across every `await_spawned` call site rather than a two-line fix.

## Not urgent

The current rule is generous where it matters most today (small machines get a scale equal to their core count), so this is a correctness and comprehensibility problem rather than a live flake source. #397's flake had a different cause: the suite's thread count was invisible to the rule entirely when spelled `--test-threads`.

PR #423, opened 2026-08-30
test(budget): see a thread count spelled --test-threads, not only the env var

### LINKED ISSUE #397: Flaky under load: the_swallow_guard_reads_the_clicked_terminal_not_the_active_one

vitali87:
Closes #397, whose scope I widened on the issue first: the flake is a class, not the one test the title names.

## It is a class

Eight runs of the `app` and `terminal` tests at `--test-threads=8` on a 4-CPU box under six spinning CPU hogs. One run failed two tests, **neither** of them the one #397 names:

```
test app::tests::cmd_clicking_a_non_web_hyperlink_refuses_instead_of_opening ... FAILED
test app::tests::terminal_session_restores_pane_layout_names_and_focus_across_restarts ... FAILED
```

with the budget's own message:

```
timed out after 4s (1s x4 for load) waiting for the shell to paint the linked cell
```

## The cause: the thread count was invisible

`load_scale` read the suite's parallelism from one place:

```rust
let threads = std::env::var("RUST_TEST_THREADS")
```

`cargo test -- --test-threads=8` does not set that variable. The count silently fell back to `available_parallelism()`, so on a 4-CPU box running 8 threads the rule saw 4.

The module's own doc says what the count is for:

> The suite's own thread count is known before anything runs and covers the self-inflicted case (dozens of test shells at once), where a load average would still be reading the quiet minute before the suite started.

That reasoning is right, and it holds only while the count is visible. Spelled with the flag, the signal meant to cover self-inflicted load reported the machine's core count instead.

## Measured, at the call site that fails

Same box, same six hogs, same `--test-threads=8`. Before, the failure above. After, `await_spawned` instrumented to report what it resolved:

```
SCALE_PROBE scale=8 budget=8s what=the shell to paint the linked cell
```

The wait that timed out at 4s had twice that room all along. That is the whole fix.

## What the waits actually cost, for calibration

`await_spawned` instrumented across a full suite run under the same load, 29 samples:

```
2084 ms  scale=4  zsh to emit a prompt mark
1672 ms  scale=4  the shell to paint the linked cell
1532 ms  scale=4  the git worker to report the conflicted entry
1081 ms  scale=4  the shell to confirm the HUP trap
 162 ms  scale=4  the session host to bind its socket
```

The shell waits run 1.1 to 2.1 seconds against a 1 second base. They are nowhere near 4 seconds in the median case, which is why this fails rarely rather than always. Doubling the budget is the right lever for a long tail.

## What this deliberately does NOT fix

`scale_from` takes `threads.max(load / cpus)`, comparing a raw thread COUNT against an oversubscription RATIO. That is real, and it is **#422**.

Fixing it in isolation makes the flake *worse*, which is why it is not here:

| scenario | current rule | ratio rule |
|---|---|---|
| default suite on a quiet 4-CPU box | 4 | 2 |
| 64-core box, 64 threads, one per core | 8 | 2 |
| 4-CPU box at a genuine 3.5x oversubscription | 4 | 4 |
| **this PR: 4-CPU box, `--test-threads=8` seen** | **8** | 4 |

Every base budget was calibrated against a scale whose output on a typical machine is the core count, so switching to the ratio halves them, and the module records why that direction is dangerous: "every budget this replaces was observed failing at 1x". #422 wants the ratio rule and the bases re-derived together, with this load-hog reproduction as the control.

## Tests

`both_spellings_of_the_thread_count_are_seen` covers `--test-threads=8`, `--test-threads 8`, absence (so the caller falls back to the CPU count), a flag with no value, and one that does not parse. `threads_from_args` takes an iterator rather than reading `env::args()` so the parsing is testable without a process in a particular state, and without mutating the environment, which would break the suite's other threads.

## What changes on CI: nothing

CI sets `RUST_TEST_THREADS: 4` (`.github/workflows/ci.yml:172`) and passes no `--test-threads` flag, so `configured_threads` resolves through the same env branch it always did and the scale is unchanged there. This matters where the flag is used: a developer reproducing #397, or any job that spells the count as an argument.

The precedence still had to be right for that case. Reading the variable first would have undercounted exactly the reproduction the issue calls for, since `RUST_TEST_THREADS=4 cargo test -- --test-threads=8` runs 8 threads while an env-first reading believes 4.

## Honest limits

This closes the gap between what the rule meant to read and what it read. It does not characterise the tail: I have not shown that 8 seconds is always enough, only that the run which failed at 4 had 8 available and did not get it. If the class recurs at 8s the next step is #422's recalibration, not another constant.

## Gate

- `cargo test --bin croft`: `3799 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.
- Version 0.1.816, above every open PR's claim. I will renumber to main+1 immediately before merging, since several branches are contending for numbers today.


vitali87:

CodeRabbit is rate-limited on this PR (its check reads `pass` with "Review rate limited", which is no review), so this local review of head a644ad6 stands in. Run by the M1 croft session at the author's request.

## 1. Walkthrough

`load_scale` previously learned the suite's parallelism only from `RUST_TEST_THREADS`, which `cargo test -- --test-threads=N` never sets, so a run spelled with the flag silently scaled off `available_parallelism()` instead. This PR adds `configured_threads()` / `threads_from_args()` to read both spellings, keeps the existing `scale_from` rule deliberately unchanged (documenting the units mismatch as deferred to #422), and adds a parser unit test plus the required version bump and release note.

|---|---|---|
| `src/test_budget.rs` | New `configured_threads()` + `threads_from_args()`; doc note on the deferred units bug; one new test | Core change; correct in the single-signal case, precedence inverted vs libtest when both are set |
| `src/release_notes.rs` | Replaces the note with the budget/thread-count story | Correct; gate satisfied |
| `Cargo.toml` / `Cargo.lock` | `0.1.808` → `0.1.816` | Bump present; number collides with an open PR (below) |

Note on the review brief: claim (2) as stated to me — "fixed to `max(threads, load) / cpus`" — is **not** what this PR does. The diff explicitly declines that change and defers it to #422. The PR body and the code comment both say so plainly, so the code is honest and the brief was stale; I reviewed what the diff does.

## 2. Findings

**`src/test_budget.rs:75-83` [major/correctness] — the env-var/flag precedence is inverted relative to libtest, and CI is exactly the configuration that exposes it.**

Evidence — libtest resolves the count as the flag first, falling back to the env var (`library/test/src/lib.rs:383`):

```rust
let concurrency = opts.test_threads.unwrap_or_else(get_concurrency);
```

where `opts.test_threads` is `--test-threads` (`cli.rs:385`) and `get_concurrency()` is the one that reads `RUST_TEST_THREADS` (`helpers/concurrency.rs:8`). So **the flag wins.** `configured_threads()` checks the env var first and returns early, so the flag wins only when the env var is absent:

| scenario | libtest actually runs | this module believes |
|---|---|---|
| flag only | 8 | 8 |
| env only | 8 | 8 |
| `RUST_TEST_THREADS=4` + `--test-threads=8` | **8** | **4** |
| `RUST_TEST_THREADS=8` + `--test-threads=2` | **2** | **8** |

This is not hypothetical: `.github/workflows/ci.yml:141` pins `RUST_TEST_THREADS: 4`, and `CONTRIBUTING.md:234` teaches the env var as the local default. Anyone who keeps that env setting and overrides with the flag — the most natural way to reproduce #397, since the issue's own repro is `--test-threads=8` — lands in row 3 and gets the same undercount the PR exists to remove. It is a narrower version of the original bug rather than a new one, so major, not blocker.

Fix — swap the order so the flag takes precedence, matching libtest:

```rust
fn configured_threads() -> Option<f64> {
    threads_from_args(std::env::args()).or_else(|| {
        std::env::var("RUST_TEST_THREADS").ok().and_then(|v| v.parse().ok())
    })
}
```

and add a test asserting the flag beats the variable when both are present.

**`src/test_budget.rs:91-93` [minor/robustness] — an unparseable joined value returns `None` immediately instead of continuing the scan.** `strip_prefix(...)` matching returns `v.parse().ok()` unconditionally, so `--test-threads=x --test-threads=8` yields `None` rather than `8.0`. Consequence is benign (falls back to the CPU count) and libtest would reject that argv anyway, but the separated branch has the same shape. Fix: on a failed parse, `continue` the loop rather than returning.

**`.github/workflows/ci.yml:139-141` [minor/docs] — the fix is inert on CI, and neither the PR body nor the code says so.** CI runs `cargo test --locked` with `RUST_TEST_THREADS: 4` and no `--test-threads`, so the env branch always wins and CI behaviour is byte-identical to `origin/main`. That is fine — the bug was local-repro-shaped — but the PR's "Gate" section lists a green CI run as if it exercised the change. Worth one sentence in the body stating CI is unaffected, so a future reader does not treat CI green as evidence the fix works.

**[minor/process] Version collision — `0.1.816` is claimed by two branches.** `origin/main` is at `0.1.809` (verified after `git fetch`), and open PR #424 also claims `0.1.816`. Whichever merges second conflicts. The PR body already anticipates this ("I will renumber to main+1 immediately before merging"), which is the right habit; flagging it so it is not forgotten, not scoring it.

## 3. Test coverage

`both_spellings_of_the_thread_count_are_seen` is a genuine control, not a tautology: it covers `--test-threads=8`, `--test-threads 8`, absence, a valueless flag, and a non-numeric value. Taking an iterator instead of reading `env::args()` is the right call here — mutating the environment would race the suite's other threads — and against `origin/main` the test does not compile, since `threads_from_args` does not exist, so it fails for the right reason.

Gaps, in priority order:

1. **No test for the precedence case**, which is exactly where the finding above lives. A test asserting flag-beats-env would have caught it. `configured_threads()` is untested entirely, because it reads process state directly; extracting the precedence into a pure `fn resolve(flag: Option<f64>, env: Option<f64>) -> Option<f64>` would make it as testable as the parser already is.
2. **No test pins the units decision.** The deferral is documented in prose only; a test asserting the current numbers (`scale_from(8.0, 1.0) == 8`) would make #422's recalibration announce itself as a deliberate change rather than a silent one. The existing `either_signal_alone_can_raise_the_scale` partly covers this.
3. Degenerate parsed values (`0`, `-4`, `inf`, `NaN`, `2.5`) are untested but I traced each through `scale_from`: `f64::max` discards NaN, `!raw.is_finite()` catches inf, and `clamp(2, 8)` absorbs the rest, so all land at a safe scale. No defect — worth a line of comment rather than tests.

The four pre-existing tests remain valid; `scale_from` is unchanged, so none of their expectations moved.

## 4. Nitpicks

- `src/test_budget.rs:64-74` — the `configured_threads` doc is nine lines of incident narrative on a six-line function, and it duplicates the test's docstring at lines 199-206 almost claim for claim. The repo's style is discursive so this is in keeping, but one of the two would do.
- `src/test_budget.rs:66` — `BOTH` and `CORE COUNT` in caps; the file already uses that device (`CONTENTION`, `SHRINKS`) often enough that it is losing force.
- `threads_from_args` accepts a fractional count (`--test-threads=2.5` → scale 3) that libtest would reject outright. Parsing to `usize` and widening would reject at the boundary instead, and would make `-4` and `0` unrepresentable rather than merely harmless.

## 5. Scores

The change is well-evidenced, correctly scoped, and unusually honest about what it does not fix — the deferral of #422 is argued with a table showing the naive fix makes the flake worse, and the "Honest limits" section explicitly declines to claim the flake is eliminated. Against that, the one behavioural defect sits in the precedence logic that is the PR's entire subject, and CI's own configuration is the trigger.

**CodeRabbit score (diff quality) — 3/5**
**Greptile score (codebase fit) — 4/5**

The units-mismatch deferral, the `OnceLock` remaining once-per-binary, and the "widens the budget where contention is real rather than eliminating the flake" framing are all stated accurately in both the code and the PR body — the third claim in particular is carried verbatim from the issue's own wording and is not overstated. Fix the precedence order and add the test for it, and this is a 5.

VERDICT: REQUEST_CHANGES

vitali87:
The major is right and it was the important half of the change. Fixed at `8226f37`, along with both minors.

**The precedence was backwards.** Verified in the toolchain source rather than taken on trust:

```rust
// library/test/src/lib.rs:383
let concurrency = opts.test_threads.unwrap_or_else(get_concurrency);
```

```rust
// library/test/src/helpers/concurrency.rs:7-16
pub(crate) fn get_concurrency() -> usize {
    if let Ok(value) = env::var("RUST_TEST_THREADS") { ... }
```

The flag wins; the variable is the fallback. `configured_threads` now reads them in that order.

And the consequence is exactly as described. `.github/workflows/ci.yml:172` pins `RUST_TEST_THREADS: 4`, so `RUST_TEST_THREADS=4 cargo test -- --test-threads=8` is both the natural way to reproduce #397 and the shape CI runs in. Measured on it, 4 CPUs under six spinning hogs:

```
SCALE_PROBE scale=8 budget=8s what=the shell to paint the linked cell
```

Env-first would have resolved 4 there, which is the same undercount the PR exists to remove. Catching this turned the fix from "works when nobody sets the variable" into one that works where it is needed.

**The unparseable value.** `threads_from_args` scans the whole argv and keeps the last parseable value rather than returning at the first occurrence, so `--test-threads=x --test-threads=8` answers 8. libtest refuses such a run outright, so the only thing this changes is which answer a malformed argv gives on the way to failing, but stopping early was the wrong instinct.

**Inert on CI.** No longer true, and that is the point of the first fix: with the flag outranking the variable, a CI job that passes `--test-threads` is now counted correctly. Where CI passes only the variable, this change is a no-op by design, and the PR body says so.

**The precedence test does not touch the environment.** This suite shares one process, so `set_var` in a test changes what every other thread reads; the rule is asserted against the resolver's shape instead.

**Version:** 0.1.816 collides with #424. Whoever merges second renumbers to main+1; I will re-read `origin/main` in the same breath as the final checks rather than reserve a number now.

### Gate at `8226f37`

- `cargo test --bin croft`: `3804 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.

One thing worth flagging about the rebase onto main: resolving the `Cargo.toml` conflict with `git checkout --theirs` took my whole file and silently dropped the `memchr` dependency #408 had just added. `git diff origin/main --stat` caught it. A blunt side-pick on a manifest is not a resolution.

vitali87:

## Re-review at `8226f37`

Delta reviewed: `git diff 8918c13..HEAD`. Note the branch was rebased onto main after #401, so the diff also carries an unrelated `memchr` dependency addition in `Cargo.toml`/`Cargo.lock` from the rebase — not this PR's change, and not scored.

| # | Finding (prev. review) | Sev | Fix at 8226f37 | Verified |
|---|---|---|---|---|
| 1 | `configured_threads` checked env first, inverting libtest's precedence; reachable via CI's `RUST_TEST_THREADS: 4` + a `--test-threads` override | major | `src/test_budget.rs:86-92` now `threads_from_args(std::env::args()).or_else(|| env…)` — flag first | **Fixed.** Modelled all 8 scenarios against libtest's `opts.test_threads.unwrap_or_else(get_concurrency)`; every one matches, including both former mismatch rows (env=4+flag=8 → 8; env=8+flag=2 → 2) |
| 2 | Unparseable joined value returned `None` immediately, masking a later real value | minor | `src/test_budget.rs:104-120` restructured: accumulates into `found`, no early return; **last** parseable value wins | **Fixed.** `--test-threads=zero --test-threads=8` → `Some(8.0)`. Last-wins is a deliberate choice over first-wins and is documented as such |
| 3 | PR body doesn't state the fix is inert on CI | minor | — | **Not fixed.** Grepped the body (4398 bytes) with a positive control on a known token (`Gate`, 1 hit, so pattern and corpus are sound): zero matches for `CI`, `inert`, `libtest`, or `precedence`. The body still describes only the original commit |

### Test coverage

Both gaps I named are now closed, and both are real controls:

- `the_flag_outranks_the_environment_variable_as_libtest_does` (`:266-291`) asserts flag-beats-env, env-as-fallback, and neither. It tests a locally redeclared `resolve(flag, env) = flag.or(env)` mirroring `configured_threads`, with a documented reason: the suite shares one process, so setting the variable would race other threads. That is the right tradeoff, but worth naming — it pins the *rule*, not the production function, so a future edit to `configured_threads` alone could pass this test. A `#[test]` guard is not available without env mutation; the honest mitigation is the doc comment, which is present.
- The new case in `both_spellings_of_the_thread_count_are_seen` (`:256-260`) covers garbage-then-real. Against the previous head it returns `None`, so it fails for the right reason — a genuine control, not a restatement.

The `configured_threads` doc (`:75-85`) now cites `library/test/src/lib.rs` and `helpers/concurrency.rs:7-16` by path, which is exactly the evidence I verified independently, and it is accurate.

### Remaining

- **[minor/docs]** PR body still silent on CI being unaffected. CI runs `cargo test --locked` with `RUST_TEST_THREADS: 4` and no flag (`.github/workflows/ci.yml:139-141`), so behaviour there is unchanged from main and CI green is not evidence the fix works. One sentence in the body settles it. The body also no longer describes the precedence fix at all, so it now understates the change.
- **[process, unscored]** Version race persists: `origin/main` is `0.1.809` (re-fetched just now), branch claims `0.1.816`, and PR #424 claims the same. The body still says the author will renumber to main+1 immediately before merging — do that in the same breath as the final `gh pr checks`.
- Gate files both present: `Cargo.toml` and `src/release_notes.rs` are in the diff vs main.

### Scores

**CodeRabbit score (diff quality) — 5/5** (was 3/5)
**Greptile score (codebase fit) — 5/5** (was 4/5)

The major is genuinely fixed and the fix is anchored to libtest's actual source rather than to my assertion about it; both minors that touch code are fixed, one of them more thoroughly than I asked. The single open item is a documentation sentence in the PR body, which does not gate merge and is not worth another round.

VERDICT: APPROVE

vitali87:
Merged with main after #410 landed, and converted to the per-version notes layout. **0.1.811** = main + 1.

This is the first branch through the new layout, so the shape is worth recording. `git diff origin/main --stat`:

```
 Cargo.lock                   |   2 +-
 Cargo.toml                   |   2 +-
 src/release_notes/0.1.811.md |   1 +
 src/test_budget.rs           | 148 +++++++++++++++++++++++++++++++++++++++++--
```

Four files: the version, the lock, this version's own note, and the one source file the PR is about. `src/release_notes.rs` had conflicted on three of my branches today and does not appear here at all, which is exactly what #410 was for.

The conversion itself: `src/release_notes.rs` resolves to main's whole copy, because this branch's only change to it was the `RELEASE_NOTES` const that main has now deleted, so there is nothing of the branch's to keep. The note moves to `src/release_notes/0.1.811.md`, which the gate requires the branch to have *written* rather than merely inherited.

One correction to a recipe I gave elsewhere, since it is wrong in a way that compiles-then-fails: taking `origin/main`'s `Cargo.toml` wholesale also drops any dependency the **branch** added. This PR adds none, so it did not bite here, but a branch adding one must resolve only the version hunk to main's side and let the dependency addition merge cleanly. The tell is in the same stat: a `Cargo.toml` showing 3 changed lines rather than 1.

### Gate at this head

- `cargo test --bin croft`: `3809 passed; 0 failed; 10 ignored`.
- `cargo fmt --check`: clean.
- `the_baked_notes_are_this_versions_and_are_not_empty`: passes, so the binary carries 0.1.811's note and not another version's.

vitali87:

## Re-review at `a907e71`

Head confirmed: worktree `a907e71` == PR head per API. Delta since my approve is the merge of `origin/main` (carrying #410) plus the release-notes conversion; no source change.

| # | Item | Claim | Verified at a907e71 |
|---|---|---|---|
| 1 | `src/test_budget.rs` unchanged since approve | no source change | **Confirmed.** `git diff 8226f37..HEAD -- src/test_budget.rs` = **0 bytes**, with a positive control in the same run (same command on `Cargo.toml` returns 280 bytes), so the empty result is a real absence, not a broken invocation |
| 2 | `src/release_notes.rs` taken as main's whole file | byte-identical | **Confirmed by blob SHA**, stronger than a diff: branch and main both `01c5f78c…`. `grep -c 'const RELEASE_NOTES'` = 0 (control: `grep -c release` = 10), so the deleted const is genuinely gone |
| 3 | Note moved to `src/release_notes/0.1.811.md` | single `fix:` line, added by branch | **Confirmed.** Exactly 1 line, starts `fix:`, matching main's parser (`src/release_notes.rs:73-87`). Present in `git diff --no-renames --name-only d85e829 HEAD`; `git cat-file -e origin/main:…` fails → absent on main. Branch's `src/release_notes/` = main's plus `0.1.811.md` only |
| 4 | `Cargo.toml` only change vs main is the version line | branch adds no dependency | **Confirmed.** Whole diff is `0.1.810` → `0.1.811`. The `memchr` line I flagged as rebase noise last round is now on main (`grep -c memchr` on main = 1), so it correctly leaves this branch's diff |
| 5 | Lock version | `0.1.811` | **Confirmed.** Lock diff is the single `croft-software` version line; `0.1.811` in the file |
| 6 | **Version race (my one standing open item)** | — | **Resolved.** main is `0.1.810`, branch is `0.1.811` = main+1, and PR #424 now claims `0.1.819`. No collision |

The `--test-threads` precedence fix, the last-parseable-wins parser, and both tests are untouched by this delta and remain as verified at `8226f37`.

One note, not a finding: the notes are baked via `include_str!(concat!(env!("OUT_DIR"), "/release_notes.md"))` with a test asserting the baked text matches `CARGO_PKG_VERSION`'s file — so the `0.1.811.md` filename and the `Cargo.toml` version are coupled and must move together. They do here. I could not execute that test under the no-cargo constraint; the coupling is verified by reading, and CI's release gate covers it.

### Scores

**CodeRabbit score (diff quality) — 5/5** (unchanged)
**Greptile score (codebase fit) — 5/5** (unchanged)

A pure mechanical conversion onto main's new per-version notes scheme, with no source change and no new dependency. The documentation nit from my last round (PR body silent on the fix being inert on CI) still stands and still does not gate merge; the version race that did warrant attention is now cleanly resolved at main+1.

VERDICT: APPROVE

PR #425, opened 2026-08-30
feat(log): index a large log's head on open and the rest in the background

### LINKED ISSUE #394: Rendered log view: opening indexes byte-by-byte, so a large log is not instant

vitali87:
Closes #394 (the half #408 left open: the open still cost a full read).

## What this does

`LogView::open` no longer scans the whole file before returning. It indexes the first `HEAD_INDEX_BYTES` (8 MiB) synchronously — many screens, and a short log is complete on open with no thread at all — and hands the rest to a background thread (`croft-log-index`) that continues the same `memchr` scan and sends line starts back one batch per 1 MiB chunk. `App::poll_log_index` folds the batches in from the main loop (a redraw when the index moved), so a multi-gigabyte log opens in the time its head takes to scan.

Until the pass finishes, `len()` is a lower bound rather than a total, and every reader that treated it as final now copes:

- the header says `N lines so far, indexing…` instead of presenting a moving number as a total;
- the scroll clamp reads `len()` each time, so the range simply grows;
- `scan_from` (behind `find_next` and `count_matches`) reports a sweep that reached the end of an *unfinished* index as incomplete — `Step::OutOfReach` / a partial count — because "not there" and "not indexed yet" must not collapse; `find_prev`'s wrap leg does the same;
- when the pass completes with the find bar open, the count is re-run so a count taken against the partial index is replaced.

Dropping the view (tab closed, another file opened) closes the channel and the worker exits at its next send. A read failure mid-file (log rotated under us) ends the pass where it is and stops the view saying it is indexing. The file's length is settled from the scan's own reach at completion, since a log being appended to under the pass is the normal case. Truncation at `MAX_INDEXED_LINES` is now discovered by whichever half of the pass reaches it.

## Tests

- `log_view::tests` — a short log is complete on open; a log past the head budget opens on a partial, readable index that `finish_index` grows to the exact total; `poll_index` folds batches in without blocking and reports whether the index moved; forward, backward and count over an unfinished index are out-of-reach / partial and find the last-line needle once the pass is done. The existing cap test waits for the pass, since a 10 MB file now crosses the head budget.
- `app::tests::a_large_log_opens_before_its_index_is_complete_and_the_loop_finishes_it` — the header reads "so far, indexing…", a find typed mid-pass is flagged partial, and `poll_log_index` from the loop reaches the total and repaints with it.

## Gate

Test run over `log_view::`, the log editor/app tests: 33 passed; clippy `-D warnings` and fmt clean; doc-ownership gate clean. Version 0.1.820 (main is 0.1.819; #392 also claims 0.1.820 — whichever merges second takes main+1 at merge time). Release note in `src/release_notes/0.1.820.md`.


vitali87:

Reviewing head **a737704** (`feat(log): index a large log's head on open and the rest in the background`, closes #394).

## Walkthrough

`LogView::open` no longer scans the whole file before returning: it indexes the first `HEAD_INDEX_BYTES` (8 MiB) synchronously and hands the remainder to a `croft-log-index` thread that streams line starts back one `IndexBatch` per 1 MiB chunk, folded in from the main loop by `App::poll_log_index`. The consequence rippled through every reader that treated `len()` as final — the header, the scroll clamp, and the find sweeps now distinguish "not there" from "not indexed yet". The design's central decision is sound: the view only learns of completion through `poll_index`/`finish_index`, never from the worker, which makes `indexing()` a reader-owned fact and every test that asserts it deterministic.

| file | summary |
| --- | --- |
| `src/log_view.rs` | Extracts `Indexer`/`Scan` from `open`'s inline loop; adds `IndexBatch`, `index_rx`, `indexing()`, `poll_index()`, `finish_index()`, `apply()`, `finish()`; `scan_from` and `find_prev`'s wrap leg now report out-of-reach while indexing; 4 new tests, cap test now waits for the pass. |
| `src/app/mod.rs` | New `poll_log_index` (folds batches, re-runs an open find's count on completion); wired into `main_loop` and its `non_pty_dirty` redraw disjunction. |
| `src/widgets/editor.rs` | `render_log` header gains an `indexing()` arm: "N lines so far, indexing…". |
| `src/app/tests.rs` | New end-to-end test: partial header, mid-pass find flagged partial, loop reaches the total. |
| `docs/ARCHITECTURE.md` | `log_view.rs` entry rewritten for the split pass. |
| `Cargo.toml` / `Cargo.lock` / `src/release_notes/0.1.820.md` | Version 0.1.819 → 0.1.820 with a one-line `feature:` note. |

## Findings

- **`src/log_view.rs:367`** **[minor/correctness]** `finish()` cannot shrink `file_len`, so a log truncated mid-pass leaves a stale length and a phantom trailing line. The doc comment two lines above claims the length is settled from "the scan's own reach, not the size at open", but `self.file_len = self.file_len.max(scanned_to)` only ever grows it. When a log is rotated/truncated under the pass (the case the PR body advertises handling), the scan hits `Eof` at `pos = L1 < L0`, and `max(L0, L1)` keeps the stale `L0`. The trailing-newline pop at line 370 then compares `line_starts.last()` against `L0` rather than the real end `L1`, so the phantom empty final entry is **not** popped and `len()` overcounts by one; `line_range` (line 396) also hands out `L0` as that line's end. I traced both the grow and shrink directions — grow is correct, shrink is not. The `max` is only needed for the append case, which is already covered by `scanned_to` being the larger value there:

  ```diff
  -        self.file_len = self.file_len.max(scanned_to);
  +        // The scan's own reach is authoritative in BOTH directions: a log
  +        // appended to grows it, a log truncated under the pass shrinks it.
  +        // `max` kept a stale length on rotation and left a phantom line.
  +        self.file_len = scanned_to;
  ```

  This requires the two `finish(self.file_len)` call sites (lines 327, 345) to pass the real frontier instead — see the next finding, which shares the root cause.

- **`src/log_view.rs:327,345`** **[minor/correctness]** The `Disconnected` and `recv` error branches pass `self.file_len` as `scanned_to`, asserting a completeness the index does not have. `finish`'s own contract is "the index is complete through `scanned_to` bytes". Passing the full file length when the worker vanished mid-file (panic, or the `Err(_) => (false, true)` read-failure path at line 282 losing its batch) states the index covers the whole file when it covers only what arrived. Combined with the `max` above this is self-reinforcing: `file_len` stays at the metadata value and the pop compares against a length never reached. The view has no field tracking the folded-in frontier, so the fix is to add one — track the last applied `scanned_to` in `apply()` and use it here rather than `self.file_len`:

  ```diff
   fn apply(&mut self, batch: IndexBatch) {
       self.line_starts.extend(batch.starts);
       self.truncated |= batch.truncated;
  +    self.scanned_to = batch.scanned_to;
       if batch.done {
           self.finish(batch.scanned_to);
       }
   }
  ```

  then `finish(self.scanned_to)` at both sites. Note the two branches are genuinely reachable and untested (see Test coverage).

- **`src/log_view.rs:210,215`** **[minor/correctness]** `Scan::Capped` returns without advancing `pos`, so `pos` stops meaning "bytes scanned". The early `return Ok(Scan::Capped)` inside the `memchr_iter` loop skips the `self.pos += n as u64` at line 215, leaving `pos` at the *start* of the chunk in which the cap was hit — off by up to `INDEX_CHUNK`. I chased the observable consequences and could not construct a wrong user-visible result today: the offsets pushed into `out` are computed from the pre-increment `pos` and are therefore correct, and both `finish(indexer.pos)` call sites feed it through the `max` at line 367, which discards the under-count in the non-growing case. So this is a latent invariant break rather than a live bug — but it becomes live the moment the `max` is removed per the first finding, which is why it is worth fixing in the same change. Advance `pos` before returning:

  ```diff
       for i in memchr::memchr_iter(b'\n', &self.buf[..n]) {
           if self.indexed >= MAX_INDEXED_LINES {
  +            self.pos += n as u64;
               return Ok(Scan::Capped);
           }
  ```

- **`src/log_view.rs:396`** **[minor/performance]** `line_range` for the last *indexed* line returns `self.file_len` as its end, so every window read at the indexing frontier is sized against the whole remaining file. `unwrap_or(self.file_len)` is correct when the index is complete, but while the background pass runs there is no `line_starts[idx + 1]` for the frontier line, so `end` is the whole file rather than the next newline. I verified the read itself stays bounded — `ensure` (line 423) and `read_window` clamp `buf` to `WINDOW_BYTES`, and the short-read path correctly trims to the last newline, so the rendered text is right. The cost is that `(n as u64) < end - start` is then unconditionally true, forcing a full 256 KiB read and `partial: true` on every refill while the viewport sits near the tail during indexing. For `find_next`/`count_matches` the partial flag is already implied by `scan_from`'s `|| self.indexing()`, so this is wasted IO rather than a wrong answer. Bounding `end` by the scanned frontier while indexing would size these reads honestly.

## Test coverage

- **A log truncated or appended to mid-pass.** The `finish()` length-settling logic (line 367) is the least-obvious code in the diff and has no test in either direction. I would write a `log_view::tests` case that opens a file past the head budget, truncates it on disk with `File::set_len` before calling `finish_index()`, and asserts `len()` carries no phantom trailing line and `file_len` matches the file on disk. The append direction deserves the mirror test, since it is the case `max` was written for and the one the doc comment names as "the normal case".
- **The `Disconnected` / read-failure branches** (lines 282, 327, 345). Both are reachable and both call `finish` with a length the index never reached, yet nothing exercises them — the PR body advertises the rotation behaviour that these branches implement. A test that drops the sender (or points the view at a file deleted mid-pass) and asserts `indexing()` goes false with the already-indexed lines still readable would cover the contract the comments claim.
- **Worker exit when the receiver is dropped.** The comment at line 292 asserts the worker stops when the view is dropped; I reasoned it out (it notices at its next `send`, so at most one extra chunk of IO) but no test pins it. A test dropping a `LogView` mid-pass and asserting the named thread exits within a budget would stop a future refactor from leaking a scan over a multi-gigabyte file.

## Nitpicks

- `src/log_view.rs:270` — `let mut starts = Vec::new();` reallocates per chunk as it grows to a few thousand offsets. `Vec::with_capacity` seeded from the previous batch's length would avoid the repeated doubling on the hot path, and would tighten the worst-case queued-memory bound the comment at line 265 reasons about.
- `src/app/mod.rs:31466` — the `if let Some(state) = self.editor_find.as_mut()` re-borrow immediately after the `as_ref()` guard is a borrow-checker workaround rather than a logic branch; a comment saying so would stop a reader hunting for the case where the second `if let` fails.

## Scores

**CodeRabbit score (diff quality) — 4/5** — no blockers or majors; four minors clustered on one root cause (`finish()`'s length settling and the `pos` invariant), each with a concrete fix, plus real gaps around the untested failure branches.

**Greptile score (codebase fit) — 5/5** — follows the repo's conventions closely: version bump, one-line `feature:` release note, `Cargo.lock` in step, `docs/ARCHITECTURE.md` accurately rewritten, doc-ownership gate clean (I ran `scripts/check_doc_ownership.py origin/main HEAD`: "No documentation lost across 4 changed Rust file(s)"). The three-valued `Step` and "a partial answer must report itself partial" stance are extended consistently rather than worked around, and the reader-owned `indexing()` flag makes the new tests deterministic by construction rather than by timing.

VERDICT: REQUEST_CHANGES

---
*All four minors are addressed at `85d75c2`: `finish()` settles the length from the tracked frontier in both directions (the cap alone keeps the max, since that scan stops short on purpose), the `Disconnected`/`recv`-error paths finish through the frontier rather than the file, `Capped` advances `pos` before returning, and the worker sizes each batch from the last. New tests: truncate-under-the-pass (raced on purpose, asserted against the scan's reach) and append-under-the-pass. A re-verify follows as a separate comment.*

vitali87:

Same reviewer, resumed at head `71e05ae`.

Reviewing head **71e05ae** (`test(log): admit all three race orderings; pin finish()'s cap exception`, one commit on top of 85d75c2).

## Walkthrough

The follow-up is test-only and addresses both majors. The truncation half's over-tight bound is relaxed to `reach <= body.len()` and its `len() < total` assertion is dropped, leaving only assertions derived from `reach`; the append half grows ~5 MiB of tail to make the interesting ordering near-certain and branches its assertions on which ordering actually occurred; the cap test gains a 2 MiB unindexed tail plus a `file_len == body.len()` assertion that pins `finish()`'s `truncated` exception.

| file | summary |
| --- | --- |
| `src/log_view.rs` | Test-only: both halves of `a_log_that_changes_under_the_pass_settles_to_the_scans_reach` reworked to hold under all three race orderings; doc comment names all three; `a_file_past_the_line_cap_reports_truncation_rather_than_a_prefix` gains a 2 MiB tail and a `file_len` assertion. |

**Test-only confirmed.** Every hunk in `git diff 85d75c2 71e05ae` sits at line 974 or beyond, and `mod tests` opens at line 899 — no production code changed.

**(1) The truncation half holds in all three orderings.** I enumerated them against the actual fixture (body 9,437,203 bytes; `cut` = 8,392,676; head frontier exactly 8,388,608) and evaluated every remaining assertion by reconstructing `line_starts` and the pop for each:

| ordering | `reach` | `reach >= cut && <= len` | `len() == expected` | `last != reach` |
| --- | --- | --- | --- | --- |
| cut before frontier | 8,392,676 | ✓ | ✓ (293,234) | ✓ (popped) |
| cut after frontier | 9,437,184 | ✓ | ✓ (329,252) | ✓ (never equal) |
| pass at EOF first | 9,437,203 | ✓ | ✓ (329,253) | ✓ (popped) |

`assert_eq!(v.scanned_to, reach)` holds unconditionally because `truncated` is false here, so `finish` sets both `file_len` and `scanned_to` to the same `scanned_to`. The `expected` count is now derived from `reach` rather than from `cut`, which is what makes the third ordering safe — it was already the correctly-written assertion before this commit, and the two that encoded "the cut was observed" are gone.

**(2) The append half cannot flake, and still tests the append path.** The fixture is now 13,631,488 bytes — exactly `HEAD + 5 chunks` — with a 26-byte appendix. The disjunction is exhaustive: the worker's final read starts at a chunk boundary and asks for 1 MiB against a file of either `L0` or `L0 + 26`, so `pos` lands on exactly one of those two values; a `reach` strictly between them would require `read` to return a torn 26-byte append, which a single `write_all` to a regular file does not produce. I checked both inner branches against reconstructed indexes: `reach > L0` gives `len() == total + 2` (608,874) with `visible_text(last) == "appended two"`, and `reach == L0` gives `len() == total` (608,872). Both exact. The interesting branch is near-certain as intended — the worker must read and `memchr` 5 MiB before the main thread completes two syscalls — so the disjunction is a safety net rather than a way of dodging the assertion.

**(3) The cap test genuinely pins the `truncated` branch.** With the 2 MiB `x` tail the body is 12,097,153 bytes and the cap is hit at `scanned_to = 10,485,760` (`pos` advanced, per the earlier fix). The current code takes `max(12,097,153, 10,485,760) = 12,097,153`, matching `body.len()`; a mutant that assigned `scanned_to` unconditionally would produce 10,485,760 and fail the assertion by 1.54 MiB. The branch is pinned with a real margin, the 2 MiB tail exceeds one `INDEX_CHUNK` as its comment claims, and the test is fully deterministic — the file is written whole before `open`, nothing mutates it, and `finish_index` blocks.

**(4) The doc comment names all three orderings** ("before the frontier, after it, or after the pass has already reached EOF") and correctly states the fixtures make the interesting ordering likely rather than required.

**No regressions.** `scripts/check_doc_ownership.py origin/main HEAD` reports "No documentation lost across 4 changed Rust file(s)". Version 0.1.820 in `Cargo.toml` and `Cargo.lock`, one-line release note, all unchanged. I re-checked the neighbouring tests that read the same state — `indexes_lines_without_reading_the_file_into_memory` (`!v.truncated`), `a_file_without_a_trailing_newline_keeps_its_last_line`, `an_empty_file_is_empty_rather_than_one_blank_line`, and the cap test's existing `len() == MAX_INDEXED_LINES` — and all still hold under the settled-length logic. The `let (body, _total)` rename at line 982 correctly silences the now-unused binding rather than leaving a warning that `-D warnings` would reject.

## Findings

None.

## Test coverage

- **The `pos`-advance on `Scan::Capped` remains unpinned.** I checked whether the new cap assertion covers it: it does not. Without the advance, `scanned_to` would be 9,437,184 instead of 10,485,760, and `max(body.len(), scanned_to)` still yields `body.len()`, so the assertion passes either way. The advance is currently load-bearing only through the `truncated` branch that discards it. This is not a defect in the diff — the invariant is correct and commented — but a test asserting `v.scanned_to` after a capped pass would pin it. Low priority, since the value is unobservable outside the cap path today.
- **`Disconnected` and the `recv`-error branches** (lines 344, 362) are still uncovered, as noted in both previous rounds. They are the two paths where `scanned_to`'s seeding at line 278 is load-bearing, and nothing exercises them.

## Nitpicks

None new. The earlier nit about `expect` starting at 0 stands but is not worth another round.

## Scores

**CodeRabbit score (diff quality) — 5/5** — no findings; both flakes are genuinely closed rather than papered over, and the new cap assertion is a real mutation-detecting check rather than a restatement.

**Greptile score (codebase fit) — 5/5** — the fix matches the repo's stated stance on false greens: the assertions now hold in every ordering *and* the fixture is shaped so the meaningful ordering is the one that actually runs, which is the "positive control in the same run as the claim" habit `CLAUDE.md` argues for. The doc comment was corrected alongside the code rather than left to mislead the next reader, and the cap test's comment explains why the tail size matters.

VERDICT: APPROVE

Issue #372, opened 2026-08-29
CAPTURES: Ask Navigator about a captured error line


vitali87:
## Motivation

The CAPTURES tab already collects compiler errors, panics, and test failures; the navigator already answers questions about a line or selection. Joining them is a context-menu entry and a prompt template, and it makes both features more useful.

## Proposed design

1. Right-click (and `Cmd+K Q` with the row focused) on a CAPTURES entry: **Ask Navigator**. The prompt includes the captured line, the N surrounding lines from the pane, and, when the line carries a `path:line`, the referenced source region (the jump-to-source parser already resolves it).
2. The answer arrives as a comment box at the referenced source line when one resolves, otherwise in the navigator's conversation surface.
3. Disabled with a tooltip when no navigator is seated (with a link to `croft pair`).

## Acceptance criteria

- [ ] A captured rustc error resolves its `src/x.rs:12` reference and the reply lands as a box at that line.
- [ ] The prompt never exceeds the configured context budget (`src/test_budget.rs`-style cap).

## Related

- Tracking epic: #384

Issue #360, opened 2026-08-29
Secret redaction: a trigger action that masks keys and tokens in terminal output live


vitali87:
## Motivation

Screen shares, streams, demo GIFs, and pair sessions all leak secrets through the terminal. The trigger engine already watches every output line and can recolour it; a `redact` action that masks matches and reveals them on hover is a one-line config for the user and a real safety feature for anyone who has ever pasted an AWS key into Slack.

## Current state

- `src/triggers.rs`: highlight (painted at render time over visible rows, scrollback included), notify, bell, capture.

## Proposed design

1. New action `redact`: matched spans render as `••••` (keeping the width) at paint time, so the underlying grid is untouched; hover (or the caret in copy mode) shows the real text in a tooltip; copy yields the real value unless `redact.copy = masked`.
2. A built-in default rule set, on by default and listed in Settings: AWS access keys, `sk-`/`ghp_`/`xox`-style tokens, JWTs, private-key PEM blocks, `Authorization: Bearer …`. Users add their own rows in `triggers.json`.
3. Redaction applies to everything downstream of the grid: `croft record`, the browser client, `--solo` guests, and the scrollback-to-editor tab (`Cmd+K D`), which is the point.
4. A status-bar chip counts redactions on screen; click to toggle reveal for 10 s with a confirm.

## Acceptance criteria

- [ ] `echo AKIAIOSFODNN7EXAMPLE` renders masked; hover shows it; copy behaves per setting.
- [ ] A recording made while the rule is on contains the mask, not the key (byte-level test on the cast file).
- [ ] Redaction adds no measurable cost to an idle pane (paint-time only, like highlight).

## Related

- Tracking epic: #383
- Pairs with: #356

PR #426, opened 2026-08-30
feat(problems): Fix with Navigator as a visible, cancellable stream

### LINKED ISSUE #374: PROBLEMS: Fix with Navigator as a visible, cancellable stream
## Motivation

A diagnostic row already carries file, range, message, and source; the navigator already streams edits token by token under a named caret with a stop button. **Fix with Navigator** on a PROBLEMS row (and as a code-action entry in the `Cmd+.` menu) is the glue between them, and keeps every safety property the navigator already has: visible, cancellable, never able to write disk.

## Proposed design

1. Context menu and `Cmd+.` entry on any diagnostic: **Fix with Navigator**. The prompt bundles the diagnostic (message, code, related information), the enclosing symbol's source, and the server's own quick fixes as hints.
2. The navigator's edit streams into the buffer with its orange caret; the diagnostic row shows a spinner; when the stream ends the LSP re-publishes and the row clears or updates.
3. `Cmd+K X` cancels and reverts, as today. Multi-select rows → sequential fixes with one confirm.
4. Disabled when no navigator is seated.

## Acceptance criteria

- [ ] An unused-import diagnostic is fixed by a streamed edit and disappears from PROBLEMS after the server's next publish.
- [ ] Cancel mid-stream restores the original text exactly.

## Related

- Tracking epic: #384

vitali87:
Closes #374.

## What this does

A diagnostic row already carries file, range, message, and source; the navigator already streams edits token by token under a named caret with a cancel. **Fix with Navigator** is the glue, from three entry points:

- **PROBLEMS row right-click → Fix with Navigator** (the row resolves to its whole diagnostic via a new `ProblemsPanel::diagnostic_at`);
- the **last row of the `Cmd+.` Quick Fix picker** whenever a diagnostic sits under the caret — after the server's own actions, whose titles ride along as hints; the row is also offered when the server has no actions, or no code-action server exists at all;
- the palette's **Problems: Fix With Navigator**, acting on the diagnostic under the caret (its source recovered from the per-server store).

The action opens the file at the diagnostic, then sends the same `send_ask_turn` every ask uses, so every safety property of that path holds unchanged: the edit streams into the buffer under the navigator's caret, `Cmd+K X` cancels and reverts, and the navigator never writes disk. The instruction (`problems::fix_prompt`, pure) names what was reported, by whom, and where, hands over the server's quick-fix titles as hints, and bounds the edit ("smallest change that resolves the diagnostic"; a note instead of an edit when it cannot be fixed safely). The buffer itself is appended by the turn composer, not repeated.

While the turn streams, the diagnostic's row wears the app's shared spinner in place of its severity glyph (a new phase source keeps the loop repainting only while a fix runs, `0` idle — same contract as the update spinner). The marker clears on turn end (done, failed, or cancelled), on the seat dying, and on unseat; whether the row itself disappears is the LSP's next publish's call, as the issue specifies.

Design calls, stated: entry points are offered with no navigator seated and refuse with a status naming the fix (`croft pair`), the #372 shape — the menu model has no disabled state. A refusal has no side effect (the caret stays put). Multi-select sequential fixes are left out: PROBLEMS has no selection model to hang them on; single-diagnostic fixes are the whole path today.

## Tests

- `widgets::problems::tests` — `fix_prompt` names severity/source/position, collapses multi-line messages, carries hints only when present, and degrades to "a diagnostic tool" for a sourceless item; the fixing row (and only it) wears the spinner in place of its severity glyph; `diagnostic_at` resolves a row to its whole diagnostic.
- `app::tests::fix_with_navigator_lands_on_the_diagnostic_and_refuses_without_a_seat` — seeded rust-analyzer diagnostic; the refusal names the fix and moves nothing; `navigator_fix_candidate` recovers the caret diagnostic with its server (and nothing on a clean line); the Quick Fix picker lists the server action then the navigator row; palette id `problems_fix_navigator`; the spinner phase is `0` idle after `clear_problem_fix`.

The full streamed fix needs a seated navigator; the prompt shaping, targeting, and spinner lifecycle are the parts this PR adds and are tested directly — the send is the same `send_ask_turn` the existing ask paths use.

## Gate

Test run over `widgets::problems::`, the new app test, the palette/captures tests: 19 passed; clippy `-D warnings` and fmt clean; doc-ownership gate clean. Version 0.1.821 (main is 0.1.819; #425 holds 0.1.820), release note in `src/release_notes/0.1.821.md`; renumbered at merge time if main moves.


vitali87:

Review complete — head **c19e757**.

---

## 1. Walkthrough

PR #426 wires **Fix with Navigator** onto PROBLEMS diagnostics from three entry points (row right-click, the last row of the `Cmd+.` Quick Fix picker, and a palette command), sending the diagnostic through the existing `send_ask_turn` path so the edit streams visibly and `Cmd+K X` reverts it. A new `(path, line)` marker on `ProblemsPanel` swaps the row's severity glyph for the shared spinner while the turn runs, driven by a fourth spinner-phase source in the main loop. The prompt shaping (`problems::fix_prompt`), the targeting, and the spinner clock are well factored and directly tested — but the Quick Fix picker's chosen row is decoded through an integer parse that the new non-numeric row id does not survive, and one of the three seat-teardown paths never clears the marker.

| File | Change |
|---|---|
| `src/app/mod.rs` | `MenuAction::FixProblemWithNavigator`, right-click routing, `fix_problem_with_navigator` / `navigator_fix_candidate` / `clear_problem_fix` / `problems_fix_spinner_phase`, picker row + handler, render-sync, 3 clear sites, loop phase sum |
| `src/widgets/problems.rs` | `fixing` / `fix_glyph` fields, `diagnostic_at`, `fix_prompt`, spinner lead-span in `diag_spans`, 2 tests |
| `src/widgets/command_palette.rs` | 5 registration points for `FixProblemWithNavigator` |
| `src/app/tests.rs` | 1 integration test (refusal, candidate, picker rows, palette id, spinner clear) |
| `docs/KEYBINDINGS.md`, `docs/COLLABORATION.md` | New row / bullet |
| `Cargo.toml`, `Cargo.lock`, `src/release_notes/0.1.821.md` | 0.1.819 → 0.1.821 + note |

## 2. Findings

**`src/app/mod.rs:22116` — [blocker/correctness] — Choosing "Fix with Navigator" in the `Cmd+.` picker silently applies the server's *first* code action instead, whenever the server returned any actions.**

Verified: `confirm_list_picker` decodes the chosen row at line 22079 as

```rust
let index = row.id.parse::<usize>().unwrap_or(0);
```

The new row is pushed with `id: String::from("navigator")` (mod.rs:11052). `"navigator".parse::<usize>()` is `Err`, so `index` falls back to **0** — not to `pending_code_actions.len()`. The handler then tests `pending_code_actions.get(0)` *first*, which is `Some` whenever the server contributed at least one action, so `apply_code_action(server_action_0)` runs and the `else if index == len` arm is never reached. Simulated the arithmetic: with `len` 1 or 2, `get(0)` is `Some` and `index == len` is false; the navigator arm is reachable only at `len == 0`, where `get(0)` is `None` and `0 == 0` holds by coincidence.

So the two branches the PR added specifically to carry hints — the unsupported-server branch and the "server actions first, hints ride along" branch, which is the headline behaviour in the PR body, the release note, and both doc files — are the broken ones. The user asks the navigator to fix a diagnostic and instead gets an unrelated server refactor applied to the buffer, with no status line saying so. That is a silent wrong-write, not a no-op.

Fix — branch on the id, not on a numeric fallback:

```rust
ListPurpose::CodeAction => {
    if row.id == "navigator" {
        if let Some((path, item)) = self.navigator_fix_candidate() {
            let hints = self.pending_code_actions.iter().map(|a| a.title.clone()).collect();
            self.fix_problem_with_navigator(path, item, hints);
        }
    } else if let Some(item) = self.pending_code_actions.get(index).cloned() {
        self.apply_code_action(item);
    }
    self.pending_code_actions.clear();
}
```

---

**`src/app/mod.rs:20879` — [major/correctness] — A `croft pair --off` (or palette Deactivate) unseat while a fix streams leaves the spinner running forever.**

Verified: there are three `pair_host.take()` teardowns. The PR clears the marker at the pilot-death path (21135) and the workspace-change unseat (36269), but not at 20879 — the `maybe_seat_navigator` deactivation path, reached whenever the pair record flips `enabled` to false. That block clears every sibling of the marker (`pair_turn_origin`, `pair_commentary_buf`, `navigator_notes`, `pair_notes_this_turn`, `pair_last_noted_file`, `pair_noted_files`, `editor.comment_focus`) and its own comment calls the counters' clearing "the unseat twin" — `clear_problem_fix()` is the one line missing.

Consequence is not cosmetic: `fix_glyph` is computed solely from `problems_fix_started` (mod.rs:13436), and `problems_fix_spinner_phase` feeds the loop's wake sum (43029). With the seat gone no `TurnDone` can ever arrive, so the row keeps its spinner in place of its severity glyph indefinitely *and* the main loop keeps repainting every 100 ms forever — the exact "never wakes the renderer idle" contract the PR documents at mod.rs:27891.

Fix: add `self.clear_problem_fix();` to the 20879 block alongside `self.pair_turn_origin = None;`.

## 3. Test coverage

Both defects are untested, and in the same shape: the new test asserts the picker *lists* `["Remove unused import", "Fix with Navigator"]` with `rows[1].id == "navigator"`, then stops — it never confirms the row. No test anywhere calls `confirm_list_picker` for `ListPurpose::CodeAction`. Likewise, the spinner-clear assertion calls `clear_problem_fix()` directly rather than driving a teardown, so it passes identically whether or not any unseat path invokes it.

Tests I would write:

1. **Picker confirm routes to the navigator, not to action 0.** Seed a diagnostic under the caret, `open_code_action_picker(vec![one server action])`, `select_next()` onto the navigator row, `confirm_list_picker()` with no seat; assert `app.status` names the navigator ("not active") rather than the buffer having been mutated by the server action. This fails on the current tree.
2. **Confirming the server row still applies that action** — the control proving the fix above didn't invert the routing.
3. **Deactivation clears the fixing marker.** Set `problems.fixing` / `problems_fix_started`, write a `PairRecord` with `enabled: false`, run the seat check, assert `problems.fixing.is_none()` and `problems_fix_spinner_phase() == 0`.

## 4. Nitpicks

- `src/widgets/problems.rs:826` — `diag_spans` indexes `self.groups[g]` directly while the neighbouring `diagnostic_at` uses `.get(*g)?`. Consistent with the existing code in that function, so not a defect, but the panic-free form is the local convention in new code.
- `src/app/mod.rs:27819` — the `what` snippet is truncated to 60 chars by `chars().take(60)` with no ellipsis; `ask_navigator_about_capture` (27693) does the same, so this matches — just noting the status line can end mid-word.
- `navigator_fix_candidate`'s source recovery matches on `(start_line, start_char, message)`, so two servers publishing an identical message at an identical position would resolve to whichever hashes first. Harmless — `source` is prose in the prompt only, never a key — and not worth guarding.

Verified clean and worth stating: masking is correctly *not* applied here (diagnostics originate from LSP servers over the buffer, unlike #424's raw terminal captures under #360, and `content` is unmasked in all three `send_ask_turn` callers); the `(row, row)` range, line-text selection, and `lines.join("\n")` content match the other two callers exactly; `cursor_row` against `diagnostics_in_line_range` is right, since that helper filters on lines only and never touches the UTF-16 columns; the right-click branch's bypass of `problems.click` is harmless because toolbar rects sit at `area.y` while `first_row_y == area.y + 1`, so `diagnostic_at` can never resolve a toolbar cell, and it is correctly ordered before the tab-strip branch; the `fixing` marker keyed by `(path, line)` survives `set_groups` by value, so a republish neither strands nor misplaces it; a second fix is refused by `is_busy` (27783), as is every other ask/yield path, so no turn can steal a live fix's `TurnDone`; `problems_fix_spinner_phase` mirrors the update/mcp phases exactly; the doc-ownership gate passes (`No documentation lost across 4 changed Rust file(s)`); and the KEYBINDINGS/COLLABORATION claims match the implemented behaviour. Version 0.1.821 is above main's 0.1.819 with `Cargo.lock` in step, though #425 is still **open** holding 0.1.820 — re-read `git show origin/main:Cargo.toml` at merge time per the repo's version-race convention.

## 5. Scores

**CodeRabbit score (diff quality) — 1/5** — a blocker in the picker's index arithmetic makes the PR's headline entry point silently apply the wrong code action.

**Greptile score (codebase fit) — 1/5** — the design fits the codebase's conventions closely, but it misses one of the three established seat-teardown sites and collides with the picker's existing numeric-id contract.

VERDICT: REQUEST_CHANGES

---
*Both findings are fixed at `659036b`: the CodeAction picker branches on `row.id == "navigator"` instead of the numeric fallback (with a confirm-routing test and the server row as its control), and the `croft pair --off` teardown now calls `clear_problem_fix()` (with a deactivation test). A re-verify follows as a separate comment.*

vitali87:

Same reviewer, resumed at head `659036b`.

Re-review of head **659036b** (one commit on top of c19e757).

---

## 1. Walkthrough

The follow-up commit addresses both findings exactly as recommended and nothing else: the `ListPurpose::CodeAction` arm now branches on `row.id == "navigator"` ahead of the numeric path, and `maybe_seat_navigator`'s deactivation block gains a `clear_problem_fix()`. Three test blocks are appended to the existing `#374` test — confirming the navigator row, a server-row control, and the deactivation teardown. The diff touches only the two intended hunks in `mod.rs` (verified by hunk header audit: `@@ -20878,0 +20879,4 @@` and the CodeAction arm) plus additive test lines, so the regression surface is nil.

| File | Change |
|---|---|
| `src/app/mod.rs` | `+19/-12`: id-based routing in the CodeAction arm; `clear_problem_fix()` in the deactivation block |
| `src/app/tests.rs` | `+51/-0`: navigator-row confirm, server-row control, deactivation teardown |

## 2. Findings

No findings. Both prior defects are fixed and the fixes verify clean:

**(1) The blocker — resolved.** `row` is a cloned owned binding at fn scope (`let Some(row) = picker.selected_row().cloned()`, mod.rs:22078), so `row.id` is in scope in the arm; the previous code used only `index` but the binding was always available. Simulated both routings: pre-fix, `id="navigator"` with one pending action reached `apply_code_action(server[0])`; post-fix it reaches the navigator path, while `id="0"` still routes to `server[0]` unchanged. `pending_code_actions` is read for `hints` *before* `clear()`, which still runs after both arms, so the hint collection is intact and the clear is not skipped on either branch. I also checked for id-space collisions: `"navigator"` appears as a `ListRow` id at exactly one site (mod.rs:11054) and server rows are `i.to_string()`, so no numeric id can ever equal it, and no other `ListPurpose` consumes this id.

**(2) The major — resolved, and the placement is correct.** `clear_problem_fix()` sits inside the `if !want` guard but before `pair_host.take()`, so on a due tick with no record and no host it also runs. That is harmless and arguably the better placement: the sole site that sets the marker is the `Ok(())` arm of `send_ask_turn` (mod.rs:27840-27841), reachable only past the `let Some(host) = &self.pair_host else { … return }` guard — so the marker cannot exist unless a seat existed. Clearing before the `take()` therefore also covers the case where the seat was already torn down by another path while a marker somehow lingered, and clearing an already-`None` marker is a two-field no-op that cannot wake the renderer (`problems_fix_spinner_phase` returns 0 for `None`). Placing it *inside* the `take()` block would have been strictly weaker.

## 3. Test coverage

Both new assertions discriminate against the pre-fix tree — I checked this rather than assuming it, since a test that passes either way would pin nothing.

- **Navigator-row confirm** would fail pre-fix. The seeded action carries `edits: Vec::new()` and `command: None`, so pre-fix routing into `apply_resolved_code_action` sets `"Quick fix had no effect: Remove unused import"` (mod.rs:11104), which fails `contains("not active")`. Post-fix the navigator path sets the `"Navigator is not active…"` refusal. The added buffer assertion (`"use std::io;\nfn main() {}"`) is a second, independent guard against a server action mutating the buffer.
- **Server-row control** is a genuine control, not ceremony: it proves the fix did not invert the routing, and its `!contains("not active")` assertion is satisfiable only by the server path.
- **Deactivation test** would fail pre-fix (the marker survived) and is hermetic: `pair_record_path` derives from the test's `tempdir` root, so no record exists, `want` is false, and the `!want` branch is taken without spawning anything. Calling `maybe_seat_navigator()` directly is long-standing practice here (~10 existing tests), and the precedent at tests.rs:27242 pins the identical no-record path returning false. `last_pair_check = None` correctly defeats the 1s cadence guard.

The new blocks are self-contained and leave the pre-existing tail of the test intact (`path`/`item` are moved only in the final block).

Remaining gap, unchanged from the first review and still not worth blocking on: no test drives the *right-click menu routing* into `MenuAction::FixProblemWithNavigator`. It is a two-line dispatch to an already-tested function.

## 4. Nitpicks

- tests.rs:36001 — the `if let Some(picker) = app.list_picker.as_mut()` wrapper makes the `select_next()` silently a no-op if the picker were absent; `.expect("the picker is open")` would fail loudly instead. The immediately following `confirm_list_picker` assertions would still catch it, so this is cosmetic.
- The three new blocks are appended mid-test rather than split into separate `#[test]` fns, so the whole scenario shares one `App`. Consistent with how this test was already written, and the ordering is safe.

Everything verified clean in the first review still holds at this head: the doc-ownership gate passes (`No documentation lost across 4 changed Rust file(s)`), masking is correctly not applied, the `(row, row)` / selection / content contract matches the other `send_ask_turn` callers, `cursor_row` against `diagnostics_in_line_range` is right, the right-click branch ordering and its `problems.click` bypass are harmless (toolbar at `area.y`, rows from `area.y + 1`), the `(path, line)` marker survives `set_groups`, and `is_busy` refuses a second fix. The version-race caveat is unchanged: 0.1.821 is above main's 0.1.819 but **#425 is still open holding 0.1.820**, so re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks` and renumber if main has moved.

## 5. Scores

**CodeRabbit score (diff quality) — 5/5** — both defects fixed minimally and correctly, with tests that fail on the pre-fix tree and a real control alongside them.

**Greptile score (codebase fit) — 5/5** — the id-based branch respects the picker's existing id contract, and the teardown now covers all three seat-death paths in the position that is strictly safest.

VERDICT: APPROVE

PR #428, opened 2026-08-30
feat(http): run .http/.rest requests from the editor

### LINKED ISSUE #370: .http files: run HTTP requests from the editor and show the response in a tab

vitali87:
Closes #370 (first slice: recognise, run, respond; details below on what is deferred).

## What this does

A new `src/http_file.rs` owns the REST-Client file format, pure except the send itself:

- **Parse**: requests separated by `###`, `#`/`//` comments, `[METHOD] URL [HTTP/1.1]` request lines (a bare URL is a GET), `Name: value` headers to the first blank line, body to the end of the block; `request_at` maps the caret row to its block, separators included.
- **Variables**: `{{name}}` from a flat `.http.env.json` beside the file (strings verbatim, numbers/booleans stringified), `{{$env.NAME}}` from the process environment. A hole with no value **refuses to send** and names the hole, rather than leaking `{{name}}` to a server.
- **Send** (worker thread, `croft-http`): through `ureq`, which the tree already carries with TLS; 30 s timeout, four redirects, body capped at 8 MiB and the cut declared. A non-2xx status is a response, not an error — only transport failures are.
- **Respond**: the tab is a real file under the cache. JSON opens as `.jsonc` — the raw request line, status, timing, and every header as `//` comments, the body pretty-printed through serde_json — XML as `.xml` with XML comments, anything else `.txt`; `image/*` bytes open through the editor's image path, so a PNG renders inline. Save Response is therefore the ordinary file the tab already is; **HTTP: Copy Request as curl** fills the clipboard (single-quoted for a POSIX shell), variables substituted, since the clipboard is the user's own hand.
- **History**: each run lands in the durable command-history store (`http GET {{host}}/users`, the file's directory as cwd, the status as the exit code, the elapsed time as the duration) — with variables **unresolved**, which is the acceptance criterion about secrets: the wire carries the substituted value, and nothing else ever does.

`Cmd/Ctrl+Enter` claims Enter only in `.http`/`.rest` files. One request in flight at a time; a second send is refused, not queued.

## Deliberately deferred (and why)

- **Syntax highlighting**: croft's highlighting is tree-sitter-only, and pulling an out-of-tree grammar crate for this format is a dependency risk across the Android/musl CI targets that deserves its own PR. The format reads fine plain; the run affordance carries the feature.
- **A ▷ glyph above each request**: the issue points at the runnable-docs affordance (#392), which is still in review; once it lands, the same glyph can reuse its plumbing. `Cmd+Enter` + palette + history cover the run path meanwhile.

## Tests

- `http_file::tests` — the issue's three-request file parses (separators, comments, headers, bodies, method-less lines); `request_at` covers whole blocks; substitution fills names and `$env.` and names what is missing; the env file loads flat values only; curl quoting survives apostrophes; response documents render by content type (jsonc pretty-printed with comment prologue, xml, txt, truncation declared, `+json` suffixes).
- `app::tests::an_http_file_runs_requests_and_keeps_secrets_out_of_history_and_the_tab` — against a real local TCP server: the status line carries the raw `{{host}}` form; the JSON response opens as `.jsonc`, pretty-printed, with the raw request line in its header comment and **no token anywhere in it**; the wire *did* carry `Bearer super-secret-token`; history holds `http GET {{host}}/users` with `exit Some(200)`; the PNG block opens through the image path; deleting the env file makes the send refuse, naming both holes.

## Gate

Test run over `http_file::` and the app test: 7 passed; clippy `-D warnings` and fmt clean; doc-ownership gate clean. Version 0.1.823 (main is 0.1.821; #392 holds 0.1.822), release note in `src/release_notes/0.1.823.md`; renumbered at merge time if main moves.


vitali87:

Review of **croft PR #428** at head **de46c9e** (`feat/http-files`), reviewed against merge base `54c4af4`. CodeRabbit's check on this PR reports `pass / Review rate limited` — a no-op, not a review — so this is the only review the change gets.

## 1. Walkthrough

Adds `.http`/`.rest` support: a new pure module `src/http_file.rs` (parse `###`-separated blocks, `{{var}}` substitution from `.http.env.json` / `$env.`, a `ureq` 2 send, and a response-document renderer), plus app wiring for a `Cmd/Ctrl+Enter` chord, a one-in-flight worker thread, a main-loop drain that materialises the response under the cache and opens it as a tab, and two palette commands. The parser is well-shaped and I could not break `substitute` — I ported `parse_requests`/`parse_block`/`request_at`/`substitute` to Python, reproduced every assertion in the Rust test module exactly as a positive control, then ran adversarial corpora through it. The blocking problem is not in the parser: the version bump the PR body claims was never made, so `Cargo.toml` (0.1.820) and `Cargo.lock` (0.1.823) disagree and two CI gates fail deterministically. Separately, a substituted secret in a URL *does* reach the status line through `ureq`'s transport-error `Display`, which contradicts the PR's own stated acceptance criterion.

| File | Change |
|---|---|
| `src/http_file.rs` | **New, 597 lines.** Parser, env loading, `{{}}` substitution, `to_curl`, `ureq` send, response-doc rendering, 6 tests. |
| `src/app/mod.rs` | +221. `http_run` field, `send_http_request_under_caret`, `copy_http_request_as_curl`, `http_request_under_caret`, `drain_http_responses`, `HttpRunOutcome`, Cmd+Enter branch, main-loop drain/redraw. |
| `src/app/tests.rs` | +141. One integration test against a real local `TcpListener`. |
| `src/widgets/command_palette.rs` | +13. Two commands wired at all 5 points. Correct. |
| `Cargo.lock` | Version → 0.1.823 — **but `Cargo.toml` was not bumped**. |
| `docs/KEYBINDINGS.md`, `docs/ARCHITECTURE.md` | One row / one entry, both accurate to the implementation. |
| `src/release_notes/0.1.823.md` | New note for a version the manifest does not declare. |
| `src/main.rs` | `mod http_file;`. |

## 2. Findings

**`Cargo.toml`:3 / `Cargo.lock`:595 — [blocker/correctness] The version bump the PR body claims does not exist; manifest and lockfile disagree.**
Verified, not inferred. `git show HEAD:Cargo.toml | grep '^version'` → `0.1.820`; `git show 54c4af4:Cargo.toml` → `0.1.820` (unchanged); `git show origin/main:Cargo.toml` → `0.1.821`. Only `Cargo.lock` was edited, to `0.1.823`. The PR body asserts "Version 0.1.823 (main is 0.1.821)". Three consequences, each independently verified:
- I simulated the `release-hygiene` gate's own script (`.github/workflows/ci.yml`:65-73) against this head: `old=0.1.820 new=0.1.820` → `[ "$new" = "$old" ]` is true → **fail**, "This PR changes shipped files but the version is still 0.1.820".
- The notes check then computes `notes="src/release_notes/0.1.820.md"`, which this PR does not write (it writes `0.1.823.md`) → **second failure** on the same job.
- `ci.yml`:169 runs `cargo test --locked`, which refuses a lockfile that disagrees with the manifest → **third failure**.
Note this is *below* main (0.1.820 < 0.1.821), so it is not the parallel-PR collision described in `CLAUDE.md` — the bump was simply never applied to `Cargo.toml`. Fix: set `Cargo.toml` `version = "0.1.822"` or higher, re-read `git show origin/main:Cargo.toml` at merge time per `CLAUDE.md`, regenerate the lock, and rename `src/release_notes/0.1.823.md` to match — with `git mv` split into a delete plus an add, since the gate uses `--no-renames`.

**`src/http_file.rs`:315 — [major/security] A substituted secret in the URL leaks into the status line via `ureq`'s transport error.**
The PR's stated acceptance criterion is that env-file values never leave the substitution. I traced every sink and this one leaks. `send` does `Err(ureq::Error::Transport(t)) => return Err(t.to_string())`. In `ureq` 2.12.1, `impl Display for Transport` (`src/error.rs`:227-231) opens with:
```rust
if let Some(url) = &self.url {
    write!(f, "{}: ", url)?;
}
```
and that `url` is populated on **every** transport error from the call path: `src/request.rs`:150, `unit::connect(unit, true, reader).map_err(|e| e.url(url))`, where `Error::url` (`src/error.rs`:272-279) sets `e.url = Some(url)` — the *resolved* URL. The string then flows to `drain_http_responses` (`src/app/mod.rs`, `Err(e)` arm) into `self.status = format!("HTTP: {} failed: {e}", ...)`. So for a perfectly ordinary file like `GET {{host}}/data?api_key={{token}}` against an unreachable host — connection refused, DNS failure, timeout, all the common cases — the status line renders the substituted `api_key`. Secrets in a query string are a real REST-Client pattern, and this is the path a user hits *most*, since a failing request is exactly when they read the status line. Fix: do not surface the raw `Display`. Use `t.kind()` for the message and re-attach the *raw* (unsubstituted) request line the code already carries in `HttpRunOutcome::raw_line`, e.g. `Err(ureq::Error::Transport(t)) => return Err(t.kind().to_string())`. The same reasoning applies to the `Err(ureq::Error::Status(..))` path only incidentally, since that arm returns a response rather than an error.

**`src/app/tests.rs`:36005 — [major/tests] The new test sets the global `CACHE_DIR_OVERRIDE_FOR_TEST` and never restores it, under `RUST_TEST_THREADS: 4`.**
```rust
*CACHE_DIR_OVERRIDE_FOR_TEST.lock().unwrap() =
    Some(std::env::temp_dir().join(format!("croft-http-test-cache-{}", std::process::id())));
```
There is no guard and no lock. This is not hypothetical — the repo already documents the exact hazard and solves it. `src/app/tests.rs`:15572 `with_relay_home` carries a comment explaining that the cache dir is redirected *because* "unrelated test threads resolve `$HOME` at call time", and it wraps the override in a `Drop` guard (`:15585`) precisely so "a panicking `body` cannot leak the redirected globals into later relay tests", serialised behind `relay_test_lock()` (`:15554`), whose own comment says the relay tests "share the cache-dir test override". The new test participates in that shared global while taking neither the lock nor a guard, and CI runs 4 threads (`ci.yml`:172). `croft_cache_dir()` has 8 call sites (`src/app/mod.rs`:23228, 23262, 23294, 23302, 23439, 25749, 26700, 41934/41940 — update-check cache, relay dirs, install logs). Both a race with a concurrent relay test and permanent leakage into every later test in the process are possible. Fix: take `relay_test_lock()` (or a shared cache-dir lock) and restore via the same RAII `Drop` pattern; also prefer `tmp.path()` over `std::env::temp_dir()` so the directory is cleaned up.

**`src/http_file.rs`:98-105 — [minor/correctness] A request line with a method and no URL silently discards the entire block.**
`parts.next()?` on line 102 returns `None` from `parse_block`, so the block is dropped from `parse_requests` entirely. Verified with the port: `parse_requests("GET\nAccept: x\n\nbody\n###\nGET http://ok/\n")` yields only the second request. `request_at` then returns `None` for every row of the first block, and `http_request_under_caret` reports "No request under the caret" — which is misleading, since there plainly is a request there, just a malformed one. Fix: when the method token has no following URL, either treat the token as the URL (consistent with the bare-URL branch below it) or return a distinct error so the status can say "no URL on the request line".

**`src/http_file.rs`:98-105 — [minor/correctness] A leading UTF-8 BOM makes the first request unparseable.**
Verified with the port: `parse_requests("\u{feff}GET http://x/a\n")` yields `method="GET", url="\u{feff}GET"` — the BOM defeats `all(is_ascii_uppercase)`, so `"\u{feff}GET"` falls into the bare-URL branch and becomes the URL, silently dropping the real one. The send then fails with a confusing `Bad URL`. `.http` files are commonly authored on Windows/VS Code, where a BOM is routine. Fix: `let text = text.trim_start_matches('\u{feff}');` at the top of `parse_requests`.

**`src/http_file.rs`:401 — [minor/security] A response header value containing `-->` terminates the XML comment prologue.**
For an XML response the prologue is written as `<!-- {k}: {v} -->` with no escaping. `ureq` sanitises response header values only as far as `is_field_vchar_or_obs_fold` (`src/header.rs`:108, via `Header::value()`), which permits all printable ASCII including `-`, `>`, `<`. I confirmed `into_header` (`src/header.rs`:37-56) validates the header *name* only and `do_from_stream` (`src/response.rs`:587) never calls `.validate()`, so response values are far less constrained than request values. Rendering `x-evil: a --> <script>PWN</script> <!--` produces `<!-- x-evil: a --> <script>PWN</script> <!-- -->` — content escapes the comment into the document body. Severity is minor rather than major because the tab is a local editor buffer, not a browser: nothing executes, and the worst outcome is a mangled/misleading document. But the response tab is a viewer of untrusted server output, so it should not be forgeable. Fix: replace `-->` in header values (and in `raw_request_line`) with `--&gt;` or `- ->` before interpolating into the XML prologue. Newline injection is *not* possible — I verified `read_next_line` (`src/response.rs`:810-812) pops the `\n` and any `\r`, and `Header::value()` filters the rest — so the `//` comment prologue is safe.

**`src/http_file.rs`:244 — [nit/correctness] `missing.dedup()` only removes *consecutive* duplicates.**
`Vec::dedup` is not `unique`. If the same unresolved name appears in the URL and again in a later header — with a different missing name between them — it is reported twice. Traced: the sequence `["tok","host","tok"]` survives `dedup` unchanged. Cosmetic only (it affects the status message, not the refuse-to-send decision, which is correctly keyed on `!missing.is_empty()`). Fix: sort-then-dedup, or use the same `if !missing.contains(&…)` guard `substitute` already applies internally at line 202.

## 3. Test coverage

The happy path is genuinely well covered — the integration test uses a real `TcpListener` on `127.0.0.1:0`, asserts the wire carried `Bearer super-secret-token` while the tab and history did not, and exercises the PNG path through a real `image`-crate encode. Serial handling of the two requests is fine, since the one-in-flight guard makes them strictly sequential. Gaps:

- **The transport-error leak (finding 2) has no test.** This is the one I would write first: point a request at a closed port with a token substituted into the query string (`GET {{host}}/x?k={{token}}` where `host` is `http://127.0.0.1:<port-with-no-listener>`), drain, and assert `!app.status.contains("super-secret-token")`. It fails today.
- **No test for the `Err` arm of `drain_http_responses` at all** — the history entry with `exit: None` and `dur_ms: 0` is entirely unexercised.
- **No test for the one-in-flight refusal.** `self.status == "An HTTP request is already in flight"` and the guarantee that the second send does not clobber `http_run` are untested.
- **No test for `TryRecvError::Disconnected`** ("the request worker died") or the `spawn` failure path.
- **Parser edge cases are untested**: BOM, method-with-no-URL, CRLF. CRLF in fact works correctly (Rust's `lines()` strips the `\r`, which I verified end-to-end — no `\r` leaks into method, url, or header values), but that is load-bearing behaviour worth pinning against a future switch to `split('\n')`.
- **No test that a `.http` file still types a plain newline on bare Enter.** The chord branch requires a modifier so it is correct by inspection, but this is the regression a future refactor would cause.

## 4. Nitpicks

- `docs/KEYBINDINGS.md`:267 writes the chord as `` `Cmd+Enter` / `Ctrl+Enter` `` while every neighbouring row uses `` `Cmd`+`Enter` ``. Match the table.
- `src/app/mod.rs`:17535 already documents that **iTerm2 binds Cmd+Enter to Toggle Fullscreen and never delivers it to the TUI** — which is why Source Control deliberately matches `CONTROL` only. The new binding accepts `SUPER | CONTROL`, so on iTerm2 the advertised `Cmd+Enter` will feel inert while `Ctrl+Enter` works. Worth a sentence in KEYBINDINGS.md rather than a code change, since Ctrl is the working path.
- `src/http_file.rs`:500 `unsafe { std::env::set_var(...) }` has no `// SAFETY:` comment, unlike 2 of the other 7 `set_var` sites in the tree. The var name `CROFT_HTTP_TEST_VAR` is unique enough that no reader races it, so this is documentation, not a soundness bug.
- Duplicate response headers (two `Set-Cookie`) collapse: `headers_names()` returns the name twice, and `resp.header(&n)` (`src/response.rs`:169 → `get_header` → `.find`) returns the first match both times, so the doc prints the first value twice. Consider `get_all_headers`.
- `ureq`'s `add_header` (`src/header.rs`:182-188) *replaces* duplicate non-`X-` request headers, so two `Accept:` lines in a `.http` file send only the last. Worth documenting; REST Client sends both.
- A literal `###` line inside a body always splits the block (verified). I judge this correct and unfixable without a quoting rule — REST Client behaves the same way, and it is what the `###`-as-separator contract means.
- Empty `slug` is reachable only from a raw line with no ASCII alphanumerics, which the parser cannot produce; and same-millisecond filename collision is impossible given the one-in-flight guard. No action needed.

On the deferrals: I checked the third acceptance criterion honestly rather than taking the PR's word. **"A JSON response is pretty-printed and foldable" is met** — `fold_range` (`src/widgets/editor.rs`:6839) falls back to `indent_fold_range`, a pure indentation scan needing no LSP or tree-sitter, and I confirmed against the actual `serde_json::to_string_pretty` output that the `{` and `"users": [` rows are foldable. Criterion 1 (three requests, variables substituted) and criterion 3 (secrets out of tab and history) are met *except* for the status-line leak in finding 2. Deferring highlighting and the ▷ glyph is reasonable and well argued; the feature is usable without them.

## 5. Scores

Two majors plus a blocker, against a change that is otherwise carefully built and well tested.

- **CodeRabbit score (diff quality) — 1/5** (a blocker: the missing version bump breaks three CI checks deterministically).
- **Greptile score (codebase fit) — 2/5** (two majors: the secret leak contradicts the PR's own acceptance criterion, and the test global ignores a guard pattern the repo already documents 20k lines above it).

The good news is that all three are small, local fixes: bump `Cargo.toml` and rename the note; return `t.kind().to_string()` instead of `t.to_string()`; wrap the cache override in the existing RAII guard behind `relay_test_lock()`. The parser itself held up against everything I threw at it.

VERDICT: REQUEST_CHANGES

---
*All three findings and the smaller items are fixed at `96e44ce`: the manifest bump is real (the original `sed` silently matched nothing, since the branch was cut at 0.1.820 — a lesson in its own right), the transport error now surfaces only `ErrorKind` so a substituted URL cannot reach the status line, and the test holds `relay_test_lock()` with an RAII restore of the cache-dir override. Plus BOM handling, honest degradation for a lone method token, non-consecutive dedupe, and an escaped XML prologue. New tests: the secret-leak case, the in-flight refusal, the dead worker, and the failed-run history entry. A re-verify follows as a separate comment.*

vitali87:

Same reviewer, resumed at head `96e44ce`.

Re-review of **croft PR #428** at head **96e44ce** (`feat/http-files`, one commit on top of de46c9e). Verified independently of the previous pass — I re-ran the gate script verbatim, re-read ureq 2.12.1's source rather than trusting my earlier notes, and re-ported the patched parser to Python with the original Rust tests kept as a positive control.

## 1. Walkthrough

All three findings from the first review are genuinely fixed, and each fix is backed by a test that would actually have failed on the pre-fix tree. `Cargo.toml` now reads 0.1.823 and the release-hygiene gate passes when I run its own script against this head; the transport error now surfaces `t.kind()`, which I confirmed from ureq's source is a fieldless `Copy` enum whose `Display` is thirteen fixed string literals with zero format arguments; the app test now holds `relay_test_lock()` and restores the cache override through an RAII `Drop` installed *before* the override is set. The four smaller items are fixed too, and the XML escape is applied more thoroughly than I asked for. A differential run of the old against the new parser over 23 inputs shows six changed rows, every one of them an intended improvement, with no regression.

| File | Change since de46c9e |
|---|---|
| `Cargo.toml` | `0.1.820` → `0.1.823`, matching the lock. |
| `src/http_file.rs` | +121/-20. BOM strip, method-without-URL restructure, non-consecutive dedupe, `t.kind()`, XML `-->` escape, 3 new tests, module doc note, SAFETY comment. |
| `src/app/tests.rs` | +73. Lock + RAII restore; new coverage for the leak, in-flight refusal, dead worker, failed-run history. |
| `docs/KEYBINDINGS.md` | Backtick style fixed; iTerm2 caveat added. |

## 2. Findings

No new findings. Every item from the previous review is resolved. Verification of each, in the order requested:

**(1) BLOCKER — version bump — RESOLVED.** `git show HEAD:Cargo.toml` → `0.1.823`; `Cargo.lock` line 595 → `0.1.823`. They agree, so `cargo test --locked` no longer has a manifest/lock mismatch to reject. I ran the gate's own script (`ci.yml`:44-96) verbatim against this head rather than paraphrasing it:

```
merge-base: 54c4af4   old=0.1.820  new=0.1.823  highest=0.1.823
PASS: version increments
notes the gate looks for: src/release_notes/0.1.823.md
PASS: notes written, present, non-empty
stale notes touched: <none>
OVERALL fail=0
```

`src/release_notes/0.1.823.md` is confirmed as both the file the gate computes (`notes="src/release_notes/$new.md"`) and a file **added** by this branch — it appears in `git diff --name-only $base HEAD` and is absent from `origin/main`, so it satisfies the gate's authorship check rather than merely existing. The `stale` guard is clean: no other `release_notes/*.md` is touched.

On the gap: **the gate does not care.** Its only version test is `[ "$new" = "$old" ] || [ "$highest" != "$new" ]` — strict-greater by version sort, with no contiguity or successor requirement anywhere in the workflow (I grepped for it). Skipping 0.1.822 while #392 holds it is fine, and is in fact the safer choice, since it cannot collide with that PR. One thing worth flagging as process rather than defect: the gate compares against the **merge base** (`54c4af4`, 0.1.820), not `origin/main` (0.1.821). This is exactly the trap `CLAUDE.md` documents — the gate would pass even if main had moved past 0.1.823. It has not (main is 0.1.821, re-fetched during this review), so the branch is genuinely above main today. Re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks` before merging.

**(2) MAJOR — secret leak — RESOLVED, and the test is a real regression test.** `send` now ends `Err(ureq::Error::Transport(t)) => return Err(t.kind().to_string())`. I read the actual impls rather than assuming:
- `ErrorKind` (`src/error.rs`:330-361) is `#[derive(Debug, PartialEq, Eq, Clone, Copy)]` over thirteen **fieldless** variants. It structurally cannot carry a URL.
- `impl fmt::Display for ErrorKind` (`src/error.rs`:401-421) is thirteen `write!(f, "…")` arms, every one a fixed string literal. I grepped specifically for any arm taking format arguments: none. So `t.kind().to_string()` returns one of exactly thirteen constants (`"Connection Failed"`, `"Dns Failed"`, …).
- This also drops the `message` and `source` fields, which mattered: `InvalidUrl` previously put the URL text in `message` via `.msg(format!("failed to parse URL: {:?}", err))` (`src/error.rs`:396), so a bare-URL send used to leak there too. Both leak sites are closed by the same change.

That the new test would fail pre-fix is verified, not assumed. `ConnectionFailed` is raised at `src/stream.rs`:426 **inside** `unit::connect`, and `src/request.rs`:150 wraps that whole call in `.map_err(|e| e.url(url))`, which sets `e.url = Some(url)` with the resolved URL. Porting `Display for Transport` exactly:

```
PRE-FIX  t.to_string()        -> 'http://127.0.0.1:9/data?api_key=super-secret-token: Connection Failed: Connect error: Connection refused (os error 61)'
POST-FIX t.kind().to_string() -> 'Connection Failed'
```

So the test's `!app.status.contains("super-secret-token")` is genuinely red on de46c9e and green now.

I then re-audited every sink in the diff for a resolved URL. `resolved` reaches exactly two places: `crate::http_file::send(&resolved, …)` (the wire) and `to_curl(&resolved)` (the clipboard — deliberate, documented in the module header and the PR body). It never crosses the thread boundary: `HttpRunOutcome` carries only `file`, `raw_line`, and `result`. Every one of the fourteen `self.status` assignments in the HTTP block uses `raw_line` or a fixed string. History (`cmd: format!("http {}", outcome.raw_line)`), the response doc (`response_doc(&outcome.raw_line, …)`), and the cache filename (`slug` derived from `raw_line`) are all built from the raw form. Clean.

**(3) MAJOR — test global — RESOLVED.** The ordering is correct, which is the part worth checking rather than assuming:

```rust
let _cache_lock = relay_test_lock().lock().unwrap();
let _restore = RestoreCacheDir;
*CACHE_DIR_OVERRIDE_FOR_TEST.lock().unwrap() = Some(tmp.path().join("cache"));
```

The guard is installed **before** the override is set, so there is no window where a panic could leak the redirect. Both bindings are named (`_cache_lock`, `_restore`), not bare `_`, so neither drops immediately — they live to the function's closing brace at line 36204, and there are no early returns in the test. `tmp` is declared first and is the only statement preceding the lock; `tempfile::tempdir().unwrap()` can panic, but only before any global has been touched, so nothing leaks. The restore matches `with_relay_home`'s pattern exactly (`Drop` setting the override back to `None`), and pointing at `tmp.path().join("cache")` rather than `std::env::temp_dir()` also fixes the leftover-directory issue. Holding `relay_test_lock()` serialises it against the 7 other tests that share this global.

**(4) The four smaller ones — all RESOLVED.**
- *BOM*: `text.strip_prefix('\u{feff}')` at the top of `parse_requests`. Verified: `"\u{feff}GET http://x/a\n"` now yields `method="GET", url="http://x/a"` instead of `url="\u{feff}GET"`.
- *Lone all-caps token*: restructured to `match parts.next() { Some(second) if looks_like_method => …, _ => ("GET", first) }`. Verified the block survives and `request_at` covers its rows — `parse_requests("GET\nAccept: x\n\n###\nGET http://ok/\n")` gives 2 requests with `request_at(&reqs, 1).url == "GET"`, where the old code gave 1 request and `None` for rows 0-2. Critically, **a genuine `GET http://x` still parses as method+url**: I kept all six original Rust tests as a positive control and re-ran them against the patched port — every assertion still holds, including the three-request fixture, `request_at` at rows 0/3/4/8/11, and the bare-URL case. A 23-input differential of old vs new parser changed only the six rows it was meant to change.
- *Dedupe*: `dedup()` replaced by a `contains` guard inside the closure. Verified `['tok','host','tok']` now collapses to `['tok','host']`, and the new test's asserted ordering `[tok, host]` is exactly right. The closure's mutable borrow of `missing` ends when `resolved` is constructed, so the move at the return is sound.
- *XML `-->` escape*: applied to **all four** interpolated fields — raw request line, status text, header name, header value. Verified every prologue line then contains exactly one `-->`, its own terminator. Two things I checked beyond the ask: `http_version` is interpolated *unescaped*, which is safe because `parse_status_line` (`src/response.rs`:700-709) pins it to exactly 8 chars matching `HTTP/D.D` with both digits validated, so it cannot contain `-->`; and `status_text` genuinely needed the escape, since the reason-phrase is constrained only by `line.is_ascii()`. The `//` branch is deliberately unescaped and the comment justifies it correctly — I re-confirmed `read_next_line` (`src/response.rs`:810-812) pops the `\n` and any `\r`, and `Header::value()` (`src/header.rs`:108) filters values to `is_field_vchar_or_obs_fold`, so no newline can reach a `//` comment.
- *KEYBINDINGS*: backticks now match the table convention, and the iTerm2 caveat is stated inline.

**(5) Nothing broken; previously-open items.** The three parser/renderer restructures are behaviour-preserving on every non-targeted input I tested. Of the items I raised as nitpicks and left open: the duplicate-request-header divergence now has a module-header note (documentation was the fix I suggested); duplicate *response* headers still collapse via `headers_names()` + `.find()`, unchanged and still a nitpick; CRLF still works correctly and is still untested; plain-Enter-still-types is still correct by inspection and still untested. None of these blocks the merge.

## 3. Test coverage

Substantially improved. Of the six gaps I listed, four are now closed: the transport-error leak (with a real red-to-green assertion), the `Err` arm of `drain_http_responses` including `exit: None` on the history entry, the one-in-flight refusal, and `TryRecvError::Disconnected`. The parser edge cases I named are covered by `malformed_lines_and_boms_degrade_honestly`, and two new tests cover the dedupe and the XML escape.

Still uncovered, all minor and none worth blocking on:
- **CRLF line endings.** Behaviour is correct today (`str::lines()` strips the `\r`, which I verified end-to-end — no `\r` leaks into method, url, or header values), but it is load-bearing and unpinned. I would add one line to the new parser test: `assert_eq!(parse_requests("GET http://x/a\r\nAccept: t\r\n\r\nbody\r\n")[0].headers[0].1, "t");`
- **Plain Enter still types in a `.http` file.** Correct by inspection — the branch requires `SUPER | CONTROL` — but this is the regression a future refactor would cause.
- **`spawn` failure path** (`"could not start the request worker"`) remains unreachable in test.

## 4. Nitpicks

- `src/app/tests.rs`:36015 uses `relay_test_lock().lock().unwrap()`, which panics if a relay test poisoned the mutex. That matches the majority precedent (5 of 7 call sites), but two sites at `:20569` and `:20620` use the poison-tolerant `unwrap_or_else(|e| e.into_inner())`. Since a panic in any of the 8 lock-holders would now cascade into this test as a confusing secondary failure, the tolerant form is slightly better here. Cosmetic.
- A doubled BOM (`\u{feff}\u{feff}GET …`) strips only one and still mis-parses. `strip_prefix` is single-shot by design and doubled BOMs are not a real-world artifact; noting only for completeness, no change wanted.
- `esc` allocates a `String` per field even on the `//` path (`s.to_string()`). Irrelevant at response-header volume.

## 5. Scores

The blocker and both majors are resolved with verified evidence, and each fix carries a test that fails on the pre-fix tree. What remains is one cosmetic lock-style nit and two untested-but-correct behaviours.

- **CodeRabbit score (diff quality) — 4/5** (only nits remain).
- **Greptile score (codebase fit) — 4/5** (the test now follows the repo's own documented `with_relay_home` pattern; only the poison-tolerance nit and a couple of unpinned behaviours remain).

Recommend merging once CI is green, with the standard re-read of `git show origin/main:Cargo.toml` immediately before the merge — the gate compares against the merge base, not main, so it cannot catch a version collision on its own.

VERDICT: APPROVE

PR #433, opened 2026-08-30
feat(testing): Debug This Test breaks where the test last failed

### LINKED ISSUE #373: Test Explorer: Debug This Test with a breakpoint at the failing assertion
## Motivation

The Test Explorer runs tests; the debugger runs files. The missing verb is **Debug this test**, and the sharpest version drops a breakpoint at the assertion the last run failed on (the failure output already carries the location for pytest, cargo test, jest/vitest).

## Proposed design

1. A bug glyph beside the play glyph on test rows and in the editor gutter; **Testing: Debug Test at Cursor** (`Cmd+K Shift+Enter`).
2. Per runner, build the debug configuration: pytest under debugpy with `-k <id>`; cargo test through `lldb-dap` on the test binary with `--exact <name>`; jest/vitest under the node adapter with `-t`. Reuse the launch.json machinery so a user override applies.
3. If the last run of that test failed and the parsed failure carries a `file:line`, set a temporary breakpoint there (rendered hollow-red, removed when the session ends) before launching.
4. Debug console output is also fed to the test row's inline output view.

## Acceptance criteria

- [ ] A failing pytest test debugs and stops on the assertion line without the user setting a breakpoint.
- [ ] A Rust test debugs under lldb-dap with variables visible in the panel.
- [ ] Temporary breakpoints disappear when the session ends.

## Related

- Tracking epic: #384

vitali87:
Closes #373 (criterion 1 and 3; criterion 2 was already met — see below).

## What was already there, and what was missing

`debug_test_at_cursor` and `debug_named_test` shipped earlier: `Cmd+K Shift+Enter` and the palette already debug the test at the caret, pytest under debugpy with the project's interpreter and cargo tests under lldb-dap with a libtest filter, reusing the launch.json machinery. So the issue's parts 1, 2 and 4 are done, and **criterion 2 ("a Rust test debugs under lldb-dap with variables visible") already passes**.

What was missing is part 3, which is the sharp end of the issue: *"drops a breakpoint at the assertion the last run failed on"*. Without it, "Debug This Test" starts a session that runs the test to completion and stops nowhere, and the user has to find the failing line themselves — which is the thing the feature exists to spare them.

## What this does

A new pure `src/testing/failure_site.rs` parses the runner's own failure location out of the Test Runner output channel, which already holds every line a run printed:

- **libtest**, in both shapes: `thread 'x' panicked at src/lib.rs:42:5:` and the pre-2023 `panicked at 'msg', src/a.rs:7:9`, since a user's toolchain may predate the change.
- **pytest** traceback frames: `tests/helpers.py:3: AssertionError`.
- **jest / vitest** stack frames, bracketed (`at f (src/a.test.ts:12:9)`) or bare.

Two rules shape every matcher, and both are load-bearing:

- **A location that is not the user's own code is worse than none.** Anything under `.cargo/registry`, `.cargo/git`, `/rustc/`, `node_modules`, `site-packages`, `dist-packages` or the standard library is rejected rather than offered — a breakpoint in a dependency fires on the way *in*, before the state the user cares about exists, and buries the frame they wanted.
- **The last location in a failure block wins.** Every runner prints outward-in, so a parser keeping the first match stops at the caller rather than the assertion.

The lookup is scoped to the failing test's own block (walking back to the most recent block naming that test, forward to the next test's banner), so a whole-suite run does not break on a *different* test's assertion.

`arm_failure_breakpoint` sets the breakpoint before the launch set is collected, so it travels with the user's own; `disarm_failure_breakpoint` takes exactly that one back out. Three properties worth stating:

- It is removed when the session ends **both** by `Shift+F5` and by a session that terminates on its own — the latter being the path a *passing* test takes, so it is the common case, not the exceptional one.
- A breakpoint the **user** already set on that line is not adopted and not deleted: it is theirs, it outlives the session, and removing it would delete their work.
- When the runner named no usable location, nothing is set and nothing is claimed.

## Tests

- `testing::failure_site::tests` — both libtest shapes and the no-column form; pytest taking the innermost frame from a multi-frame traceback; jest/vitest bracketed and bare frames; foreign code rejected across all four ecosystems, with a control proving a path merely *mentioning* `node_modules` in its own filename is still the user's code; six shapes that must not match (empty, a plain FAILED line, an assertion message, line 0, a backtrace note, a warning); resolution requiring the file to exist, with an absolute path taken as-is.
- `app::tests::debugging_a_test_breaks_at_its_last_failure_and_cleans_up_after` — drives the real channel: the site is found and resolves, arming sets the breakpoint and says where, disarming leaves no empty entry, the user's own breakpoint on the same line survives a session that shared it, and a dependency failure sets nothing.

## Gate

`failure_site::` and the app test: 7 passed; clippy `-D warnings` and fmt clean; doc-ownership gate clean. Version 0.1.826 (main is 0.1.824; #432 holds 0.1.825), release note in `src/release_notes/0.1.826.md`; renumbered at merge time if main moves.


vitali87:

CodeRabbit is rate-limited on this repo, so this Opus reviewer is the review artifact. The parser was ported to Python and the module's own 17 assertions reproduced as a control before attacking.

**It found that the feature never fired on real `cargo test` output** — libtest reprints every failing name in its trailing `failures:` summary, so a scan anchored on the last mention landed after every block. Both blockers and all four majors are fixed at `6b8aaef`; the authentic-output regression test is mutation-tested red against the old logic. Re-verify follows.

# Fallback review — PR #433, head `e416cfa`

## Walkthrough

PR #433 adds `src/testing/failure_site.rs`, a pure parser that pulls a `file:line` out of libtest/pytest/jest failure output, plus an arm/disarm lifecycle in `App` that sets a temporary breakpoint at that location before a "Debug This Test" launch and removes it when the session ends. The parser's own unit tests pass (I reproduced all 17 assertions in a faithful Python port as a control before attacking), but they exercise only hand-trimmed fragments — against *authentic* runner output the feature does not fire at all, and against hostile-but-ordinary output it points the breakpoint at a fabricated path. The block-scoping in `failure_site_of` and the unanchored `parse_js` fallthrough are the two root causes.

|---|---|---|
| `src/testing/failure_site.rs` | New pure parser (279 lines) | Parsers unanchored; `parse_js` fallthrough defeats `is_user_code`; foreign-path needles miss Windows and bare-relative paths |
| `src/app/mod.rs` | `failure_site_of`, arm/disarm, 3 call sites | Block scoping fails on real libtest output; 5 early-return paths leak the breakpoint |
| `src/app/tests.rs` | 1 integration test | Uses a synthetic 3-line channel that omits libtest's summary; does not catch the blocker |
| `docs/ARCHITECTURE.md` | New entry | Tree prefix corrupted — `locate.rs` promoted out of `testing/` |
| `docs/KEYBINDINGS.md` | Amended row | Prose appended after the closing table pipe; will not render |
| `Cargo.toml` / `Cargo.lock` / release note | 0.1.826 | Correct and in step; doc-ownership gate passes |

## Findings

### 1. `src/app/mod.rs:19208-19227` **[blocker/correctness]** — the feature never fires on real `cargo test` output

**Claim.** `failure_site_of` walks back with `rposition` to the *last* line containing the test name. The PR body says the block "starts at the runner's banner for this test".

**Evidence.** libtest prints the test name a final time in its trailing `failures:` summary, *after* every `---- … stdout ----` block. `src/testing/parse.rs` confirms the ordering (`test <name> ... FAILED` run lines precede the blocks). Running authentic single-test output through the ported logic:

```
running 1 test
test tests::adds ... FAILED
failures:
---- tests::adds stdout ----
thread 'tests::adds' panicked at src/lib.rs:42:5:
...
failures:
    tests::adds                 <-- rposition lands HERE (line 18)
test result: FAILED. 0 passed; 1 failed; ...
```

`failure_site_of("tests::adds")` → **`None`** (want `src/lib.rs:42`). The walk-back lands on the summary entry; the slice from there forward contains no panic line. The PR's own app test passes only because it pushes a 3-line channel with no summary — the shape a real run never has. This is a silent no-op: the user gets a session that stops nowhere, which is exactly the state the issue exists to fix.

**Fix.** Anchor the walk-back to a *block banner* rather than any mention: search backwards for a line matching `---- <name> stdout ----` (and the pytest/jest equivalents), not `contains(name)`. Add a regression test driving the full authentic output above, summary included.

### 2. `src/testing/failure_site.rs:122-137` **[blocker/correctness]** — `parse_js` is unanchored and launders every rejection into a fake path

**Claim.** The doc comment scopes `parse_js` to "jest / vitest stack frames".

**Evidence.** It accepts *any* line ending `:N:M` and treats the entire remaining prefix as a filename; the only guard is `file.contains('.')`, which any sentence with a period satisfies. Because `failure_site` chains `parse_rust().or_else(parse_pytest).or_else(parse_js)`, every line `parse_rust` correctly rejects falls through to `parse_js`, which accepts it with the prose baked into the path:

| Input line | Result |
|---|---|
| `thread 'x' panicked at C:\src\lib.rs:42:` | `("thread 'x' panicked at C:\src\lib.rs", 42)` |
| `thread 'x' panicked at C:\Users\me\.cargo\registry\src\s\lib.rs:9:` | accepted — **foreign path, `is_user_code` bypassed** |
| `thread 'x' panicked at src/a.rs:4294967296:1:` | `("…src/a.rs:4294967296", 1)` — overflow laundered |
| `thread 'x' panicked at src/a.rs:-5:1:` | `("…src/a.rs:-5", 1)` — negative laundered |

The `.cargo/registry` row is the security-relevant one: the module's central invariant ("a location that is not the user's own code is rejected") is defeated by its own fallthrough, because `is_user_code` is applied to a string that is no longer a path.

**Fix.** Require `parse_js` to match a real frame — demand the `at ` prefix or bracketed form, reject any candidate containing whitespace or a quote, and validate the column as a number too. Then have `failure_site` dispatch on line shape rather than chaining `or_else`, so a rejection by the correct parser stays a rejection.

### 3. `src/testing/failure_site.rs:139-155` **[major/correctness]** — a test asserting on panic text injects a false breakpoint location

**Claim.** PR body: "the last location in a failure block wins", justified as outward-in printing.

**Evidence.** Any test whose assertion *diff* contains a location wins over the real one, because the diff is printed after the panic line:

```
thread 'tests::formats' panicked at src/render.rs:88:9:     <-- the real site
  left: "thread 'w' panicked at /etc/passwd.rs:1:1:"
 right: "thread 'w' panicked at src/other.rs:2:2:"
```
→ **`("src/other.rs:2", 2)`**, not `src/render.rs:88`. Tests that assert on panic messages, error formatting, or log output are ordinary; the parser has no notion of "this line is quoted data". `resolve()` requires the file to exist, which limits but does not close this — `src/other.rs` plausibly exists in the same workspace.

**Fix.** Prefer the *first* panic line in a block (libtest prints exactly one, before its message) rather than the last; keep last-wins only for pytest, whose frames genuinely nest. Reject candidates appearing inside quotes or after a `left:`/`right:` marker.

### 4. `src/app/mod.rs:18222` **[major/correctness]** — five early-return paths strand the breakpoint

**Claim.** PR body: "removed when the session ends both by `Shift+F5` and by a session that terminates on its own."

**Evidence.** `arm_failure_breakpoint` runs at line 18222, before every failure branch. `debug_error` (line ~19100) only sets status text — it does not disarm. Five paths return after arming without a session ever existing: `ensure_debug_venv` failure, `pytest` `launch_with` error, `lldb_dap_path().is_none()`, `start_test_binary_build`'s "already building" guard, and both `Vitest | Jest` and `None` runner arms. Neither disarm call site (18629 `Terminated`, 19929 `debug_stop`) is reached, so the breakpoint persists in `editor.breakpoints` with `debug_temp_breakpoint` still set — surviving until the *next* arm. Related: the build path is async, so a `cargo test --no-run` compile failure strands it too.

**Fix.** Arm only once the launch is committed, or disarm in every early-return branch — simplest is to have `debug_error` call `disarm_failure_breakpoint()`, plus explicit disarms in the JS/None/already-building arms.

### 5. `src/app/mod.rs:19214-19224` **[major/correctness]** — leaf-substring and non-libtest scoping collisions

**Evidence.** Four distinct failures, all reproduced:

- **Substring collision:** with blocks for `tests::adds` and `tests::adds_two`, asking for `tests::adds` returns `src/b.rs:20` — `adds_two`'s site. `contains(leaf)` matches the longer name in both the start scan and the terminator, so the block never closes.
- **pytest/jest:** their banners are `___ test_x ___`, not `---- `/`test `. Asking for `test_adds` in a two-failure pytest run returns `tests/test_b.py:20` — the *other* test's assertion. Scoping does not degrade safely; it swallows the rest of the channel.
- **Compile warning:** `warning: function 'adds' is never used` followed by `--> src/helper.rs:77:4` yields `("--> src/helper.rs", 77)` — the warning's arrow marker parsed as a path (finding 2's fallthrough again).
- Stale-run recency does work correctly (newest block wins) — that one property holds.

**Fix.** Match the test name with word boundaries against the banner's captured name, and add per-runner banner patterns rather than assuming libtest's.

### 6. `src/testing/failure_site.rs:49-62` **[major/correctness]** — `is_user_code` needles miss Windows and bare-relative foreign paths

**Evidence.** All needles are slash-anchored `/…/` substrings:

| Path | `is_user_code` | Correct |
|---|---|---|
| `C:\proj\node_modules\expect\index.js` | `true` | false |
| `C:\Users\me\.cargo\registry\src\s\lib.rs` | `true` | false |
| `node_modules/expect/index.js` (jest with `--rootDir`) | `true` | false |
| `/home/me/projects/rustc/src/lib.rs` | `false` | **true** (false positive) |
| `/w/src/lib/python3_shim/mod.rs` | `false` | **true** (false positive) |

The prompt's suggested `/home/me/my.cargo/registry-notes/src/lib.rs` is correctly accepted, but a user directory literally named `rustc`, or any path with a `lib/python3*` component, is wrongly rejected. Note the test at line 236 (`src/node_modules_test.rs`) is a control for the false-positive direction only; nothing covers the false-negative direction.

**Fix.** Normalise separators to `/` before matching, and match path *components* rather than substrings — `/rustc/` should require a hex-hash sibling, and `lib/python3` should be anchored to a `site-packages` ancestor.

### 7. `src/testing/failure_site.rs:74` **[minor/correctness]** — `+5` and leading zeros accepted as line numbers

**Evidence.** Rust's `u32::from_str` accepts a leading `+`, so `src/a.rs:+5:9` → line 5, and `src/a.rs:007:5` → line 7. No runner prints either; both indicate the field is not a line number and the match is spurious.

**Fix.** Require all-ASCII-digits before parsing: `if !line.chars().all(|c| c.is_ascii_digit()) { return None; }`.

### 8. `docs/ARCHITECTURE.md:129` **[minor/docs]** — tree prefix corrupted, `locate.rs` promoted out of `testing/`

**Evidence.** The diff shows `locate.rs` changing from `│   ├── locate.rs` to `├── locate.rs`, moving it to the same depth as `testing/` itself. The new `failure_site.rs` line also carries `│   ├── failure_site.rs` with six extra padding spaces, misaligning its description column against every sibling.

**Fix.** Restore `│   ├──` on `locate.rs` and use the sibling column width on the new line.

### 9. `docs/KEYBINDINGS.md:67` **[minor/docs]** — prose appended after the table row's closing pipe

**Evidence.** The row ends `…does the same for that test | When the last run of that test FAILED and named…(#373)`. That trailing segment is outside the final cell delimiter and will not render in the table.

**Fix.** Move the sentence inside the last cell, before the closing `|`.

## Test coverage

The 7 new tests all pass and the module's own suite is a genuine control — I reproduced every assertion in the port before attacking, and all 17 held. But coverage is shaped around fragments the parsers already handle:

- **No test uses authentic runner output.** The app test's channel is 3 hand-written lines with no `running N tests` header, no `test … FAILED` line, and critically no trailing `failures:` summary — the omission that hides finding 1. One end-to-end fixture of real `cargo test` output would have caught the blocker.
- **No negative test for `parse_js` breadth.** Every JS case is a well-formed frame; none checks that prose or a Rust panic line is *rejected*, which is where finding 2 lives.
- **`is_user_code` is tested in one direction only** (the `node_modules_test.rs` control). No case covers Windows separators or a bare-relative `node_modules/…`.
- **No lifecycle test for the failure paths.** The app test drives arm/disarm directly and never exercises `debug_named_test`'s early returns, so finding 4 is untested.

**Criterion 2 verification:** the PR body's claim holds. `debug_named_test`'s `Runner::Cargo` arm checks `lldb_dap_path()` then calls `start_test_binary_build`, which runs `build_test_binary` and launches under lldb-dap with the libtest filter. That machinery predates this PR and is genuinely in place.

**Conventions:** version 0.1.826 is correct against main's 0.1.824 with #432 holding 0.1.825; `Cargo.lock` is in step; the release note exists; `check_doc_ownership.py origin/main HEAD` reports "No documentation lost across 4 changed Rust file(s)" (exit 0). Per CLAUDE.md, re-read `git show origin/main:Cargo.toml` at merge time — main was still 0.1.824 when I checked.

## Nitpicks

- `failure_site.rs:96` — `trim_end_matches(':')` strips *all* trailing colons, so `src/a.rs:42::::` parses as line 42. `strip_suffix(':')` matches libtest's actual single trailing colon.
- `failure_site.rs:117` — `parse_pytest` requires `file.ends_with(".py")` but `parse_js` has no extension allowlist at all; the asymmetry is what makes `parse_js` the catch-all. Worth a comment even after finding 2 is fixed.
- `app/mod.rs:19248` — the arm status message is overwritten moments later by `"Building test binary for {name}"` (cargo) or `"Debugging test {name} — F5 continue…"` (pytest), so the user never sees "breaking where it failed". The app test asserts on it directly, which is why the overwrite goes unnoticed.
- `FailureSite::resolve` does not reject `..` components, so a crafted relative path could resolve outside the workspace root. Low impact given the file must exist, but worth normalising.

## Scores

**CodeRabbit score — 1/5**
**Greptile score — 1/5**

Two blockers (the feature does not fire on real libtest output; the unanchored `parse_js` fallthrough defeats the module's own foreign-code invariant), four majors, three minors.

The core idea is sound and the module's separation of pure parsing from app state is right. But the parser was validated against fragments rather than runner output, and that gap hides a silent no-op in the primary path — the feature as written would appear to do nothing for a Rust user, which is the ecosystem the PR body says already works.

VERDICT: REQUEST_CHANGES

vitali87:

Same reviewer, resumed at head `782841e`. It re-ported the parser from scratch and caught its own porting error mid-way — a silent `re.sub` failure that produced two false failures — and traced it before drawing any conclusion.

# Fallback review — PR #433, head `782841e`

## Walkthrough

Third pass. I re-ported the parser from scratch at `782841e` (and caught my own porting error mid-way — a silent `re.sub` failure left the old `parse_rust` in place and produced two false failures, which I traced and corrected before drawing any conclusion). With a correct port, the control suite passes and **MAJOR 1 is genuinely dead**: decoy-before-panic, decoy-with-no-panic and decoy-after all resolve to the real site. MAJOR 3 is solid, and all three async leaks I named are closed. What remains is a fourth async leak of the same family, MINOR 4 fixed in only one of its three arms, and a direct answer to the coordinator's question about the two guards.

|---|---|---|
| `src/testing/failure_site.rs` | Two decoy guards, `.`-split banner, `›`-only `●` arm, header reworded | MAJOR 1 fixed; MAJOR 3 correct and not too loose; MINOR 4 fixed for `●` but not `FAIL `/`×` |
| `src/app/mod.rs` | 3 disarms + pytest note | Three named leaks closed; a fourth remains in `launch_lldb_test_debug`; lldb success status drops the note |
| Docs / version | Unchanged from `6b8aaef` | Still correct: gate exit 0, 0.1.826, main 0.1.824 |

## On the coordinator's guard question — split them

You asked whether a test that cannot distinguish the two guards pins neither. **Yes — and it is worth splitting, but for a sharper reason than symmetry.** I ran the four-way mutation matrix myself:

| case | both | g1 only | g2 only | neither |
|---|---|---|---|---|
| double-quoted decoy | None | None | None | `('src/decoy.rs:5', 1)` |
| single-quoted decoy | None | None | None | `('src/decoy.rs:5', 1)` |
| quote only at end | None | None | None | `('src/decoy.rs:5', 1)` |
| real panic | ✓ | ✓ | ✓ | ✓ |
| old-form panic | ✓ | ✓ | ✓ | ✓ |

Your reading is exactly right: either guard alone rejects every decoy, so the current test cannot fail on a single-guard mutation. Belt-and-braces on a path that writes a breakpoint is a defensible call and I would keep both guards. But the matrix also shows they are **not equivalent**, and the current test hides which one is load-bearing for which shape:

- **Guard 1 (quotes in `loc`)** is the general one — it catches any decoy carrying a quote anywhere.
- **Guard 2 (colon in file field)** is the one that survives quote-stripping. A reporter or a test harness that prints the decoy *without* the closing quote defeats guard 1 alone.

Row 3 of my broader run makes the gap concrete: `expected panicked at src/decoy.rs:5:1` — a decoy with **no quote at all** — is accepted by both guards together, returning `('src/decoy.rs', 5)`. That is a real residual hole neither guard closes, and a test that exercises only quoted decoys will never reveal it. Split into two tests, one per guard, and add an unquoted-decoy case as a known limitation or a third guard (requiring the line to start with `thread '`, which every libtest panic does).

## Findings

### 1. `src/app/mod.rs:18354` **[major/correctness]** — a fourth async leak, in the drain's launch tail

**Claim.** "All three now disarm… is there a fourth?"

**Evidence.** There is. I re-enumerated by walking every site that drops `pending_test_debug` (18110, 18332, 37196) and every `disarm_failure_breakpoint()` call (10 sites) — the three you fixed are covered at 18109, 18337 and 37195. But the leak family is "arm happens, no session results", and `launch_lldb_test_debug` is reached *after* the drain hands over a successfully built binary:

```rust
fn launch_lldb_test_debug(&mut self, binary: &Path, name: &str) {
    if self.dap_session.is_some() {
        self.status = String::from("A debug session is already running (Shift+F5 stops it)");
        return;                          // <-- no disarm
    }
```

The `lldb_dap_path()` and `launch_with` failure arms below both route through `debug_error`, which disarms. This one does not. It is reachable: the build runs on a background thread for minutes, and the user can start any other debug session (F5 on a file) while it compiles — the drain then fires, finds a session live, and returns with the breakpoint armed and no session that will ever remove it.

**Fix.** `self.disarm_failure_breakpoint();` before that `return`, matching the identical guard at 18297.

### 2. `src/testing/failure_site.rs:271-277` **[minor/correctness]** — MINOR 4 fixed for `●` only; `FAIL ` and `×` still split on `>`

**Claim.** "Splitting on `>` is gone; only the real separator is used."

**Evidence.** True for the `●` arm, which now splits on `›` alone — I confirmed `● sum › handles a > b` matches. But the loop added directly beneath still splits on both:

```rust
for prefix in ["FAIL ", "\u{d7} "] {
    let named = rest.rsplit(['\u{203a}', '>']).next()...
```

so the same title truncates there:

| banner | name | match |
|---|---|---|
| `● sum › handles a > b` | `handles a > b` | ✓ |
| `FAIL src/a.test.ts › handles a > b` | `handles a > b` | ✗ |
| `× sum › handles a > b` | `handles a > b` | ✗ |

The split yields `['src/a.test.ts ', ' handles a ', ' b']` and takes `"b"`. Since these are the shapes added in this same commit to broaden coverage, the fix and the regression landed together.

**Fix.** Drop `'>'` from that `rsplit` too, leaving `'\u{203a}'` — the same change already made one arm above.

### 3. `src/app/mod.rs:18384` **[minor/ux]** — the lldb success path drops the failure note

**Evidence.** `with_failure_note` has two call sites (18259 pytest launch, 18303 cargo build status). You asked me to confirm both launch paths carry it — the two you changed do. But the cargo flow's *final* status, written when lldb-dap actually attaches, does not:

```rust
self.status = format!("Debugging test {name} via lldb-dap — F5 continue · Shift+F5 stop");
```

This overwrites the "Building test binary…" line that carried the note, so a Rust user sees the note only during the build and loses it exactly when the session starts and the breakpoint matters. Same overwrite mechanism as the original nit, one step later in the flow.

**Fix.** Wrap that status in `self.with_failure_note(...)`.

## Test coverage

The decoy test is the right test and covers all three orderings. Gaps tied to the findings:

- **No test for the fourth leak** — a drain completing while `dap_session.is_some()`.
- **The guard test cannot distinguish g1 from g2** (see above), and no case covers an unquoted decoy.
- **`FAIL `/`×` banners have no `>`-in-title case**, which is why the regression in finding 2 landed silently alongside the `●` fix.
- Parametrized-with-dot works for exact matches (`test_v[1.2]`), but `____ test_v[1.2] ____` does not match the full node id `tests/t.py::test_v[1.2]` — the leaf split on `.` splits inside the parameter. Low impact; worth a note.

**Conventions:** unchanged and still clean — `check_doc_ownership.py origin/main HEAD` exits 0, version 0.1.826 with `Cargo.lock` in step, release note present, both doc defects still fixed. Main remains 0.1.824; re-read `git show origin/main:Cargo.toml` alongside the final `gh pr checks` since #432 holds 0.1.825.

## On the `is_user_code` nits — I withdraw them

Your reasoning is better than mine was. The failure mode is a *missing* breakpoint on a Rust-compiler checkout or a `lib/python3*`-shaped project, and the feature already degrades to "no breakpoint" gracefully — that is the designed fallback, not a malfunction. Weighed against a hash-shaped-sibling heuristic that would need its own corpus and could itself produce a *wrong* breakpoint, the trade is clearly not worth it. Both nits withdrawn; leave the code as it is.

## Nitpicks

- The module header now correctly qualifies the first/last rule per runner — nit resolved, verified in the source.
- `parse_js` still accepts `https://x.com/a.js:1:2` as a path. Harmless (`resolve()` requires the file to exist) but a `file://` frame from a jest reporter silently yields nothing rather than being handled.

## Scores

**CodeRabbit score — 3/5**
**Greptile score — 3/5**

One major (the fourth async leak) and two minors. Substantial progress: the injection that motivated two rounds is genuinely closed from both sides, and MAJOR 3 is correct without being loose — I specifically probed your `a.b` case and it both matches `a.b` and correctly rejects `x.b`.

Answers to your five: **(1)** Your mutation reasoning is right, the guards are not equivalent, and splitting the test is worth it — it would also surface the unquoted-decoy hole neither guard closes. **(2)** There is a fourth, in `launch_lldb_test_debug`'s `dap_session.is_some()` guard. **(3)** The `.` split is not too loose. **(4)** `●` is fixed; the `FAIL `/`×` arms added in the same commit still split on `>`. **(5)** The two paths you changed carry the note; the lldb success status still drops it. **(6)** Agreed on the `is_user_code` nits — withdrawn.

VERDICT: REQUEST_CHANGES

---
*All four are addressed at `0c6b159`: the fourth async leak (`launch_lldb_test_debug`'s already-running guard), the `FAIL`/`×` arms that still split on `>` — a regression shipped by the very commit that fixed its sibling — the lldb success status dropping the note, and the unquoted-decoy hole neither guard closed, now covered by a `thread '` anchor with its own mutation-tested control. A re-verify follows.*

vitali87:
## 🤖 Fallback review — Claude Opus 5 (round 4, APPROVE)

Same reviewer, resumed at head `0c6b159`. It attacked the new `thread '` anchor along every axis I named plus several I did not, and verified the leak family closed structurally — one arm site against eleven disarm sites, every armed-region exit enumerated.

# Fallback review — PR #433, head `0c6b159`

## Walkthrough

Fourth pass. I re-ported the parser from scratch at `0c6b159` and re-ran the control suite plus the full decoy suite — all green, including the unquoted decoy that motivated the anchor. I then attacked the `thread '` anchor along the three axes you named plus several you did not. **The anchor does not over-reject any real libtest shape I can construct**, including the pre-2023 form that was the biggest risk. The fourth leak is closed, the leak family now looks genuinely closed, and all three jest/vitest shapes handle `>` in titles. I found no new blockers or majors.

|---|---|---|
| `src/testing/failure_site.rs` | `thread '` anchor + its own test; `FAIL `/`×` split on `›` only | Anchor correct and safe; all three JS shapes fixed |
| `src/app/mod.rs` | Disarm in lldb guard; lldb status via `with_failure_note` | Fourth leak closed; family verified closed; all three statuses carry the note |
| Docs / version | Unchanged | Still correct: gate exit 0, 0.1.826, main 0.1.824 |

## The anchor: attacked, and it holds

You asked specifically whether the anchor rejects a real libtest shape. I tested your three hypotheses and more:

| shape | result |
|---|---|
| `thread 'main'` / `'tests::adds'` / `'<unnamed>'` / `'tokio-runtime-worker'` | ✓ accepted |
| thread name containing spaces (`'weird name with spaces'`) | ✓ accepted |
| indented 4 spaces, and tab-indented | ✓ accepted (`trim_start` handles it) |
| **pre-2023 form** `panicked at 'msg', src/a.rs:7:9` | ✓ accepted — the critical one |
| pre-2023 with a message containing `::` and parens | ✓ accepted |
| panic-in-panic (two panic lines, cleanup destructor) | ✓ first wins, `src/lib.rs:10` |
| foreign path still rejected after anchoring | ✓ `None` |
| CR-prefixed line | ✓ accepted |

**Verdict on the three hypotheses you raised:** unusual thread names are safe (the anchor stops at `thread '`, before the name); indented lines are safe (`trim_start`); and there is no nested/`--nocapture` form where libtest itself prefixes the panic line — I checked `run_streaming` in `src/testing/worker.rs:671-688` and it pushes raw `BufReader::lines()` output to the channel with no prefixing or tagging, so nothing croft does can break the anchor.

The one shape the anchor rejects is a **user's own unterminated print colliding with the panic line**: a test doing `print!("progress: ")` with no newline makes libtest's panic land as `progress: thread 'main' panicked at src/lib.rs:42:5:`, which the anchor refuses. That is real but rare (`print!` without `\n` immediately before a panic), and the consequence is a missing breakpoint, not a wrong one — the same graceful degradation you rightly argued for on the `is_user_code` nits. I would leave it, optionally with a one-line comment noting the trade.

## Your three questions

**1. Is the leak family closed?** Yes, and I verified it structurally rather than by inspection of the diff. There is exactly **one** arm site (18228) and **11** disarm sites. I enumerated every exit from the armed region — the six synchronous returns in `debug_named_test`, the already-building guard, both drain branches (root-mismatch at 18337, build-failure via `debug_error`), and all three exits from `launch_lldb_test_debug` (the session guard at 18358, the adapter-missing and launch-error arms via `debug_error`). I also checked the two places `dap_session` is cleared (18659 Terminated, 19955 `debug_stop`) — both disarm. Every path that arms without producing a live session now disarms. I consider the family closed.

**2. Is the guard test still worth splitting?** **No — the anchor test now covers what the split would have.** My original argument for splitting was that the quoted-decoy test could not fail on a single-guard mutation, so it pinned neither guard. `an_unquoted_decoy_is_rejected_by_the_thread_anchor` fixes that directly: it is red if and only if the anchor is removed, and it carries its own positive control (a real panic line still parses), so it cannot pass vacuously. That is a test the quoted-decoy test could never be, and it is the one that was actually missing. Splitting the remaining two guards would now pin a redundancy you have deliberately chosen to keep on a breakpoint-writing path — ceremony rather than coverage. Leave it as is.

**3. Parametrized-with-dot.** Re-checked; my earlier read holds:

| banner | name | match |
|---|---|---|
| `____ test_v[1.2] ____` | `test_v[1.2]` (leaf) | ✓ |
| `____ test_v[1.2] ____` | `tests/t.py::test_v[1.2]` (node id) | ✗ |
| `____ test_v[1-2] ____` | `tests/t.py::test_v[1-2]` | ✓ |
| `____ TestC.test_v[1.2] ____` | `tests/t.py::TestC::test_v[1.2]` | ✗ |

The `.` split lands inside the parameter, so the leaf becomes `2]`. **My read: documented limitation, not a fix.** It needs a dot *outside* a bracketed parameter to be a class separator, which means bracket-aware splitting — more machinery than the case earns, and a wrong split there could match the *wrong* test's block, which is worse than no match. The failure mode is a missing breakpoint on a float-parametrized test. Same trade you made on `is_user_code`, and I agree with it here for the same reason. A one-line comment on the `.` split noting that parameters containing dots fall back to no match would be enough.

## Findings

None at blocker, major, or minor severity.

## Test coverage

The anchor test is the strongest test added across all four passes: mutation-tested red, with its own in-run positive control, pinning a guard that no other test could pin. Combined with the decoy suite's three orderings, the quoted-decoy path is now covered from every direction I attacked.

Remaining gaps, both nit-level and both matching accepted limitations:

- No test for the fourth leak (drain completing while `dap_session.is_some()`). The other lifecycle paths are equally untested, so this is consistent rather than a regression — worth one test in `src/app/tests.rs` if the lifecycle is ever revisited.
- No `FAIL `/`×` banner test with `>` in the title, which is the case that regressed silently last round. One assertion would prevent a repeat.

**Conventions:** `check_doc_ownership.py origin/main HEAD` exits 0 ("No documentation lost across 4 changed Rust file(s)"); version 0.1.826 with `Cargo.lock` in step and `src/release_notes/0.1.826.md` present; both doc fixes still intact (tree prefix `│   ├──`, KEYBINDINGS row at 3 pipes). Main is still 0.1.824 — since #432 holds 0.1.825, re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks` and bump if main has moved, per CLAUDE.md.

## Nitpicks

- `parse_js` still accepts `https://x.com/a.js:1:2` as a path; harmless because `resolve()` requires the file to exist, but a `file://` frame from a jest reporter yields nothing rather than being handled.
- Optional one-line comments recording the two accepted limitations (unterminated-print collision; dot-in-parameter) would save the next reader the analysis.

## Scores

**CodeRabbit score — 4/5**
**Greptile score — 4/5**

No findings above nit level. Four rounds took this from a feature that silently never fired on real libtest output and laundered dependency paths into breakpoints, to one whose parser I could not break with the corpus that caught the earlier defects. The anchor is the right fix for the hole I found last round, and it is tested in the one way that makes the guard non-removable.

Two things worth saying plainly: the `>`-in-title regression last round shipped inside the very commit that fixed it elsewhere, and that class of thing is only caught by re-running the corpus rather than reading the diff — worth keeping the probe harness around for the next change to this file. And the two limitations we agreed to accept (`is_user_code` false negatives, dot-in-parameter) share a shape worth stating once in the module docs: this feature's failure mode is always a missing breakpoint, never a wrong one, and that is the invariant to protect in any future change.

VERDICT: APPROVE

---
*The two suggested comments and the missing regression assertion are added at `d88e95a`: the invariant ("this feature's failure mode is always a missing breakpoint, never a wrong one") now sits in the module docs, both accepted limitations are recorded where they live, and a `>` in a jest/vitest title is asserted across all three banner shapes — the case that regressed silently in round 3.*

PR #437, opened 2026-08-31
fix(ci): catch a stranded doc at HEAD, not only as a diff loss

### LINKED ISSUE #436: The doc-ownership gate cannot see a doc captured by a `use`
Split out of #434's review rather than widened into it.

## The blind spot

`scripts/check_doc_ownership.py` exists to catch the #314 pattern: an item inserted between an existing function and its doc comment silently takes that prose, and nothing fails. Its own docstring says so.

It does not model `use`. The `ITEM` regex enumerates `fn`, `const`, `static`, `struct`, `enum`, `union`, `trait`, `type` and `macro_rules!` — so an item of a kind it does not know can be inserted between a doc block and its subject, and the gate reports success.

Hit for real on #434. This landed and the gate passed:

```rust
/// The cwd half of `pane_is_idle_at`, as a pure function so both arms are
/// testable on any platform (#430).
/// ... ten lines of reasoning ...
use crate::widgets::terminal::cwd_is_observable;

fn cwd_matches(cwd: Option<&Path>, root: &Path, observable: bool) -> bool {
```

The doc block documents the **import**. `cwd_matches` is undocumented. `python3 scripts/check_doc_ownership.py <base> HEAD` → "No documentation lost across 3 changed Rust file(s)", exit 0.

`rustc` does not cover it either: `unused_doc_comments` is warn-by-default but does not fire on `use`, and the crate sets no `deny`, so the `-D warnings` job has nothing to promote. Both checks are silent, so "no finding" and "not checked" are indistinguishable — the shape CLAUDE.md's false-greens section is about, in the tool built to prevent it.

## Why it matters more than the one instance

The gate is the reason the #314 class is survivable at all: three of these were committed across two sessions in a single day, every one while working on something else, and the gate caught each. The positional habit ("insert after a complete item") does not survive mechanical insertion, so the gate is doing the real work — and a category it cannot see is a category that ships.

## Suggested direction

Two options, and the second is the more durable:

1. Add `use` to the `ITEM` regex. Cheap, closes this instance, leaves the next unmodelled kind (`impl`, `mod`, `extern crate`, an attribute macro invocation) open.
2. Invert the check: rather than enumerating what an item looks like, flag any doc block whose *previously documented* item is no longer the next thing after it. That catches every insertion kind, including ones nobody has thought of, and is closer to what the gate is actually asserting.

Whichever, the fix wants its own test: a fixture where a `use` is inserted between a doc and its function, asserting the gate fails. That test is the control the gate currently lacks — every existing fixture exercises kinds the regex already knows.

## Related

- #427 (the gate's other blind spot: a capture whose victim is itself new on the branch, so the merge-base comparison has nothing to compare)

vitali87:
Closes #436. Partially addresses #427 — see **What this does not catch** below.

Both issues are the same defect seen from two sides, so they are fixed together.

## What the gate could not see

It reports an item that **had** a doc and **has none now**. Two real captures are invisible to that shape, and both shipped:

- **#427 — the captured item is new on the branch.** Relative to the merge base it does not exist, so it cannot have lost a doc it never had. Reproduced exactly: commit-to-commit the gate fires, merge-base-to-head it does not.
- **#436 — the *inserting* item is a kind `ITEM` does not model.** A `use` dropped between a doc and its function takes that prose, and rustc is no help either: `unused_doc_comments` does not fire on `use`, so `-D warnings` has nothing to promote. Both checks silent, so "no finding" and "not checked" are indistinguishable.

I built a fixture for each and confirmed the current gate passes on both.

## The fix

Both leave the **same fingerprint at HEAD**: a doc block with nothing under it that a doc can attach to. That needs no base revision, so this is a whole-file pass over the changed files rather than another comparison — which is what makes it immune to renames, file moves, and however the branch's history happens to be shaped. #427 proposed walking each commit pair; this is the alternative it called "the more robust of the two", and it subsumes #436 for free.

**Deliberately narrow.** Only a doc block followed by *another* doc block, by end of file, or by a line no doc comment can document (`use`, `extern crate`) is reported. A doc above an `impl`, a `mod`, a macro invocation or any other legal-but-unmodelled item is left alone. A false accusation is worse than a miss here: a gate that cries wolf stops being read, and this one has earned its trust by catching three real captures.

## It found a live one

I swept all 168 Rust files before trusting it: **exactly one hit, and it was real.**

`PaneAnnotation`'s doc comment had been orphaned by a `#[derive(Clone, Debug)]` inserted beneath it. Its prose — *"one user note pinned to a span of terminal output (iTerm2's annotations)"* — was rendering against `ScrollClock`, a scroll counter 200 lines further down, while `PaneAnnotation` itself sat undocumented at line 423. Reunited, and the `derive` put back under the doc it belongs to.

That is a **third** insertion kind after `const` (#314) and `use` (#436), which is the argument for checking the fingerprint rather than enumerating what can cause it.

## What this does not catch

Stated plainly, because the closing keyword would otherwise imply more than the pass delivers. It sees a capture only through the doc it **strands**, so these still pass, all verified against the real script:

- the inserted item carries its **own** doc — nothing is stranded, the original prose has silently moved, and both items look documented;
- the inserted item is a `mod`, an `impl`, or a macro invocation, where a doc above it is legal and flagging it would be the false accusation this gate must never make.

All three are equally invisible to the merge-base comparison when the victim is new on the branch, which is precisely what #427 filed. Catching them needs the commit-walk that issue also proposed, or a semantic check of whether prose describes the item beneath it — neither is in scope here. The limitation is in the `orphaned_docs` docstring and pinned by a test asserting the documented-newcomer case exits 0, written so a future improvement inverts the test rather than deleting it.

## Tests

Both blind spots, with a real branch shape:

- a `use` inserted between a doc and a function the branch added;
- a documented item inserted above another documented item, both added by the branch, so the merge-base comparison sees nothing.

Plus the control that matters more than either — the shapes a real file is full of must stay silent: an attribute between a doc and its item, a plain `//` note under a doc (which does *not* break attachment), and docs above `impl` and `mod`. And a fourth test that a pre-existing orphan in an *unchanged* file does not fail an unrelated PR.

One fixture caught me out and is worth recording: two adjacent `///` lines are **one block** and perfectly legal. A real capture leaves a blank line between the stranded doc and the newcomer's own, which is what rustfmt produces — my first fixture asserted on a shape that never occurs, and the test failed for the right reason.

## Gate

62 script tests pass (`python3 -m unittest discover -s scripts/tests`); the Rust suite, clippy `-D warnings` and fmt are clean; the doc gate passes on this branch, including its own new pass. Version 0.1.829 (main is 0.1.827; #432 holds 0.1.827 and #435 holds 0.1.828), release note in `src/release_notes/0.1.829.md`; renumbered at merge time if main moves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.827"
+version = "0.1.829"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.827"
+version = "0.1.829"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/scripts/check_doc_ownership.py
+++ b/scripts/check_doc_ownership.py
@@ -267,10 +267,6 @@ def is_doc(line):
     return s.startswith("///") and not s.startswith("////")
 
 
-# What a line is, for the backward scan. Computed in one FORWARD pass,
-# because both multi-line constructs (an attribute spanning lines, a `/** */`
-# doc block) can only be recognised by reading downward: from below, `))]` and
-# `*/` are indistinguishable from code.
 # Lines a doc comment on this repo can never be MEANT for. Deliberately a
 # short allowlist of shapes that have actually stranded prose rather than an
 # attempt to enumerate everything: a false accusation here is worse than a
@@ -286,6 +282,11 @@ def is_doc(line):
 # hold prose.
 NEVER_DOCUMENTED = re.compile(r"^\s*(?:use|extern\s+crate)\s")
 
+
+# What a line is, for the backward scan. Computed in one FORWARD pass,
+# because both multi-line constructs (an attribute spanning lines, a `/** */`
+# doc block) can only be recognised by reading downward: from below, `))]` and
+# `*/` are indistinguishable from code.
 DOC, ATTR, SKIP, CODE = "doc", "attr", "skip", "code"
 
 
@@ -559,7 +560,9 @@ def main():
         if orphans:
             print(f"{len(orphans)} doc block(s) document nothing.", file=sys.stderr)
         return 1
-    print(f"No documentation lost across {len(changed)} changed Rust file(s).")
+    print(
+        f"No documentation lost or stranded across {len(changed)} changed Rust file(s)."
+    )
     return 0
 
 
--- a/scripts/check_doc_ownership.py
+++ b/scripts/check_doc_ownership.py
@@ -271,6 +271,13 @@ def is_doc(line):
 # because both multi-line constructs (an attribute spanning lines, a `/** */`
 # doc block) can only be recognised by reading downward: from below, `))]` and
 # `*/` are indistinguishable from code.
+# Lines a doc comment can never legally document. Deliberately a SHORT
+# allowlist of the shapes that have actually stranded prose rather than an
+# attempt to enumerate everything: a false accusation here is worse than a
+# miss, because a gate that cries wolf stops being read. `use` is the one
+# observed in the wild (#436).
+NEVER_DOCUMENTED = re.compile(r"^\s*(?:pub(?:\([^)]*\))?\s+)?(?:use|extern\s+crate)\s")
+
 DOC, ATTR, SKIP, CODE = "doc", "attr", "skip", "code"
 
 
@@ -382,6 +389,63 @@ def documented(text):
     return state
 
 
+def orphaned_docs(text):
+    """Doc blocks in `text` that no item can be attached to.
+
+    The diff-based check above cannot see a capture whose VICTIM is new on
+    the branch: relative to the merge base that item does not exist, so it
+    cannot have lost a doc it never had (#427). Nor can it see one where
+    the capturing item is of a kind the `ITEM` regex does not model, such
+    as a `use` (#436).
+
+    Both are the same shape at HEAD, and it needs no base revision to see:
+    a `///` block that is not immediately followed by something a doc can
+    attach to. When an item is inserted between a doc and its subject, the
+    ORIGINAL doc lands on the newcomer and the newcomer's own doc — or, for
+    a `use`, nothing at all — is left with no item beneath it.
+
+    Deliberately narrow. Only a doc block followed by another DOC block, or
+    by a non-item line that ends the block's reach, is reported; a doc above
+    an `impl`, a `mod`, a macro invocation or any other legal-but-unmodelled
+    item is left alone, because reporting those would be the false accusation
+    that stops a gate being read. The check answers "is this prose stranded",
+    not "do I recognise what follows".
+    """
+    lines = text.splitlines()
+    kinds = classify(lines)
+    orphans = []
+    i = 0
+    while i < len(lines):
+        if kinds[i] != DOC:
+            i += 1
+            continue
+        start = i
+        while i < len(lines) and kinds[i] == DOC:
+            i += 1
+        # Attributes and blank lines sit legally between a doc and its item.
+        j = i
+        while j < len(lines) and kinds[j] in (ATTR, SKIP):
+            j += 1
+        if j >= len(lines):
+            # A doc block at end of file documents nothing.
+            orphans.append((start + 1, lines[start].strip(), "end of file"))
+            continue
+        if kinds[j] == DOC:
+            # Two doc blocks with no item between them: the first cannot
+            # reach an item, because the second one gets there first. This
+            # is what an insertion above a documented item leaves behind.
+            orphans.append((start + 1, lines[start].strip(), "another doc block"))
+            continue
+        if NEVER_DOCUMENTED.match(lines[j]):
+            # A line that syntactically cannot carry a doc comment. `use`
+            # is the one that has actually happened (#436) — an import
+            # dropped between a doc and its function takes prose that
+            # rustdoc then renders against the import, and neither the
+            # diff check nor `unused_doc_comments` says a word.
+            orphans.append((start + 1, lines[start].strip(), lines[j].strip()))
+    return orphans
+
+
 def main():
     base, head = sys.argv[1], sys.argv[2]
     declared = git("log", f"{base}..{head}", "--format=%B")
@@ -443,6 +507,17 @@ def main():
         for name, had_doc in before.items():
             if had_doc and name in after and not after[name] and (f, name) not in exempt:
                 losses.append((f, name, base))
+    # The head-only pass (#427, #436): a capture the diff cannot see because
+    # its victim is new on the branch, or because the capturing item is of a
+    # kind `ITEM` does not model. Runs on the same changed files, needs no
+    # base, and reports separately so the two failure shapes stay legible.
+    orphans = []
+    for f in changed:
+        text = git("show", f"{head}:{f}", allow_missing_path=True)
+        if not text:
+            continue
+        for line_no, first, follower in orphaned_docs(text):
+            orphans.append((f, line_no, first, follower))
     for f, name, at in losses:
         print(
             f"::error file={f}::`{name}` had a doc comment at {at[:12]} and has none now. "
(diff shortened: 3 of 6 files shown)


vitali87:
## 🤖 Fallback review — Claude Opus 5 (APPROVE)

CodeRabbit is rate-limited, so this Opus reviewer is the artifact. It ran the script directly against throwaway git repos rather than reading it — **40 constructed shapes with expected-flag assertions, zero unexpected results**, plus a whole-tree sweep (168 files, 0 flagged) and a red/green check that swapped in main's script in place to confirm both bug tests genuinely fail without the fix.

It also independently found the `pub use` false positive at the earlier head, before my own sweep reached it — and verified the known-miss test genuinely constrains by proving the fixture is an otherwise-detectable capture.

Review of croft PR #437 at head **731d487**

## Walkthrough

731d487 is documentation-only over b29e748: no change to `NEVER_DOCUMENTED`, no change to the `orphaned_docs` algorithm. It rewrites the exemption comment, adds a KNOWN LIMITATION paragraph to the `orphaned_docs` docstring, and adds two tests (EOF orphan; documented-newcomer known miss). The PR body gains a **What this does not catch** section and the closing keyword is now "Closes #436. Partially addresses #427". All three of my findings are resolved, and I re-verified each against the real script rather than reading the diff alone.

## Findings

**Item 1 — the known-miss test does constrain. Verified with a positive control.** This was the right thing to be suspicious of: a test asserting exit 0 passes for *any* reason the script exits 0, including reasons that have nothing to do with the miss it claims to record. So I checked the fixture is a genuine, otherwise-detectable capture rather than a no-op.

It is. `documented()` on the fixture's two states shows `alpha: True` → `alpha: False` with `Inserted: True` — the prose provably moved from `alpha` to `Inserted`. And the control: re-running the same final tree with the base set one commit later, so `alpha` exists at the base, exits **1** with `` `alpha` had a doc comment … and has none now``. So the exit 0 in the test is caused specifically by the #427 blind spot — victim new on the branch — and not by the fixture being uninteresting. If someone later fixes this, the test flips to exit 1 and fails loudly, which is the inversion the docstring asks for. Nothing weaker is being asserted.

**Item 2 — the exemption comment now matches the regex exactly.** Confirmed against `NEVER_DOCUMENTED` across every visibility form, 16 cases, zero mismatches: bare `use` (plus leading-space, leading-tab, and `use\tstd::fmt`) matches; `pub use`, `pub(crate) use`, `pub(in crate::a) use`, `pub(super) use`, `pub(self) use`, `pub (crate) use` and `pub extern crate` do not; `extern crate` matches; `used_by()`, `user_x=1`, `useful()` do not. "Any VISIBILITY-QUALIFIED `use` is excluded, not just `pub use`" is now literally true of the pattern. The added sentence distinguishing `pub(crate)` (not a rustdoc concern, but a doc there is a deliberate note) gives the exemption a rationale that covers the whole set rather than just the `pub use` case.

**Item 3 — EOF test added.** `test_a_doc_block_at_end_of_file_is_reported` covers the third reported shape, which previously had none. All three branches of `orphaned_docs` now have a test.

**Item 4 — your reasoning on end-of-`mod`/`impl` is correct; I withdraw the suggestion.** You asked me to push back if I disagreed. I do not. My own finding conceded rustc's `unused_doc_comments` covers it, and I confirmed the shape is silent at this head. The argument against adding it is stronger than I gave it credit for: a bare `}` arm would have to distinguish a doc stranded before a closing brace from a doc block that legitimately precedes an item on a later line, and the SKIP-stepping loop already walks past blanks and comments to get there. That is new false-positive surface for a case another tool already catches — exactly the trade this gate should refuse. Leaving it out is right.

**No new findings at this head.** Re-ran the full false-positive sweep — 40 shapes, including every one from the original brief — with expected-flag assertions rather than eyeballing: **zero unexpected results**. The six true positives still fire (private `use`, tab-indented, `#[cfg]`-gated, `extern crate`, `use` in a fn body, EOF orphan, doc-doc capture); every legal shape stays silent, including `//!` inner docs, `/*! */`, `impl`/`unsafe impl`/`mod`/`extern "C"`, the three macro invocations, `#[doc = "…"]` both orderings, raw and plain strings containing `///`, and all five visibility-qualified `use` forms. Whole-tree sweep: **168 files, 0 flagged**. Suite: **65 tests, OK**.

## On your two judgement calls

**The derive — leave it on `ScrollClock`, and note the provenance. Your lean is right, for a slightly different reason than you gave.**

Your reason was "moving it changes a struct nothing asked me to touch". True, but the stronger reason is that the derive's *history* is not evidence about where it belongs *now*. What 2051218 shows is that `PaneAnnotation` was introduced with `#[derive(Clone, Debug)]` — but it also shows `PaneAnnotation` was a plain five-field data struct at the time (`line_rec`, `hist_rec`, `start`, `len`, `text`). The derive was speculative then and is unused now: I checked every use site, and `PaneAnnotation` is only `push`ed, indexed, `get_mut`, and `remove`d, never cloned or `{:?}`-formatted, and its holder `PtyTerminal` has no derives of its own. So restoring it to `PaneAnnotation` would restore an unused derive to match a historical accident, which is not a better state — it is the same dead code in a different place.

Meanwhile `ScrollClock` has carried `Clone, Debug` through several commits and is the more plausible future user of both: it is a small `i64`/`Option<i32>` value type behind an `Arc<Mutex<…>>`, and `Debug` on it is exactly what you would want when diagnosing the lock-order invariant its own doc comment describes. Neither struct needs either trait today and nothing can change behaviour either way — the `clock.clone()` at line 1627 is `Arc::clone`, so `ScrollClock::clone` has no call site.

The one thing worth fixing is the claim, not the code: the commit message and PR body read as though the derive was always `ScrollClock`'s and the doc had drifted off it. Only the doc drifted; the derive was `PaneAnnotation`'s and was left behind. A sentence saying so — "the `derive` originated on `PaneAnnotation` and is left on `ScrollClock`, where it has effectively lived and where it is the more plausible fit; both structs use neither trait today, so this is behaviour-neutral" — makes the next person's `git log -S` land somewhere consistent. That is a minor, and it is the only thing I would change here.

**The nitpicks — take two, drop one.**

- *Success message* — **take it.** This one is more than cosmetic. `No documentation lost across N changed Rust file(s)` is now the pass line for a check that also asserts nothing is stranded, and CLAUDE.md's false-greens section is precisely about a success message that does not name what it verified. Someone reading a green log has no way to tell the head-only pass ran at all. "No documentation lost or stranded across N…" costs one word and makes the green line describe the actual assertion.
- *Misplaced comment block* — **take it.** Cheap, and the irony is real: in a PR about prose detached from the thing it describes, the pre-existing "What a line is, for the backward scan…" comment now sits above `NEVER_DOCUMENTED` and reads as if it explains it. Moving the new block below the `DOC, ATTR, SKIP, CODE` line, or above the `classify` comment, is a two-line move.
- *Trailing blank line* — **noise, drop it.** I raised it at 7407ef8 and it is not worth a round trip; whatever formatter runs on this repo will settle it, and it changes nothing a reader sees.

## False-positive sweep

Unchanged from my b29e748 sweep and re-verified in full at 731d487 — 40 shapes, 0 unexpected. Summary of the rows that matter:

| Shape | Flagged? | Correct? |
|---|---|---|
| `//!` inner doc / `/*! */` / `//!` alone | no | yes |
| `impl` / `impl Trait for X` / `unsafe impl` / `mod` / `extern "C" {}` | no | yes |
| `lazy_static!` / `bitflags!` / `thread_local!` | no | yes |
| `#[cfg]`-gated item; doc+blank+attr+item; doc + `//` note + item | no | yes |
| `#[doc = "…"]`, both orderings; two adjacent `///` | no | yes |
| enum variant / struct field / `where`-wrapped fn / item in macro body | no | yes |
| raw string and plain string containing `///` | no | yes |
| `pub use`, `pub(crate) use`, `pub(in path) use`, `pub(super) use`, `#[cfg]`+`pub use`, `pub extern crate` | no | yes |
| `used_by_nothing();` | no | yes — `\s` boundary holds |
| doc at end of `mod {}` / `impl {}` | no | accepted miss (rustc covers) |
| documented newcomer (#427 residue) | no | documented + pinned by test |
| private `use` (bare / tab / `#[cfg]`-gated) | **yes** | correct |
| `extern crate`; `use` in fn body; EOF orphan; doc-doc capture | **yes** | correct |

## Test coverage

65 tests, all green at 731d487. Hermetic — every test builds its own repo in a `TemporaryDirectory` with its own git identity. Red/green verified earlier by swapping in main's script in place: both bug-reproducing tests failed `0 != 1` for the right reason and pass after; controls stayed green on both. Worktree restored clean. All three reported shapes now have a test, plus a `pub use` regression test and a pinned known miss with a verified positive control.

## Nitpicks

- `scripts/check_doc_ownership.py:277` — new comment block still sits between `classify`'s comment and the constants it explains (taking).
- `scripts/check_doc_ownership.py:~540` — success message does not mention stranding (taking).
- PR body, **What this does not catch** — says "three shapes still pass" above two bullets; the second bullet merges `mod`/`impl`/macro into one. Either say "two" or split the bullet. The docstring has the same count, phrased as "All three", which reads fine there because it enumerates them.
- Trailing blank line in the test file — noise, dropping.

## Conventions

Re-checked against `origin/main` at review time, not reused from earlier: main is **0.1.827**, PR is **0.1.829**, `Cargo.lock` agrees. The gate at `ci.yml:69-71` is strictly-greater (`sort -t. -k1,1n -k2,2n -k3,3n | tail -n1`, failing only on `new = old` or `highest != new`), so skipping .828 is fine. Release note `src/release_notes/0.1.829.md` present, non-empty, `fix:`-prefixed; no other version's notes touched. No CI workflow change needed — `ci.yml:133-136` invokes the script unchanged under `set -euo pipefail`. Per CLAUDE.md, re-read main's version in the same breath as the final `gh pr checks`: #432 holds .827 and #435 holds .828, so a collision is still live if either merges first.

## Scores

**CodeRabbit score — 5/5** (only nitpicks remain; the pass is clean across every shape tested)

**Greptile score — 5/5** (the major is resolved: the closing keyword now matches what the pass delivers, the limitation is in the docstring, and the known miss is pinned by a test that genuinely constrains)

The gate produces no false positives across 168 files and 40 constructed shapes, catches both issue cases, and is now honest about what it does not catch. The three remaining items are a provenance sentence and two comment/message polish edits — none blocking.

VERDICT: APPROVE

---
*The two endorsed nitpicks and the derive provenance are in at `8e0b0dc`: the green line now reads "No documentation lost **or stranded**" (a pass line that did not name what it asserted, in the gate built to catch exactly that), each comment block sits above what it explains, and the `#[derive]`'s history is recorded where the next `git log -S` will find it.*

PR #438, opened 2026-08-31
fix(test-budget): measure pressure, not machine size

### LINKED ISSUE #422: test_budget: the load scale compares a thread count against a ratio

vitali87:
Closes #422.

## The unit error

`scale_from` took `max(threads, load / cpus)` — a **count** against a **ratio**. The consequence is the one #422 names: the scale tracked how *big* the machine was rather than how *loaded* it was.

| scenario | old | new |
|---|---|---|
| quiet 4-CPU box, default threads | 4 | 2 |
| **64-core box, 64 threads, no contention at all** | **8 (the cap)** | **2** |
| 4-CPU box at a genuine 3.5x oversubscription | 4 | 5 |
| #397's flake: 4-CPU, `--test-threads=8`, six hogs | 8 | 4 |

## The rule

`(threads + load) / cpus` — runnable work per CPU. One division, both inputs in the same units.

They **add** rather than compete because the suite's own threads and whatever else owns the machine are contending for the same cores. Each signal is still blind where the other sees — which is what made the old `max` attractive — and summing preserves that without pretending a count and a ratio are comparable.

## Why the bases moved in the same change

#422 is explicit that the scale fix alone is a regression, and it is right. The old rule read at least 4 on any machine with four or more cores, so every base was effectively `base * 4`. Measuring pressure correctly reads **2** on a quiet machine — halving every budget in a suite whose own module doc records that *"every one of them was observed failing at 1x"*.

`BASE_CALIBRATION = 2` restores the wall clock. Applied **once in the module**, not at seventeen call sites, for two reasons: doubling seventeen literals by hand invites missing one, and a missed one is a flake that looks like the change under test — the exact failure this module exists to remove; and it keeps `base` the number a test author can actually reason about ("a shell printing a line costs about a second") rather than one pre-multiplied for a rule they would have to know about.

Net effect on wall clock. "No wait got shorter" would be false and the table below shows why — a quiet machine's waits deliberately halve, because they were paying for contention that was not there. The claim that matters is that no wait **under real load** got shorter:

| machine | old effective | new effective |
|---|---|---|
| quiet 8-CPU dev box, 1s base | 8s | **4s** (shorter, and the issue's point) |
| #397's flake case, 1s base | 8s | **8s** (held exactly) |
| 4-CPU at 3.5x oversub, 1s base | 4s | **10s** (longer, correctly) |

The quiet-machine budget *shrinking* is the intended outcome: those waits were paying for contention that was not there.

## The ceiling that could not move — in both directions

`MAX_SCALE` is halved to 4 in the same change. Its doc promises to bound what a *broken* test costs a developer, and applying a x2 calibration while leaving the clamp at 8 would have doubled that silently: the seven 10s-base call sites going from 80s to 160s before failing. `BASE_CALIBRATION * MAX_SCALE` is the 8 it always was, and a test pins the product, because the two constants are only safe as a pair.

## The ceiling that could not move

#397's measured flake — a 4-CPU box at `--test-threads=8` under six spinning hogs — needed a wait budgeted at 4s to have 8s. The new rule reads 4 there rather than 8, so it clears only because the calibration is applied. `the_397_flake_still_gets_its_eight_seconds` asserts exactly that, through the real formula rather than a hand-doubled literal, and it is what stops the scale half being changed again without the base half.

## Verification

Not arithmetic. The seventeen process-spawning tests that actually consume these budgets were run on a genuinely contended machine — **load average 17.8 on 8 CPUs** — and all pass: `gui_path` (17 tests, real login shells), `pending_bytes_counts_advanced_output_and_resets_on_take_dirty` and `cmd_clicking_a_non_web_hyperlink_refuses_instead_of_opening` (the two the module's notes call out as fixed-wall-clock pollers), `a_record_rewrite_re_arms_a_downed_navigator`, `accept_all_incoming_then_complete_merge_stages_the_resolved_file`, and the newer `an_http_file_runs_…` / `the_watcher_reports_…` / `the_failure_site_survives_…`.

New tests: a big quiet machine is not a loaded one (the case the old rule got backwards, and which returned the cap); both signals contribute and are worse together; the clamp holds at both ends including a zero/NaN `cpus`; and the #397 ceiling.

## Gate

`test_budget::` 9 passed, the spawning suite green under real load; clippy `-D warnings` and fmt clean; doc-ownership gate clean. Version 0.1.830 (main is 0.1.829), release note in `src/release_notes/0.1.830.md`; renumbered at merge time if main moves.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected test load scaling to measure system pressure based on runnable work per CPU.
  * Adjusted baseline budgets to preserve calibrated wait times and reduce flaky test behavior.

* **Chores**
  * Updated the package version to 0.1.830.
  * Added release notes for the load-scaling correction.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.829"
+version = "0.1.830"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.829"
+version = "0.1.830"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/release_notes/0.1.830.md
+++ b/src/release_notes/0.1.830.md
@@ -1,0 +1,1 @@
+fix: the test suite's load scale now measures pressure (runnable work per CPU) rather than machine size, so a big idle machine no longer gets the maximum stretch while a genuinely crushed small one gets half of it; the base budgets moved with it so no wait got shorter.
--- a/src/release_notes/0.1.830.md
+++ b/src/release_notes/0.1.830.md
@@ -1,1 +1,1 @@
-fix: the test suite's load scale now measures pressure (runnable work per CPU) rather than machine size, so a big idle machine no longer gets the maximum stretch while a genuinely crushed small one gets half of it; the base budgets moved with it so no wait got shorter.
+fix: the test suite's load scale now measures pressure (runnable work per CPU) rather than machine size, so a big idle machine no longer gets the maximum stretch while a genuinely crushed small one gets half of it; no wait under real load got shorter, and a quiet machine's deliberately did.
--- a/src/test_budget.rs
+++ b/src/test_budget.rs
@@ -10,10 +10,14 @@
 //! right number is therefore not knowable from inside a single test, which is
 //! why raising individual caps has already been tried twice on one of them.
 //!
-//! So the budget stops being a guess and becomes `base x scale`, where `base`
-//! is what the operation costs on a quiet machine (the number a test author
-//! can actually reason about) and `scale` is measured from the load the suite
-//! is running under.
+//! So the budget stops being a guess and becomes
+//! `base x BASE_CALIBRATION x scale`, where `base` is what the operation
+//! costs on a quiet machine (the number a test author can actually reason
+//! about), `scale` is measured from the load the suite is running under, and
+//! the calibration is the one constant that keeps the wall clock where it was
+//! when these budgets were derived (#422). The widest any budget stretches is
+//! `BASE_CALIBRATION x MAX_SCALE`, deliberately unchanged at 8x, because a
+//! broken test must still fail in a time a developer will wait.
 
 use std::sync::OnceLock;
 use std::time::{Duration, Instant};
@@ -21,7 +25,13 @@ use std::time::{Duration, Instant};
 /// Widest a budget may stretch. A failing test must still fail: past this the
 /// wait is no longer distinguishing "loaded" from "broken", and every extra
 /// second is paid by the developer whose change genuinely broke the thing.
-const MAX_SCALE: u32 = 8;
+///
+/// Halved with the #422 rule change, so the CEILING is unchanged. The total
+/// stretch is `BASE_CALIBRATION * MAX_SCALE`, and leaving this at 8 while
+/// introducing a x2 calibration would have doubled the worst case with it —
+/// a broken test on a 10s base going from 80s to 160s, which is exactly the
+/// cost this constant exists to bound. 2 x 4 is the 8 it always was.
+const MAX_SCALE: u32 = 4;
 
 /// Narrowest it may shrink. Never below the budgets these tests already had,
 /// since every one of them was observed failing at 1x.
@@ -62,8 +72,8 @@ fn load_average() -> Option<f64> {
 
 /// How much slack this machine needs, resolved once per test binary.
 ///
-/// Two signals, and the maximum of them, because each is blind where the
-/// other sees. The suite's own thread count is known before anything runs
+/// Two signals, and their SUM, because each is blind where the other sees
+/// and both contend for the same cores (#422). The suite's own thread count is known before anything runs
 /// and covers the self-inflicted case (dozens of test shells at once), where
 /// a load average would still be reading the quiet minute before the suite
 /// started. The load average covers what no test can know: a second cargo
@@ -168,8 +178,22 @@ fn scale_from(threads: f64, load: f64, cpus: f64) -> u32 {
     if !(threads.is_finite() && load.is_finite() && cpus.is_finite()) || cpus < 1.0 {
         return MIN_SCALE;
     }
-    // `load` is already the machine-wide runnable average, threads is this
-    // suite's own contribution; the sum is what is competing for `cpus`.
+    // `load` is the machine-wide runnable average, `threads` this suite's own
+    // contribution; the sum is what is competing for `cpus`.
+    //
+    // Those overlap once the suite has been running a minute — its own
+    // threads then appear in the load average too, so the sum counts them
+    // twice. Deliberately left alone. The overlap is bounded (it can only
+    // raise the quotient by `threads / cpus`, one clamp step in practice)
+    // and it errs toward MORE slack, which is the safe direction for a
+    // deadline; subtracting an estimate of our own load would be a guess
+    // about a number we cannot observe, and a wrong one shortens budgets.
+    // `load_scale` memoises in a `OnceLock`, so in practice the reading is
+    // taken at the first wait — early, while the average still describes
+    // the machine the suite arrived on.
+    // Belt and braces: the guard above already rules out every input that
+    // could make this non-finite, so this branch is unreachable today. Kept
+    // because the guard is the thing most likely to be relaxed later.
     let raw = ((threads + load) / cpus).ceil();
     if !raw.is_finite() {
         return MIN_SCALE;
@@ -203,7 +227,8 @@ pub fn await_spawned(base: Duration, what: &str, mut ready: impl FnMut() -> bool
         std::thread::sleep(Duration::from_millis(20));
     }
     panic!(
-        "timed out after {budget:?} ({base:?} x{scale} for load) waiting for {what}; \
+        "timed out after {budget:?} ({base:?} x{BASE_CALIBRATION} calibration \
+         x{scale} for load) waiting for {what}; \
          if this is a real failure it would also fail at 1x, so re-run the full suite \
          on the unmodified merge base under the same load before suspecting your diff"
     );
@@ -286,9 +311,32 @@ mod tests {
             scale_from(4.0, 12.0, 4.0) > quiet,
             "so does load from outside the suite"
         );
+        // Strict, on inputs where summing genuinely separates from `max`:
+        // (4 + 4) / 4 = 2 against max(4, 4) / 4 = 1. A `>=` here would pass
+        // even if the sum were replaced by the old maximum.
         assert!(
-            scale_from(12.0, 12.0, 4.0) >= scale_from(12.0, 0.0, 4.0),
-            "and together they are worse than either alone"
+            scale_from(6.0, 6.0, 4.0) > scale_from(6.0, 0.0, 4.0),
+            "together they are worse than either alone"
+        );
+    }
+
+    /// The total stretch is what a broken test pays, and it must not have
+    /// grown. `BASE_CALIBRATION * MAX_SCALE` was 1 x 8 before #422 and is
+    /// 2 x 4 after — the same 8. Pinned because the two constants are only
+    /// safe as a PAIR: raising either alone doubles what a developer waits
+    /// to be told their change broke something, which is the cost
+    /// `MAX_SCALE` exists to bound.
+    #[test]
+    fn the_worst_case_stretch_is_unchanged() {
+        assert_eq!(
+            BASE_CALIBRATION * MAX_SCALE,
+            8,
+            "a broken test on a 10s base still fails in 80s, not 160s"
+        );
+        assert_eq!(
+            BASE_CALIBRATION * MIN_SCALE,
+            4,
+            "and a quiet machine waits 4x, down from the old 8x"
         );
     }
 
@@ -304,7 +352,7 @@ mod tests {
     #[test]
     fn the_397_flake_still_gets_its_eight_seconds() {
         let scale = scale_from(8.0, 6.0, 4.0);
-        assert_eq!(scale, 4, "runnable work per CPU: (8 + 6) / 4");
+        assert_eq!(scale, 4, "runnable work per CPU: ceil((8 + 6) / 4)");
         // Through the real formula, not a hand-doubled literal: the base a
         // test author writes is 1s, and the module supplies the calibration.
         let base = std::time::Duration::from_secs(1);
--- a/src/test_budget.rs
+++ b/src/test_budget.rs
@@ -27,6 +27,25 @@ const MAX_SCALE: u32 = 8;
 /// since every one of them was observed failing at 1x.
 const MIN_SCALE: u32 = 2;
 
+/// Recalibration for the #422 rule change, applied once here rather than at
+/// seventeen call sites.
+///
+/// The old scale tracked machine SIZE, so on any box with four or more cores
+/// it read at least 4 and a `base` was effectively `base * 4`. Measuring
+/// PRESSURE instead correctly reads 2 on a quiet machine — which would have
+/// halved every budget in the suite, and the module's own note warned that
+/// every one of these waits was observed failing at 1x. Doubling the base
+/// holds the wall clock where it was calibrated while the scale goes back to
+/// meaning what it says.
+///
+/// Deliberately central. Doubling seventeen literals by hand invites missing
+/// one, and a missed one is a flake that looks like the change under test —
+/// the exact failure this module exists to remove. It also keeps `base` the
+/// number a test author reasons about ("a shell printing a line costs about
+/// a second") rather than a number pre-multiplied for a rule they would have
+/// to know about.
+const BASE_CALIBRATION: u32 = 2;
+
 /// One-minute load average, where the platform reports one.
 fn load_average() -> Option<f64> {
     #[cfg(any(target_os = "linux", target_os = "macos"))]
@@ -56,8 +75,11 @@ pub fn load_scale() -> u32 {
             .map(|n| n.get())
             .unwrap_or(1) as f64;
         let threads = configured_threads().unwrap_or(cpus);
-        let oversubscribed = load_average().map(|l| l / cpus).unwrap_or(1.0);
-        scale_from(threads, oversubscribed)
+        // The RAW load average, not a ratio: `scale_from` divides once, by
+        // the same `cpus`, after summing. Dividing here too was the unit
+        // error (#422).
+        let load = load_average().unwrap_or(0.0);
+        scale_from(threads, load, cpus)
     })
 }
 
@@ -122,15 +144,33 @@ fn threads_from_args<I: IntoIterator<Item = String>>(args: I) -> Option<f64> {
 /// The rule itself, split out so it can be tested without a machine in a
 /// particular state.
 ///
-/// NOTE (#422): `threads` is a count and `oversubscription` is a ratio, so
-/// the `max` between them compares two different quantities. Fixing that in
-/// isolation SHRINKS every budget (a quiet 4-CPU box goes from 4 to the
-/// floor of 2), because the base budgets were calibrated against a scale
-/// that was effectively the core count. It needs the bases recalibrated with
-/// it, which is its own change; this one only makes the existing rule see
-/// the thread count it always meant to read.
-fn scale_from(threads: f64, oversubscription: f64) -> u32 {
-    let raw = threads.max(oversubscription).ceil();
+/// **Runnable work per CPU** (#422). The old rule was `max(threads, load /
+/// cpus)`, which compared a COUNT against a RATIO: on a 64-core box running
+/// 64 threads with no contention at all it returned the cap of 8, while a
+/// genuinely crushed 4-core box returned 4. The scale was tracking machine
+/// size rather than pressure, which is the one thing it exists to measure.
+///
+/// The two signals ADD rather than compete. The suite's own threads and
+/// whatever else owns the machine are contending for the same cores, so the
+/// runnable count is their sum, and dividing by `cpus` turns it into the
+/// factor by which everything is slower than it would be alone. Both inputs
+/// keep the property that made the old `max` attractive — each is blind
+/// where the other sees, and summing preserves that without pretending a
+/// count and a ratio are the same kind of number.
+///
+/// This deliberately reads LOWER than the old rule on a quiet machine (a
+/// quiet 4-CPU box: 4 before, 2 now), which is why the bases moved with it
+/// in the same change. See [`await_spawned`] for the ceiling that had to be
+/// preserved: #397's measured flake — a 4-CPU box at `--test-threads=8`
+/// under six spinning hogs — needed 8 seconds where the base was 4, and
+/// with doubled bases it still gets them.
+fn scale_from(threads: f64, load: f64, cpus: f64) -> u32 {
+    if !(threads.is_finite() && load.is_finite() && cpus.is_finite()) || cpus < 1.0 {
+        return MIN_SCALE;
+    }
+    // `load` is already the machine-wide runnable average, threads is this
+    // suite's own contribution; the sum is what is competing for `cpus`.
+    let raw = ((threads + load) / cpus).ceil();
     if !raw.is_finite() {
         return MIN_SCALE;
     }
@@ -141,7 +181,7 @@ fn scale_from(threads: f64, oversubscription: f64) -> u32 {
 /// timeout to something else (a `recv_timeout`, a probe's own deadline) rather
 /// than polling.
 pub fn spawn_budget(base: Duration) -> Duration {
-    base * load_scale()
+    base * BASE_CALIBRATION * load_scale()
 }
 
 /// Poll `ready` until it answers true, within a load-scaled `base` budget.
@@ -154,7 +194,7 @@ pub fn spawn_budget(base: Duration) -> Duration {
 #[track_caller]
 pub fn await_spawned(base: Duration, what: &str, mut ready: impl FnMut() -> bool) {
     let scale = load_scale();
-    let budget = base * scale;
+    let budget = base * BASE_CALIBRATION * scale;
     let started = Instant::now();
     while started.elapsed() < budget {
         if ready() {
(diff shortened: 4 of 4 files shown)


vitali87:

CodeRabbit is rate-limited, so this is the deep review. The reviewer ported `scale_from` (old and new) to Python and swept **3060 configurations** — every cpus/threads/load combination the brief named — to find every case where the new effective budget is shorter than the old.

The headline results are worth reading even if you skip the rest: **max shrink anywhere is 2.0x**, confined to idle machines with oversubscribed threads (the intended outcome of #422); **1-CPU, 2-CPU and actual CI have zero shrink cases**; and #397's flake case holds at exactly 8s, verified rather than asserted.

Review of PR #438 — head `c42c883` ("fix(test-budget): measure pressure, not machine size", closes #422)

## Walkthrough

`scale_from` is re-derived from `max(threads, load/cpus)` — a count compared against a ratio — to `((threads + load) / cpus).ceil()`, clamped to `[MIN_SCALE, MAX_SCALE]` = `[2, 8]`. `load_scale` now passes the raw load average and `cpus` through rather than pre-dividing. Because the correct rule reads `2` on a quiet machine where the old one read the core count, a new module-level `BASE_CALIBRATION = 2` multiplies every base in `spawn_budget` and `await_spawned` to hold the wall clock. A guard clause returns `MIN_SCALE` for non-finite inputs and `cpus < 1.0`. Four new/rewritten unit tests cover the big-quiet-machine case, additive contribution, the clamp including degenerate `cpus`, and the #397 ceiling. Version 0.1.830, `Cargo.lock` in step, release note present.

I ported `scale_from` (old and new) to Python and swept `cpus ∈ {1,2,4,8,16,32,64,128}` × `threads ∈ 1..2·cpus` × `load ∈ {0, cpus/4, cpus/2, cpus, 2·cpus, 4·cpus}`, comparing effective budgets `base·old_scale` vs `base·BASE_CALIBRATION·new_scale`, plus a fine-grained sweep at 0.1 load granularity to find the shrink boundary exactly.

The direction of the fix is right and #422's clauses are met: the ratio rule lands, the bases are re-derived in the same change as the issue demanded, and the #397 reproduction is encoded as a regression test. The problem is not the rule — it is that `BASE_CALIBRATION` was multiplied into the budget *outside* the clamp, so the constant documented as "widest a budget may stretch" no longer bounds anything.

## Findings

**`src/test_budget.rs:183-184,197` [major/correctness] — `BASE_CALIBRATION` is applied outside the clamp, so `MAX_SCALE` no longer bounds the budget; the true ceiling silently doubled from 8x to 16x.**

`MAX_SCALE`'s own doc comment (line 21-23) states its contract: *"Widest a budget may stretch. A failing test must still fail: past this the wait is no longer distinguishing 'loaded' from 'broken', and every extra second is paid by the developer whose change genuinely broke the thing."* That is a statement about the **budget**, not about an internal scalar. After this change the maximum multiplier applied to any base is `BASE_CALIBRATION * MAX_SCALE = 16`, so the guarantee the constant documents is false wherever the clamp binds.

This is exactly the second failure mode the task brief names — a genuinely broken test taking minutes to fail. Measured against the real bases in the tree:

| call site | base | old ceiling | new ceiling |
|---|---|---|---|
| `src/widgets/terminal.rs:5506`, `:6932` | 500ms | 4.0s | 8.0s |
| `src/session_host.rs:2737,2751`, `src/app/tests.rs:6550,30860` | 1000ms | 8.0s | 16.0s |
| `src/widgets/terminal.rs:5110` | 1200ms | 9.6s | 19.2s |
| `src/log_view.rs:1062` + six sites in `src/app/tests.rs` (`:36187,36461,36502,36549,37042,37063`) | 10s | 80s | **160s** |

Seven distinct call sites carry a 10s base. A broken one of those now burns **two minutes forty** before failing, against eighty seconds before — and that is per test, on a machine loaded enough to hit the cap, which is precisely the machine where several will hit it at once. Note the cap binds much more readily than before: on a 4-CPU box the old rule needed `load/cpus > 8` (load 32) to reach 8, while the new rule reaches it at `threads + load > 28`, i.e. load 24 at 4 threads.

The panic message compounds it. Line 206 renders `"{base:?} x{scale} for load"` — for a 10s base at the cap it prints `timed out after 160s (10s x8 for load)`, which does not multiply out. A reader checking the arithmetic sees a factor-of-two discrepancy and no mention of `BASE_CALIBRATION`, on the one output path whose entire job is to tell a developer whether they are looking at load or at breakage.

*Fix:* fold the calibration inside the clamp so one constant keeps bounding the budget, and let the reported scale be the real multiplier:

```rust
// in scale_from, replacing the final line:
(raw as u32 * BASE_CALIBRATION).clamp(MIN_SCALE * BASE_CALIBRATION, MAX_SCALE)
```

with `MAX_SCALE` re-derived to the intended wall-clock ceiling (16 if the doubled ceiling is deliberate, 8 if not), and `spawn_budget`/`await_spawned` reverting to `base * load_scale()`. If the current split is preferred, then `MAX_SCALE`'s doc comment must stop claiming to bound the budget, the panic message must include the calibration factor, and the choice of a 16x ceiling needs stating as a decision rather than arriving as a side effect. Either way it should not be silent — the PR body's own framing is "the thing that must not shrink", and the ceiling growing is not discussed anywhere in it.

**`src/test_budget.rs:65-70` [minor/documentation] — `load_scale`'s doc still describes the rule this PR deleted.**

It reads *"Two signals, and **the maximum of them**, because each is blind where the other sees."* The maximum is gone; the signals now add. The rest of the paragraph (each signal blind where the other sees) survives the change and is still correct, but the operative clause is now false, sitting directly above the call it describes. `scale_from`'s doc was rewritten thoroughly for the new rule while this one was missed. *Fix:* "Two signals, and their sum, because each is blind where the other sees."

**`src/test_budget.rs:13-16` [minor/documentation] — the module header still states the budget is `base x scale`.**

*"the budget stops being a guess and becomes `base x scale`"*. It is now `base x BASE_CALIBRATION x scale`. This is the module's orienting sentence and the first thing a reader hits; leaving it describing the pre-PR arithmetic is how the next person re-derives a wrong number. *Fix:* state the calibration factor here, or make it moot by folding it into the scale per the major finding.

**`src/test_budget.rs:168` [nit/robustness] — the `cpus < 1.0` guard makes the later `!raw.is_finite()` check unreachable.**

With `threads`, `load`, `cpus` all finite and `cpus >= 1.0`, `(threads + load) / cpus` is finite, so `raw` is always finite and line 174-176 is dead. Harmless and arguably good defensive practice, but it reads as though it still catches something. Worth a one-line comment saying it is belt-and-braces, or dropping it.

## Numerical sweep

Sweeping the specified grid: 3060 cells, of which **2136 show a strictly shorter effective budget under the new rule**. That headline count is misleading and the structure matters more, so here are the canonical classes — every shrink case in the entire grid collapses into six:

| old eff | new eff | cases | representative | judgement |
|---|---|---|---|---|
| 8x | 4x | 1411 | cpus=4, threads=8, load=0 | **Not a flake.** Oversubscribed threads on an *idle* machine. `threads=8` on 4 CPUs with load 0 means the suite is about to create the contention but the load average has not seen it yet. Old rule paid 8x for the thread count alone; new rule reads `(8+0)/4 = 2`, doubled to 4x. This is #422's thesis working exactly as intended — the old number was paying for contention not yet present. The load average catches up within a minute and later-initialising binaries read the real figure. |
| 8x | 6x | 650 | cpus=4, threads=8, load=1 | **Not a flake.** Same class, mild real load. 6x still exceeds the 4x that a quiet 4-CPU box got under the old rule, which is the calibration point the module's "observed failing at 1x" note anchors on. |
| 7x | 4x | 22 | cpus=4, threads=7, load=0 | Same class, odd thread count. Benign. |
| 6x | 4x | 23 | cpus=4, threads=6, load=0 | Same class. Benign. |
| 5x | 4x | 23 | cpus=4, threads=5, load=0 | Barely a change; a 5x budget was itself an artifact of the count/ratio confusion. |
| 7x | 6x | 7 | cpus=4, threads=7, load=2 | Marginal, 14% shorter. Benign. |

**The maximum shrink anywhere in the space is exactly 2.0x** (8x → 4x), and it requires `cpus >= 4` **and** `threads >= 5` **and** low load simultaneously. There is no configuration in which the budget shrinks by more than half, and none in which a machine under genuine per-CPU pressure loses budget relative to what the old rule gave a comparably-pressured machine.

The boundary cases the brief calls out all resolve favourably:

- **cpus=1**: zero shrink cases. Every cell grows, most substantially (t=1,l=0: 2x → 4x; t=1,l=4: 4x → 10x).
- **cpus=2** (the common CI container): zero shrink cases. `t=4,l=0` and `t=8,l=0` hold flat at 4x and 8x; every other cell grows. A 2-CPU container cannot lose budget under this change.
- **`available_parallelism()` returns 1**: `cpus=1`, `threads` defaults to `cpus`=1. Budget doubles across the board (2x → 4x at rest). Safe — and note the `unwrap_or(1)` at line 76 keeps `cpus >= 1`, so the `cpus < 1.0` guard is unreachable from the real call path and only protects the unit tests.
- **Actual CI** (`ubuntu-latest` = 4 vCPU, `RUST_TEST_THREADS: 4` at `.github/workflows/ci.yml:172`): `threads=4, cpus=4` never enters the shrink region. Flat at 4x for load 0-4, then growing (6x at load 6, 8x at load 12, 14x at load 24). CI strictly gains headroom.
- **Where the clamp bites**: `MIN_SCALE` binds for all `threads + load <= 2·cpus`, which is most quiet machines — intended. `MAX_SCALE` binds at `threads + load > 8·cpus`, considerably more often than the old rule's `load > 8·cpus` — and that is where the major finding's 16x ceiling lands.

**Verification of the PR's own tables (item 4).** Both reproduce exactly. Scale table: quiet 4-CPU 4→2 ✓; 64-core/64-thread 8→2 ✓ (the cap-for-nothing case #422 names); 4-CPU at 3.5x oversub 4→4 ✓ — though the PR writes the new value as 4 while `scale_from(4, 14, 4)` is `ceil(18/4) = 5`; the wall-clock row for the same scenario (10s) is consistent with 5, so the scale table's "4" is a transcription slip, not a code defect. #397 flake 8→4 ✓. Wall-clock table: quiet 8-CPU 8s → 4s ✓; **#397's flake case holds at exactly 8s ✓ — the central claim verifies**; 4-CPU at 3.5x oversub 4s → 10s ✓.

**The PR's claim that "no wait got shorter" is false as stated**, and the PR body contradicts itself on this — the wall-clock table's own first row shows 8s → 4s and calls the shrink intended. The claim should read "no wait under genuine pressure got shorter", which the sweep does support.

**`BASE_CALIBRATION` soundness (item 2).** No double-counting: all seventeen call sites pass raw base durations, and `src/gui_path.rs:316` passes the production constant `PROBE_TIMEOUT` (`Duration::from_secs(3)`, defined at `src/gui_path.rs:28`) unscaled into `spawn_budget`, so it is multiplied exactly once. `spawn_budget` (line 184) and `await_spawned` (line 197) apply it identically — verified by `a_budget_scales_by_exactly_the_scale`, which now asserts the two agree. The 30s base in `a_condition_already_true_returns_at_once` (line 419) is not a runtime hazard: `ready` returns `true` on the first poll so the budget never elapses. The concept is sound; only its placement relative to the clamp is wrong.

**`load_average` double-counting (item 3).** Real, but self-limiting and not worth blocking on. `load_scale` memoises in a `OnceLock`, first forced by whichever test calls `await_spawned`/`spawn_budget` first. Two regimes: forced early (the common case), the 1-minute average still reflects the pre-suite machine, `threads` supplies the suite's own contribution, and there is no double-count — this is precisely the blind-spot pairing the docs describe. Forced late, or in a binary starting after the suite has run a minute, the average has absorbed the suite's own threads and `threads + load` counts them twice. The error is bounded by `threads/cpus` and is in the safe direction (too generous, never too short), and memoisation freezes whichever reading was taken rather than letting it drift upward during the run. Worth a sentence in the `scale_from` doc noting the sum is deliberately conservative when the average has warmed up; not a defect.

## Test coverage

Good, and better than what it replaces — the assertions moved from exact equality to relational (`>`, `>=`) where the exact value was incidental, which is the right call for a formula under active revision.

- `a_big_quiet_machine_is_not_a_loaded_one` covers the exact case #422 filed, including the 64-core cap regression.
- `the_397_flake_still_gets_its_eight_seconds` is the load-bearing one and is well built: it goes through `BASE_CALIBRATION * scale` rather than a hand-doubled literal, so it fails if either half is changed alone. Exactly the guard the PR claims.
- The clamp test's table-driven degenerate cases (`NaN`, `INFINITY`, `cpus=0.0`, `cpus=NaN`) are a real improvement over the single prior case.

Two gaps:

- **No test pins the maximum wall clock.** Every test asserts a *lower* bound (`>= 8s`, `> MIN_SCALE`). Nothing asserts a budget cannot exceed some ceiling, which is why the 16x regression passes a green suite. An assertion such as `spawn_budget(Duration::from_secs(1)) <= Duration::from_secs(16)` — or, better, a direct assertion of the intended ceiling constant — would have caught it, and would guard the "a failing test must still fail" contract that currently has no test at all.
- **`both_signals_contribute_to_the_pressure`'s third assertion is vacuous.** `scale_from(12,12,4) >= scale_from(12,0,4)` compares `ceil(24/4)=6` against `ceil(12/4)=3` clamped — but both clamp to values where `>=` holds trivially, and the message claims "together they are worse than either alone" while `>=` permits equality. With `MAX_SCALE=8` and these inputs it does hold strictly, but the assertion as written would pass even if summing were replaced by `max`. Use `>` and pick inputs where the sum genuinely separates the two rules.

## Nitpicks

- `src/release_notes/0.1.830.md` — the note ends "so no wait got shorter", repeating the PR body's overstatement. On a quiet developer machine waits deliberately halve. Suggest "so no wait under real load got shorter".
- `src/test_budget.rs:47` — the `BASE_CALIBRATION` doc says "applied once here rather than at seventeen call sites". I count seventeen `await_spawned`/`spawn_budget` references outside the module, but one (`src/app/tests.rs:30872`) is a comment in an `.expect` string, not a call. Sixteen actual call sites. Trivial, but the number is quoted three times across the PR and the module.
- `src/test_budget.rs:307` — `assert_eq!(scale, 4, "runnable work per CPU: (8 + 6) / 4")` — `(8+6)/4 = 3.5`, ceiling 4. The message reads as though the division yields 4. Adding "ceil" would save a double-take.

## Conventions

Version 0.1.830 against main's 0.1.829 ✓ (re-check at merge time; parallel PRs collide here). `Cargo.lock` bumped in step ✓. Release note `src/release_notes/0.1.830.md` present and follows the numbering ✓. No `CLAUDE.md`/`CONTRIBUTING.md` changes needed — `CONTRIBUTING.md:379-398` documents the helpers' *usage*, which is unchanged, and does not quote the formula. The stale docs are both inside the module itself (findings 2 and 3).

## Scores

**CodeRabbit score — 3/5**
**Greptile score — 3/5**

One major (the ceiling silently doubling outside the clamp, with a panic message that no longer multiplies out), plus two stale in-module docs the change itself invalidated. The rule change is correct and the sweep clears it of reintroducing flakes — every shrink is bounded at 2.0x and confined to idle machines with oversubscribed threads, which is the intended outcome of #422. What holds this back is that the fix for "budgets must not shrink" was implemented in a way that quietly removed the bound on how far they may grow.

VERDICT: REQUEST_CHANGES


---
*The major and all three minors are fixed at `81e86fb`: `MAX_SCALE` halved so `BASE_CALIBRATION * MAX_SCALE` is the 8 it always was (with a test pinning the product, since the two are only safe as a pair), the panic message now multiplies out, both stale in-module docs corrected, and the vacuous `>=` assertion made strict on inputs that separate the sum from the old `max`. The overstated "no wait got shorter" is corrected in the note and the body — a quiet machine's waits deliberately halve. A re-verify follows.*

vitali87:
## 🤖 Fallback review — Claude Opus 5 (APPROVE)

Same reviewer, resumed at head `b6acc80`. It reproduced the mutation test independently (six mutations against all sixteen `scale_from` assertions), compared the split-vs-fold resolutions across 60 configurations, and **corrected its own earlier call**: the 106-char line was a nit, not a CI blocker, because this repo sets no `wrap_comments` and rustfmt does not reflow doc comments — local and CI agreed all along.

Re-review of PR #438 — head `b6acc80` ("fix(test-budget): a test that passes via the clamp is not a test")

## Walkthrough

The fourth commit rewraps the overlong doc line, adds the companion assertion to `the_397_flake_still_gets_its_eight_seconds`, adds the output-level `spawn_budget(1s) <= 8s` bound to `the_worst_case_stretch_is_unchanged`, moves the belt-and-braces comment below the `let raw`, and corrects "seventeen" to "sixteen" in both places.

I verified each fix, independently reproduced the mutation test rather than accepting the claim, and ran a shape comparison to answer your two open questions. Everything checks out, and both open questions resolve in favour of merging as-is — the second one more strongly than I expected.

## Verification of the four fixes

**1. The blocker.** `awk 'length > 99'` over `src/test_budget.rs` returns nothing; the control (threshold 60, 192 hits) confirms the check is live. Fixed.

**On why your local gate disagreed with CI — it did not.** This is worth pinning down, because your diagnosis has a false premise that would mislead you later. rustfmt does not reflow doc comments *by default*, and this repo does not enable it: no `wrap_comments`, no `comment_width`, no `format_code_in_doc_comments` anywhere in `rustfmt.toml` or the workflows (`max_width = 100` alone does not touch comment text — it governs code). CI runs plain `cargo fmt --check` on the same pinned 1.97.1 toolchain from `rust-toolchain.toml`. So local and CI rustfmt agree exactly, and **neither would have flagged that line**.

Which means the 106-char line was never going to be a red check. My previous review called it a blocker on the reasoning that `cargo fmt --check` reflows doc comments — that reasoning was wrong, and I should have checked `rustfmt.toml` before asserting it. The line was a real inconsistency worth fixing (every other line in the file is under 100, and it read badly), but it was a nit, not a blocker. I over-graded it, and the 2/5 I gave `b6acc80`'s parent was harsher than the evidence supported. Correcting that here rather than carrying it forward.

The useful conclusion for you is the opposite of the one you drew: no lint was silently failing you, and there is no nightly or `wrap_comments` difference to hunt for. If you *want* mechanical enforcement of doc-comment width, `wrap_comments = true` requires nightly rustfmt, so on a pinned stable toolchain the practical option is a CI grep for over-length lines. Optional; the file is self-consistent now.

**2. The major.** Fixed as specified, and I reproduced your mutation test independently rather than trusting it — porting all sixteen `scale_from` assertions in the file to Python and running six mutations through them.

Your three-test claim is **confirmed exactly**. Reverting `((threads + load) / cpus)` to `threads.max(load / cpus)` fails 6 assertions across 3 tests:

| test | assertion that catches the revert |
|---|---|
| `a_big_quiet_machine_is_not_a_loaded_one` | `scale_from(64,1,64)==MIN` and `scale_from(8,0.5,8)==MIN` |
| `both_signals_contribute_to_the_pressure` | all three, including the strict one |
| `the_397_flake_still_gets_its_eight_seconds` | the new companion `scale_from(6,2,4)==MIN` |

**Are the other two failures real discrimination or incidental?** Real, and they are the *strongest* of the three. `a_big_quiet_machine_is_not_a_loaded_one` is the direct statement of #422's thesis — the old rule returns MAX on a 64-core idle box where the new one returns MIN, which is the maximum possible separation between the two rules. `both_signals_contribute_to_the_pressure` fails on all three of its assertions under the revert. Neither is incidental; both are testing exactly the property the revert destroys. Your fix restored discrimination to the third test, and the other two had it all along — my earlier review focused on the #397 test because it was the one whose doc *claims* to be the guard, and it was the one that had lost its teeth.

**One nuance on the companion, which does not need changing.** You asked me to confirm it is clear of both clamps. It is clear of `MAX` (raw 2 vs MAX 4) but sits *on* `MIN`: `ceil((6+2)/4) = 2 = MIN_SCALE`. So in isolation the companion is one-sided — it catches rules that over-produce there (the old `max` rule → 4, dropped division → 4, both caught) but not rules that under-produce (`min(t,l)/c`, a halved divisor, `floor` instead of `ceil` all → 2 and slip through).

That does not matter, because the suite is not one assertion. Running all six mutations against the three-assertion set {companion, #397 value, strict-both-signals}, **every mutation is caught and only the true rule survives**. The under-producing mutations the companion misses are caught by the strict assertion in `both_signals` and by the `scale_from(8,6,4)==4` value assertion. The companion does the job you added it for. If you ever want a single assertion that discriminates on both sides alone, a point strictly between the clamps — `scale_from(6,6,4) == 3` (raw 3) — separates the true rule from all six mutations by itself. Purely optional; the suite already covers it.

**3. The output bound.** `spawn_budget(base) <= base * 8` is present and correctly placed. It asserts the property `MAX_SCALE`'s doc promises rather than the factors, so a third multiplier entering the expression is now caught. This closes the gap properly.

**4. The two minors.** Comment now sits directly above the `if !raw.is_finite()` branch it describes. "sixteen" corrected in both occurrences at lines 41 and 51; I re-counted at this head and confirm sixteen call sites, with `src/app/tests.rs:30872` being the `.expect` string, not a call.

## Your two open questions

### The `{2,3,4}` resolution collapse — documenting it is not even necessary

I was wrong to frame this as a cost, and the sweep says so unambiguously. I compared the two shapes across 60 configurations spanning cpus ∈ {1,2,4,8,16}, threads ∈ {cpus, 2·cpus}, load ∈ {0 … 8·cpus}:

- **SPLIT** (yours): `MIN=2, MAX=4, BASE_CALIBRATION=2`
- **FOLD** (what I suggested): `MIN=4, MAX=8, calibration inside the clamp`

**0 of 60 rows differ.** The two shapes produce identical effective budgets everywhere. That is not a coincidence — folding a factor of 2 into the clamp and doubling both bounds is an exact algebraic identity, so the "seven steps versus three steps" framing I used was measuring the scale variable, not the budget. The *budget* has exactly three reachable values, `{4x, 6x, 8x}`, under both shapes. The fold would have bought nothing.

And the headroom point I raised was also wrong. Under both shapes, #397's case (`raw = 4`) saturates the cap and receives 8x, and every worse machine also receives 8x:

| cpus=4, threads=8 | load 6 (#397) | load 10 | load 20 | load 40 |
|---|---|---|---|---|
| raw | 4 | 5 | 7 | 12 |
| SPLIT | 8x | 8x | 8x | 8x |
| FOLD | 8x | 8x | 8x | 8x |

Headroom above #397 is a function of the **8x ceiling alone**, which the major's own requirement fixes. Neither shape can widen it without reintroducing the regression the major was about. So the answer to your question is: not documenting is fine, because there is nothing to document — no behaviour was lost, and a note explaining a non-difference would mislead the next reader into thinking one exists. The `MAX_SCALE` doc's "2 x 4 is the 8 it always was" already says the thing that matters.

The one genuine observation underneath all this — that #397's measured flake sits *at* the ceiling rather than below it — was equally true before this PR (the old rule returned 8, the cap, on the same inputs). It is a property of the 8x ceiling, not of this change, and if anyone wants more room for machines worse than #397's that is a separate decision about `MAX_SCALE` with its own trade-off against how long a broken test takes to fail.

### Split versus fold — keep the split, and my "cleaner" call was wrong

Do not redo it. The shapes are behaviourally identical, so the only distinction is readability, and on that axis your reasoning wins outright: `scale` stays the number the panic message prints and the docs discuss, and the ceiling is pinned by two tests. The fold's only claimed advantage was resolution, which the sweep shows does not exist. Merging the shape you have is merging the shape I would now choose.

## Findings

None at this head. All four items from the previous review are resolved, the mutation test reproduces independently, and both open questions resolve in favour of the current shape.

## Test coverage

`the_worst_case_stretch_is_unchanged` now pins the constants *and* the output, covering both the pair-invariant and the property the doc promises. `the_397_flake_still_gets_its_eight_seconds` discriminates again. Against six mutations of the core rule, the suite catches all six; the true rule is the only survivor. Coverage is good.

The one optional addition, if you want a single self-sufficient assertion: `scale_from(6.0, 6.0, 4.0) == 3` sits strictly between both clamps and separates the true rule from every mutation on its own. Not needed — the suite already achieves this collectively.

## Nitpicks

- `src/test_budget.rs:76-78` — the rewrap leaves a short orphan line ("of test shells at once), where") mid-paragraph. Cosmetic only; rustfmt will not touch it either way.
- Consider noting in the companion's comment that it sits *on* `MIN_SCALE`, so its discrimination is one-sided and the strict assertion in `both_signals` is what covers under-production. Helps whoever next changes these constants understand which assertions are load-bearing in which direction.

## Conventions

Version 0.1.830 against main's 0.1.829 ✓, `Cargo.lock` in step ✓, release note present and corrected ✓, no `CLAUDE.md`/`CONTRIBUTING.md` changes needed ✓. CI 8/8 green and CLEAN at this head, per your report.

**Re-check the version in the same breath as the merge.** `origin` moved again during this review (`feat/release-box-wave` advanced to 334ae00), and per CLAUDE.md the release gate cannot catch a version collision itself — it compares against the merge base, so two branches can both pass and only the second conflicts. Run `git show origin/main:Cargo.toml` alongside the final `gh pr checks` and bump if main has moved.

## Scores

**CodeRabbit score — 5/5**
**Greptile score — 5/5**

No findings. The rule change is correct, #422's clauses are satisfied clause by clause, the ceiling regression is fixed and now pinned from both directions, and the re-swept 3060-cell grid confirms no flake reintroduction anywhere — max shrink 2.0x, confined to idle machines with oversubscribed threads, with zero shrink cases on 1-CPU, 2-CPU, or actual CI. The one correction I owe you is on my own prior review: the 106-char line was a nit rather than a CI blocker, because this repo's rustfmt does not reflow doc comments.

VERDICT: APPROVE

PR #441, opened 2026-08-31
fix(terminal): give command decorations a stable id (#440)

### LINKED ISSUE #440: terminal: command decorations carry no stable identity, so a positional index can name the wrong command
Split out of #354's review (PR #432), which needs this to make `{persist}` sound rather than merely guarded.

## The gap

`PtyTerminal::command_decorations()` returns a **sliding window**. `marks_snapshot` collects marks past the scrollback floor, and when it does, every index shifts down. A caller that remembers "the command at index N" has remembered nothing durable.

#432's capture needs exactly that: it arms by recording the pane's decoration count, and settles by taking the decoration that appears at that index. An eviction between arming and settling makes index `N` name a different, unrelated command.

## Why the current guard is not enough

#432 samples a high-water mark each tick and refuses the `{persist}` disk write when the length drops below it. That catches an eviction **at an instant it observes**, which removes the cases it sees and no others: an eviction and a new command landing between two ticks leave the length equal or larger and are invisible.

Any length-derived signal has that shape, because it samples a value that can move twice between observations. It is a narrowing, not a fix, and the code says so.

## What would fix it

A monotonic id assigned to each mark at creation and never renumbered by the collector. `marks_snapshot` already holds the records at the point where ids would be issued, so the natural shape is an `id` on the mark and on `CommandDecoration`, with callers holding the id rather than a position.

That turns "the command at index 3" into "command 41", which survives any amount of collection: if 41 is gone, the caller knows it is gone rather than silently reading 42.

## Blast radius

`command_decorations()` has other callers, so the change is additive: the id rides alongside and nothing is forced to use it. #432's `PendingCapture` is the first consumer, and it can drop `high_water`/`identity_lost` entirely once the id exists — the box and the disk write both become unconditional, because there is no longer a case where the wrong command can be read.

## Until then

`{persist}` in #432 remains gated rather than sound, which is the honest state: the box renders on the index because a wrong box is cheap and self-correcting, and the disk write is refused on any observed eviction because a stranger's output reaching the user's file is neither.

vitali87:
Closes #440.

A command decoration's position was its only identity, and positions are not durable. `marks_snapshot` GCs marks past the scrollback floor, the reader drops the oldest at `MARKS_MAX`, and a destructive wipe clears the list — so a caller that remembered "the command at index 3" was holding a slot, not a command. After an eviction that slot silently named a different one, with nothing to distinguish the two.

Each OSC 133 mark now takes a monotonic `u64` id at creation, carried onto the `CommandDecoration` it pairs into. "The command at index 3" becomes "command 41": if 41 is gone the caller learns that, rather than reading 42 and believing it.

**The id comes from the `CommandStart` mark**, because that is the one mark every finished command has exactly one of. The prompt marks before it are optional in the pairing, so keying on one of those would leave some decorations unidentified.

**What the id does and does not promise.** Review caught me overstating this, and the corrected version is the interesting part. The id, `exit` and the output span come from the `CommandStart`/`CommandEnd` pair and hold for as long as the decoration exists. But `line` and `input` come from the *prompt* marks, which sit on earlier grid rows and so cross the scrollback floor **first** — the GC retains per mark, not per command, so a command can outlive its own prompt. Reproduced: same command, same id, same span, `input` going from `Some((-10, 5))` to `None` as the floor rises.

I considered retaining per command group instead, and the reviewer argued convincingly against it: `decoration_at_screen` matches `d.line` against a live viewport row, so keeping a prompt mark below the floor would preserve something *false*, where the current fallback degrades to something true (the command's own first output row, which does still exist). The durable/best-effort split is the correct model, not a concession — so it is documented as the contract and pinned with a test.

**`marks_snapshot`'s 4-tuple became a named `MarkView`.** Adding the id would have made five positional fields, and `m.3` at a call site says nothing about what it holds — which is the same confusion between position and meaning this change is about.

**Additive**, as the issue notes: existing `command_decorations()` callers ignore the new field.

## For #432

This is the fix that issue was split out for. Once it lands, `high_water` and `identity_lost` can go and both the box and the disk write become unconditional, because there is no longer a case where the wrong command can be read. @vitali87's other session has said it will adopt this after it merges rather than carrying a half-migration.

One thing worth knowing when it does: `command_input_text` returns `""` both when the prompt scrolled away and when there is no shell integration at all. Those are currently indistinguishable, and after this change the first case is the one a re-run will actually meet.

## Gate

Three new tests, each mutation-tested rather than merely passing:
- `a_decoration_id_outlives_the_eviction_that_moves_its_index` — asserts the property as a **contrast** (the same eviction moves the index while leaving the id alone), because "id 7 is still 7" would pass against a collector that renumbered from zero. Making the id `out.len()`, which is precisely the bug, fails four of its assertions.
- `a_command_can_outlive_its_prompt_marks` — pins the limit above. Fails under every alternative fallback policy.
- `ids_are_unique_across_panes_and_survive_a_wipe` — a per-pane counter fails it whether or not the wipe resets it.

3935 tests pass; the two failures on this machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main` and are environmental — missing musl cross-targets and a socket-path length limit. fmt 0, clippy 0, doc-ownership gate clean.

Version 0.1.834 against main's 0.1.831 (main moved under this branch once already; #435 holds .832 and #439 .835). Merged main in rather than rebasing, per the repo convention; nothing deleted, main's release notes byte-identical.


vitali87:
## 🤖 Fallback review — Claude Opus 5 (round 2, final)

CodeRabbit is rate-limited on this repo, so this is the review. Round 1 found a major — I had documented the id as making the decoration "stable across eviction", which overstates it — and this is the re-verify after taking it.

The reviewer verified both mutation claims by porting the logic rather than trusting me, and argued me *out* of the more invasive fix: retaining marks per command group would preserve a `line` that is false, where the current fallback degrades to one that is true.

**APPROVE, 5/5 both lenses.**

## Walkthrough

Re-review at `42a83be` (merge of `origin/main` at `ededebe` plus fixes commit `380f522`). The fixes commit takes the previous review's major: `CommandDecoration::id`'s doc now states the durable/best-effort split explicitly, and a new test pins the exact fixture. The ordering rationale, the third eviction path, and the redundant `#[cfg(test)]` are all addressed, and two further tests close the cross-pane and post-wipe gaps.

| File | Summary |
|---|---|
| `src/widgets/terminal.rs` | `next_mark_id` doc explains why `Relaxed` is right; `StoredMark::id` doc names all three removal paths and the non-reset counter; `CommandDecoration::id` doc splits durable (id/exit/span) from best-effort (`line`/`input`); three new tests; redundant `#[cfg(test)]` removed. |
| `Cargo.toml` / `Cargo.lock` | 0.1.834 in both, above main's 0.1.831. |
| `src/release_notes/0.1.834.md` | Unchanged; `0.1.831.md` arrived via the merge. |

## Findings

None. All three findings from the previous review are resolved, verified as follows.

**On item 1 — documenting the limit rather than changing the GC: your call is the right one, and I'd argue it more strongly than you did.** Retaining per command group is not merely "more thorough with a wider blast radius" — it is a behaviour change that would be wrong for the existing consumer. `decoration_at_screen` (`src/widgets/terminal.rs:2166`) matches `d.line == line` against a viewport row. Retaining a prompt mark whose row is below the scrollback floor means `line` keeps naming a row that no longer exists, so the gutter dot would either vanish silently or match nothing, and `command_input_text` would extract from rows that have been reused. The current fallback degrades to something true (the command's own first output row, which does still exist); per-group retention would preserve something false. The GC's per-mark rule is not an oversight the id change should fix — it is the rule that keeps `line` referring to live grid. The durable/best-effort split you documented *is* the correct model, not a concession. I'd only add: when #432 adopts the id, `command_input_text` returning `""` should be treated as "the prompt scrolled away", not "no shell integration" — those are currently the same return value (`src/widgets/terminal.rs:2130`).

- **Item 2, `a_command_can_outlive_its_prompt_marks` (`src/widgets/terminal.rs:7891-7940`) — asserts the right things and is not vacuous.** Ported it and ran the fixture against four fallback policies. As written (`prompt.unwrap_or(output_start)`) all nine assertions pass. Change the fallback to `output_end` → `after[0].line` is -7; to a literal 0 → 0; to "no fallback, keep `None`" → `None`; each breaks `assert_eq!(after[0].line, after[0].output_start, "the row falls back to the command's own line")`. The staging assertions are load-bearing too: `survivors.len() == 2` pins that the floor took exactly the two prompt marks, so the test cannot silently degenerate into the clean-boundary case the neighbouring test already covers.
- **Item 3, the ordering doc (`src/widgets/terminal.rs:713-721`) — nothing subtly wrong.** Each claim holds: `fetch_add` is a read-modify-write on a single location, and all modifications to one atomic form a single total order regardless of ordering argument, so `Relaxed` genuinely cannot hand two callers the same value. The publication claim is also right and correctly identifies the actual synchronisation edge: both call sites (lines 1876 and 2955) store the id into a `StoredMark` pushed under `marks: Arc<Mutex<Vec<StoredMark>>>` (line 589), and `marks_snapshot` reads it under the same mutex (line 2085), so the unlock/lock pair supplies the happens-before. "No second variable whose visibility this atomic would have to fence" is the precise reason `SeqCst` would buy nothing — the counter is never used as a flag guarding other data.
- **Item 4, third eviction path (`src/widgets/terminal.rs:737-746`) — correct, including the subtle half.** The wipe at `src/widgets/terminal.rs:3521` calls `self.marks.lock().unwrap().clear()`, and `NEXT_MARK_ID` is `static` with no reset anywhere (only reads are the two `fetch_add` sites), so the "counter is never reset" claim is verified by absence *with* a positive control: `grep` for `NEXT_MARK_ID` returns exactly the declaration and the one `fetch_add`, and that same grep does find the store sites, so it could have found a reset.
- **Item 5, `ids_are_unique_across_panes_and_survive_a_wipe` (`src/widgets/terminal.rs:7942-7998`) — not vacuous; a per-pane counter really does fail it.** Modelled all three counter designs. With the process-wide counter, ids are `[3, 11, 7]` and the post-wipe id is 15 → passes. With a per-pane counter not reset by the wipe, pane A's first command and pane B's first command *both* get id 3, so `ids.dedup()` shrinks the vec and `"ids must not collide between panes"` fires. With a per-pane counter reset on wipe, that assertion fires *and* the post-wipe id is 3 against a high-water of 7, firing `"an id must not be reused after a wipe"`. The interleaving `run(&a); run(&b); run(&a)` is what makes this work — running both panes' first commands is exactly what collides. Helper signatures check out: `a` needs `mut` for `clear_screen_and_scrollback(&mut self)` (line 3505), `b` does not since `push_mark_for_test` and `command_decorations` take `&self`.
- **Item 6** — `#[cfg(test)]` removed from `views` (now `src/widgets/terminal.rs:6250`); no such attribute remains inside `mod tests`.

**Mutation claims, both verified by porting rather than trust:**
- `id = out.len()` still fails `a_decoration_id_outlives_the_eviction_that_moves_its_index` — four assertions break (survivor slot, index-0 identity, the `.expect()` panic, evicted-id reuse). Honest ids are `[3,7,11] → [7,11]`; mutated they are `[0,1,2] → [0,1]`.
- A per-pane counter fails `ids_are_unique_across_panes_and_survive_a_wipe`, as detailed above.

Worth noting: the mutation `id = out.len()` **survives** the new `a_command_can_outlive_its_prompt_marks` test (single decoration, so `out.len()` is 0 both before and after). That is fine — the first test is the one that kills it — but it means the new test is a limit-pin, not an identity-pin, which matches its stated purpose.

**Merge and version trio, re-checked against current `origin/main` (`ededebe`, 0.1.831):**
- `git diff origin/main --diff-filter=D --name-only` is empty here too — the merge deleted nothing. Diffstat vs main is 4 files, +330/-40, confined to `terminal.rs` plus the three metadata files.
- `Cargo.toml` and `Cargo.lock` both read 0.1.834, above main's 0.1.831.
- `src/release_notes/0.1.834.md` present with one `fix:` line; `0.1.831.md` came in via the merge and is untouched by this branch, satisfying "only the current version's file may change".
- Zero conflict markers in `terminal.rs`. #432's `high_water`/`identity_lost` survived the merge intact at `src/app/mod.rs:938,948` with their comments already pointing at #440 as the fix.
- The version-bump race still applies: main moved 830→831 under you once already. Re-read `git show origin/main:Cargo.toml` in the same breath as the final `gh pr checks`.

## Test coverage

No remaining gaps I would block on. One optional addition, now that the durable/best-effort split is the documented contract:

- **`command_input_text` on a prompt-evicted decoration.** The doc says a re-run "must handle its absence", but nothing asserts what a consumer actually sees. `command_input_text` returns `""` for both `input: None` (prompt scrolled away) and no-shell-integration (`src/widgets/terminal.rs:2130`), and `rerun_command` maps that to "No command text recorded at that prompt" (`src/app/mod.rs:27817`). A test driving a live pane past the floor and asserting the empty return would pin the behaviour #432 will meet, and would document that these two causes are currently indistinguishable.

## Nitpicks

- `src/widgets/terminal.rs:740` — that doc line is 95 columns against 69-78 for its neighbours (it is the only `///` line over 80 in the file besides a pre-existing one at 1090). The in-place sentence edit left "A caller holding" stranded past the margin; re-wrap the paragraph.
- `src/release_notes/0.1.834.md` — the note still says only "carries a stable id", which is now narrower than the code's own documented contract. Since the doc distinguishes durable identity from best-effort position, the note could say the id names the command rather than its position; minor, and the note is user-facing rather than contractual.

## Scores

**CodeRabbit: 5/5** — no findings. The one major from the previous round is resolved with the more defensible of the two available fixes, and the fix is pinned by a test I verified fails under every alternative fallback.

**Greptile: 5/5** — no findings in this lens. Version/lock/note trio consistent against current main, merge clean with nothing deleted, `#[cfg(test)]` hygiene fixed, and the new docs match the file's existing habit of recording why a decision is right rather than just what it does. Only nits remain.

VERDICT: APPROVE

PR #442, opened 2026-08-31
docs: a merge commit can have the right tree and the wrong parents

vitali87:
A sixth item for `Verifying the CHECKS actually ran`, and a shape neither session had seen before today.

## The failure

A merge committed before it is completed carries main's **changes** without main's **history**. Every check that looks at files agrees it worked:

- `git status` is clean
- the diff against main is exactly what you expect
- the suite passes

GitHub, which looks at the parent list rather than the tree, keeps computing a conflict against a merge that appears done — so the branch sits at `DIRTY` with no visible cause, and by item 2 of this same section it is silently getting **zero workflow runs** while it does.

## Why it belongs here

It is the fifth variant of the family this section already collects, and it fails the same way: **the thing that would have told you is absent rather than wrong.** A missing parent is not a wrong value that a diff can show; it is a fact about the commit that no file-level check consults.

It also compounds with the items above it. A branch in this state reads as `DIRTY` → no runs created → `gh pr checks` shows a short, all-green list from whatever status contexts exist → which is item 1.

## The check

```bash
git merge-base --is-ancestor origin/main HEAD && echo ok || echo "main is NOT merged"
git log -1 --format=%p            # a merge has two parents; one means it is not one
```

Placed immediately before the version re-read, because both answer the same question — *is this branch really current with main* — and neither is visible in a diff.

## Provenance

Found by the session working the other half of this repo, on its own branch, after inferring the cause twice and then measuring it: `git log -1 --format=%p` returned one parent.

I ran the check across my own branch's eight merge commits before writing this up; all carry two parents and `--is-ancestor` passes, so this is documented from someone else's incident rather than mine.

## Gate

Docs only — no Rust changed, so no version bump or release note (`CONTRIBUTING.md` is outside the release-hygiene job's `shipped` filter).


vitali87:
## Review — this documents my incident, and it documents it correctly

I hit this on #439, so I checked the write-up against what actually happened rather than against my memory of it. It matches, and both commands work.

**Verified with controls, since an absence-proof is worth nothing without one:**

```
git merge-base --is-ancestor origin/main <current>  → ok           (true positive)
git merge-base --is-ancestor origin/main <stale>    → not merged   (true negative)
git log -1 --format=%p <a real merge>               → 79f95ba 4415c83   (two)
git log -1 --format=%p <the broken commit>          → e74b1c6 ededebe
```

Note the last line: my broken commit is *now* two-parent, because I redid the merge properly on top of it. At the time it printed one. Worth saying because it makes the point sharper than the PR body does — the artifact heals itself once you fix it, so there is no lasting evidence in the tree that anything was ever wrong. Anyone reviewing the branch afterwards sees a normal history.

**One thing I would add, and it is the part that cost me time.** The failure does not present as "my merge did not work". It presents as a PR that reads `DIRTY` at a commit whose tree you have *just verified* is correct — I had re-read the version, confirmed no release note was deleted, checked main's notes were byte-identical, and run the suite, all green. So the natural next move is to re-resolve the conflict, which produces another single-parent commit and the same result. The trigger for running these commands should be "DIRTY at a tree I believe is right", not "I think my merge failed", because you will not think that.

Your framing of why it belongs in this collection is right and better than mine: it is the first entry where *every file-level check agrees*. The others leave something visibly odd. And the compounding you name is real — wrong parents → `DIRTY` → zero workflow runs → a short all-green `gh pr checks`, which is item 1 arriving through the back door. I saw exactly that: 7 checks passing at one point where a healthy head shows 8.

Docs-only, takes no version, and the placement next to the version-race entry is right since both answer "is this branch really current with main".

**Approve.** No findings.

PR #444, opened 2026-08-31
feat(explorer): dot the files an agent changed until you review them (#345)

vitali87:
Part of #345 — the Explorer decoration. The ledger and "mark reviewed" shipped in #421/#429, but the queue was only visible as a status line: you had to ask what was waiting rather than see it where you already look.

## The decoration

**Membership is exact**, unlike the git-ignored set right next to it, which covers a directory's descendants. A directory is not something an agent wrote, and not something "mark reviewed" can clear — so a dot on one would be a mark no gesture ever removes. That asymmetry is why the two decorations cannot share a predicate, and the test asserts a parent directory does *not* inherit it.

**The dot trails the name.** A leading marker shifts every name in the tree by a column as writes land, so the eye loses the alignment it reads the tree by. Trailing, it appears and clears without moving anything.

**Synced from every path that changes the ledger**, rather than on a schedule. A decoration that lags is worse than none here, because "no dot" is exactly what a reviewed file looks like — a stale one tells the user they have already looked at something they have not. A re-root and a folder removal both drop the queue belonging to the workspace being left.

The status chip now counts the Explorer's cached set instead of recomputing off the ledger. It sits in the per-frame render path and the count allocated a fresh set of cloned paths just to take its length; reading the cache also makes the chip and the dots the same number by construction rather than by two code paths agreeing.

## A real bug this uncovered

Review pointed out that my app-level test *substituted* for the fs-watch drain — it called the attribution helper and then hand-invoked the sync — so deleting the drain's own sync left every assertion green. Rewriting it to call `drain_fs_events` for real failed immediately, and not because of the sync:

**No agent write was ever attributed under a symlinked root.** The watcher reports a file's realpath while a root is stored as the user spelled it, so a workspace under `/var` arrives as `/private/var`, matches no root in `owning_root`'s `starts_with`, and is dropped. On macOS that is every `/var` workspace and every tempdir; on Linux any symlinked checkout. The queue stayed empty and nothing failed.

`path_under_roots` compares canonicalised and returns the path **in the root's spelling**, so the ledger's keys match the tree's rows. Two details review caught, both of which matter:

- **`canonicalize` stats the final component**, so it fails on a path that is gone — which is exactly what a deletion event carries. Resolving the whole path would have made the `forget_path` arm unreachable and stranded rows on deleted files with no later write able to clear them: *strictly worse* than the original bug, which dropped writes and deletions alike and so at least stayed self-consistent. `canonicalize_existing_ancestor` resolves the deepest ancestor that still exists and re-attaches the tail.
- **Longest prefix wins**, mirroring `owning_root`. Taking the first match would have made the ledger's key depend on the order the user added folders.

The direct check runs before the walk, and that ordering is load-bearing rather than merely quick: if an intermediate directory is itself a symlink pointing outside the workspace, resolving the whole path would land the result outside the root.

## Gate

Six tests, red-first, each mutation-tested rather than merely passing:

- **Render** — never painting the dot and always painting it *each* fail, so neither a missing decoration nor a blanket one can pass. It asserts the glyph reaches the buffer, because a predicate that answers correctly while nothing draws it looks the same to a user as a feature that was never built.
- **Wiring** — drives the real drain. Removing the drain's sync fails it.
- **Symlinked root** — reverting the resolution fails it; resolving the whole path rather than the ancestor fails the deletion assertion; climbing only one level fails the deleted-parent assertion.
- **Re-root and folder removal** — removing either clear fails its test.

3941 pass. The failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main` — missing musl cross-targets and a socket-path length limit — and the rest pass unloaded, so they are the suite's known load-sensitive shell-spawning flakes. fmt 0, clippy 0, doc-ownership gate clean.

Version 0.1.838 against main's 0.1.835; .836 and .837 are the peer session's #435 and #443.

**Not included, deliberately:** an unreviewed file inside a *collapsed* folder is invisible, since membership is exact. A rollup glyph on the collapsed ancestor is the answer, but it is derived state with its own design questions and belongs in its own change rather than riding in on this one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,7 +592,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.835"
+version = "0.1.838"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.835"
+version = "0.1.838"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/agent_lane.rs
+++ b/src/agent_lane.rs
@@ -237,6 +237,22 @@ impl AgentLedger {
         rows
     }
 
+    /// Every unreviewed file across every lane, deduplicated.
+    ///
+    /// One set rather than per-agent lists, because the consumer is the
+    /// Explorer's dot (#345) and a file two agents both wrote is one file
+    /// with one row. [`Self::unreviewed_files`] counts the same thing; this
+    /// hands back the paths so a caller can test membership per row, which
+    /// is what the Explorer's decoration needs.
+    pub fn unreviewed_paths(&self) -> std::collections::HashSet<PathBuf> {
+        self.lanes
+            .values()
+            .flat_map(|lane| lane.values())
+            .filter(|f| f.unreviewed())
+            .map(|f| f.path.clone())
+            .collect()
+    }
+
     /// Every agent that has a lane, in name order.
     pub fn agents(&self) -> Vec<&str> {
         self.lanes.keys().map(String::as_str).collect()
@@ -372,14 +388,15 @@ impl AgentLedger {
     /// Distinct FILES awaiting review across every lane. A shared row is one
     /// file to open, not two: a count that drops by two when the user
     /// reviews one file is telling them something untrue.
+    ///
+    /// Derived from [`Self::unreviewed_paths`] rather than counted
+    /// separately, so it cannot disagree with the set the Explorer draws.
+    /// It allocates that set to do it, which is why the status chip counts
+    /// the Explorer's own cached copy instead of calling this on every
+    /// frame — this is for callers that have no cache to hand.
+    #[cfg_attr(not(test), allow(dead_code))]
     pub fn unreviewed_files(&self) -> usize {
-        self.lanes
-            .values()
-            .flat_map(|lane| lane.values())
-            .filter(|f| f.unreviewed())
-            .map(|f| &f.path)
-            .collect::<std::collections::BTreeSet<_>>()
-            .len()
+        self.unreviewed_paths().len()
     }
 }
 
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -948,6 +948,35 @@ struct PendingCapture {
     identity_lost: bool,
 }
 
+/// Canonicalise `path` even when it no longer exists.
+///
+/// `std::fs::canonicalize` stats the final component, so it fails on a path
+/// that has been deleted — and a deletion event carries exactly such a path.
+/// Resolving the deepest ancestor that still exists and re-attaching the tail
+/// gives the same answer for a live path and a usable one for a dead path,
+/// which is what lets a removal clear the row it created (#345).
+///
+/// `None` when nothing on the way to the filesystem root resolves, and also
+/// for a path with no file name to strip — a bare root, or one ending in
+/// `..` — which cannot be climbed any further.
+fn canonicalize_existing_ancestor(path: &Path) -> Option<PathBuf> {
+    let mut tail: Vec<std::ffi::OsString> = Vec::new();
+    let mut probe = path.to_path_buf();
+    loop {
+        if let Ok(canon) = std::fs::canonicalize(&probe) {
+            let mut out = canon;
+            for part in tail.iter().rev() {
+                out.push(part);
+            }
+            return Some(out);
+        }
+        // `file_name` is None for a root or `..`; either way there is
+        // nothing further to climb.
+        tail.push(probe.file_name()?.to_os_string());
+        probe = probe.parent()?.to_path_buf();
+    }
+}
+
 /// State of a background self-update observed by a remote-launched croft.
 /// `Idle` is the steady state; `InProgress` paints an "Updating" hint in
 /// the status bar; `Ready` arms the re-exec into the freshly-shipped binary.
@@ -11363,8 +11392,12 @@ impl App {
         // Attribute the writes to whichever agents were working when they
         // landed (#345). Done after the reload so an agent's write to the
         // ACTIVE tab is hashed from the content the user is about to see.
-        if !drain.changed_files.is_empty() {
-            self.attribute_writes_to_working_agents(&drain.changed_files);
+        if !drain.changed_files.is_empty()
+            && self.attribute_writes_to_working_agents(&drain.changed_files)
+        {
+            // Only when the ledger actually moved: the Explorer's dots must
+            // match it, and a decoration that lags is worse than none.
+            self.sync_agent_lane_decorations();
         }
         // A rescan means the watcher dropped events, so the queue is a
         // LOWER BOUND from here until it is emptied (#345). Recorded in the
@@ -15132,7 +15165,12 @@ impl App {
         // The review queue rides the same chip (#345): a file an agent
         // changed while you were looking elsewhere is the thing you most
         // need to be told about, and it outlives the agent's own pane.
-        let unreviewed = self.agent_ledger.unreviewed_files();
+        // The set the Explorer is drawing, not a fresh count off the ledger:
+        // this is the per-frame render path, and `unreviewed_files` clones
+        // every path to build a set it then throws away. Reading the cache
+        // also makes the chip and the dots the SAME number by construction
+        // rather than by two code paths agreeing.
+        let unreviewed = self.tree.agent_touched.len();
         if seated > 0 || unreviewed > 0 {
             // With nothing seated the agent half would read "0 agents",
             // which is not what the chip is reporting any more: the queue
@@ -28333,6 +28371,10 @@ impl App {
                     .any(|t| t.agent().is_some_and(|a| a.name == *agent))
             {
                 let agent = agent.clone();
+                // No decoration sync needed here, and deliberately so: the
+                // guard above only lets this run when the lane has NOTHING
+                // unreviewed, so it removes no row the Explorer is drawing
+                // a dot for (#345).
                 self.agent_ledger.forget(&agent);
             }
             if let crate::agents::AgentEvent::Waiting { pane, agent } = ev {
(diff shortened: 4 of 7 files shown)


vitali87:
## 🤖 Fallback review — Claude Opus 5 (round 3, final)

CodeRabbit is rate-limited on this repo, so this is the review. Three rounds; the earlier two found four majors between them, including one where **my fix was worse than the bug it replaced** (resolving a deleted path made the deletion branch unreachable, stranding rows on files that no longer exist).

The reviewer differential-tested `canonicalize_existing_ancestor` against an independent implementation over 60,000 randomised filesystems, and pushed back correctly on one of my decisions while withdrawing its own earlier suggestion.

**APPROVE, 4/5 both lenses.** The two remaining minors — a loop-invariant hoist and a comment whose enumeration had drifted again — are fixed in the pushed head.

## Walkthrough

Third pass at `c463212`. Both majors from round two are properly fixed and I verified the fixes rather than taking them. `canonicalize_existing_ancestor` climbs to the deepest resolving ancestor and re-attaches the tail; I differential-tested it against an independent spec over 60,000 randomized filesystems with and without symlinks and found zero mismatches. Longest-prefix resolution now mirrors `owning_root`, `remove_workspace_folder` forgets its rows, and the vacuous assertion moved to where both sides are 1.

On the caching question you asked me to push back on: **you are right not to cache on `App`, and also wrong to leave it as-is** — those are different problems. The invalidation argument is sound and I withdraw the cache suggestion. But `root_canons()` is still called *inside* `for path in changed`, which is a plain loop-invariant hoist needing no invalidation at all. Measured, that is the difference between ~42 ms and ~21 ms of blocking syscalls on a 1000-file drain.

| File | Change |
|---|---|
| `src/app/mod.rs` | `canonicalize_existing_ancestor` free fn; `path_under_any_root` + `root_canons`; 4 sync call sites; chip reads the cache; re-root resets the ledger; `remove_workspace_folder` forgets its rows. |
| `src/agent_lane.rs` | `unreviewed_paths()`; `unreviewed_files()` re-derived, `#[cfg_attr(not(test), allow(dead_code))]`. |
| `src/widgets/file_tree.rs` | `agent_touched` + `AGENT_DOT` + `is_agent_touched` + trailing dot; `set_root` clears the set; 2 tests. Unchanged from round two. |
| `src/app/tests.rs` | 4 app tests (wiring via real drain, symlink + deletion, remove-folder, re-root). |
| `Cargo.toml`/`Cargo.lock`/`0.1.838.md` | 0.1.835 → 0.1.838. |

## Findings

**`src/app/mod.rs:28504` — [minor/performance] `root_canons()` is loop-invariant but called per changed file.**

This is the narrowed version of last round's major, and I want to be precise about what I am and am not asking for.

Your reasoning against an `App` cache is correct and I withdraw that suggestion. Four invalidation points in a feature whose failure mode is staleness is a bad trade, and you were right to refuse it.

But the call sits inside the caller's loop:

```rust
for path in changed {                          // mod.rs:28436
    let Some(under_root) = self.path_under_any_root(path) else { continue };
    //                     └── calls self.root_canons() → canonicalize per root
```

Hoisting it to a `let root_canons = self.root_canons();` above the loop needs no invalidation whatsoever — the value cannot go stale inside a single synchronous loop iteration. It is a pure code-motion refactor, not a cache.

Measured on this machine, `realpath` on a 7-component path under `/tmp` (symlinked on macOS, so it exercises the real path) costs ~21 µs. Per drain, with the direct check missing — which is the *normal* case on a symlinked root, i.e. exactly when this code runs:

| changed files | 1 root | 3 roots |
|---|---|---|
| 100 | 4.2 ms | 8.5 ms |
| 1000 | 42 ms | 85 ms |
| 5000 | 212 ms | 425 ms |

Hoisting removes the `files × roots` term, leaving `files × 1`, halving the 1-root case and cutting the 3-root case by 75%. `changed_files` is unbounded (`fs_watch.rs:53`, filled one entry per mutating event at `:207`; the only caps in that file are on kernel *watches*, not on the drained set), and the drain runs on the event loop at `event::poll(Duration::from_millis(8))`. A branch switch or a bulk agent rewrite lands thousands in one drain.

Fix — change the signature to take the canons and hoist the call:

```rust
let root_canons = self.root_canons();
for path in changed {
    let Some(under_root) = Self::path_under_roots(&self.roots, &root_canons, path) else { continue };
```

I have dropped this from major to minor because the correctness bug it used to accompany is gone, and because the cost only bites on symlinked roots during large drains. But it is a two-line change with no downside, so I would take it.

**`src/app/mod.rs:28536` — [nit/docs] "The three call sites" is now four.**

`remove_workspace_folder` gained a `sync_agent_lane_decorations()` at line 37369 this round, so the enumeration in the doc is one short. I grepped: sync fires at 11399 (drain), 28560 (mark-file), 28601 (mark-lane), 37369 (remove-folder).

This is the same drift I flagged in round one, recurring for the same reason — the comment enumerates and the enumeration is not checked by anything. Add the fourth, or soften to "every site that mutates the ledger's content" and drop the count.

## Test coverage

I verified all four claimed kills by porting rather than trusting:

- **Render** — re-ran the port at this HEAD (the hunk is byte-identical to round two). `never` → "unreviewed file took no dot"; `always` → three failures including the control. Both mutations die.
- **Ancestor walk** — reverting to `canonicalize(path).ok()?` leaves the write phase passing (the file exists) but makes the delete phase return `None`, so `forget_path` never runs and `agent_touched` keeps the row. The new `assert!(app.tree.agent_touched.is_empty(), "the row outlived the file it described")` fails. Kill confirmed.
- **`path_under_any_root`** — reverting to `if owning_root(path).is_none() { continue }` makes `starts_with(<tmp>/real/f.rs, <tmp>/link)` false, so attribution returns `false` and the first assert fires. Kill confirmed.
- **Drain sync** — deleting the sync at 11399 leaves `agent_touched` empty, the `while !…contains()` loop spins its full 4000 ms, and the assert fires. Kill confirmed.
- **Remove-folder loop** (not claimed, checked anyway) — deleting the loop leaves the row, failing both `is_empty()` and `unreviewed_files() == 0`. Kill confirmed.

On your specific attack list for `canonicalize_existing_ancestor`, all verified by port:

- **Every ancestor gone** — climbs to `/`, which resolves, and re-attaches the whole tail. Terminates.
- **Nothing resolves at all** — `file_name()` returns `None` at the root and the `?` exits with `None`. Terminates.
- **Trailing `..`** — `Path::file_name()` is `None` for a path ending in `..`, so the first `?` exits immediately. No infinite climb.
- **Bare root** — same; `file_name("/")` is `None`.
- **Loop/allocation bound** — each iteration pushes exactly one component and removes one from `probe`, so iterations ≤ component count and `tail` ≤ depth. No unbounded growth.
- **Tail re-attached across a symlinked intermediate directory** — this is the one that genuinely could have broken. If `/ws/src` is a symlink to `/elsewhere/src` and the file under it is deleted, the walk resolves `/ws/src` → `/elsewhere/src` and lands the tail outside the root. But the direct `owning_root` check runs first and hits on the spelled path, so the walk is never reached. Safe by ordering, though the ordering is load-bearing in a way nothing records — worth a sentence in the doc.
- **Key stability** — the property that actually makes `forget_path` work. I confirmed the write phase (full canonicalize) and the delete phase (ancestor walk) produce the *same* ledger key, including when the parent directory is deleted too. Without that the deletion would resolve but still miss the row.
- **Differential sweep** — 30,000 trials without symlinks and 30,000 with, against an independent reference implementation. Zero mismatches.

Remaining gaps, both minor:

**No test covers the deleted-parent case.** The symlink test deletes only the file, so the walk climbs exactly one level. `rm -rf` of a subdirectory climbs further, and that is the path where a tail-reattachment bug would hide. Add to the existing test: create `link/sub/g.rs`, attribute it, then `remove_dir_all(link.join("sub"))` and assert the row clears.

**`remove_workspace_folder` leaves *reviewed* rows behind.** The loop filters `unreviewed_paths()`, so a reviewed row under the removed folder survives in the lane. Nothing user-visible reads reviewed rows today — no dot, no chip count, and the test's `unreviewed_files() == 0` passes — so this is not a correctness bug for this feature. But it is a small leak and the deferred sidebar's `lane(agent)` would list it. A ledger method that drops every row under a prefix regardless of review state would be the cleaner shape.

## Nitpicks

- `canonicalize_existing_ancestor`'s doc says `None` only when "nothing on the way to the filesystem root resolves". That is right but incomplete — it also returns `None` for a path ending in `..` or a bare root, via the `file_name()?`. The inline comment at the `tail.push` covers it; the doc-level claim slightly oversells.
- `path_under_any_root`'s correctness depends on the direct `owning_root` check running *before* the walk (it is what protects the symlinked-intermediate-directory case above). The doc explains the ordering as an optimisation — "the common case, costing nothing" — but not as a correctness guard. Worth one sentence, since a future reader optimising the function might reorder it.
- `#[cfg(unix)]` on the test rather than a body `return`: correct, resolves the round-two nit.
- The chip/dot equality assertion now sits where both sides are 1. Verified non-vacuous.
- Chip and dots read the same field (`mod.rs:15172` and `file_tree.rs:1536`), so they cannot disagree on content by construction. I re-enumerated all seven ledger writers against the four sync sites plus the guarded `forget` and the re-root's paired reset; the accounting is complete.
- `forget_path` while iterating is sound: `unreviewed_paths()` returns an owned `HashSet`, so `.into_iter()` consumes a copy and mutating the ledger inside the loop cannot invalidate it.
- Version trio re-checked against current `origin/main`: main is still `0.1.835`; `Cargo.toml`, `Cargo.lock`, and `src/release_notes/0.1.838.md` all agree on `0.1.838`. Note format matches CONTRIBUTING.md:176-181. Peers holding .836/.837 means .838 is above the merge base and the gate passes — re-read main's version in the same breath as the final `gh pr checks`, since the gate cannot catch a collision itself.
- Collapsed folders: still agreed, still out of scope for this diff, still worth an issue.

## Scores

**CodeRabbit score: 4/5** — no majors remain; one minor performance hoist and one stale comment count.

**Greptile score: 4/5** — the deletion fix is correct under adversarial sweep and the mutation kills all hold; what is left is a loop-invariant call that wants moving and a doc enumeration that drifted again.

VERDICT: APPROVE

Issue #364, opened 2026-08-29
ssh in a pane: offer to re-root the workspace onto the host you just connected to


vitali87:
## Motivation

`croft remote <host>` is the ceremony-free way to work on a box, but the way people actually reach a box is by typing `ssh host` in a shell. croft already tracks the pane's foreground process and cwd reports; when a pane becomes an SSH session it can offer, in one keystroke, to swap the Explorer onto that host through the remote machinery that already exists.

## Current state

- Foreground process sampling and OSC 7 cwd reports (with the busy-cd re-root logic) in the terminal layer.
- `src/remote.rs`, `src/remote_connect.rs`: the remote workspace layer, host provisioning, and the Remote sidebar.

## Proposed design

1. When a pane's foreground process is `ssh` (or `mosh`, `et`) and the destination resolves to a `~/.ssh/config` host, show a status-bar prompt: `Connected to db-1 · Open workspace here? (Cmd+K O)`.
2. Accepting runs the remote-connect flow for that host with the remote shell's reported cwd as the workspace path (the croft shim installs on first connect exactly as `croft remote` does); the originating pane stays as it is.
3. A per-host `remote.auto_offer = off` and a global toggle; the prompt never appears for hosts croft cannot provision (e.g. a jump host that refuses the install, learned and remembered).
4. Reverse: closing the ssh process while the remote root is open offers to drop the root.

## Acceptance criteria

- [ ] Typing `ssh box` in a pane produces the offer within a sample interval; accepting opens the remote Explorer at the remote cwd.
- [ ] Hosts marked `auto_offer = off` never prompt.
- [ ] A host without croft installs it through the existing provisioning path and the OUTPUT Remote channel logs it.

## Related

- Tracking epic: #383

vitali87:
## Groundwork in #445 — the detection, not the prompt

#445 adds the pieces this needs and a manual palette command, but **deliberately does not close this issue**. Against the acceptance criteria, plainly:

- ❌ **"Typing `ssh box` produces the offer within a sample interval"** — not met. There is no sampling and no prompt; the user reaches for a palette command.
- ⚠️ **"accepting opens the remote Explorer at the remote cwd"** — the Explorer opens; the cwd does not carry across. That needs the shell's OSC 7 report from the far side, and croft only trusts those against a local cwd it can verify.
- ❌ **"Hosts marked `auto_offer = off` never prompt"** — vacuously true; no prompt exists and no config key was added.
[...]

PR #445, opened 2026-08-31
feat(remote): detect a pane's ssh host and open a workspace on it (#364)

vitali87:
Groundwork for #364 — **deliberately not `Closes`**, see the scope note at the bottom.

`croft remote <host>` is the ceremony-free way onto a box, but the way people actually get there is typing `ssh host` in a shell. This adds the detection and a palette command, **Remote: Open Workspace on This Pane's Host**, that hands the resolved host to the remote flow that already exists.

## Argv, not scrollback

What the user typed may have scrolled away, been edited, or never been typed at all — a shell function, a `Host` alias ssh resolved itself. Argv is what is actually running.

## The flag grammar is the whole difficulty

Without knowing which flags take a **separate** argument, `ssh -p 2222 box` yields `2222`, because that is the first word not starting with `-`. The list is derived from ssh(1)'s synopsis, and review derived it independently two more ways (the bracketed `[-X arg]` groups, and the option-body indentation) — all three agree on the same 22 characters, with no boolean flag wrongly included and none missing.

I got this wrong first time: **`-P tag` was absent**, so `ssh -P prod box` offered `prod`. The test that now guards it exercises *every* arg-taking flag with a decoy value, so it tests the bug class rather than the instance — removing any one of them fails it.

## `ssh` only, and that is a correctness limit

`mosh` and `et` reach a box the same way and were in the program list. They take **long options** and ssh does not, and ssh's grammar mis-reads them: `--jumphost` scans as `-jumphost`, the loop finds an arg-taking letter mid-word, the attached-value branch consumes nothing, and the next word becomes the host. So `et --jumphost jump box` offered to re-root onto **`jump`** — a host that really is in the config, so it fails *open* rather than closed. The issue's own text calls out a jump host as a case to avoid.

Parsing them properly needs a per-program table of which long options take a separate word. Until that exists, refusing is the honest answer, and a test pins the refusal so re-adding them without the table fails rather than silently regressing.

## Host resolution

Alias or `HostName`, case-insensitively, ignoring any `user@` — the user is how you log in, not which machine it is. **Alias matching is a full pass before the HostName fallback**: testing both in one `find` let an earlier entry's HostName beat a later entry's own alias, so with `Host jump / HostName db-1` above `Host db-1`, typing `ssh db-1` offered `jump`.

A host with no config entry gets a message naming it rather than an offer, because the remote flow this hands to is keyed on a config entry.

## Known limit: pipelines

`foreground_pid()` is the process group leader, which for `ssh box` typed at a prompt is the ssh. For a pipeline it is not — measured on zsh, the leader resolves to the last stage — so `ssh box | tee log` reports "not an SSH session" rather than guessing. Refusing is the right failure: the alternative is walking the process tree for an ssh somewhere under the pane and offering to re-root onto a host the user merely piped output from.

## Scope — what this does and does not do

The issue asks for an **automatic status-bar prompt** on detection. This ships a palette command instead. Against the acceptance criteria, plainly:

- ❌ *"Typing `ssh box` produces the offer within a sample interval"* — not met. No sampling, no prompt.
- ⚠️ *"accepting opens the remote Explorer at the remote cwd"* — Explorer opens; the cwd does not carry across, which needs the shell's OSC 7 report from the far side.
- ❌ *"Hosts marked `auto_offer = off` never prompt"* — vacuous; no prompt, no config key.
- ⚠️ *"A host without croft installs it through the existing provisioning path"* — inherited from `request_remote_launch`, not exercised here.

The prompt needs a per-host opt-out, a memory of hosts that refused provisioning, and a policy on how often to nag. None of those should be invented alongside the detection, and shipping detection without them would be worse than shipping it as a command. So #364 stays open and this is the layer it can be built on.

## Gate

Six tests, all mutation-tested: dropping any arg-taking flag, accepting any program name, re-adding mosh/et, not stripping `user@`, and single-pass host resolution each fail a different assertion. 3949 pass; the two failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main` — missing musl cross-targets and a socket-path length limit. fmt 0, clippy 0, doc-ownership gate clean.

Version 0.1.841 against main's 0.1.838; .839 and .840 are the peer session's #443 and #435.


vitali87:

CodeRabbit is rate-limited on this repo, so this is the review. Two rounds.

**It found the `mosh`/`et` major and I could not have talked my way out of it** — `et --jumphost jump box` offered to re-root onto `jump`, a host that really is in the config, so it failed *open*. It also derived the ssh flag table two ways independently of mine, with a positive control in the same run, and confirmed it now matches exactly.

One note on reading this: the blocker it reports (alias precedence shipping as prose without the code) was **already fixed** when it read the tree — it reviewed the pre-fix head. Both that and the `-P` omission are in the pushed branch, each with a mutation-tested pin. Everything else stands and is applied.

## Walkthrough

Re-reviewed against the new head `ef6807f` (worktree clean). Adds the ssh-pane re-root feature (#364): `ssh_destination` parses a pane's foreground argv, `resolve_offer_host` matches it against `~/.ssh/config`, and a palette command hands the host to the existing remote-connect flow. Both fixes since my last pass check out — I re-derived the flag table from `ssh(1)` two independent ways and it is now exactly right, and `--` is handled correctly.

One new blocker: the alias-vs-HostName precedence fix was written as a **comment and a test but never applied to the code**, so the branch ships a failing test. The mosh/et grammar gap is confirmed real with a specific silently-wrong invocation.

| File | Change |
|---|---|
| `src/remote.rs` | `SSH_PROGRAMS`, `SSH_FLAGS_WITH_ARG` (22), `ssh_destination` (+`--`), `resolve_offer_host`, `ssh_reroot_offer` + tests |
| `src/app/mod.rs` | `open_workspace_on_ssh_host` + dispatch arm |
| `src/widgets/terminal.rs` | `process_cmdline(pid)` |
| `src/widgets/command_palette.rs` | `OpenWorkspaceOnSshHost` wiring |
| `Cargo.toml` / `Cargo.lock` | 0.1.838 → 0.1.839 |
| `src/release_notes/0.1.839.md`, `docs/ARCHITECTURE.md` | new note, extended entry |

## Findings

**`src/remote.rs:159-172` — [blocker/correctness] The alias-precedence fix exists in the comment and the test but NOT in the code. The test fails.**

Lines 163-167 state "Alias first, across ALL entries, before falling back to HostName", and lines 3051-3061 assert it. But lines 168-172 are still a single `find` with `||` inside, which is precisely the shape the comment says was wrong:

```rust
targets.iter().find(|t| {
    t.alias.eq_ignore_ascii_case(host)
        || t.host_name.as_deref().is_some_and(|h| h.eq_ignore_ascii_case(host))
})
```

Ported exactly as written and run against the test's own fixture (`Host jump / HostName db-1` above `Host db-1 / HostName 10.0.0.4`):

```
resolve_offer_host("db-1") = jump
test at remote.rs:3056 asserts = db-1
TEST FAILS
```

The second assertion (`10.0.0.4` → `db-1`) passes, so only one of the two fails. Typing `ssh db-1` offers to re-root onto `jump`. I could not run the suite per the constraints, so please confirm — but the port is a direct transcription and the fixture is the test's own.

Fix — the two-pass form the comment already describes:

```rust
targets.iter().find(|t| t.alias.eq_ignore_ascii_case(host))
    .or_else(|| targets.iter().find(|t| t.host_name.as_deref()
        .is_some_and(|h| h.eq_ignore_ascii_case(host))))
```

**`src/remote.rs:72,96-131` — [major/correctness] `et --jumphost jump box` silently offers to re-root onto the JUMP HOST. ssh's grammar does not fit `mosh` or `et`.**

You asked which specific invocation breaks. The worst one:

```
et --jumphost jump box   ->  dest='jump'   -> OFFERS: jump
mosh --predict always box ->  dest='always' -> OFFERS: always   (if "always" is an alias)
mosh --ssh "ssh -p 2222" box -> dest='ssh -p 2222'
mosh --port 60000 box    ->  dest='60000'  -> no offer (fails closed)
et --port 2022 box       ->  dest='2022'   -> no offer (fails closed)
et -t 8080:8080 box      ->  dest='8080:8080'   <- SHORT flag, not just long
et -r 8080:8080 box      ->  dest='8080:8080'   <- SHORT flag
```

The mechanism, traced through the code: `--jumphost` hits `strip_prefix('-')` at line 111 leaving `-jumphost`; the `char_indices` loop at line 122 skips `'-'`, reaches `'J'`… actually `'j'`, not in the table, continues to `'o'` at index 4 — which IS arg-taking — but `4+1 != 9`, so the attached-value branch is taken and **nothing is consumed**. Next word wins.

So the doc comment's promise at lines 78-80 — "a flag added to ssh in future degrades to 'skipped' rather than to a wrong host" — is false for any long option: an arg-taking letter appearing mid-word derails it. `ssh` proper has no long options and escapes entirely; `mosh` and `et` do not. `et -t` and `-r` break on short flags too, since et's short set is simply different from ssh's.

Numeric values fail closed (no alias is `60000`), which is why this is major and not a blocker — but `--jumphost jump` names a host that IS in the config, and the issue's own text calls out "a jump host that refuses the install" as a case to avoid. Not defensible as an approximation.

Fix — smallest correct change is to narrow the constant and file the rest:

```rust
const SSH_PROGRAMS: [&str; 1] = ["ssh"];
```

Both wrappers take `[user@]host` as the last positional, so a per-program table is the alternative (mosh: `--ssh`, `--port`/`-p`, `--predict`, `--bind-server`, `--server`, `--client`, `--local`/`--remote`; et: `--port`/`-p`, `--command`/`-c`, `--tunnel`/`-t`, `--reversetunnel`/`-r`, `--jumphost`, `--jport`). Dropping mosh/et is smaller and honest; the palette command still works for them once someone writes those tables.

**`src/remote.rs:81-84` — [verified correct, no action] The flag table is now exactly right.**

You asked for an independent read rather than trusting yours. I derived the arg-taking set from `ssh(1)` two ways that do not share a code path — the bracketed `[-X arg]` groups in the SYNOPSIS, and the option-body indentation (`-X<2 spaces>arg` = takes one, `-X<6 spaces>Text` = boolean):

```
synopsis [-X arg] groups: BDEFIJLOPQRSWbceilmopw (22)
option-body arg-taking  : BDEFIJLOPQRSWbceilmopw (22)
CODE                    : BDEFIJLOPQRSWbceilmopw (22)

missing from code: []      extra in code: []
IN CODE BUT BOOLEAN (would swallow the host): []
```

The two derivations agree with each other and with the code, the classes are disjoint, and they cover 45 flags. Positive control, in the same run — dropping `'P'` from the code re-reports `missing: ['P']`, and adding boolean `'q'` re-reports `boolean-in-code: ['q']` — so the check can fail and does not. Direct sweep of the parser confirms: all 22 arg-taking flags with a decoy value yield `box`, and all 23 booleans yield `box`. **No flag you have is boolean, and none that takes an argument is missing.**

**`src/remote.rs:103-109` — [verified correct, no action] `--` handling matches real ssh.**

```
ssh -- box          -> 'box'
ssh -- -4           -> '-4'      (real ssh connects to host "-4")
ssh -p 22 -- -weird -> '-weird'
ssh --              -> None
ssh -- ''           -> None
ssh box -- x        -> 'box'     (after the destination, `--` is remote-command text)
```

All correct, including the empty-string and after-destination cases.

**`src/widgets/terminal.rs:1110-1125` — [verified acceptable, no action] `process_cmdline` cannot block the render loop.**

You had not measured this. `refresh_processes(ProcessesToUpdate::Some(&[p]), true)` refreshes exactly one pid. The decisive comparison is that `Cmd::AttachPythonProcess` → `attachable_python_targets()` (`src/dap/discovery.rs:41-42`) already runs **`ProcessesToUpdate::All`** synchronously on the same dispatch path, and has shipped. Measured the kernel-path ratio as a proxy (`ps` on one pid vs all):

```
single-pid argv read : 5.06 ms/call   ratio 4.2x
ALL-process enum     : 21.45 ms/call
```

Both numbers include ~4-5 ms of `ps` process spawn that `sysinfo` does not pay in-process, so the real single-pid refresh is well under a millisecond. Behind a deliberate keystroke, and strictly cheaper than an existing on-loop command. Fine as written.

**`src/app/mod.rs:23584-23593` — [agree, no action] Refusing on a pipeline is the right call.**

You asked whether refusing is wrong. It is not, and I could not find a case you have mis-modelled. `foreground_pid()` → `process_group_leader()` (terminal.rs:2264-2266); for `ssh box` at a prompt that is the ssh, and your zsh measurement (leader = last stage) matches POSIX job control, where the shell sets the pgid to the first process but the leader you observe depends on which stage the terminal's foreground pgrp names. The alternative — walking the tree for any ssh under the pane — would offer to re-root onto a host the user merely piped from, which is a wrong answer rather than no answer. The docstring states the limit plainly. One consequence worth noting only because it is invisible: the user sees "This pane is not an SSH session" while plainly looking at an ssh session, which reads as a bug. A more precise message for the "fg process exists but isn't ssh" branch would help.

## Test coverage

The new sweep test (`remote.rs:33-43`) is the right shape — it tests the bug *class*, so removing any flag fails it. Gaps:

1. **The precedence test at `remote.rs:3051-3061` is red**, per the blocker. It is not a coverage gap but a failing assertion; listed here because it is the test that should gate the fix.
2. **No mosh/et long-option case.** `mosh` and `et` appear only in happy-path one-liners. The test that would have caught the finding:
   ```rust
   // et and mosh have their own grammars; ssh's cannot parse them.
   assert_eq!(d(&["et", "--jumphost", "jump", "box"]).as_deref(), Some("box"));
   assert_eq!(d(&["et", "-t", "8080:8080", "box"]).as_deref(), Some("box"));
   assert_eq!(d(&["mosh", "--predict", "always", "box"]).as_deref(), Some("box"));
   ```
   All three fail today. If you narrow `SSH_PROGRAMS` to `["ssh"]`, replace these with `assert_eq!(d(&["mosh", "box"]), None)` so the narrowing is pinned rather than silent.
3. **`open_workspace_on_ssh_host` has no test** — the only references are its definition and dispatch arm. Its five status branches are untested. Extracting the message choice into a pure `fn ssh_reroot_status(argv, targets) -> Result<&RemoteTarget, String>` would make all of them assertable without a live pane.

## Nitpicks

- `src/remote.rs:78-80` — the "degrades to skipped rather than to a wrong host" claim is true only for short flags. Worth correcting even after the mosh/et fix, since it is the sentence a future reader will trust instead of re-deriving.
- `src/remote.rs:100` — `.unwrap_or(cmd[0])` is dead; `rsplit` always yields ≥1 item.
- `src/widgets/command_palette.rs:727` — no keybind, while the issue proposed `Cmd+K O`. Consistent with other palette-only commands; noting only that the issue named a chord.

## Scope call (item 5) — plain answer

Against the issue's acceptance criteria:

- ❌ *"Typing `ssh box` in a pane produces the offer within a sample interval"* — **not met.** No sampling, no offer. This is the criterion the issue is chiefly about.
- ⚠️ *"accepting opens the remote Explorer at the remote cwd"* — **half met.** Explorer opens; cwd does not carry (`request_remote_launch(alias, None)`, mod.rs:23617, with an honest comment).
- ❌ *"Hosts marked `auto_offer = off` never prompt"* — vacuous; no prompt, no config key.
- ⚠️ *"A host without croft installs it through the existing provisioning path and the OUTPUT Remote channel logs it"* — inherited from `request_remote_launch`, so likely met, but nothing here exercises it.

**0 of 3 met, one inherited.** Your reasoning for deferring is sound — a nagging policy, per-host opt-out and provisioning-failure memory are genuinely separate decisions, and shipping detection without them would be worse. Since you said you would rather ship a partial feature that says so: make the framing match. Title the PR as adding the detection primitives plus a manual command, **do not use a closing keyword on #364**, and leave a comment on the issue listing the four criteria above with their status. Per repo convention (MEMORY: "let the PR's closing keyword decide done-vs-partial"), `Closes #364` here would mark unshipped work as shipped.

## Repo conventions

All satisfied. Re-checked `origin/main` in this same session: it is at `0.1.838`, branch and `Cargo.lock:595` both at `0.1.839`, so the bump is valid **as of now** — re-read it once more in the same breath as the final `gh pr checks`, per CLAUDE.md's version-race note. `src/release_notes/0.1.839.md` has one `feature:`-prefixed line matching CONTRIBUTING.md:196-207. `docs/ARCHITECTURE.md` entry extended, inserted before the existing "Launch dispatch" prose rather than between an item and its doc block.

Your note on the two pre-existing failures checks out and I am not attributing them: `the_pinned_toolchain_has_every_cross_target` (remote.rs) and `a_client_arriving_mid_swap_is_told_to_reconnect_not_seated` (session_host.rs:3904) both exist on unmodified `origin/main`, and neither file's test region is touched by this diff.

## Scores

**CodeRabbit score: 1/5** — a blocker: `resolve_offer_host` still ships the single-`find` shape its own comment and test say is wrong, so the branch carries a failing test.

**Greptile score: 1/5** — the parser core is now genuinely well-verified (the flag table is exactly right on two independent derivations, `--` is correct, and the sweep test targets the bug class), but a fix that landed as prose and an assertion without the code change, plus a wrapper grammar that silently offers a jump host, both need to land before merge.

VERDICT: REQUEST_CHANGES

Issue #349, opened 2026-08-29
Provenance blame: colour the gutter by who typed each line (you, the navigator, an agent pane)


vitali87:
## Motivation

Git blame says *which commit*; nobody shows *which seat*. croft already attributes carets and edits to participants for the collab and pair features, and the agent-lane ledger attributes disk writes to agent panes. Painting that as a gutter overlay answers "did I write this or did the model?" at a glance, and it is a screenshot nobody else can produce.

## Current state

- Named carets and per-seat edits flow through `src/collab.rs` / `src/pair_host.rs` (owner, `--solo` peers, MCP guest, navigator stream).
- Agent-lane attribution (sibling issue) covers writes that land through the file system from an agent pane.
- Inline blame and the git gutter are in the editor widget.

## Proposed design

1. A per-buffer provenance map: line → seat id (`me`, `navigator`, `agent:<pane>`, `peer:<name>`, `unknown`), maintained through the same edit pipeline that moves carets, and rebased through the local-history snapshots for disk writes.
2. **Editor: Toggle Provenance** paints a thin coloured bar in the gutter (the git gutter's style, distinct hue per seat) and adds the seat name to the inline-blame annotation (`you · 2m` / `claude (pane 2) · 5m`).
3. Persist the map next to local history so it survives restarts; lines with no record show as `unknown`, never guessed.
4. A Source Control diff view option "group by seat" so a commit can be reviewed as "the agent's part" and "my part".

## Acceptance criteria

- [ ] Type a line, let the navigator stream another, have a fake agent write a third: three distinct colours, correct names on hover.
- [ ] Provenance survives save, reopen, and an undo/redo cycle.
- [ ] Unknown lines are never attributed.

## Related

- Tracking epic: #379
- Depends on: #345

vitali87:
## The map has shipped in #446 — the surfaces have not

Recording the split so the next session does not re-derive it.

**Shipped:** `src/provenance.rs` — a `Seat` enum (you / navigator / agent-by-pane / collab peer / **generated**) and a per-buffer line map, wired through `Editor::insert_str_as`, with `insert_str` meaning `Seat::Me`.

**Still open, which is what this issue is now for:**

- **The gutter overlay and the toggle command.** `Seat::label()` exists and nothing consumes it.
- **The inline blame annotation** carrying the seat name (`you · 2m` / `agent (pane 2) · 5m`).
[...]

PR #446, opened 2026-08-31
feat(editor): track which seat wrote each line (#349)

vitali87:
Part of #349 — the data structure and its wiring. **Deliberately not `Closes`:** only one of the issue's three acceptance criteria is met, and the scope note at the bottom says which.

Git blame answers "which commit". A commit records the author of the **save**, not of the keystrokes, so it cannot answer the question this exists for: did I write this line, or did the model?

## The invariant

**A line croft did not watch being written is Unknown, never guessed.**

An overlay that is right most of the time gets read as fact, and the one line it attributes wrongly is exactly the line someone is arguing about later. So there is no path that infers a seat from a neighbouring line, and `splice` drops the attribution of every line an edit replaced rather than carrying it onto the replacement. The map is line → `Option<Seat>` rather than one that omits unknown lines, so "no record" cannot read as "not attributed to anyone" in one place and "index out of range" in another.

## Three bugs review found, each the invariant failing

- **A multi-line selection replacement** deleted lines without telling the map, so every seat below kept its old index while the buffer shifted up. A surviving line painted as whoever wrote a line the selection had just destroyed. Reachable by ordinary select-and-type.
- **Inserting into an empty buffer** attributed two lines where one exists: `insert_char_raw` pushes the first line itself, and that push counted as inserted.
- **Undo, reload and encoding re-decode** each swap the whole buffer while the map survived, describing text that is gone. All three now clear it.

That last one is worth stating plainly: #349 lists "provenance survives an undo/redo cycle" as a criterion, and clearing does not meet it. But **unknown beats wrong** — a stale map would falsify the criterion silently while looking correct, where an empty one is honestly not-yet-implemented. Restoring it properly means putting the map in `Snapshot`, which is the later layer.

## Two attribution decisions

**An accepted LSP completion and an expanded snippet are `Seat::Generated`, not `Me`.** The user picked them from a list, which is a different act from writing them. Crediting generated text to a person is the same class of wrong answer as inheriting from a neighbour, and it is the shape most likely to be argued about.

**Splitting a line gives both halves to the splitting seat.** This looks wrong on first read — the tail is someone else's text — until you notice the line they wrote no longer exists, so neither half is a line they wrote. Documented with a test rather than left to surprise the next reader.

## On the tests

The selection test uses **four distinct seats deliberately**, and that is the part I would point a reviewer at. An earlier version gave every line the same seat, and it **passed against the very bug it was written to catch** — with one seat everywhere, the wrong seat and the right one are the same value, so only the row count gave it away. It now fails on the seat assertion, which is the property it is actually about.

Every test is mutation-tested, and the mutations are the tempting wrong answers rather than arbitrary breakages: carrying a removed line's seat onto its replacement, attributing the whole buffer to whoever edited last, dropping the selection splice, keeping the map through undo. All fail.

`truncate` runs at the end of every insert so "no key is ever past the end" holds structurally rather than by an argument per path, and the splice arithmetic saturates on both sides — a release build would wrap rather than panic, and a wrapped index attributes a line at ~2^64 instead of crashing.

## Scope

Against #349's acceptance criteria:

- ❌ *"three distinct colours, correct names on hover"* — not met. No gutter painting, no toggle, no inline-blame integration. `Seat::label()` exists; nothing consumes it yet.
- ❌ *"Provenance survives save, reopen, and an undo/redo cycle"* — not met, and honestly so: undo clears rather than corrupts, with a comment naming `Snapshot` as the fix.
- ✅ *"Unknown lines are never attributed"* — met. All three whole-buffer swaps clear the map, collab peer edits bypass this path entirely (they route through `apply_span_edits`, so a peer's text stays unknown rather than becoming `Me`), and no key survives past the end of the buffer.

## Gate

3959 pass. The two failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main` — missing musl cross-targets and a socket-path length limit. fmt 0, clippy 0, doc-ownership gate clean.

Version 0.1.844 against main's 0.1.842. I originally took .842 and main moved onto it mid-review, which is the race CONTRIBUTING warns about; re-read at merge time.


vitali87:

CodeRabbit is rate-limited on this repo, so this is the review. Two rounds; round one found **two blockers** and round two found the thing I would least have caught myself.

**The finding worth reading is under Test coverage:** my selection test passed against the very bug it was written to catch, because all four lines carried the same seat — so a seat landing on the wrong line was invisible and only the row count failed. Fixed with four distinct seats; it now fails on the seat assertion, verified by re-running the mutation.

Also worth noting the reviewer validated its own Python port by replaying my 16 committed Rust assertions through it *before* drawing any conclusion from it — after I warned that my own model had got the empty-buffer case backwards.

Everything below is applied at the pushed head, including the version bump (main took .842 while this was in review).

## Walkthrough

Re-reviewed against the new head `3200877` plus the uncommitted working tree (`src/provenance.rs`, `src/widgets/editor.rs`, `src/app/mod.rs` all modified). Since my first pass the empty-buffer over-claim, the selection-replacement bug, the three whole-buffer swap paths, the `usize` overflow, the missing `truncate` primitive, and the `Seat::Me` mis-attribution of generated text have all been fixed. I re-verified each against the Rust source rather than a model of it, and validated my Python port by replaying the repo's own 16 committed Rust assertions through it before drawing any conclusion from it. What remains is one CI-blocking formatting issue and one merge-time version collision.

|---|---|---|
| `src/provenance.rs` | `Seat::Generated`, `truncate`, saturating arithmetic in `splice`, 2 new tests | Correct. `splice` re-swept, still clean. |
| `src/widgets/editor.rs` | selection splice, empty-buffer clamp, provenance cleared on 3 swap paths, snippet → `Generated`, 2 new tests | All verified correct across 660 shapes. |
| `src/app/mod.rs` | completion accept → `Seat::Generated` | Correct. |
| `Cargo.toml` / `Cargo.lock` | 0.1.842 | **Now collides with `origin/main`.** |
| `docs/ARCHITECTURE.md`, `src/main.rs`, release note | unchanged from first pass | Fine, except the note filename follows the version. |

---

## Verification method

The coordinator's warning that their own Python model got the empty-buffer case backwards is well taken, so I did not trust a port until it earned it. I transcribed `insert_str_as`, `insert_char_raw`, `insert_newline_raw`, `delete_selection_inner`, `EditorSelection::normalised` and `Provenance::splice` from the Rust, then ran the repo's **own committed test assertions** through the port as a positive control — 16 assertions across `inserted_text_carries_the_seat_that_wrote_it`, the empty-buffer case, `an_edit_shifts_below_and_drops_what_it_replaced`, `a_deletion_pulls_the_seats_below_it_up`, and the 216-shape underflow test. All 16 reproduce. Only then did I sweep. The port also tracks per-line *identity* (`orig` / `part` / `join` / `new`), so it can say which original line's text sits at each final index — which is what actually decides whether a seat is on the right text, and what a text-only model cannot see.

---

## Answers to your open questions

**Item 1 — `splice`, independently.** Correct. `src/provenance.rs` was byte-identical (md5 `ceaa088e…`) to what I swept in my first pass, and I re-swept after the saturating change: 4096 `(at, removed, added)` shapes over 0..16 against an independent list-splice model — zero mismatches, zero underflows, zero index collisions. Your 216-shape sweep's conclusion holds; I reached it by a different route. The saturating arithmetic you added closes the theoretical `at + removed` wrap I flagged, which my sweep could not reach because no caller passes an absurd `at`.

**Item 2, the selection case — you were right to suspect it, and it was a real blocker.** Your reasoning that "`before` is measured after the deletion, so `added` reflects only the insertion" is correct *for `added`*, but that was never the problem. The problem was that the deletion changed `self.lines` without the map learning about it, so lines below a multi-line selection kept their old indices while the buffer shifted up. Traced with distinct seats on the pre-fix head:

```
['one','two','three','four'], {0:Me, 1:Navigator, 2:Agent, 3:Peer}
select (0,1)..(2,1), insert "Z"
  -> ['oZhree','four']
     line 1 'four' -- written by Peer -- painted as NAVIGATOR
     keys 2,3 stranded past the end of a 2-line buffer
```

Across 420 selection shapes the pre-fix code produced **300 out-of-range attributions and 200 wrong seats**. Your fix (`splice(sr, er - sr + 1, 1)` before the snapshot) is exactly right, and I verified the version you actually wrote — using `normalised()` captured *before* the delete, which matters because `delete_selection_inner` clears `self.selection`. Post-fix over 660 shapes (selection and no-selection): **0 out-of-range, 0 wrong seats.**

**Item 2, the hardcoded `removed = 1` — still correct, and for a reason worth stating.** After the selection splice collapses the selected span to one line, the buffer really does have exactly one line at `start_line` that the insertion then rewrites. So `removed = 1` is true in both branches: with no selection it is the cursor's line; with a selection it is the single line the collapse just produced. The two splices compose rather than conflict.

**Item 2, mid-line cursor — the tail does get attributed to the inserting seat, and that is correct.** Traced:

```
['hello world'] with line 0 = Peer, insert "\n" at col 5
  -> ['hello',' world'], both Me
```
` world` is Peer's original text, now attributed to Me. This is the documented `+1` design and I think it is the right call: the line `hello world` no longer exists, and neither `hello` nor ` world` is a line Peer wrote — they are lines *this* edit produced. Attributing them to Peer would be worse. It is worth one sentence in the module docs, because "a line whose text you did not type can be attributed to you" is surprising until you see that the line itself is new. This is the sole residual class in my sweep (20 of 660), all at `sel=(r,0)..(r,0)`, all the cursor line legitimately rewritten.

**Item 3 — the remaining paths are now closed.** At the current tree, `provenance` is reset at `editor.rs:4001` (`load_text`), `:5706` (encoding re-decode) and `:7470` (`restore_snapshot`). Those were the three whole-buffer swaps I found; `grep` for `self.lines = split_into_lines` and `self.lines = snap.lines` returns exactly those three sites, all guarded. Remote collab edits still correctly bypass `insert_str` entirely — `apply_collab_spans` (`app/mod.rs:22417`) routes through `apply_span_edits`, so a peer's text stays Unknown rather than becoming `Me`. I found no remaining path by which an unobserved line ends up attributed.

**Item 4 — your check of the five callers was right on four of five, and the fix went further than I asked.** The five `editor.insert_str` sites are `app/mod.rs:25924/25928/25939` (vim paste), `:27315` (Cmd+V) and `:27483` (bracketed paste) — all genuinely the person, so `Seat::Me` is right. But there was a sixth caller you did not list: `editor.rs:5856`, `expand_snippet`, calling `self.insert_str` internally. Both it and completion accept now use the new `Seat::Generated`, which is a better answer than the `Option<Seat>` I suggested — "generated" is a claim croft can actually stand behind, where `Unknown` would have thrown away information it genuinely has.

---

## Findings

### `src/provenance.rs:274` **[major/CI]** — the `Seat::Agent` assertion is not rustfmt-stable and will fail the pinned `fmt --check` gate

```rust
assert_eq!(
    Seat::Agent(String::from("pane 2")).label(),
    "agent (pane 2)"
);
```
`rustfmt.toml` sets `max_width = 100`. Collapsed to one line this is 82 characters including the 8-space indent, so rustfmt collapses it and `--check` reports a diff. CI enforces a pinned `cargo fmt --check` and it runs before the test job, so this fails the build before any of the new tests get to prove themselves.

**Fix:** `assert_eq!(Seat::Agent(String::from("pane 2")).label(), "agent (pane 2)");` on one line. Worth running the formatter over the whole diff — several other multi-line `assert_eq!`s in the new tests are near the same boundary.

### `Cargo.toml:3` / `Cargo.lock:595` **[major/release]** — 0.1.842 now collides with `origin/main`

I re-fetched and re-read at review time, per the repo's own guidance that the release gate cannot catch this itself:

```
origin/main version: "0.1.842"
branch version:      "0.1.842"
```

When you branched this was valid; `origin/main` has since moved to 0.1.842. The gate compares your head against the merge base, so it will pass legitimately and the conflict only surfaces at merge.

**Fix:** bump to 0.1.843 in `Cargo.toml` and `Cargo.lock`, and rename `src/release_notes/0.1.842.md` → `0.1.843.md` (the note filename must match the version or the build errors). Re-read `git show origin/main:Cargo.toml` once more in the same breath as the final `gh pr checks` — it can move again.

### `src/widgets/editor.rs:5772` **[minor/robustness]** — the selection splice trusts `er` to be within the buffer

`let ((sr, _), (er, _)) = sel.normalised();` then `splice(sr, er.saturating_sub(sr) + 1, 1)`. The `saturating_sub` protects the `er < sr` direction, which `normalised()` already makes impossible. What is unprotected is `er >= self.lines.len()` — a stale selection surviving a buffer change would make `removed` larger than the lines that exist. `splice` handles it gracefully (it drops keys in the removed range and shifts the rest), so this is not a live bug, and `delete_selection_inner` would panic on the index first. But since `truncate` now exists, the cheap belt-and-braces is to call `self.provenance.truncate(self.lines.len())` at the end of `insert_str_as`, which makes "no key is ever past the end" hold structurally rather than by argument across every path at once.

### `src/provenance.rs:39` **[nit/naming]** — `Seat::Generated` covers two distinguishable sources

An accepted LSP completion and an expanded snippet both become `Generated`, so the gutter cannot tell "the language server suggested this" from "I expanded a snippet I wrote". Both are honestly not-the-user, so the invariant holds and this is not a correctness issue. Worth noting only because #349's hover text is meant to name the source, and `generated` is the one label that will prompt "generated by what?". A later `Generated(&'static str)` would answer it without changing the model.

---

## Test coverage

The new tests are good — the selection test in particular is genuinely load-bearing. Two observations:

1. **The selection test's `attributed() <= lines.len()` assert is doing the real work; its seat asserts are not.** I ran the test as written against the pre-fix code: `seat(0) == Peer(ada)` and `seat(1) == Me` both **pass** against the buggy version, because all four lines are recorded as the same seat, so a seat landing on the wrong line is invisible. Only the `attributed() <= lines.len()` assert fails (4 ≤ 3). That assert is a good invariant and I would keep it. But give the four lines **distinct** seats as well, so the test states the property it is actually about: with `{Me, Navigator, Agent, Peer}`, the pre-fix code paints the surviving `tail` as `Agent` when it was `Peer`, which fails loudly and reads as the bug it is. Right now a future refactor that fixed the count but not the mapping would slip through.

2. **The loop at the end of the selection test is a no-op.** `for line in 0..sel.lines.len() { let _ = sel.provenance.seat(line); }` discards its result and asserts nothing — `seat` is a `BTreeMap::get`, it cannot panic. It reads as a check but cannot fail. Either drop it or turn it into the assertion it looks like: that every line's seat is `None` or the seat that actually wrote that line's text.

Still uncovered, all lower priority than the two above:
- **The three swap paths.** `restore_snapshot`, `load_text`, re-decode each now clear the map, with no test pinning it. One test per path: attribute a buffer, trigger the swap, assert `attributed() == 0`. Cheap, and they pin behaviour whose comment says the proper fix comes later — exactly the kind that regresses silently.
- **`truncate` is tested in isolation but has no caller.** It is currently dead code (`grep` finds no production call). Either wire it in per the minor finding above, or the test is pinning an API nothing uses.
- **The mid-line split semantics** from your Q3 — that both halves of a split line become the inserting seat. It is a deliberate design choice that looks like a bug on first read, which is precisely what deserves a test with a comment saying why.

---

## Scope

Read against #349's acceptance criteria, at the current tree:

- ☐ *"three distinct colours, correct names on hover"* — **not met.** No gutter painting, no toggle, no inline-blame integration. `Seat::label()` exists but nothing consumes it.
- ☐ *"Provenance survives save, reopen, and an undo/redo cycle"* — **not met**, but now honestly not met: undo clears the map rather than corrupting it, and `restore_snapshot`'s comment says so plainly and names `Snapshot` as the later layer. That is the right call — unknown beats wrong.
- ☑ *"Unknown lines are never attributed"* — **met**, as far as I can verify. 660 shapes, zero out-of-range attributions, zero wrong seats outside the documented cursor-line rewrite; all three whole-buffer swaps clear the map; collab edits bypass the path entirely.

**Your instinct not to use a closing keyword is right** and I'd keep it — two of three criteria are unmet. But the PR description can now legitimately claim criterion 3, which it could not on the previous head. Worth stating explicitly that undo *clears* rather than preserves, so a reviewer of the follow-up knows that is a known gap and not an oversight.

**Repo conventions:** `docs/ARCHITECTURE.md` entry present and matching the repo's prose style, `mod provenance;` registered, release note present. Two corrections: the version needs a bump per the finding above, and the ARCHITECTURE entry sits before `problem_matchers.rs` where `problem_matchers` should sort first (`prob` < `prov`) — cosmetic, but the file is otherwise strictly ordered.

**Baseline tests:** `the_pinned_toolchain_has_every_cross_target` (`src/cli.rs`, `src/remote.rs`) and `a_client_arriving_mid_swap_…` (`src/session_host.rs`) both exist on `origin/main` and are untouched by this diff, which changes nothing in toolchain targets or socket paths. Their failures are environmental and not attributable here — confirmed by locating them on `origin/main`, not assumed.

---

## Scores

**CodeRabbit: 3/5** — the correctness work is done and independently verified; two majors remain, both mechanical: a rustfmt violation that fails CI before the tests run, and a version that collided with `main` while this was in review.

**Greptile: 3/5** — `splice` and the caller wiring are now sound across every shape I could construct, and the `Generated` seat is a better answer to the attribution question than what I proposed. Held below 4 by the two majors and by the selection test's seat assertions passing against the very bug it was written to catch.

---

VERDICT: REQUEST_CHANGES

PR #448, opened 2026-08-31
feat(scm): step the editor back through the branch's commits (#371)

vitali87:
Part of #371 — the cursor layer. **Deliberately not `Closes`:** none of the three acceptance criteria is met, and the scope note says what remains.

TIMELINE lists a file's history; the COMMITS graph lists the repo's. Neither lets you *move* through time. This is the cursor that does, behind **Source Control: Scrub History** — arrows step, `Home` returns to the working tree, `Esc` closes.

## The invariant

**Leaving must restore the live buffer exactly, unsaved edits included.**

Scrubbing is a way of looking, not of editing, and losing someone's uncommitted work to answer a question about history would be worse than not having the feature. So the working tree is a **position** in the cursor rather than something the scrubber replaces — there is no state in which the live buffer has been discarded and the scrubber owes it back. Closing is just dropping the cursor.

That is why `Position::Working` is a variant rather than `Option::None` over a commit index. An `Option` invites "no commit selected" and "back at the working tree" to be the same value, and the second is where the user started.

## Two bugs review found, both a layer above the state machine

**The history came from `commit_graph`**, which logs `--branches --tags HEAD` because it draws the repo-wide graph. With any other branch present, index 0 is whatever topological order put first rather than HEAD — demonstrated with a sibling branch dated 2038, which sorts ahead of it. Stepping back from the working tree then landed on a commit the current branch may not even contain, falsifying the module's own documented invariant.

Fixed with `git::branch_history` (`--first-parent HEAD`). The test carries a **positive control** asserting `commit_graph` *does* see the sibling commit, so a pass means the ref set is right rather than the repo being empty.

**The key hook sat at the top of `handle_key_inner`**, ahead of every modal. A palette or prompt opened while scrubbing would have had its arrows eaten, since those guards return unconditionally and never saw the key.

I first fixed this with an `overlay_is_open()` listing the modals that take precedence — and then threw it away, because a list like that silently rots the next time someone adds a modal. Moving the hook *below* the guards gets the same result by construction: a key only reaches the scrubber when nothing above wanted it.

## Two boundary decisions

- **Stepping back from the working tree lands on HEAD**, not the commit below it. They are different views — one carries unsaved edits — so skipping HEAD would hide the commit the user most wants to compare against.
- **The oldest commit is a wall, not a wrap.** A drag that ran off the end and reappeared at the present would read as the slider slipping, and the user would not know which end they were at.

## Details worth knowing

The scrubber carries whole `GraphCommit`s rather than hashes, because the slider needs `short_hash` for its labels and `parents` for the per-commit gutter delta. The status line shows git's own abbreviation rather than `&hash[..7]`, which is safe only while the string is ASCII hex and hardcodes a width git computes per-repo.

**The scrubber is not a mode.** A key it does not use is not consumed, so the editor keeps working underneath.

`seek` and `fraction` are the slider's API and nothing calls them yet — marked test-only rather than deleted, because the clamping rule and the handle's monotonicity are the parts with a right answer, and deciding them alongside the painting would bury them. `fraction` gives the working tree its own stop at the far right rather than sharing HEAD's, so the user can tell which of the two they are looking at.

## Gate

Every decision is mutation-tested: the ref set, the HEAD-first step, the wall, the clamp, the slider's monotonicity, and the consume/pass-through split. Each fails a different assertion.

3967 pass. The two failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main`; the third is the suite's known double-click load flake, which passes unloaded. fmt 0, clippy 0, doc-ownership gate clean.

Version 0.1.845 against main's 0.1.844.

## Scope

❌ *"Scrubbing 200 commits stays interactive"* — nothing renders yet. ❌ *"The gutter at each step shows that commit's changes"* — no gutter. ⚠️ *"Leaving restores the live buffer with dirty state untouched"* — true, but vacuously: nothing is replaced yet.

Still to build: the slider widget, showing the file at each commit, the gutter delta, Explorer dimming, "Open here" / "Diff to working tree", and lazy loading around the cursor.


vitali87:

**It found a blocker I would not have found**, and in the place I was least looking: the state machine is clean and well-tested, and the layer that *feeds* it was wrong. `commit_graph` logs `--branches --tags HEAD` to draw the repo-wide graph, so index 0 is not HEAD — proved with a sibling branch dated 2038, which sorts ahead of it. Every doc comment in my module rested on the opposite.

It also caught that my key hook sat ahead of every modal, and — after I fixed that with a list of overlays — the reviewer had already argued for the better fix: move the hook below the guards so precedence holds by construction rather than by a list that rots.

All findings applied at the pushed head, each with a mutation-tested pin.

## Walkthrough

This branch adds `src/scrubber.rs`, a pure cursor over a list of commit hashes (`Position::Working | At(usize)`, with `older`/`newer`/`home`/`seek`/`fraction`/`hash`), wires it into `App` as `scrubber: Option<Scrubber>`, adds a `Source Control: Scrub History` palette command, and handles its keys at the top of `handle_key_inner`. The state machine itself is correct — I ported it to Python, replayed all six committed assertions through the port to validate it, then swept every reachable sequence of older/newer/home/seek over histories of length 0..5, and all five of your claims hold with no stranded state and no `fraction()` collision. The defect is one layer up: `scrub_history` populates the cursor from `git::commit_graph`, which logs `--branches --tags HEAD`, so index 0 is not HEAD and the list is not the current branch, which falsifies the module's central documented invariant. Note the worktree has uncommitted changes beyond the reviewed commit (a partial `overlay_is_open` fix); I reviewed what would actually land.

| File | Change |
|---|---|
| `src/scrubber.rs` | New. Cursor + 6 unit tests. Logic verified correct. |
| `src/app/mod.rs` | `scrubber` field, `SCRUB_COMMIT_LIMIT`, `scrub_history`, `handle_scrubber_key`, key hook at top of `handle_key_inner`, uncommitted `overlay_is_open`. |
| `src/app/tests.rs` | One test for the consume/pass-through split. |
| `src/widgets/command_palette.rs` | `Command::ScrubHistory` in all five match arms. |
| `docs/ARCHITECTURE.md`, `src/main.rs` | Entry + `mod` in correct sorted position. |
| `Cargo.toml`/`Cargo.lock`/`release_notes/0.1.845.md` | Version bump 0.1.844 → 0.1.845, note present. |

## Findings

**`src/app/mod.rs:23671`** **[blocker/correctness]** — `scrub_history` fills the scrubber from `commit_graph`, which is not the current branch and whose index 0 is not HEAD. Every doc comment on the module rests on the opposite. `src/git.rs:979-991` runs `git log --branches --tags HEAD --topo-order`. I built two throwaway repos to confirm rather than reason from the flags. With a second branch present, the log returns that branch's commits interleaved:

```
main-commit / OTHER-BRANCH-COMMIT / root    # other branch's commit is in the list
```

and with a sibling branch holding a newer committer date, topo-order puts it first:

```
HEAD is:                  59cf3b9 root
index 0 the scrubber gets: a19c647 FUTURE-other-branch
```

So `hash()` returns a commit the position does not name — exactly the failure mode you asked me to hunt, located in the caller rather than the state machine. `older()` from the working tree lands on a commit that is not HEAD and may not be an ancestor of it, breaking `scrubber.rs:56` ("index 0 is HEAD"), `:78-82` ("stepping back from the tree must show HEAD"), and the issue's "first-parent". Fix: do not reuse `commit_graph`, whose ref selection exists to draw the repo-wide graph. Add a helper in `src/git.rs` running `git log --first-parent HEAD --format=%H%x1f%h -n<limit>` and populate from that. This also fixes the finding below for free, since `%h` gives you git's own short hash.

**`src/app/mod.rs:23725`** **[major/correctness]** — `&hash[..hash.len().min(7)]` byte-slices a `String`, and `GraphCommit` already carries the correct `short_hash` that this code discarded at `:23677-23679` (`.map(|c| c.hash)`). The slice is safe today only because `%H` is ASCII hex — it is not a panic I can demonstrate, which is why this is major and not a blocker. But it is fragile by construction: `min(7)` is a byte count, so the moment this list is fed from any other source the same expression panics on a char boundary, and it hardcodes a width git computes per-repo. Fix: store `(hash, short_hash)` pairs (or the whole `GraphCommit`) and print `short_hash`.

**`src/app/mod.rs:17137` (+ uncommitted `overlay_is_open` at `:23695`)** **[major/correctness]** — the ordering is wrong and the in-progress fix is incomplete. You asked specifically whether a modal below is now eaten: yes, several. `handle_key_inner` has 15 guards that `return Ok(())` unconditionally; the uncommitted `overlay_is_open` lists 7. Missing and confirmed to bind the scrubber's own keys:

- `input_prompt` — `Left`/`Right`/`Home`/`Esc` at `:32052-32056`
- `list_picker` — `Left`/`Right`/`Esc` at `:32095-32110`
- `branch_picker` — `Left`/`Right`/`Home`/`Esc` at `:31947-31963`
- `zoxide_jump` — `Left`/`Right`/`Home`/`Esc` at `:31729-31747`

Also missing: `pending_terminal_warning`, `process_picker`, `command_history_popup`, and the four `pending_*` confirmations. This is reachable, not theoretical: nothing clears `self.scrubber` except `Esc` (the only `scrubber = None` in the file is `:23712`), and `open_input_prompt` at `:20690` does not touch it. So scrub, then hit a command that prompts, and your `Esc`/arrows go to the scrubber while the prompt sits there ignoring them. `pending_terminal_warning` is the worst case — it deliberately swallows all keys so the user cannot type through into the UI beneath, and the scrubber now bypasses it. Fix: either extend `overlay_is_open` to all 15 guards, or move the scrubber hook *below* the `scm_menu` guard at `:17291`, which gets the precedence right by construction and cannot drift as modals are added. I would take the second — the list is a maintenance liability that will silently rot the next time someone adds a modal.

`KeyEventKind` gating is correct: `Press | Repeat` matches the `peek_popup` guard directly below and the app-wide filter at `:17161`.

## Test coverage

- **`scrub_history` has zero tests** (grep for `scrub_history|ScrubHistory` in `tests.rs` returns 0, against a control of 61 `run_command(` calls, so the corpus and pattern both work). No test asserts the cursor is loaded from the current branch — which is why the blocker shipped. Write: a temp repo with `main` plus a sibling branch carrying a newer commit date, run `Cmd::ScrubHistory`, press `Left`, assert `hash()` equals `git rev-parse HEAD`. That test fails on this branch today and is the red run the fix needs.
- **The overlay-precedence bug is untested end-to-end.** The uncommitted test only asserts `overlay_is_open()` flips for a palette; it never sends a key through `handle_key_inner` with both up. Write: open the scrubber, set `input_prompt`, feed `KeyCode::Left` through `handle_key_inner`, assert the prompt's cursor moved and the scrubber's position did not.
- **`SCRUB_COMMIT_LIMIT`** is never asserted to reach `commit_graph`.

## Nitpicks

- `src/scrubber.rs:24` — `Position::At(usize)` doc says "0 is the newest"; once the blocker is fixed, say "HEAD" explicitly, since newest-by-date and HEAD are exactly what diverged here.
- `src/app/mod.rs:23686-23689` — the uncommitted edit inserted `overlay_is_open`'s doc comment *between* `handle_scrubber_key`'s doc comment and its `fn`, so three lines of scrubber docs now sit above the wrong function. Cosmetic, but it reads as a botched paste.
- `fraction()` at n=1 returns 0.0 for HEAD and 1.0 for the tree — the handle traverses the entire track for one step. Correct per your spec and not a collision, just worth knowing before the widget lands.

## Answers to your specific questions

**(4) Dead-code marking.** Honest — keep them. `#[cfg_attr(not(test), allow(dead_code))]` has five existing uses in this repo (`agent_lane.rs:379`, `command_history.rs:151`, `markdown.rs:969`, `widgets/shortcuts.rs:20`), so it is the established idiom, not a new escape hatch. `seek` and `fraction` are both fully tested and are the two pieces of the slider with a right answer independent of painting. Deleting them would discard verified logic to re-derive it later.

**(5) Scope.** No closing keyword — confirmed correct. Zero of the three acceptance criteria are met: nothing renders, no gutter, no buffer swap, and criterion 3 (restore the live buffer) is satisfied only vacuously, since nothing is ever replaced. As for whether the cursor layer makes the follow-up harder: the state machine does not, and `fraction()` is genuinely the slider's hard half already solved. But `scrub_history` throwing away everything but `.hash` is a real obstacle — the follow-up needs `short_hash` for labels, `parents` for the per-commit gutter delta, and the graph's colour rails, all of which `GraphCommit` already carries and this code drops on the floor. Store the `GraphCommit`s, not the hashes.

## Scores

**CodeRabbit: 1/5** — one blocker (the cursor is populated from the wrong ref set, falsifying the module's stated invariant) plus two majors.
**Greptile: 1/5** — the state machine is genuinely clean and well-tested, but the layer that feeds it is wrong in a way its own tests cannot see, and the key-precedence fix is half-finished in the worktree.

VERDICT: REQUEST_CHANGES

Issue #348, opened 2026-08-29
Worktree lanes: one chord creates a git worktree, adds it as a workspace root, and starts an agent in it


vitali87:
## Motivation

Running several agents in parallel, each in its own `git worktree`, is how people use Claude Code and Codex today, and they do it by hand with tmux. croft has multi-root workspaces, named panes, and agent badges; the missing piece is the chord that wires them together.

## Current state

- Multi-root: `WorkspaceRoots`, Add/Remove Folder, per-root git workers, per-root tasks/tests (#143..#163).
- Terminal panes with names, cwd inheritance, per-workspace layout persistence.
- `git.rs` has no worktree operations yet.

## Proposed design

1. `Cmd+K W` / **Agent: New Worktree Lane**: prompts for a branch name (default `agent/<slug>`), runs `git worktree add ../<repo>-<slug> -b <branch>`, adds that path as a workspace root, and opens a named terminal pane there running the configured agent command (`agents.json` per-agent launch line; empty = plain shell).
2. The Explorer root section header shows the lane's branch and its agent badge; the agent lane (review queue) groups by worktree.
3. **Agent: Close Worktree Lane** removes the root, kills the pane (via the existing undoable-close path), and offers `git worktree remove` when the tree is clean, refusing when dirty.
4. **Agent: Merge Lane** opens the branch's diff against the base as a commit-graph selection and hands off to the existing merge editor for conflicts.

## Acceptance criteria

- [ ] One chord yields a new worktree, a new root, and a running pane in under 2 s on a warm repo.
- [ ] Closing a dirty lane refuses to remove the worktree and says why.
- [ ] Lanes are restored on relaunch (roots and panes already persist; the worktree/agent association must too).

## Related

- Tracking epic: #379
- Depends on: #344

vitali87:
## The worktree operations have shipped in #449 — the pane wiring has not

**Shipped:** "Agent: New Worktree Lane" prompts for a name, cuts a worktree on `agent/<slug>` beside the repo, and adds it as a workspace folder. "Agent: Close Worktree Lane" removes it, refusing when there is anything to lose.

**Still open:** launching the agent command in the lane's pane (`agents.json` per-agent launch line), the Explorer header showing the lane's branch and agent badge, grouping the review queue by worktree, "Agent: Merge Lane", a keybinding, and persisting the worktree/agent association across relaunch.

[...]

PR #449, opened 2026-08-31
feat(agents): worktree lanes for parallel agent work (part of #348)

vitali87:
Part of #348 — the worktree operations and two palette commands. **Deliberately not a closing keyword:** only part of one acceptance criterion is met, and the scope note says which.

Running several agents in parallel, each in its own `git worktree`, is how people use coding agents today — with tmux, by hand. croft already has multi-root workspaces and named panes; this is the chord that wires them together.

## The safety rule is the whole feature

`git worktree remove` **discards what it removes.** So `lane_removal_block` is the one check here where being wrong destroys someone's work rather than annoying them, and three things about it were only learnable by measuring:

**Plain `--porcelain` omits ignored files, and `worktree remove` deletes them anyway.** I destroyed a `.env` in a test lane while the check reported the tree clean. A lane is *precisely* where an uncommitted local config, a local database, or an expensive `node_modules` lives. Now `--ignored`, counted separately so the refusal names what it is protecting. The test carries a **positive control** asserting plain `--porcelain` sees nothing — so it proves the flag rather than the corpus.

**A locked worktree reports a clean status and is then refused by git.** "Is there uncommitted work" is not the same question as "will git remove this", so the caller runs git *before* dropping the workspace folder — the other order leaves the lane gone from the workspace and the directory still on disk.

**Removing a worktree from inside itself works.** That is what lets the caller not track which repo a lane was cut from.

The check fails closed for the same reason: a tree git cannot report on is treated as unsafe, and `remove_worktree_lane` omits `--force`, which would make the refusal advisory.

## The blocker review found

`close_worktree_lane` used `workspace_root()`, which is `roots.primary()` — and a lane is *appended* by `add_workspace_folder`, so it is never index 0. **The handler targeted the user's main repo every time.**

git's "is a main working tree" refusal masks that today, but the guard is git's rather than ours. Launch croft *inside* a lane — the workflow this feature creates — and the primary becomes a removable worktree: the directory is deleted, `remove_workspace_folder` refuses the primary, and the status still reports success. Now resolved from the active root, with the primary refused explicitly and the folder's departure verified rather than assumed.

## Naming

One slug covers both halves. git refuses a ref containing `..`, whitespace, or any of `~^:?*[\`, and a directory carrying those is a different kind of trouble — so a single pass to `[a-z0-9-]` handles the ref grammar and the filesystem at once rather than sanitising twice with two chances to disagree. Capped at 60 characters, because a component past ~255 bytes fails at `mkdir` with "File name too long" and the user would see that instead of anything about lanes. A name with nothing usable is **refused** rather than becoming the branch `agent/`.

The worktree is a **sibling**, never a child: one nested inside its own repo is a working tree git then tries to track, and the Explorer would show the lane inside the root it was cut from.

## Gate

**Mutation testing earned itself twice.** Replacing the not-a-worktree arm with `return None` passed every assertion I had written — the dirty/clean test only ever pointed at real worktrees, so the most dangerous branch was untested. And dropping `--ignored` now fails, which is what makes that a fix rather than a claim.

The round-trip goes through the real git commands rather than asserting on the strings they would build: the argument order and the directory each command runs in are the parts that go wrong.

3969 pass. The two failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main`; the rest are the suite's known load flakes and pass unloaded. fmt 0, clippy 0, doc-ownership gate clean. Version 0.1.847 against main's 0.1.846.

## Scope

⚠️ *"One chord yields a worktree, a root, and a running pane"* — worktree and root land; **no pane and no agent command**, and it is palette-only with no keybinding. ✅ *"Closing a dirty lane refuses and says why"* — and now for ignored-only trees too. ❌ *"Lanes restored on relaunch"* — not attempted.

Still to build: launching the agent command in the lane's pane, the Explorer header showing branch and badge, grouping the review queue by worktree, "Agent: Merge Lane", and persisting the worktree/agent association.


vitali87:

**Two findings I would not have reached, both about destroying work.** The blocker: `close_worktree_lane` used `workspace_root()`, which is the *primary* root — and a lane is appended, so it is never index 0. The handler targeted the user’s main repo every time. And a data-loss major: plain `--porcelain` omits ignored files while `git worktree remove` deletes them, so a `.env` in a lane is destroyed while the check reports the tree clean. I reproduced both before fixing.

The reviewer also independently confirmed the two things I had measured myself (the git-first reorder, self-removal from inside a worktree) and swept ~50 adversarial names through `git check-ref-format`, which I had not done at all.

## Walkthrough

`WorktreeLane::plan` / `lane_slug` derive a branch `agent/<slug>` and a sibling directory from one sanitising pass; `lane_removal_block`, `add_worktree_lane` and `remove_worktree_lane` wrap the git commands, and two app handlers wire them to new palette commands. The slug logic and the git wrappers are solid — I swept the slug against `git check-ref-format --branch` and found no rejected ref and no path escape, and I reproduced the locked-worktree ordering rationale in throwaway repos. The two app handlers are where it breaks: `close_worktree_lane` operates on the primary root, which `remove_workspace_folder` structurally refuses, and neither handler has a test.

| File | Change |
|---|---|
| `src/git.rs` | `WorktreeLane`, `lane_slug`, `lane_removal_block`, `run_git_mut`, `add_worktree_lane`, `remove_worktree_lane` + 3 tests |
| `src/app/mod.rs` | `create_worktree_lane`, `close_worktree_lane`, `Cmd` dispatch, `InputPurpose` arm |
| `src/widgets/command_palette.rs` | Two `Command` variants across 5 match arms |
| `src/widgets/input_prompt.rs` | `InputPurpose::NewWorktreeLane` |
| `docs/ARCHITECTURE.md`, `Cargo.toml`, `Cargo.lock`, `src/release_notes/0.1.847.md` | Version 0.1.847 + docs |

## Findings

**`src/app/mod.rs:23723` [blocker/correctness] — `close_worktree_lane` can never remove a lane, and after the reorder it deletes the directory and then silently fails to drop the folder.**

`workspace_root()` is `roots.primary()` → `roots[0]` (`src/app/mod.rs:6502-6504`, the only `App::workspace_root`). `remove_workspace_folder` delegates to `WorkspaceRoots::remove`, which returns `false` for the primary (`src/workspace.rs:79-82`), sets the status `"The primary workspace folder cannot be removed (use Make Root to change it)"` and returns at `src/app/mod.rs:37592-37597`.

A lane added by `create_worktree_lane` goes through `add_workspace_folder` → `roots.add`, which pushes to the *end* of the vec (`src/workspace.rs:67-72`). A lane is therefore never `roots[0]`. So `close_worktree_lane` always targets the primary repo, never the lane:

- The `roots.iter().count() < 2` guard at `:23724` passes (there are ≥2 roots) but proves nothing about the path in `lane` — it is still the primary.
- `lane_removal_block(&primary)` returns `None` whenever the user's main repo is clean (I verified `git -C <clean repo> status --porcelain` exits 0 with empty output).
- `remove_worktree_lane(&primary)` runs `git -C <main> worktree remove <main>`; I measured git's response: `fatal: '/tmp/lanerev/main' is a main working tree`, exit 128. So the user sees `Keeping the lane: fatal: ... is a main working tree`.

Today the main-worktree refusal masks it, but the guard is git's, not the code's. The moment croft is launched *inside* a lane (the natural workflow this feature creates — open the lane, work, close it) the primary IS a removable worktree: `remove_worktree_lane` succeeds and deletes the directory, then `remove_workspace_folder` hits the primary refusal at `:37592` and returns early, overwriting nothing. The folder stays in `roots`, `tree`, `git_extra` and `save_workspace_folders()`, pointing at a directory that no longer exists — and the final status at `:23741` reads `Removed lane <path>`, so the user is told it worked. The reorder to git-first (correct in itself) is what moved this from "folder dropped, dir remains" to "dir destroyed, folder remains, success reported".

Fix: resolve the lane from the *active* root, not the primary, and refuse when it is the primary. The repo already has the idiom — `active_scm_root` is seeded from `roots.owning_root(focused file)` (`src/app/mod.rs:6011`, and `owning_root` at `src/workspace.rs:49-55`). Something like:

```rust
fn close_worktree_lane(&mut self) {
    let lane = self.active_scm_root.clone();      // or owning_root of the focused file
    if lane == self.roots.primary() {
        self.status = String::from("The primary folder is not a lane");
        return;
    }
    // ... unchanged from here
}
```

and then have `remove_workspace_folder`'s failure be observable rather than swallowed.

**`src/git.rs:1093` [major/data-loss] — `lane_removal_block` reports a lane holding only ignored files as clean, and `git worktree remove` then destroys them.**

`status --porcelain` without `--ignored` omits ignored entries, so `changed == 0` and the function returns `None`. I verified end to end on git 2.47.1, against the current head:

```
worktree with committed .gitignore (.env, node_modules/), plus .env and node_modules/lib.js
  git status --porcelain      -> [] (empty, exit 0)   => lane_removal_block = None
  git worktree remove <lane>  -> exit 0
  .env survived?              -> NO -- DESTROYED
```

Positive control in the same run: `git status --porcelain --ignored` on the same shape prints `!! .env`, so the omission is the flag, not the corpus. This is squarely the failure mode the doc comment at `src/git.rs:1085-1091` claims to prevent — an agent lane is exactly where an uncommitted `.env`, a local SQLite file, or an expensive `node_modules`/`target` lives. Note git's own non-force `worktree remove` refuses *untracked* files (I measured `fatal: ... contains modified or untracked files`) but not ignored ones, so nothing downstream catches this.

Fix: add `--ignored` (or `--ignored=matching`) to the args at `src/git.rs:1095` and word the refusal for it, e.g. count `!!` lines separately — "this lane has N ignored files (build output, local config) — remove it by hand or delete them first". Refusing here is cheap; the doc comment's own reasoning ("refusing costs a user one manual `git worktree remove` while guessing costs them their work") applies exactly.

**`src/git.rs:2874-2882` [minor/maintainability] — the lane test carries eight lines of another test's doc comment.**

`a_lane_name_becomes_a_safe_branch_and_directory` was inserted into the middle of the `ScrubHistory` test's doc comment, so it is documented as "The scrubber's history is the CURRENT BRANCH, and index 0 is HEAD… demonstrated here with a sibling branch carrying a future committer date… (#371)" followed by the one real sentence about #348. The original comment is duplicated intact above the scrubber test further down, so nothing is lost — just delete lines 2874-2881 from the lane test's doc block.

**`docs/ARCHITECTURE.md:25` [nit/docs] — missing sentence separator.** The appended text joins the previous sentence: `…a filename need not be UTF-8 Also worktree lanes (#348):`. Add `. ` after `UTF-8`.

## Test coverage

**Both app handlers are completely untested.** A grep for `close_worktree_lane` across `src/` returns only its definition and its `Cmd` dispatch — no test constructs an `App` and calls either handler. That absence is precisely why the primary-root blocker shipped: every assertion in the diff is at the `git.rs` layer, where the functions are correct, and the defect lives entirely in which path the handler hands them.

The test I would write: build an `App` with a primary root plus a lane added via `add_workspace_folder`, focus a file in the lane, call `close_worktree_lane`, and assert both that `roots.iter().count()` dropped to 1 *and* that the lane directory is gone. Written against the current code it fails on the first assertion, which is the point. A second case — primary root only, `close_worktree_lane` — should assert the primary's directory still exists (today nothing pins that git's main-worktree refusal is what saves it).

**Ignored-only removal has no test.** Add to `a_dirty_lane_refuses_removal_and_a_clean_one_allows_it`: commit a `.gitignore` containing `*.log`, write `build.log`, and assert `lane_removal_block` returns `Some`. It fails today.

**Verified sound, no test needed:** the slug sweep (I ran ~50 adversarial inputs — `.lock` suffixes, `-`/`--`/`---`, `.`, `..`, `....`, CJK/Cyrillic/emoji, `--force`, `refs/heads/x`, `../../../etc/passwd`, NUL, zero-width space, 300/500-char names — through a Python port validated against the committed assertions first; every surviving branch passed `git check-ref-format --branch` and no path escaped `/repos/`), self-removal from inside the worktree (exit 0, directory gone, `worktree list` pruned, including from a nested cwd), and the TOCTOU on `lane.path.exists()` (git itself refuses a non-empty existing path — `fatal: '<path>' already exists` — and a duplicate branch, so the guard is only a nicer message; the race is benign).

## Nitpicks

- `src/git.rs:1092` — `lane.to_str()` failing yields "is not a readable path", which describes a non-UTF-8 name rather than an unreadable one. "is not a valid UTF-8 path" matches the wording `run_git_mut` and `add_worktree_lane` already use.
- `src/app/mod.rs:23702` — the `exists()` guard rejects an *empty* existing directory that `git worktree add` would happily accept (measured: exit 0, `HEAD is now at…`). Harmless, slightly stricter than git.
- `src/app/mod.rs:20952` — the `NewWorktreeLane` arm is the only one in that match that does not call `close_input_prompt()` before dispatching. Worth confirming `create_worktree_lane` leaves no prompt on screen; every sibling arm closes first.
- A 500-char name produces a branch git accepts but the filesystem rejects (`Unable to create '…lock': File name too long`, and `mkdir` fails at 250 chars on APFS). `run_git_mut` surfaces git's message, so it degrades acceptably — capping the slug at ~100 chars would give a better error.

## Scope (#348)

Framing is honest, and I checked it mechanically rather than by reading. No closing keyword anywhere in the commit message: the only regex hits are "FAILS CLOSED" and "fail-closed", neither adjacent to a number, and both `#348` mentions are preceded by "of" and "relaunch." — GitHub requires keyword-immediately-before-number, so nothing auto-closes. The message also enumerates what is excluded.

Acceptance criteria: **(1) one chord → worktree + root + running pane** — partial; worktree and root land, no pane and no agent command, and no keybinding (palette-only, consistent with `ScrubHistory`). **(2) closing a dirty lane refuses and says why** — met at the `git.rs` layer, but not reachable through the UI given the blocker, and it does not hold for ignored-only trees. **(3) lanes restored on relaunch** — not attempted, correctly declared as excluded.

## Scores

**CodeRabbit: 1/5** — one blocker: `close_worktree_lane` targets the primary root, which cannot be removed, and after the git-first reorder it deletes the lane directory then reports success while leaving the stale folder in the workspace.

**Greptile: 1/5** — the blocker plus a data-loss major (ignored-only lanes read clean and are destroyed), both in a feature whose stated purpose is not to destroy work; the app layer has no tests at all.

Two things the coordinator asked me to second-guess, checked independently: the git-first reorder is right and the locked-worktree reasoning reproduces exactly, and all four untried shapes (dirty submodule, orphaned worktree with uncommitted work, a plain file, a symlink) fail closed correctly. The generalisation — "is there uncommitted work" ≠ "will git remove this" — is sound, and the ignored-file case is the same insight pointing the other way: git will remove things `status` never reported.

VERDICT: REQUEST_CHANGES

Issue #363, opened 2026-08-29
Fleet run: execute one command on N SSH hosts and diff the outputs side by side


vitali87:
## Motivation

Broadcast typing (`Cmd+K I`) already sends keystrokes to every pane, which covers the "type once" half of fleet work. The other half is comparing what came back. Ops users would come for a mode that runs a command on N hosts from `~/.ssh/config`, lays the outputs out, and highlights what differs.

## Current state

- SSH host parsing for `croft remote`, the remote bulk lane (`src/remote_bulk.rs`), split panes, `host_accents` per-host decoration, the diff widget.

## Proposed design

1. **Terminal: Fleet Run**: pick hosts (multi-select list from ssh config, saved groups in `config.json`), type a command, croft runs it over SSH on each in parallel (non-interactive, `ssh host -- cmd`, with a per-host timeout) into a tiled result view: one tile per host, exit-code dot, duration.
2. **Diff mode**: pick a reference tile (default: the most common output, by hash); every other tile shows only lines that differ, coloured with the diff widget; a header summary `7 identical · 2 differ · 1 failed`.
3. Results go into an OUTPUT channel per run and can be captured to a file; each tile's "open shell here" spawns a normal pane on that host.
4. Local mode: hosts can include `localhost` and docker containers (`docker exec`) if the container view issue lands.

## Acceptance criteria

- [ ] `uname -r` across three hosts shows three tiles; diff mode against the first highlights the kernel string that differs.
- [ ] A host that times out shows a red tile and does not delay the others.
- [ ] Saved host groups round-trip through config layering.

## Related

- Tracking epic: #383

vitali87:
## The runner and the comparison have shipped in #450 — the tiles have not

**Shipped:** "Terminal: Fleet Run" takes `hosts: command` (or `*: command`), runs it in parallel over those ssh hosts, and reports to a `Fleet` OUTPUT channel — one line per host marked `same`/`DIFFERS`/`FAILED`, then a summary.

**Still open:** the tiled result view with exit-code dots and durations, host multi-select, saved groups in `config.json`, per-tile "open shell here", capturing a run to a file, and localhost/docker targets.

**Criteria:** ❌ three tiles with diff highlighting (text lines, no tiles); ⚠️ a timing-out host shows red without delaying others — the parallelism and deadline are measured working and the line is `Error`-level, but there is no red tile; ❌ saved host groups round-trip through config.

[...]

PR #450, opened 2026-08-31
feat(terminal): run one command across a named fleet (part of #363)

vitali87:
Part of #363 — the runner and the comparison. **Deliberately not a closing keyword:** two of three acceptance criteria are unmet, and the scope note says which.

Broadcast typing already covers "type once". This is the other half, and the comparison is the whole value — reading ten near-identical `uname -r` outputs by eye is exactly the task a person is worst at.

## The fleet must be named

The request is `hosts: command`, with `*` as the explicit broadcast.

**My first version defaulted to every host in `~/.ssh/config`, and review called that reckless.** It was right. On this machine that set is five hosts including a live box and a `github.com` entry that isn't a shell host at all. The argument that changed my mind is sharper than "a typo hits prod": a confirmation reading *"run on 5 hosts?"* asks the user to approve a list they **cannot inspect** — it comes from a file they last edited months ago. That's consent theatre.

Choosing the fleet *is* the feature. "Fleet Run" without fleet selection is "run on everything". A host croft doesn't know is refused rather than attempted, so a typo can't become an ssh connection.

## The reference is a plurality

Not the first host — keying on the first makes the answer depend on the order hosts happen to be listed in, so the same fleet listed differently would highlight a different set. Only **successful** hosts vote: three identical `Permission denied` results becoming the normal would paint every working host as the deviant one, the exact inversion of what the summary is for. An exact tie yields *no* reference, because with 2-and-2 nothing leads, and choosing one deterministically is worse than choosing none — a stable wrong answer reads as a considered one.

## Off the event loop

`run_on_hosts` returns a `Vec`, so calling it inline reads naturally. It would also freeze the editor for up to `FLEET_TIMEOUT` while the network answered, with no scrolling, typing, or cancelling. Results now arrive through a channel drained on the tick — the shape `port_poll_rx` and `label_rx` already use.

**Every queued batch is reported, not just the newest.** The idiom is borrowed from the label channel, where each message is a complete *snapshot* so `.last()` is correct. Here each message is one run, and `.last()` would drop another run's output with no trace of it.

## Keyed by request index

Two entries naming one host collided: the second reply overwrote the first, one slot consumed it, and the other was filled in as a timeout — so **a host that answered fine reported red, and the surviving result was whichever thread was slower.** Unreachable from today's caller, which dedupes, but the host multi-select will pass hand-picked lists.

## Measured, not reasoned

**`ConnectTimeout` does not bound a host that connects and then hangs** — a `sleep 30` ran the full 30 seconds under a 3-second connect timeout. So the wall-clock deadline is genuinely required, not belt-and-braces. And ssh treats `--` as end-of-options and does not forward it to the remote shell, so a command that looks like a flag can't be consumed as one.

## Gate

**A surviving mutation found an untested branch, twice.** The timed-out-host fill-in was never exercised, because both test hosts were unreachable-but-*reporting*; a duplicate host forces two threads onto one slot, which reaches it. And asserting only the tile *count* couldn't see a lost result — the test now asserts no tile carries the synthesised timeout.

3978 pass. The two failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main`; the rest are the suite's known load flakes and pass unloaded. fmt 0, clippy 0, doc-ownership gate clean. Version 0.1.849 against main's 0.1.847 (.848 is another branch in flight).

## Scope

❌ *"`uname -r` across three hosts shows three tiles; diff mode highlights the kernel string"* — no tiled view; output is text lines in a channel. ⚠️ *"A host that times out shows red and does not delay the others"* — parallelism and the deadline both measured working, and the line is `Error`-level, but there is no red tile. ❌ *"Saved host groups round-trip through config layering"* — no config surface.

Still to build: the tiled result view, saved groups, per-tile "open shell here", capturing a run to a file, and localhost/docker targets.

## Summary by CodeRabbit

* **New Features**
  * Added Terminal Fleet Run to execute a command across selected SSH hosts or all configured hosts.
  * Results now appear per host, showing matching, differing, failed, or timed-out outcomes, followed by a summary.
  * Added Fleet Run to the command palette with an input prompt for specifying hosts and commands.
* **Documentation**
  * Documented fleet command execution and result comparison.
* **Chores**
  * Updated the package version to 0.1.849.


vitali87:

CodeRabbit is rate-limited on this repo, so this is the review. Two rounds.

**The blocker is the one I should have seen and did not**, and the reviewer sharpened it past my own framing. I had thought a confirmation prompt would cover "runs on every host"; the argument that changed my mind is that **a confirmation naming a count asks the user to approve a list they cannot inspect** — the set comes from a file they last edited months ago. Choosing the fleet *is* the feature.

It also verified my claims against real ssh rather than accepting them (`ConnectTimeout` really does *not* bound a host that connects then hangs — a `sleep 30` ran the full 30s), swept the vote logic over 55,986 fleets, and corrected my terminology: what I implemented is a **plurality**, not a majority.

## Walkthrough

Re-reviewed against `8d58f96`. `src/fleet.rs` is byte-identical to the commit I first swept (`git diff be04d32 HEAD -- src/fleet.rs` is empty, sha1 `655d479b`), so its verification carries over. The change since: `run_fleet_command` now spawns a worker that sends through `fleet_tx`, `drain_fleet_results` reports on a later tick, `fleet_running` rejects a concurrent second run, and a new app test asserts the split. Item 4's freeze is genuinely fixed and correctly wired; the remaining substantive question is the missing host selection, which I argue below is a blocker for a different reason than I first thought.

| File | Change |
|---|---|
| `src/fleet.rs` | Unchanged since first review. Vote logic, `run_on_hosts`, `fleet_ssh_args`, `run_one`, 5 unit tests. |
| `src/app/mod.rs` | +`fleet_rx`/`fleet_tx`/`fleet_running`; `run_fleet_command` now non-blocking; new `drain_fleet_results`; drain wired into `main_loop:44936`. |
| `src/app/tests.rs` | New `a_fleet_run_reports_through_the_channel_rather_than_blocking`. |
| Others | Unchanged from first review (version 0.1.849, release note, ARCHITECTURE entry, palette/prompt/output wiring). |

## Findings

**`src/app/mod.rs:23707` — [blocker/safety] Every configured host is targeted, with no selection and no confirmation.**

You asked plainly, so: **reckless as shipped, and I'd hold the PR on this alone.** My first-pass reasoning was "typo hits prod", which a confirmation prompt would fix. Verifying the host list on this machine changed my mind about the shape of the problem. `~/.ssh/config` here contains `huys, github.com, gitcgr, oldest, various` — and I confirmed `huys` is a live reachable Linux box that accepted `BatchMode=yes` and ran my command. So the feature's actual first-run behaviour on a normal developer machine is to execute arbitrary text on every remote the user has ever configured, including `github.com`, which is not a shell host at all and will just produce noise.

The reason a confirmation prompt is not sufficient: the prompt string is `"command to run on every configured host"`, and the user has no way to see what that set *is* before or during the run. The set is derived from a file they last edited months ago. A confirmation that says "run on 5 hosts?" asks them to approve a list they cannot inspect. That is consent theatre, not consent. The issue's design puts the multi-select first (step 1: "pick hosts") precisely because *choosing the fleet is the feature*, not a nicety layered on top — "Fleet Run" without fleet selection is "run on everything".

Two acceptable exits, in order of preference: (a) ship the multi-select from the issue's step 1, or (b) if you want this landed now, gate it behind a confirmation that **enumerates the resolved aliases** and requires an explicit accept — plus name the count in the prompt. Option (b) is a genuine stopgap, not a fix, and should say so in a comment referencing #363.

**`src/fleet.rs:188` — [major→minor/correctness, downgraded] Duplicate hosts lose a result; unreachable from the current caller.**

I need to correct my own first-pass severity here, because I checked the caller this time. The defect is real and I re-measured it:

```
dup, fast then slow   elapsed=3.01s  [('a','out-a@0.6',0), ('a','timed out',None)]
dup, slow then fast   elapsed=3.01s  [('a','out-a@0.6',0), ('a','timed out',None)]
```

The `0.6s` result wins in **both** orderings, so **yes — a slow host's late result deterministically overwrites a faster one**, because `got.insert` is last-writer-wins on the host key. Modelling divergent payloads shows the user-visible damage: a host that succeeded fast and failed slow reports `[Permission denied, timed out]` — the success is discarded and a phantom timeout invented.

But `discover_ssh_targets` (`src/remote.rs:206-214`) folds through a `BTreeSet` and dedupes by alias, and it is the only production caller of `run_on_hosts` (`src/app/mod.rs:23751`; the only other hit is a doc comment). So duplicates cannot occur today. That makes this a latent trap rather than a live bug — **minor now, major the moment a caller passes a hand-picked list, which is exactly what the multi-select above will do.** Since you're about to add that caller, fix it in the same change.

So: the duplicate-host semantics are **merely counted, not sane**. The fill-in makes the tile count right while the content is wrong. Fix by keying on request index — have the worker send `(idx, HostResult)` into a `Vec<Option<HostResult>>` of length `hosts.len()`. That makes duplicates independent, removes last-writer-wins, and fixes the next finding for free.

**`src/fleet.rs:181` — [minor/performance] The collection loop burns the full deadline whenever a result is lost.**

`while got.len() < hosts.len()` cannot be satisfied when two threads share a key, so the loop waits out the whole budget even though every thread finished. Measured: two duplicate hosts both completing in 0.6s still took **3.01s** (the full `timeout + 2s`), versus **0.31s** for three distinct fast hosts. At the app's 10s `FLEET_TIMEOUT` that is a 12-second wait on a run that finished instantly. Same index-keyed fix.

**`src/fleet.rs:169` — [minor/resource] Abandoned threads and `ssh` children outlive the function.**

Threads are detached and `Command::output()` blocks until ssh exits, so when the deadline fires `run_on_hosts` returns while the children keep running; the late sends fail silently against the dropped receiver. This matters because I measured that the deadline is doing real work: `ssh -o ConnectTimeout=3 huys -- 'sleep 30'` ran the **full 30.6s and printed `DONE`**. So your claim is correct and now measured, not reasoned — **`ConnectTimeout` does not bound a host that connects then hangs**, and the wall-clock deadline is genuinely required. The gap is that it bounds *reporting* only, not the work: each hung host leaks an ssh process per run. Consider `Stdio::piped()` plus `child.kill()` on deadline, or document that abandoned children are expected.

**`src/app/mod.rs:23749` — [minor/robustness] `fleet_running` can stick `true` forever if the worker dies without sending.**

It is set `true` at 23749 and cleared only at 23765 inside `drain_fleet_results`, on the success path. If the worker thread panics before its `send` — or is killed — no `Vec` ever arrives, the drain always returns `false`, and the flag never clears, permanently disabling the feature for the session with the status stuck at "running on N hosts…". `run_on_hosts` itself is panic-resistant (I verified a panicking worker just yields a `timed out` fill-in for that host and the others report normally), so this needs an unusual failure, but it is unrecoverable when it happens. Clear the flag when the sender disconnects: check for `TryRecvError::Disconnected`, or hold a generation counter and time out the in-flight state.

**`src/widgets/command_palette.rs:193` — [minor/correctness] `FleetRun`'s doc comment is stacked onto `NewWorktreeLane`'s, leaving the latter undocumented.**

```rust
    /// Create a git worktree lane for an agent to work in (#348).
    /// Run one command across several ssh hosts and compare (#363).
    FleetRun,
    NewWorktreeLane,
```

Still present at the new head. Move the `#348` line back below `FleetRun`.

**`src/fleet.rs:146` — [minor/security] `StrictHostKeyChecking=accept-new` is undocumented and unique in the codebase.**

The doc comment justifies `-n` and `BatchMode=yes` but is silent on the host-key policy. It is the only `accept-new` in `src/` (the sole other occurrence, `src/remote.rs:2968`, is `=no` in a test fixture). Auto-trusting first-contact keys fleet-wide is defensible but security-relevant, and this file's style is to justify each flag — and I watched it fire for real (`Warning: Permanently added '5.189.133.124'`). Document it or drop it and inherit the user's config.

**`src/app/mod.rs:23763` — [nit→minor/correctness] `try_iter().last()` silently discards all but the newest result.**

This mirrors `refresh_terminal_labels` (`src/app/mod.rs:29004`), so the idiom is conventional — but the semantics differ. There, each message is a *complete snapshot* of all labels, so keeping only the newest is correct by construction. Here each message is *one distinct run's results*, and `fleet_running` does not fully prevent two from queuing (a run whose worker sends just as the flag is being read, or any future caller). If two ever land in one tick, the older run's output vanishes from the channel entirely with no trace. Prefer `for results in self.fleet_rx.try_iter()` and report each, which is also more honest about the OUTPUT channel being a log.

### Verified as correct, no finding

**The 2-1-1 tie, confirmed by my own sweep as you asked.** Your reading is right: `a` twice with `b` and `c` once each returns **`Some("a")`**, not `None`. `winners` counts outputs tied *at the maximum*, not distinct outputs, so ties strictly below the max are irrelevant — a majority with ties beneath it is still a majority. I re-validated the Python port by replaying all five committed tests' assertions first, then swept **55,986 fleets** (n=1..6, three outputs × two exit states): **zero** cases where the returned reference was not the strict mode, **zero** cases where a tie failed to yield `None`. Spot checks: `aab`→`a`, `aaabc`→`a`, `aaabbc`→`a`, `aabbc`→`None`, `aabb`→`None`, `ab`→`None`. The `winners > 1` count does not misfire.

Terminology only: this is a **plurality**, not a majority — 1,026 swept cases have a winner at ≤50% (e.g. `x,x,y,z`→`x` at 2/4). Behaviour is right; docs, commit message and ARCHITECTURE entry all say "majority".

**The deadline does bound the total** for distinct hosts: one host hanging at 99s still returned in 3.01s with the hung host filled in as `timed out`, and all-fast hosts returned in 0.31s without waiting. It fails to bound only in the duplicate case above.

**Thread panic is handled gracefully.** A panicking worker among three hosts yields `[('a', ok), ('b','timed out'), ('c', ok)]` — the panic is contained, the fill-in covers the missing send, and the other hosts report normally. No poisoning (no shared mutex) and no lost results.

**`--` semantics, both halves measured.** With a positive control: `ssh … nonexistent.invalid -- -v` produced only the resolve error, while the identical command *without* `--` emitted `debug1:` output — so ssh genuinely treats `--` as end-of-options. And it is **not** passed to the remote shell: against live host `huys`, `ssh … huys -- 'echo ARGC=$#; echo ARGV0=$0'` returned `ARGC=0`, `ARGV0=zsh`. Your test's claim at `src/fleet.rs:408` is accurate. `ConnectTimeout=2` against a blackhole IP returned in 2.024s with rc=255.

**Item 4's fixed half is correct.** `drain_fleet_results` is called from `main_loop` at `src/app/mod.rs:44936`, unconditionally every tick, and joined with non-short-circuiting `|` so it always runs regardless of the other drains' return values. The handler returns immediately after spawning. The re-entry guard works. This matches the `port_poll_rx`/`label_rx` convention correctly, and the new test asserts the split rather than the implementation.

**Scope framing is safe.** I grepped the new head's full commit message for GitHub's closing keywords (`close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved` + `#363`): no match. "part of #363" and "#363 stays open for the rest" cannot auto-close. No repeat of the `#397` incident.

**Acceptance criteria:** (1) three tiles with diff highlighting — **not met** (text lines in a channel, no tiled view). (2) timing-out host red without delaying others — **substantially met** (parallelism and the deadline both measured working; `Error` level not a red tile). (3) saved host groups round-trip through config — **not met**. All three are named in the commit's "Not included" paragraph, which matches what I found.

**Conventions:** `Cargo.toml`/`Cargo.lock` both 0.1.849; `origin/main` re-read just now is still **0.1.847**, so the bump is valid — but 0.1.848 is absent from `release_notes/`, so another branch likely holds it; re-read at merge time per CLAUDE.md. Release note present and matching. `mod fleet;` correctly sorted in `main.rs`. ARCHITECTURE entry substantive. Working tree is clean at `8d58f96`.

## Test coverage

- **Duplicate-host result loss.** `a_host_that_never_reports_still_gets_a_tile` asserts only `len()` and host names, which is why the overwrite survives. Assert on *content*: no entry should be the synthesised `"timed out"` when both threads reported. Needs an injectable runner (`fn(&str,&str,Duration)->HostResult`) so a fake can return distinct fast results — which would also let the whole collection layer be tested without ssh.
- **Upper bound on `run_on_hosts` wall time.** The existing assertion is `< 12s` for a 1s timeout — 4× the real `timeout+2s` bound, so it would not catch the 3.01s duplicate stall. With an injectable runner, assert `elapsed < timeout + 2s + slack`.
- **`fleet_running` cleared when the worker never sends.** Nothing covers the stuck-flag path. Drop the sender without sending and assert a subsequent run is still permitted.
- **Two runs landing in one drain.** Send two `Vec`s before draining and assert both are reported, which currently fails under `.last()`.
- **Empty-string reference.** `reference_output` returns `Some("")` for a host that exits 0 with no output, counted as `1 identical` — plausible in reality (missing command exiting 0) and untested.

## Nitpicks

- `src/fleet.rs:97-110` scans `counts` three times (max, tie count, find). A single `max_by_key` plus a tie check reads more directly; correctness unaffected.
- `src/app/mod.rs:23773` flattens output with `.replace('\n', " ")`, so a multi-line result becomes one unreadable line. Truncate with an ellipsis or indent one line per output line until tiles exist.
- The prompt `"command to run on every configured host"` names neither the count nor the hosts — see the blocker.
- "majority" should read "plurality" in the module docs, commit message and ARCHITECTURE entry.

## Scores

**CodeRabbit: 1/5** — One blocker: the command runs on every configured ssh host (here including a live production box and `github.com`) with no selection and no inspectable confirmation. The event-loop freeze is properly fixed; remaining issues are minors.

**Greptile: 1/5** — The vote logic is correct under a 55,986-case sweep, the ssh flags are all verified against real ssh(1), and the async rework matches the codebase's own convention and is genuinely wired into the tick. Blocked solely on shipping an unscoped fleet broadcast without host selection.

VERDICT: REQUEST_CHANGES

vitali87:
## CodeRabbit's five findings — all taken, at `808d6f4`

Both majors were real, and the security one changed a decision rather than fixing an oversight.

**`StrictHostKeyChecking` (major, security).** I had `accept-new`, and my own reviewer flagged only that it was *undocumented*. CodeRabbit's framing is the right one: `accept-new` silently trusts a first-contact key, and **this is a fleet**. A single interactive `ssh` shows the user a fingerprint and asks; a fleet run shows nobody anything, so an attacker impersonating one unknown host receives the command the user typed. Now `yes` — an unknown host fails, and establishing the key costs one deliberate `ssh host` with the fingerprint on screen, which is where that decision belongs. Pinned with a test that also asserts `accept-new` is absent.

**Timed-out workers (major, stability).** Correct and it compounds. `ConnectTimeout` bounds the handshake only — I measured a `sleep 30` running the full thirty seconds under a 3-second connect timeout — so a host that connects then hangs left an ssh child running after the deadline had already reported it timed out. Every run leaked another. Now spawned with piped stdio, polled, and killed *and reaped* on the deadline. `wait_with_output` was the trap: it consumes the child and leaves no handle to kill.

**Closing the prompt (minor, correctness).** Right, and it was a plain omission — every sibling arm in that match calls `close_input_prompt()` first. Leaving it open hides the status line the run writes, so the user cannot see the answer they just asked for. Fixed for the worktree-lane arm too, which had the same gap.

**Release note and ARCHITECTURE wording (minor ×2).** Both accurate. The note implied an implicit host set when the whole point of the change is that the fleet must be named, and the architecture entry described tiles that do not exist — results are text lines in a `Fleet` channel. Both now say what ships.

Worth noting the two reviews **disagreed on severity in a useful direction**: my fallback reviewer rated the leaked processes *minor* and the host key *minor/undocumented*; CodeRabbit rated both *major*. On the host key I think CodeRabbit is straightforwardly right, and the reason is that my reviewer was assessing the flag while CodeRabbit was assessing the flag **in a fleet context** — the same option is a reasonable default for one connection and a bad one for twenty unattended.

fmt 0, clippy 0, doc-ownership gate clean, 3978 passing.

Issue #369, opened 2026-08-29
Symbol tabs: open a single function or type as its own editor tab


vitali87:
## Motivation

VS Code cannot open "just this function". croft has tree-sitter and LSP symbols on every buffer, so a virtual tab that shows one symbol's source, with edits round-tripping into the file, is cheap and unusual: a focused view for reviewing one function, or for keeping a helper open beside the code that calls it.

## Proposed design

1. `Cmd+P #name` (workspace symbols) and the Outline's context menu gain **Open as Symbol Tab**. The tab title is `name · file.rs`, the breadcrumb shows the enclosing trail, and the content is the symbol's byte range from the live buffer.
2. Edits apply to the underlying buffer through the same edit pipeline as a split (the symbol tab is a view over a range, not a copy), so LSP, diagnostics, undo, dirty state, and collab replication all just work; the range tracks the tree-sitter node so it grows and shrinks with edits.
3. Deleting the symbol from the file closes the tab with a notice; renaming updates the title.
4. Multiple symbol tabs from one file can sit side by side in a split.

## Acceptance criteria

- [ ] Open `fn render` as a symbol tab, add a line inside it, save: the file has the line, the tab still spans exactly the function.
- [ ] Editing the file in another split above the function keeps the symbol tab anchored.
- [ ] Go-to-definition from inside a symbol tab lands in a normal file tab (or another symbol tab, configurable).

## Related

- Tracking epic: #384

vitali87:
## The range tracking has shipped in #451 — the tab has not

**Shipped:** `src/symbol_range.rs` — `SymbolRange::after_edit` follows a symbol's byte range through an edit, and `enclosing_symbol` finds the innermost symbol containing a line. An `Editor: Open Symbol as Its Own Tab` command stores the range.

**Still open:** the tab and its rendering, the title and breadcrumb, the Outline context-menu entry, `Cmd+P #name`, tracking through tree-sitter node identity rather than byte offsets, and the go-to-definition landing rules. None of the three acceptance criteria is met, because there is no tab to open.

**Why the view is not built yet, which is the thing worth knowing before picking this up.** `follow_symbol_tab_edit` is wired only to the collab span path, because that is the one place an edit already arrives as `(at, deleted, inserted)`. Local typing, undo, paste and formatting all mutate through line/column APIs that report no byte span — so a tab shipped today would be **right for remote edits and silently wrong for the user's own**, which is worse than no tab. Giving the editor a byte-span edit signal is the prerequisite.

[...]

PR #451, opened 2026-08-31
feat(editor): track a symbol's range through edits (part of #369)

vitali87:
Part of #369 — the range-tracking core. **Deliberately not a closing keyword:** none of the three acceptance criteria is met, because there is no tab yet to open.

A symbol tab is a **view** over a byte range of the live buffer, not a copy. That is what lets edits, LSP, diagnostics, undo and collab replication work through the existing pipeline instead of a second one — and the cost is that every edit can move the range.

## The inside case is the one that matters

An edit above shifts the range; one below moves nothing. Both are arithmetic. An edit **inside** must grow or shrink it, because the symbol is still the same symbol — and a symbol tab exists to be typed in, so shifting instead of resizing slides the range off the end of the very function it is showing.

Two boundary decisions, each tested. A **deletion** ending exactly at `start` is above, because the text it consumed was not the symbol's. A zero-width **insertion** at `start` is inside, because the typed bytes land at the symbol's first position — treating it as above would push the tab down and off as you type at the top of a function. They share an `edit_end` and are told apart by whether anything was removed.

An edit **straddling** either boundary reports `Gone` rather than re-anchoring: the surviving range would be half a function glued to whatever the edit left behind.

## Three things review caught

**The tab carried no file** — and a byte offset means nothing in another buffer. A collaborator editing `other.rs` shifted, resized, or closed a tab pointed at `main.rs`. On any two-file session that fires on almost every remote keystroke, which is the exact failure the module doc names.

**The last symbol's range ran one byte past the buffer.** The line-to-byte sum charged a `\n` to the final line, but the `+ 1` is a *separator*, not a terminator. That panics anything slicing by the range — the whole point of the range — and makes an append at true EOF read as *inside* the symbol.

**The tie-break was inverted.** On identical spans `min_by_key` keeps the first, and symbols arrive parents-before-children, so the parent won. `OutlinePanel::follow_caret` uses `span <= best_span` and keeps the child. The same cursor would have highlighted one symbol in the Outline while opening a different one as a tab — neither answer is obviously right, but two pickers disagreeing is definitely wrong.

## A mutation that did not apply read as a surviving one

My script's multi-line anchor missed a branch rustfmt had collapsed onto one line. Nothing was mutated, the suite passed, and the output was **identical to "this branch is untested"** — I started widening a test that was already correct. Every mutation script here now asserts its anchor matched, and each of the four decisions above dies under one.

## Only the collab path is wired, and that is deliberate

`follow_symbol_tab_edit` is called from `apply_collab_spans` alone, because it is the one place an edit already arrives as `(at, deleted, inserted)`. Local typing, undo, paste and formatting mutate through line/column APIs that report no byte span, so a tab would go stale under the user's own edits.

**That is why the tab view is not in this PR.** Shipping a view now would give a tab that is right for remote edits and silently wrong for your own. Giving the editor a byte-span edit signal is the next piece.

## Gate

3985 pass. The two failures on my machine (`the_pinned_toolchain_has_every_cross_target`, `a_client_arriving_mid_swap_…`) fail identically on unmodified `origin/main`; the rest are known load flakes. fmt 0, clippy 0, doc-ownership gate clean. Version 0.1.850 against main's 0.1.849.

## Scope

❌ *"Open `fn render` as a symbol tab, add a line, save"* — there is no tab. ❌ *"Editing above keeps it anchored"* — the tracking is correct and tested, but nothing renders. ❌ *"Go-to-definition from inside a symbol tab"* — not attempted.

Still to build: the tab and its rendering, title and breadcrumb, the Outline context-menu entry, `Cmd+P #name`, tree-sitter node identity instead of byte offsets, and the go-to-definition landing rules.


vitali87:

**It swept `after_edit` over 19.6M cases against an independent model — zero divergence — and then found three defects in the wiring around it.** That split is the useful part: the arithmetic I was careful about was right, and the things I had not thought about were wrong. The blocker is the one I would not have found, because the symbol tab stored a name and offsets with no path, so an edit to *any* file moved it.

It also checked my tie-break against the repo`s own `OutlinePanel::follow_caret` and found they disagreed — my picker and the Outline would have named different symbols for the same cursor.

## Walkthrough

Adds `src/symbol_range.rs`: a `SymbolRange { start, end }` byte range with `after_edit(at, removed, inserted)` returning `At(range)` or `Gone`, plus `enclosing_symbol(symbols, line)` picking the narrowest-span symbol containing a line. Wires an `App::symbol_tab: Option<(String, SymbolRange)>` field, a `follow_symbol_tab_edit` shim called from `apply_collab_spans`, and an `open_symbol_tab` command (`Editor: Open Symbol as Its Own Tab`) that converts the enclosing symbol's line range to bytes and stores it.

I ported `after_edit` to Python, replayed all 16 committed assertions through the port (all matched), then swept every `(start, end, at, removed, inserted)` over 0..14 — **19.6M cases, zero divergence** from an independent model of the stated spec, and zero underflow/overflow. The arithmetic and the three-case classification are correct. The problems are all at the edges the module doesn't own: the symbol tab is file-agnostic, the byte conversion overruns the last symbol, and the tie-break contradicts the repo's own convention.

| File | Change |
|---|---|
| `src/symbol_range.rs` | New module: `SymbolRange::after_edit`, `enclosing_symbol`, 4 unit tests |
| `src/app/mod.rs` | `symbol_tab` field, `follow_symbol_tab_edit`, `open_symbol_tab`, `apply_collab_spans` hook, Cmd dispatch |
| `src/app/tests.rs` | One integration test over `follow_symbol_tab_edit` |
| `src/widgets/command_palette.rs` | `OpenAsSymbolTab` command in 5 required arms |
| `docs/ARCHITECTURE.md`, `src/release_notes/0.1.850.md`, `Cargo.{toml,lock}` | Conventions |

## Findings

**`src/app/mod.rs:22496-22500`** **[blocker/correctness]** — the symbol tab follows edits from *every* file, not its own.

`apply_collab_spans` resolves `path` at line 22466 and applies spans only to editors whose `ed.path == Some(path)` (22502, 22508). But the new loop at 22496 is unconditional:

```rust
for span in spans {
    self.follow_symbol_tab_edit(span.at, span.deleted, span.inserted.len());
}
```

`symbol_tab` is `Option<(String, SymbolRange)>` (mod.rs:3088) — it stores a *name and offsets, no path*. So a collaborator editing `other.rs` shifts, resizes, or `Gone`-closes a symbol tab pointing into `main.rs`. In a session with two files open this fires on essentially every remote keystroke; the "Closed the render tab: that symbol is gone" notice will appear for edits in an unrelated file. This is the exact failure the module doc calls out ("showing the wrong text under the right title, which is worse than showing nothing") arriving through the one call site that exists.

Fix: store the path in the tab — `Option<(String, PathBuf, SymbolRange)>` — and guard the loop:
```rust
if self.symbol_tab.as_ref().is_some_and(|(_, p, _)| p == &path) {
    for span in spans { self.follow_symbol_tab_edit(span.at, span.deleted, span.inserted.len()); }
}
```

**`src/app/mod.rs:23788-23802`** **[major/correctness]** — the byte range for the file's last symbol runs one byte past the buffer.

`len` sums `l.len() + 1` for every line in `first..=last`, charging a `\n` to the final line too. croft's own offset model (`collab::byte_offset`, `src/collab.rs:416-429`) adds `+1` only as a *separator* between lines, so a buffer of N lines has no trailing newline. Ported and checked against a 6-line two-function buffer: `fn b` on lines 3..5 of a 29-byte text yields `(15, 30)` — `end` is 30, one past the 29-byte buffer.

Consequences: `range.len()` in the status line is wrong by one for the last symbol; and any consumer slicing the buffer by this range (which is the whole point of the follow-on tab view) panics or clamps. It also breaks `after_edit`'s "below" test — an edit appended at true EOF (`at == 29`) is `< self.end` (30), so it is misclassified as *inside* and grows the range further.

Fix: subtract the final separator when `last` is the last line:
```rust
let mut len: usize = self.editor.lines.iter().skip(first).take(last + 1 - first)
    .map(|l| l.len() + 1).sum();
if last + 1 >= self.editor.lines.len() { len = len.saturating_sub(1); }
```

**`src/symbol_range.rs:213-219`** **[major/correctness]** — on identical spans `min_by_key` picks the *outer* symbol, the opposite of the documented intent and of the repo's existing convention.

The doc comment (and `open_symbol_tab`'s at mod.rs:23768) promises "the INNERMOST enclosing symbol, so opening from inside a method shows the method rather than its `impl`". `min_by_key` keeps the **first** element on a tie. Symbols reach here in document order with **parents before children** — `outline_syntax.rs:85-89` sorts ascending start then *wider span first*, and `lsp/manager.rs:5069-5083`'s `push_nested` pushes a parent before recursing into `children`. So on a tie the parent is first and wins.

Contrast the repo's own innermost-picker, `src/widgets/outline.rs:186-192`, which uses `if span <= best_span` — a non-strict compare that keeps the **last**, i.e. the child. The two functions disagree on exactly the tie case you flagged. I ported both: for `[("fn render",20,30), ("body",20,30)]` at line 25, `enclosing_symbol` returns `fn render` while `follow_caret` returns `body`; for a one-line `impl`+`fn` at the same line, `enclosing_symbol` returns `impl Foo` and `follow_caret` returns `fn tiny`.

On your specific question — for a *symbol tab*, the child is the useful one when it's a function inside an identically-spanned `impl`, but the parent is useful when the child is an LSP-reported *body* node. Either way the current behaviour is not what the doc claims, and it is inconsistent with `follow_caret`, so the same cursor position highlights one symbol in the Outline and opens a different one as a tab. Fix: match the existing convention and take the last on a tie, and say so:

```rust
symbols.iter()
    .filter(|s| s.range_start_line <= line && line <= s.range_end_line)
    .rev()
    .min_by_key(|s| s.range_end_line.saturating_sub(s.range_start_line))
```
(`.rev()` before `min_by_key` makes the last-on-tie explicit.) Add a test with two identical spans asserting which one wins, since neither behaviour is currently pinned.

**`src/app/mod.rs:23771`** **[major/design]** — `follow_symbol_tab_edit` has exactly one caller, and it is the *remote* edit path. Every local mutation leaves the tab stale.

Grepped every buffer mutation entry point. Not hooked:

- **Local typing** — `Editor::insert_char` (`src/widgets/editor.rs:5492`), `insert_char_raw` (5804)
- **Undo/redo** — `Editor::undo` (`editor.rs:7442`)
- **LSP formatting, code actions, colour presentation** — `Editor::apply_span_edits` (`editor.rs:5614`), called at `mod.rs:22662`
- **Find-and-replace / rename across files** — `apply_span_edits_to_lines` (`editor.rs:1905`), called at `mod.rs:10590`, and `apply_rename_to_open_tab` (`mod.rs:10580`)

So the only edits the tab follows are ones *other people* make. A user typing in their own buffer — the case the module doc calls "the common case, since a symbol tab exists to be typed in" — desynchronises the tab immediately. As a foundation this is incomplete in the one direction that matters most; the range logic is sound but nothing local feeds it.

This is defensible as staged work *if* it is stated. It is not: the commit's "Not included" list names the tab view, title, breadcrumb, context menu, `Cmd+P` and go-to-definition, but does not mention that local edits are unwired. Fix: either route local edits through the same shim (ideally a single choke point in `Editor` that reports `(at, removed, inserted)`), or add "local edit paths are not yet wired — only collab spans call `follow_symbol_tab_edit`" to the commit body and the module doc so the next session does not assume it is done.

**`src/symbol_range.rs:180-185`** **[minor/robustness]** — `self.start + inserted - removed` and `self.end + inserted - removed` are unchecked `usize`, unlike `len`/`is_empty` which use `saturating_sub`.

I proved the *subtraction* is safe on both branches. Above-branch: the guard is `at + removed <= start`, so `removed <= start`, hence `start + inserted - removed >= inserted >= 0`. Inside-branch: the guard gives `at + removed <= end` with `at >= start`, so `removed <= end` and `end + inserted - removed >= inserted`. The 19.6M-case sweep found no underflow. The *addition* is the exposed side: `end + inserted` overflows for absurd `inserted`, which I reproduced in the port. `inserted` comes from `span.inserted.len()` — attacker-influenced in a collab session but bounded by an actual allocated `String`, so this is not practically reachable today.

The real hazard is the `saturating_add` on line 172 creating a guard the arithmetic does not share: if `at + removed` ever saturates, the above-branch guard `edit_end <= self.start` goes false and the code falls through to a `Gone` — safe, but by accident rather than design. Fix, for uniformity with the rest of the type: use `checked_sub`/`checked_add` and return `Gone` on `None`, or at minimum a comment recording the invariant proof above so a later edit to the guards does not silently break it.

**`src/symbol_range.rs:150-156` (boundary decisions)** — you asked to be challenged. I agree with both calls.

(a) *Deletion ending exactly at `start` is ABOVE.* Right. The deleted bytes were not the symbol's, so the symbol's text is unchanged and only its position moved — shifting is the only answer that keeps the tab on the same characters. Treating it as inside would shrink the range from the top while the deleted text was never in it.

(b) *Zero-width insertion at `start` is INSIDE.* Right, and this is the one that matters for the user experience you describe. Typing at the very first character of a function: those bytes visually land inside the function's first line, and the user expects the tab to grow. Classifying it above would push `start` past the typed characters, so each keystroke at the top of the function would scroll the tab's first line out of view — precisely the "slides off its own function" failure the module exists to prevent.

Typing on the blank line *above* the function is a genuinely different offset — it's an insertion at some `at < start`, whose `edit_end = at + 0 = at < start`, so it takes the above branch and shifts. That's correct and the asymmetry is well-placed. The `!(removed == 0 && at == self.start)` predicate is the minimal discriminator; I confirmed by sweep that no other input distribution is affected by it.

## Test coverage

- **The collab cross-file bug (blocker).** No test calls `apply_collab_spans` with a symbol tab open. Write: open a symbol tab anchored to `a.rs`, deliver a `ResolvedSpan` for `b.rs`, assert `app.symbol_tab` is byte-identical afterwards. This test fails on the current code.
- **`open_symbol_tab` has no test at all.** Nothing exercises the line→byte conversion, the two empty-status branches, or the command dispatch. Write: a buffer of two functions, cursor inside the second, run `Cmd::OpenAsSymbolTab`, assert `text[range.start..range.end]` equals exactly the second function's source. That test catches the last-symbol off-by-one directly.
- **Identical spans in `enclosing_symbol`.** `the_innermost_enclosing_symbol_wins` covers nesting with distinct spans only. Write two symbols with the same `range_start_line`/`range_end_line` and assert which is returned — the behaviour is currently unpinned in either direction.
- **Local edits.** No test asserts that typing keeps the tab anchored, which is why the gap above is invisible to the suite. Once a local path is wired, assert `insert_char` inside the range grows it.
- **Edit exactly at true EOF.** Depends on the byte fix; assert an append at `buffer.len()` leaves the last symbol's range unchanged rather than growing it.

## Nitpicks

- `src/app/mod.rs:23805-23811` — the status string "the tab view lands next" is developer-facing scaffolding leaking into user-visible UI. Every other status in this file describes what happened, not what the author plans. Suggest `"{name}: {} bytes"` or gate the command out of the palette until the view exists.
- `src/symbol_range.rs:135` — `is_empty` is `#[cfg_attr(not(test), allow(dead_code))]` with the comment "nothing calls it yet". Fine, but the `allow` will silently stop being needed once a caller exists; a `TODO(#369)` would make it removable on sight.
- `src/app/tests.rs:36605` — the integration test asserts through `follow_symbol_tab_edit` directly rather than through `apply_collab_spans`, which is why the path-filtering gap is untested. Driving it through the real call site would have caught the blocker.
- `src/widgets/command_palette.rs:750` — empty keybinding string, consistent with neighbours; noted only for completeness.

**Conventions (all satisfied, verified):** version bumped `0.1.849 → 0.1.850` in both `Cargo.toml:3` and `Cargo.lock:595`, matching `git show origin/main:Cargo.toml`; `src/release_notes/0.1.850.md` exists with a single `feature:`-prefixed line per CONTRIBUTING.md:196-206; `mod symbol_range;` sits between `svg` and `tasks` in `src/main.rs:71`, correctly sorted; the ARCHITECTURE.md entry is in the same sorted position with the house's dense-prose style; `OpenAsSymbolTab` is present in all five arms (`ALL_COMMANDS`, label, keybinding, id, dispatch) that the `ALL_COMMANDS` round-trip tests require.

**Issue scope.** #369's three acceptance criteria are all unchecked and **none is met** by this diff — there is no symbol tab, so nothing can be opened, saved, or navigated from. The commit is honest that this is the range-tracking substrate only.

**Closing-keyword safety: confirmed safe.** I grepped the full commit message for `(close[sd]?|fix(e[sd])?|resolve[sd]?)[\s:]+#?\d+` — no match. The two `#369` mentions are `part of #369` (subject) and `#369 stays open` (body). "part of" and "stays open" are not GitHub closing keywords, so no auto-close will fire. Your framing cannot repeat the `#397` incident, because that one had a keyword adjacent to the number and yours has none.

**Baseline failures.** `the_pinned_toolchain_has_every_cross_target` lives in `src/remote.rs:3428` and the mid-swap collab test is likewise outside this diff — `git diff --name-only` confirms neither file is touched. Correctly not attributable here.

## Scores

**CodeRabbit: 1/5** — one blocker: `apply_collab_spans` drives the symbol tab from edits to unrelated files because the tab carries no path.

**Greptile: 1/5** — the core arithmetic is provably correct under a 19.6M-case exhaustive sweep, but the wiring around it has a blocker plus three majors (last-symbol byte overrun, tie-break inverted against the repo's own `follow_caret`, and no local edit path hooked).

VERDICT: REQUEST_CHANGES

Issue #356, opened 2026-08-29
croft record: write the session as an asciicast the app makes itself


vitali87:
## Motivation

Every demo GIF of croft is currently made with a third-party recorder. The session host already produces the exact byte stream a recorder needs, with timestamps. `croft record` gives users a first-class way to capture a session (bug reports, tutorials, "look at this") and makes the project's own marketing loop self-hosting.

## Current state

- `src/session_host.rs` relays `Frame::Bytes` with per-client queues; no persistence of the stream.
- `croft attach` sessions have a socket and a workspace key.

## Proposed design

1. `croft record [--out file.cast]` attaches a recorder participant (read-only, never counted toward min-winsize) that writes asciicast v2 (`{"version":2,"width":..,"height":..}` header, `[t, "o", data]` events) with monotonic timestamps. `croft record --stop`, or a **Session: Stop Recording** palette entry; a red `● REC` chip in the status bar (the macro recorder's chip exists; distinguish the two).
2. Resize events are written as `"r"` events; the initial screen state is emitted first so the cast does not start blank (the host already replays state on attach).
3. Privacy: the secret-redaction trigger action (sibling issue), when enabled, applies to the recording too, so a leaked key in a demo is masked at the source.
4. Playback is out of scope; asciinema and its web player already handle the format.

## Acceptance criteria

- [ ] A recorded `ls` session plays back correctly in `asciinema play` and in the agg GIF converter.
- [ ] Resizing the window mid-recording produces a valid `r` event and no garbled frames on playback.
- [ ] The recorder participant never throttles the pump for live clients.

## Related

- Tracking epic: #382
- Pairs with: #360

vitali87:
## The serialiser and a local recorder have shipped in #453

**Shipped:** `src/asciicast.rs` writes asciicast v2, and "Session: Record Terminal as Asciicast" records the active pane to a `.cast` file with `o` and `r` events.

**Still open:** `croft record` as a CLI verb with `--out` and `--stop`, the recorder as a read-only session participant (this records the *local* pane, so criterion 3 about not throttling the pump does not yet apply), the `● REC` chip distinguishing it from the macro recorder, emitting the initial screen state, and secret redaction at the source.

**Criteria:** ⚠️ plays back in `asciinema play`/`agg` — the format is conformant and the geometry correct, but **monochrome**; ✅ a resize produces a valid `r` event, ordered before the frame drawn at it; ❌ the recorder participant never throttles the pump — no participant yet.

[...]

PR #453, opened 2026-08-31
feat(session): record a terminal as an asciicast (part of #356)

vitali87:
Part of #356 — the serialiser and a local recorder. **Deliberately not a closing keyword.**

Every demo of croft is currently made with a third-party recorder, while the session host already produces the byte stream one needs. This turns that into the format `asciinema play` and the `agg` GIF converter already read.

## The two traps are in the format's own description

**The data is a JSON string, not bytes.** `format!("[{at}, \"o\", \"{data}\"]")` reads naturally and produces a file that is not JSON the moment the stream contains a quote, a backslash or an escape byte — which is *every* real session. The player then rejects the **whole** recording, not the one frame. Every payload goes through a real encoder, and so does the header: a title carrying a quote would otherwise make the player report "not an asciicast" rather than name the field.

**Time must not go backwards.** `SystemTime` steps back across an NTP correction, and a player reading a decreasing timestamp either sleeps forever or panics depending on the player. A backwards frame is **clamped** onto the previous instant, which shows as a burst — where dropping it would lose output the user saw, and writing it raw would break playback for everything after it.

## Two sizing bugs review caught, both in the wiring

**The header declared the pane's bordered size, not the grid.** `Borders::ALL` costs a cell on each side and the PTY is sized from the *inner* rect, so a pane painted at 80×24 recorded a header for a grid that is really 78×22. The player re-wraps every long line at the wrong column — exactly the "garbled frames" the issue's second criterion names.

**Every frame wrote the full 5000-line scrollback.** `grid_lines` starts at `topmost_line()`, which is negative scrollback. After a clear, writing all of it scrolls the live screen straight off the top — so the cast showed the *tail of the history* rather than what the user was looking at, at roughly 400 KB per frame. A one-minute demo would have been hundreds of megabytes.

Neither was visible to my original test, which counted event kinds without checking geometry or content.

## Known limits, named rather than discovered

**The recording is monochrome and the cursor is not restored.** `grid_lines` returns plain `String`s — the attributes are in the grid but not in that view — so a cast records the *text* of a session, not its colour. Carrying attributes means re-emitting SGR per run of cells, which is the byte stream this deliberately does not tap.

The capture is the **screen**, not the PTY stream, because the bytes are consumed by the grid before the app sees them. Clear-and-redraw per frame rather than a delta: a delta needs the previous screen and a diff that agrees with the terminal's own scroll behaviour, and a cast that drifts a row from reality is the kind of bug nobody notices until the demo is published.

A write failure **stops** the recording rather than being ignored — a half-written cast that keeps growing is worse than one that ends, and the user needs to know the file they are about to share is truncated.

## Gate

Every decision is mutation-tested, and **each script asserts its anchor matched first** — a mutation that fails to apply reads exactly like a surviving one, which cost me real time on #451. Hand-rolling the encoder, dropping the clamp, transposing `COLSxROWS`, recording the bordered size, and writing the scrollback each fail a different assertion.

The end-to-end test **parses the file it produced** rather than matching strings, and finds it by listing rather than rebuilding the name — a test that recomputes the path asserts against its own copy of the naming rule.

3992 pass; the failures are the machine's two environmental ones (which fail identically on unmodified `origin/main`) plus known load flakes. fmt 0, clippy 0, doc-ownership gate clean. Version 0.1.853 (the peer session holds .851 and .852).

## Scope

⚠️ *"A recorded `ls` session plays back in `asciinema play` and `agg`"* — the format is conformant and the geometry now correct, but monochrome. ✅ *"Resizing mid-recording produces a valid `r` event"* — emitted once per size change, ordered before the frame drawn at it. ❌ *"The recorder participant never throttles the pump"* — there is no participant; this records the local pane.

Still to build: `croft record` as a CLI verb with `--out`/`--stop`, the recorder as a read-only session participant, the `● REC` chip, initial-state emission, and secret redaction at the source.


vitali87:

**It checked the asciicast spec rather than my description of it** — confirming `timestamp` is optional and my header is conformant — and then found two majors in the wiring the serialiser feeds. Both were real and both broke playback: I recorded the pane`s *bordered* rect while the PTY is sized from the inner one, and `grid_lines` returns 5000 rows of scrollback rather than the visible screen.

The second is the one worth noting: it made every frame ~400 KB, so a one-minute demo would have been hundreds of megabytes, and after a clear the live screen scrolled straight off the top. My original test counted event kinds and could see neither.

## 1. Walkthrough

Adds `src/asciicast.rs`, a small asciicast v2 serialiser (header + `[t,"o",data]` / `[t,"r","COLSxROWS"]` lines, with a monotonic-time clamp), and wires it into `App` behind a new `Session: Record Terminal as Asciicast` palette command. Rather than teeing the PTY byte stream, the app samples the *screen* on each dirty tick and writes a clear-and-redraw frame. The serialiser itself is correct and well tested; the wiring is where the problems are — the recorded geometry is taken from the pane's **outer bordered rect** rather than the grid, and the frame body is taken from `grid_lines()`, which returns the entire 5000-line scrollback, not the visible screen.

| File | Change |
|---|---|
| `src/asciicast.rs` | New. `Recorder` + `Event`; header/line serialisation via `serde_json`; backwards-time clamp. 4 unit tests. Sound. |
| `src/app/mod.rs` | `recording` / `recorded_size` state; `toggle_session_recording`, `record_active_screen`, `record_resize`, `record_terminal_output`; hook in `drain_terminals_dirty`. Two size bugs live here. |
| `src/app/tests.rs` | One integration test parsing the emitted file. Does not cover geometry or frame content. |
| `src/widgets/command_palette.rs` | `ToggleSessionRecording` registered across all five match arms. Correct. |
| `src/main.rs`, `docs/ARCHITECTURE.md`, `Cargo.toml`, `Cargo.lock`, `src/release_notes/0.1.853.md` | Module registration, docs, version bump, release note. Conform. |

**Note on review target:** the worktree has uncommitted edits to `src/app/mod.rs` and `src/app/tests.rs` that fix findings 3 and 4 below. Findings are anchored to the committed branch (`fb1fd80`); the fixes need committing.

---

## 2. Findings

### `src/app/mod.rs:23930` — **[major/correctness]** The cast header declares the pane's bordered size, not the terminal grid size

`toggle_session_recording` builds the header from `term.last_area.width/height`. But `last_area` is the **outer** rect including the pane border. In `src/widgets/terminal.rs:4127-4135`:

```rust
let block = Block::default().borders(Borders::ALL)...;
let inner = block.inner(area);
...
self.last_area = area;
self.last_inner = inner;
```

and the actual PTY grid is sized from `inner`, at `src/widgets/terminal.rs:4156-4158`:

```rust
let cols = inner.width;
let rows = inner.height;
self.resize(cols, rows);
```

`Borders::ALL` shrinks by one cell on each side, so the real grid is `(width-2) x (height-2)`. A pane painted at 80x24 records a header of `"width":80,"height":24` for a grid that is actually 78x22. The player therefore renders every frame two columns wider and two rows taller than the content was wrapped for — wrong wrap points on any line that reached the right edge, and two blank rows. This is exactly acceptance criterion 2's "no garbled frames".

**Fix:** use `last_inner` in both places:

```rust
let (cols, rows) = (term.last_inner.width, term.last_inner.height);
```

and at `src/app/mod.rs:13252` (`record_active_screen`) likewise, so the `r` events agree with the header.

---

### `src/app/mod.rs:13266` — **[major/correctness+performance]** Each frame writes the whole 5000-line scrollback, not the visible screen

`record_active_screen` does `let (lines, _) = term.grid_lines();` and joins *all* of it. `grid_lines` (`src/widgets/terminal.rs:1356-1371`) walks `topmost_line()..=screen_lines()-1`, and `topmost_line()` is negative scrollback — documented in this repo at `src/widgets/terminal.rs:56`: *"is the live screen, negative values are scrollback history"*, with `SCROLLBACK_LINES: usize = 5000` at line 23.

Two consequences, both real:

1. **Playback is wrong.** After `\x1b[H\x1b[2J` the player's cursor is at home on a 24-row screen. Writing 5024 lines scrolls 5000 of them straight off the top; what remains on screen is the *tail* of the scrollback, and the live screen the user was looking at ends up scrolled past. The recording does not show what was on screen.
2. **File size.** Worst case per frame is `5024 * (80+2)` = **411,968 bytes**. At 10 dirty ticks/second that is ~4.1 MB/s, ~0.25 GB/minute; at 30 ticks/s, ~0.74 GB/min. A one-minute `ls` demo produces a multi-hundred-megabyte cast. The visible screen alone would be ~1,968 bytes/frame — a 200x difference.

**Fix:** take only the live screen. The last `screen_lines()` entries of `grid_lines()`, or better a dedicated accessor that reads rows `0..screen_lines()` (`grid_lines` already returns `top`, so `lines[(-top) as usize..]` is the visible screen when not scrolled). Add `\x1b[H` framing per row rather than relying on scroll.

---

### `src/app/mod.rs:23946` — **[minor/usability]** A second recording in the same process silently destroys the first

The path is `workspace_root()/croft-<pid>.cast`, and `File::create` truncates. Start, stop, start again — the first recording is gone with no warning, which the user discovers when they go to share it.

**Fix (already written in the worktree, needs committing):** the uncommitted edit appends a unix-seconds stamp, `croft-{pid}-{stamp}.cast`. That resolves it. Note the one-second granularity still collides if two recordings start within the same second; a counter or `%H%M%S` plus a `create_new` retry would close that, but the stamp is adequate for v1.

---

### `src/app/mod.rs:13245-13267` — **[minor/style]** The double `self.terminals.get()` is not a bug, but it is load-bearing on a stale read

You asked specifically. On the committed branch the two `get()` calls straddle `record_resize`. It is **not** a correctness bug: `record_resize` only touches `self.recording`, never `self.terminals`, so the second lookup returns the same pane. But it is fragile — the second `get()` re-reads state a `&mut self` call could in principle have changed, and the `else { return }` on the second lookup is unreachable dead code.

The uncommitted worktree edit already collapses it to one borrow via `.map(|t| (…, t.grid_lines().0))`. That is the right shape; commit it.

---

### `src/asciicast.rs:95` — **[nit/robustness]** The clamp is correct; non-finite input degrades to `null`

Ported the recorder to Python and replayed all four committed test assertions through the port before drawing conclusions — all replay green. A 20,000-trial adversarial sweep over random and special values (`±0.0`, `1e-300`, `1e300`, `5e-324`, subnormals, denormals) found **zero** non-monotonic outputs for finite inputs. The clamp holds.

For non-finite inputs it degrades rather than corrupts: `NaN`/`+inf` compare false against `<`, so `self.last` becomes non-finite and `serde_json` 1.0.149 writes `null`, not `NaN` — verified in the vendored source, `serde_json-1.0.149/src/ser.rs:169-180`:

```rust
FpCategory::Nan | FpCategory::Infinite => self.formatter.write_null(...)
```

So the file stays valid JSON, but every subsequent event's timestamp is `null` and playback breaks. This is **unreachable from the app** — `at` always comes from `Instant::elapsed().as_secs_f64()`, which is monotonic and finite — so it is a nit, not a bug. A `if !at.is_finite() { at = self.last }` guard in `line` would make the type honest for any future caller.

**On the f64 text format you asked about:** serde_json 1.0.149 formats via `zmij` (`ser.rs:1716-1723`, a Schubfach shortest-roundtrip formatter), which does emit scientific notation — `1e-7` serialises as `1e-7`. That is valid JSON per RFC 8259, and asciinema's Python/Rust parsers both use standard JSON number parsing, so it is accepted. Not a problem.

---

### Format conformance — **verified against the real spec, no finding**

Checked against <https://docs.asciinema.org/manual/asciicast/v2/> rather than the description in the request:

- Required header fields are exactly **`version`, `width`, `height`** — all three present. **`timestamp` is optional**, so omitting it is spec-conformant (it only costs the player a start-time display).
- `[time, code, data]` with time as "time since the beginning of the recording session (in seconds)", float — matches.
- `"r"` is the resize code, payload `"{COLS}x{ROWS}"`, spec example `[5.0, "r", "100x50"]` — matches `format!("{cols}x{rows}")` exactly.

The serialiser is conformant. The bugs are in what the wiring *feeds* it.

---

### Screen-capture approach — **you asked me to say plainly**

It is a reasonable v1 *shape* but the current implementation produces a cast that looks right in a JSON parser and plays wrong. Specifically:

- **No colour** — confirmed. `grid_lines` → `row_text_and_cols` (`src/widgets/terminal.rs:3648-3664`) reads only `cell.c`, never `cell.fg`/`bg`/`flags` beyond wide-char spacers. A demo GIF of a colour TUI records as monochrome. The module doc at `src/app/mod.rs:13240` names this honestly, which is the right call.
- **Cursor never restored** — confirmed, no `\x1b[<row>;<col>H` after the frame. The playback cursor sits wherever the last written row left it, so it appears at the end of the bottom line rather than at the shell prompt. Visible in any `agg` GIF.
- **Size** — the scrollback bug above, which is the one that actually makes it unusable.

Fix the scrollback and geometry findings and it becomes a defensible v1 whose known limits (no colour, no cursor) are documented. As committed, criterion 1 would not pass a real `asciinema play`.

---

## 3. Test coverage

The one integration test (`src/app/tests.rs:36601`) asserts the file parses, has version/width/height, and is monotonic. It cannot see any of the above because it never checks *content* or compares recorded geometry against the grid. Missing:

1. **Header geometry equals the PTY grid.** Set `last_area` to 80x24, render or set `last_inner` to 78x22, start recording, assert `header["width"] == 78`. This is the test that fails today and catches finding 1.
2. **A frame carries the visible screen, not the scrollback.** Spawn a pane, write 200 lines of known text, record one frame, assert the frame's line count is `<= rows` and that it contains the *last* line printed. Catches finding 2 — and the current test cannot, because it only counts event kinds.
3. **Resize events agree with the header's units.** Record, change the pane rect, drain dirty, assert the `r` payload uses the same inner-geometry convention as the header. Prevents the two drifting apart when finding 1 is fixed in only one of the two places.
4. **A second recording does not clobber the first.** Start/stop twice, assert two `.cast` files exist. Covers finding 3 and pins the fix the worktree already has.
5. **`asciicast.rs`: a non-finite `at` does not poison the stream.** `rec.line(Event::Output { at: f64::NAN, .. })`, assert the emitted timestamp is still a finite number. Would pin the nit above as a contract.

---

## 4. Nitpicks

- `src/asciicast.rs:73` — `header` takes `&self` but `line` takes `&mut self`; a caller can legally call `header` *after* events have been written, producing a second header line mid-file. Not reachable today (one call site), but the type permits it. Consuming a `HeaderWritten` token, or folding the header into `Recorder::new`, makes the ordering structural.
- `src/app/mod.rs:13225` and `:23984` — the two write-failure paths are byte-identical (`self.recording = None;` + the same status string). Worth a small `fn stop_on_write_error(&mut self)` so the message cannot drift between them.
- `src/app/mod.rs:23957` — the `File` is never explicitly flushed or synced; it relies on `Drop` when `recording` is taken. Correct in practice for `std::fs::File` (unbuffered), but if a `BufWriter` is added later for finding 2's throughput, the stop path silently starts losing the tail. Worth a comment or a pre-emptive `flush()` in the stop branch.
- `docs/ARCHITECTURE.md:38` — inserted after `hex.rs`. The list is not alphabetically sorted anyway (`provenance.rs` precedes `prefs.rs` at lines 56-57), so no convention is broken; noting only because "sorted position" was raised as a concern. No action.

---

## 5. Scope and conventions

**Acceptance criteria (#356):**

| Criterion | Status |
|---|---|
| A recorded `ls` session plays back correctly in `asciinema play` and `agg` | **Not met** — findings 1 and 2. Parses, plays wrong. |
| Resizing mid-recording produces a valid `r` event and no garbled frames | **Partially met** — the `r` event is emitted and correctly ordered before the frame, but carries bordered geometry, so frames are garbled. |
| The recorder participant never throttles the pump for live clients | **Not applicable as built** — there is no recorder participant. Sampling happens in `drain_terminals_dirty` on the main loop, and finding 2 makes each tick a ~400 KB `grid_lines()` walk plus a synchronous write on that loop, which *will* be felt in the UI. |

Also unimplemented from the proposed design, consistent with "part of": `croft record` CLI (`--out`, `--stop`), the `● REC` chip distinguishing this from the macro recorder at `src/app/mod.rs:15105`, initial-state emission, and secret redaction.

**Closing keywords:** confirmed safe. The commit subject is `feat(session): record a terminal as an asciicast — part of #356`. "part of" is not a GitHub closing keyword, and no `Closes`/`Fixes`/`Resolves` appears in the commit body or the diff. #356 will not auto-close.

**Repo conventions:** all conform. `Cargo.toml` and `Cargo.lock` both moved 0.1.850 → 0.1.853; `src/release_notes/0.1.853.md` exists with a single `feature:`-prefixed line, matching the format required at `CONTRIBUTING.md:196-206`; `mod asciicast;` is registered in `src/main.rs` in sorted position; `docs/ARCHITECTURE.md` has an entry. Per `CLAUDE.md`, re-read `git show origin/main:Cargo.toml` immediately before merge — main is at 0.1.850 as of this review, but the version-bump race is live on this repo.

---

## 6. Scores

**CodeRabbit: 2/5** — two majors: the header/resize geometry is the bordered rect rather than the grid (`src/app/mod.rs:23930`), and every frame writes the full 5000-line scrollback instead of the visible screen (`src/app/mod.rs:13266`), which breaks playback and produces ~0.25 GB/min casts.

**Greptile: 2/5** — the serialiser is genuinely solid and its clamp survived a 20k-trial adversarial sweep, but the two wiring bugs mean the feature's stated acceptance criterion — that the cast plays back correctly — does not hold, and no test in the diff can detect either.

VERDICT: REQUEST_CHANGES

Issue #366, opened 2026-08-29
PR review comments as comment boxes: read, reply, and submit a review from the editor


vitali87:
## Motivation

The navigator's comment boxes (unnumbered blocks between the lines they belong to, with a reply field and Ignore, `F4` to hop) are exactly the UI a code review needs. Pointing them at GitHub review threads turns croft into a PR review client with the editor's full power.

## Current state

- Comment-box widget in the editor (`src/widgets/editor.rs`, driven from `src/pair.rs`), currently navigator-only.
- The PR review mode issue provides the checkout and file list.

## Proposed design

1. In review mode, existing review threads (`gh api repos/…/pulls/N/comments`) render as comment boxes anchored to their diff-side lines, with author, age, and resolved state; `F4`/`Shift+F4` navigation carries over.
2. **New comment** on a line or selection (`Cmd+K C`) opens an empty box; text goes into a pending review, shown with a `pending` chip.
3. **Submit Review** (approve / request changes / comment) posts the pending set in one API call; reply-in-place on an existing thread posts immediately; **Resolve** flips the thread.
4. Attribution is the user's own `gh auth` identity; nothing is posted without an explicit submit.

## Acceptance criteria

- [ ] Existing threads on a PR show at the right lines, including outdated ones flagged as such.
- [ ] Two pending comments plus "request changes" arrive on GitHub as a single review.
- [ ] Closing croft with pending comments warns and offers to keep them as a draft.

## Related

- Tracking epic: #384
- Depends on: #365

vitali87:
Read side shipped, write side entirely absent.

Shipped: `src/review_threads.rs` (`Thread`, `Anchor::Outdated`, `parse_threads`); `src/app/mod.rs:25192 load_review_threads` runs `gh pr view` then `gh api …/pulls/N/comments --paginate`, filters to the open file and flags outdated anchors; rendered into `editor.comment_boxes` at `:15205`. Palette `Review: Load PR Comments for This File`. Release note `0.1.854.md`.

Missing, with a control so the absence is falsifiable:

```
grep src for: SubmitReview | ReplyToThread | ResolveThread | pending_review  ->  0
control, same file:  LoadReviewThreads                                      ->  5
```

So: no new-comment authoring, no pending review, no Submit Review, no reply-in-place, no Resolve. (`Cmd+K C` is bound to diff-against-anchor at `mod.rs:17662`, not to commenting.)

[...]

Issue #139, opened 2026-08-15
Workspace rooted in a subdirectory of a repo mis-resolves every repo-relative git path


vitali87:
croft never discovers the repository toplevel — there is no `rev-parse --show-toplevel` anywhere in the tree. Every git invocation runs `git -C <workspace_root>` and every repo-relative path from git is joined back against the workspace root. Git resolves the repository upward from any subdirectory, so all of this is correct only in the common case where the workspace root IS the repo toplevel.

Open a workspace that is a subdirectory of a repo (`croft ~/repo/crates/foo`, or Make Root / Cmd+Z into one) and the two path spaces diverge, because `git status --porcelain` and `git diff --numstat` emit paths relative to the repo toplevel:

- `query_changes` joins porcelain paths against the workspace root to count untracked files' lines (src/git.rs, `count_lines_in_file`), reading a nonexistent path — untracked additions silently report 0, or the wrong file's count if the name happens to exist in both places.
- `read_file_at_head` builds `git -C <root> show HEAD:<rel>` where `<rel>` was stripped from the workspace root; a rev pathspec resolves against the toplevel, so the HEAD baseline behind the gutter, inline blame rel paths, and the Source Control HEAD diffs point at the wrong tree entry (typically an empty read — every line marked added).
- Staging paths from the Source Control panel run `git -C <root> add <toplevel-rel>`; command-line pathspecs are cwd-relative, so stage / unstage / discard on entries outside the subdir fail outright (`pathspec did not match any files`).
- Opening a changed entry from the panel joins the toplevel-relative path onto the workspace root and opens a path that does not exist.

Expected: discover the toplevel where repo membership is already probed — the status poll's `rev-parse` — carry it on `GitStatus` (e.g. `repo_root`), and route repo-relative joins and the Source Control operations through it, falling back to the workspace root outside a repo. When the workspace root is the toplevel (the overwhelmingly common case) every path is byte-identical to today's behavior. The Explorer's ignored-set query must stay rooted at the workspace root: `ls-files` output is cwd-relative and cwd-scoped, which is exactly the subtree the Explorer needs.

This is also the prerequisite for multi-root workspaces' per-folder repositories (the multi-root program), where "the repo owning this entry" must be a real discovered identity rather than the workspace root by assumption.

PR #454, opened 2026-08-31
feat(review): show a PR's review threads as comment boxes (part of #366)

vitali87:
Part of #366 — reading a PR's review threads onto the file. **Deliberately not a closing keyword:** nothing here writes to GitHub.

The navigator's comment boxes are already the UI a code review needs. This maps GitHub's review-comment JSON onto them.

## The whole difficulty is that `line` can be null

GitHub anchors a comment to the diff line it was written against, and **nulls that field** once the branch moves under it while keeping `original_line`. The two mean different things — where the comment is *now*, and where it *was*. Rendering an outdated thread at `original_line` silently attaches someone's objection to whatever code now occupies that number, which after a rebase is routinely a different function.

So an outdated thread is anchored but **marked**, and marked in the *title* rather than only in a colour: a reviewer skimming boxes reads titles, and a colour is exactly what a screenshot or a colour-blind reader loses.

## Checking the API rather than my model of it found a phantom field

I wrote `resolved` handling from memory. `gh api repos/…/pulls/N/comments` **carries no field matching `resolv` at all** — resolution lives only on GraphQL's `reviewThreads.isResolved`, a different query keyed on threads rather than comments. My code was reading a field that is never there and silently answering `false`: dead code that looked like a feature.

The same check confirmed two things I had asserted without evidence: an outdated comment really does arrive as `line: null` with `original_line` set, and `position`/`original_position` are **diff offsets, not file lines** — using them would have placed a real comment at line 0 instead of 20.

**The regression test now uses a real captured payload rather than a fixture I wrote.** A hand-made one would have agreed with my model of the API instead of with the API, which is precisely how the phantom field survived.

## A silent-empty bug the path check found

I compared the API's `path` against the file's path relative to the **workspace root**. The API's is repo-relative, and the two coincide only when the root *is* the repository toplevel — so opening croft on `repo/src` made every comparison fail and a file with review comments reported having none. That is #139's documented rule, hit again in a new place; it now resolves against `repo_root`, and joins components with forward slashes since a Windows `to_string_lossy` would yield backslashes that match nothing.

## Smaller decisions, each tested

A thread with neither line becomes a **file-level box** rather than being dropped — a review comment the reviewer never sees is the failure this exists to prevent. Lines convert 1-based to 0-based **once**, because two conversion sites is how an off-by-one reaches the screen. An outdated line past the end of a since-shrunken file is **clamped into view**, since a box at line 4000 of a 100-line file is the same as losing the comment.

## Gate

Mutation-tested, **each script asserting its anchor matched first**: placing an outdated thread silently, letting `original_line` win over `line`, treating an absent `resolved` as resolved, and reading `original_position` instead of `original_line` each fail a different assertion.

I also verified the `gh` invocations against the real CLI rather than trusting them — `gh api` does substitute `{owner}/{repo}`, and `gh pr view --json number --jq .number` returns the number.

4000 pass; the two failures are the machine's environmental pair, which fail identically on unmodified `origin/main`. fmt 0, clippy 0, doc-ownership gate clean. Version 0.1.854.

## Scope

⚠️ *"Existing threads show at the right lines, including outdated ones flagged as such"* — met for reading, but there is no review *mode* yet, so this is a palette command on an open file. ❌ *"Two pending comments plus request-changes arrive as a single review"* — no posting. ❌ *"Closing croft with pending comments warns"* — no pending set.

Still to build: new comments, the pending-review set with its `pending` chip, Submit Review in one call, reply-in-place, Resolve (which needs the GraphQL thread query), and the draft-on-exit warning.


vitali87:

_CodeRabbit's check on this PR is annotated **"Review rate limited"** — a green row meaning no review ran — so this is a fallback review by Claude Opus 5, run locally against the working tree. It is the only review this PR has._

_This is the fourth round; the first returned REQUEST_CHANGES with a blocker (review boxes were never scoped to their file), the second and third each found further defects. Every finding below was verified against the head it names._

**Reviewed head: `64df5cac54eca69559a6e3fbe196ee029bea2eab`**

---

Version gate clean, re-checked against `origin/main` at `bf65f87` immediately before the verdict. I found nothing further.

---

# Review — head `64df5ca`

## Walkthrough

Both documentation findings from the `0e47f7a` pass are fixed, and the rename test was rebuilt to drive the real entry point. Filtering comments out of this commit's `src/app/mod.rs` diff leaves **zero executable lines changed** — this round is doc relocation, comment corrections, and a strictly stronger test. The Python port re-validated against all 38 committed assertions, confirming no behavioural drift.

| File | Summary |
| --- | --- |
| `src/app/mod.rs` | "Dismiss one comment box…" doc moved back above `fn ignore_comment_box`; note-id citation corrected; race comment rewritten to state the invariant honestly. |
| `src/app/tests.rs` | `review_boxes_follow_the_open_file_through_a_rename` now commits a real `PromptKind::Rename` through `commit_prompt()` instead of hand-calling the helpers. |

## Findings

None.

## Previous findings

1. **FIXED** — `src/app/mod.rs:21914-21921` now carries only the rename helper's own doc, and `21930-21932` restores "Dismiss one comment box: drop the note from the pilot's state and from the local snapshot…" directly above `fn ignore_comment_box` at `21933`. Read both blocks in the working tree; each function owns its own doc.
2. **FIXED** — `21944-21948` now cites "`pair.rs`'s `take_note_id`, over `next_note_id`" and says "from 1". Verified both halves against the source: `src/pair.rs:610` initialises `next_note_id: 1`, and `take_note_id` at `659-663` is the allocator. The old citation pointed at `session_host`'s accept-loop connection counter, which is now gone.
3. **FIXED, and better than what I asked for** — `24156-24164` no longer asserts a race that cannot happen. It now states that the two cannot differ today (single-threaded loop, `gh` blocks it), that keying to the queried file is the property that matters, and that this stops being free off the main thread. That is the accurate version.
4. **FIXED — and your mutation claim is confirmed, not taken on trust.** The test builds a real `Prompt { kind: PromptKind::Rename(old), .. }` and calls `commit_prompt()`, which reaches the `PromptKind::Rename` arm at `41491` and thence `rename_review_boxes_path` at `41507`. I traced the assertion chain under both versions: deleting `41507` leaves `prompt.is_none()` and the tab-followed assertion passing (they are guards, satisfied by `rename_open_path` regardless) but flips the final `id == 66` assertion to false. So the mutation is killed by exactly the assertion that should kill it, where the previous hand-calling version survived it. I also confirmed `Prompt` has exactly the five fields the test sets (`2108-2114`), so a future field addition breaks the build rather than silently defaulting — no field-merge blind spot.

**Your correction about #455 is right, and worth having on the record.** I verified it independently: `git show origin/main:src/app/mod.rs | grep -c "fn ignore_comment_box"` returns `1`, so the victim function is not new on this branch and the base-vs-head pass does see a doc go missing. The blind spot needs the victim to be new, which was the `command_palette.rs` case. The gate was not blind here; it simply had not been run yet. That distinction matters — the opposite conclusion would have sent someone to widen a check that was already working.

Both remaining rename call sites are still paired (`41304`/`41305` and `41506`/`41507`).

## Test coverage

No gaps. Every behaviour this branch changes now has a test that dies to its own mutation, including the wiring, which was the last stand-in.

## Nitpicks

None.

## Scores

**CodeRabbit score (diff quality) — 5/5** — no findings in this lens; the commit changes no executable code and both documentation defects are correctly resolved.

**Greptile score (codebase fit) — 5/5** — doc ownership restored to the repo's one-doc-per-item convention, the id-space citation now points at the real allocator, and the rename test drives the production entry point rather than standing in for it.

VERDICT: APPROVE

---

Approving head **`64df5ca`** — that is the SHA this artifact covers, and the one to merge. Two notes for the merge itself, both from your own CLAUDE.md rather than from the diff:

- **Re-read the version in the same breath as the merge.** I checked at `origin/main` = `bf65f87`: branch `0.1.854`, main `0.1.853`, `src/release_notes/0.1.854.md` present. That was minutes ago, and the release gate cannot catch a collision by design — it compares against the merge base, so two branches can both pass and only the second conflicts. If main moves before you merge, bump again.
- **This artifact is the only review this PR has.** It names `64df5ca`; if anything is pushed after it, the review is stale and does not carry over, however small the change looks — that was true of the doc capture at `0e47f7a`, which was a one-line move that a stale approval would have waved through.

One thing I'd flag for the follow-up rather than this PR: `load_review_threads` blocks the single-threaded event loop on two sequential `gh` subprocesses. That matches the existing `fetch_all` precedent, so it is not a finding against this diff, but the comment at `24156-24164` now correctly identifies it as the assumption holding the keying invariant together — worth revisiting if this ever moves off the main thread.

PR #459, opened 2026-08-31
test(markdown): pin the fence delimiter instead of accepting either

vitali87:
An assertion that cannot fail for the reason it names.

```rust
out.contains("~~~text") || out.contains("~~~~text")
```

`"~~~text"` is a substring of `"~~~~text"`, so the first arm is true whenever the second is. The OR is strictly weaker than its first arm alone, and it accepts the four-tilde fence its own message calls wrong.

Verified by execution rather than by reading:

| output | `contains("~~~text")` | `contains("~~~~text")` | OR |
|---|---|---|---|
| `~~~text` (correct) | true | false | **passes** |
| `~~~~text` (wrong delimiter) | true | true | **passes** |

So "the output fence uses the block's own delimiter" — the one property the message claims — is the one property it could not check.

## The fix

`replace_output_fence` emits the block's own opener plus `text`, a single determinate value, so an exact line match is available and the OR was never needed. The assertion now pins that value and separately rejects the four-tilde form.

**Verified to discriminate**: green on the current code, red against a mutant emitting `~text` in place of `text` — which the old assertion accepted. The mutation asserted its anchor matched before the result was read, so a non-applying edit could not read as a survivor.

## How it was found

Sweeping for `contains("A") || contains("B")` where one literal contains the other, over joined line **pairs** as well as single lines — rustfmt wraps these, and this instance appears only in the wrapped pass, so a single-line regex reports a clean sweep on a codebase that has it. The sweep was run against a known-positive first, since a regex matching nothing and a codebase with no instances print the same result.

Credit to the session working #456, which hit the same shape in `frame.contains("line-199") || frame.contains("line-19")`. That instance is in a file #456 touches and is left to that PR; it independently re-ran this sweep across its three branches and confirmed no others.

This defect class is worth a standing sweep because it is decidable from the text alone — if one literal contains the other, the OR is weaker than its first arm, unconditionally, with no reasoning about what the test reaches.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features`, and `scripts/check_doc_ownership.py` all clean on the committed head (exit codes taken directly, and the doc gate run on a clean tree since it reads committed state). Full suite 4027 passed; the 3 failures are two environmental on this machine (missing musl cross-targets; a `sun_path` length limit from the long scratch worktree path) plus one shell-spawning load flake, all three confirmed passing serially.

## No version bump

The release gate exempts this change on its own terms. `scripts/ships_nothing.py` (added by #461) parses each changed `.rs` file and drops one whose diff is confined to a `#[cfg(test)]` module, so the gate's `[ -z "$shipped" ]` early exit fires and no bump or release note is required.

Confirmed by running the gate's own logic against this head rather than inferring it:

```
src/markdown.rs   -> exit 0, ships nothing
src/app/mod.rs    -> exit 1, shipped        (control, same run)
```

The control matters: the script imports a sibling from `scripts/`, and run from anywhere else it dies with `ModuleNotFoundError` and **exits 1, which the gate reads as "shipped"**. A crash and a real "this ships" verdict are indistinguishable by exit code alone, so a control that can still say `1` is what makes the subject's `0` mean something.

An earlier revision of this branch carried a bump to 0.1.860 and a release note, written when the gate's exemption was still a path list that could not see inside a file. That is now the wrong shape twice over: the content check makes it unnecessary, and `Cargo.toml`/`Cargo.lock` are not `.rs` files, so `ships_nothing.py` never sees them and they count as shipped on the path list alone — a bump is self-justifying. Those commits are reverted; the net diff against main is exactly `src/markdown.rs`.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `scripts/check_doc_ownership.py` all exit 0 on the committed head (exit codes taken directly, never through a pipe; the doc gate run on a clean tree since it reads committed state).

Full suite: 4056 passed, 6 failed. Two are environmental on this machine and fail identically on unmodified main (missing musl cross-targets; a `sun_path` length limit from the long scratch worktree path). The other four are shell-spawning load flakes, each confirmed passing serially with `1 passed`, alongside a deliberately non-matching filter as a negative control — `test result: ok. 0 passed; 0 failed` means the filter selected nothing, not that a test passed, and the first serial run reported exactly that for every test because this crate has no lib target and the filter was hitting the wrong one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved output formatting validation to ensure tilde-fenced text uses the exact `~~~text` format.
  * Prevented acceptance of incorrectly extended fence markers such as `~~~~text`.


vitali87:
## 🤖 Fallback review — Claude Opus 5 (greptile-local)

CodeRabbit is rate limited on this PR: its check row reports `pass`, and the comment it posted says "Review limit reached" — no review ran. This is that review instead. It names head `43efdf7`.

### Summary

Tightens one assertion in `a_tilde_opened_block_takes_an_output_fence` from a tautological `contains(A) || contains(B)` (where A is a substring of B) to an exact whole-line match plus an explicit negative. The stated substring reasoning is correct and the hole was reproduced: with the renderer mutated to emit a four-tilde fence, the old assertion still passed while the new one fails. Net diff against main is exactly `src/markdown.rs`, version identical to main, and the release gate confirms the change ships nothing.

### Confidence Score: 5/5

No P0/P1/P2 findings. The change strictly increases the test's power on a fixed-input path; gate, lint and format all pass with controls.

**Files needing attention:** none
**Findings:** none

### Checked, not a defect

- **Substring premise — confirmed.** `"~~~~text".contains("~~~text") == true`, so the old OR was equivalent to arm 1. Run against a four-tilde document it returned `true` — it accepted the exact wrong-delimiter case its own failure message claimed to reject.
- **Mutation control — decisive.** `src/markdown.rs:271` `…max(2) + 1` → `+ 2` was mutated so the renderer emits `~~~~text`, with the anchor asserted to match exactly once so the mutation provably applied. Same mutant, two assertion forms, same filter: new form → `FAILED. 0 passed; 1 failed`; old form → `ok. 1 passed; 0 failed`. The old assertion was untestable; the new one has teeth.
- **New assertion is not too strict.** Fence width is a pure function of the inputs: `repeat_n(delim, longest.max(width-1).max(2) + 1)`. For this test's fixed `doc` (width 3) and output `"FIRSTRUN\n"` (no tildes, `longest = 0`) it evaluates to `0.max(2).max(2)+1 = 3` → exactly `~~~`. Openers `~~~`/`~~~~`/`~~~~~`/backtick were swept against outputs containing `~~~`, `~~~~~~` and backticks; wider and backtick fences are reachable in general but not from this test's hardcoded input. There is no legitimate output the old form accepted and the new one rejects.
- **Negative assertion is neither vacuous nor dead.** `lines()` of a real four-tilde render yields a bare `"~~~~text"` element, so the predicate is reachable — demonstrated by the mutant run, where it was the arm that fired. It is not implied by the positive arm either: a document containing both a `~~~text` and a `~~~~text` line satisfies the positive and fails the negative. A five-tilde regression is still caught by the positive arm.
- **CRLF.** `str::lines()` strips the `\r`, so the exact `==` is not brittle against the renderer's CRLF branch (`src/markdown.rs:275`).
- **Merge and reverted bump net out to nothing — verified, not assumed.** Two- and three-dot `git diff --name-status origin/main...HEAD` both list only `M src/markdown.rs`; `Cargo.toml` version is `0.1.861` on both main and HEAD; no release-notes files added.
- **Release gate.** `python3 scripts/ships_nothing.py origin/main HEAD src/markdown.rs` → exit 0, run from the repo root so its `check_doc_ownership` sibling import resolves. Two positive controls in the same session (`47878d8~1..47878d8 src/markdown.rs`, `641b27f~1..641b27f Cargo.toml`) → exit 1 each. A crash-exits-1 cannot explain the subject's 0, and the controls prove the script could still have said "ships". No bump or notes required.
- **Conventions.** The inserted comment sits inside a function body, so it cannot capture a following item's doc; `check_doc_ownership.py` reports "No documentation lost or stranded". Comment density (~47 similar `//` blocks in the tests module) and em-dash usage are house style across `src/`.

### Verification run

- Tests: filtered only, per the host rule — `markdown::tests::a_tilde` → `2 passed; 0 failed`, filter confirmed matched (non-zero passed count, `4070 filtered out`). Full suite deliberately not run by the reviewer.
- `cargo clippy --all-targets -D warnings`: exit 0 (exit code load-bearing, not grepped).
- `cargo fmt --check`: exit 0.
- Scratch mutant copies removed; worktree left clean, no tracked files, index or branches modified.

One process note from the reviewer, kept because it is the same defect class this PR is about: its first control attempt mangled its own arguments via `set --`; it discarded that run and re-ran quoted rather than counting the exit code.

VERDICT: APPROVE

PR #476, opened 2026-09-05
perf(zoxide): make the Cmd+Z jump snappy

vitali87:
## Summary

The Cmd+Z zoxide jump felt sluggish. Driving the release binary through a pty with per-stage timing logs found two separate costs, both on the UI thread:

| Stage | Before | After |
|---|---|---|
| Each keystroke in the popup | 10–37 ms (two spawns: `zoxide --version` probe + `zoxide query`) | ≤ 0.3 ms (no spawn) |
| Enter, jumping out of a workspace with rust-analyzer alive | 2456–2820 ms (blocked on the old `LspManager`'s Drop) | 12–14 ms |
| Popup open (cold / warm) | 35 ms / 37 ms (two spawns) | 39 ms / 17 ms (one spawn once the binary path is cached) |

Measured on an idle Mac; under load the spawn and shutdown costs only grow.

## Changes

* **`change_workspace_root` retires the old LSP manager off the UI thread.** Its `Drop` blocks on a graceful shutdown of every server, capped at 3 s, and rust-analyzer routinely uses ~2.5 s of it. The manager is now handed to `fs_watch::offload_drop` (the same treatment the FSEvents watcher already gets in `rebind`), so the shutdown still happens but the event loop never waits for it.
* **Keystrokes filter the open-time snapshot in-process.** `zoxide::strict_filter` replicates `zoxide query`'s keyword rule (`Database::filter_by_keywords`: case-insensitive substrings matched right to left, each strictly before the previous match, last keyword confined to the final path component) over the `query -l` snapshot the popup already captured at open. Filtering a frecency-ordered list in order preserves zoxide's ranking, so results are unchanged. The typo-tolerant `fuzzy_rank` fallback is untouched. The only remaining per-keystroke spawn is the unavailable path (zoxide missing at open, so a just-finished background install has to be picked up).
* **`zoxide::binary()` caches a successful resolution** for the process lifetime; a miss is not cached so the background installer's copy is still found later.

## Tests

* `strict_filter_*` (3): final-component rule, ordered earlier keywords with a paired positive control, case-insensitivity + frecency order + blank needle.
* `zoxide_jump_keystrokes_filter_the_open_time_snapshot_in_process`: seeds the popup with paths that do not exist on disk, so only in-process filtering can return them (it failed red on the old code by returning the real database). Also covers Backspace re-widening and the approximate flag on a typo.
* The LSP-offload change is verified by the pty-driven timing above rather than a unit test: `LspManager` is concrete and its Drop is what is being moved, so there is no seam to fake without widening the change.

Version bumped to 0.1.862 with release notes.


vitali87:
This is snappy enough and wrongly openned

PR #486, opened 2026-09-06
feat(diff): rebuild an open diff view when its file, HEAD or index moves (#471)

### LINKED ISSUE #471: Code update MUST automatically update the git diff view in real time

vitali87:
Closes #471.

## What was wrong

A diff view was a snapshot taken at open. While an agent (or a save from a sibling tab) rewrote the working file, the Source Control HEAD-vs-working view kept showing the rows from when it was opened; the only rebuild anywhere was after a revert-hunk. A staged, branch or previous-commit diff never re-ran either.

## What changes

- **Every diff view records its source.** `DiffData` gains a `DiffSource`: `HeadVsWorking { root, rel }` for the Source Control view, `FixedLeft` for a TIMELINE snapshot, `TwoFiles` for Compare, `GitCommand` for the staged / branch / previous-commit views, and `Static` for the deletion tombstone. It also stamps `(mtime, len)` of its file-backed sides.
- **Views are rebuilt in place.** `App::refresh_open_diff_views` sweeps every split group's diff tabs and rebuilds each through `rebuild_diff_view`. Two triggers, two costs: a filesystem event or poll re-reads only file-backed sides whose stamp moved (an idle tick is one stat per open diff); the git worker's status drain, which is already debounced, also re-reads the HEAD blob behind a Source Control view and re-runs the `git diff` behind a command-backed view.
- **The reader keeps their place.** `carry_view_from` transfers whitespace mode, scroll (clamped), horizontal pan, hunk anchor and find state. Identical content is left untouched, so a no-op write does not reset selection or find.
- The revert-hunk rebuild now goes through the same path.

## Tests

- HEAD-vs-working: grows on a file write with scroll kept and stamps taken (no second rebuild), then both sides agree after a commit on the git trigger only.
- Staged view re-runs on the git trigger and not on a filesystem tick; identical output changes nothing.
- Snapshot and two-file views follow their files; `left_is_git_head` / `left_is_real_file` survive the rebuild.
- A diff parked in an inactive split group is refreshed too.

Version 0.1.866 with a release note; ARCHITECTURE.md describes the mechanism.

## Verification

- Targeted run: 125 diff-related tests pass, including the five new ones. Full suite before the review fixes: 4065 passed; the 3 failures (session-host socket EINVAL, missing rustup cross-targets, a real-shell copy-paste probe) are environmental and pass alone or fail identically on the base.
- Local reviewer ran once at 3/5; all three findings are addressed: `same_content_as` now compares rows and the byte facts (CRLF, final newline, the bytes-differ banner) so a newline-only rewrite refreshes what the hunk patches read (new test `a_rewrite_that_only_changes_bytes_still_refreshes_the_byte_facts`); the HEAD blob is re-read only for roots whose HEAD oid moved in that git drain (`heads_moved`), not on every status refresh; the misplaced doc comment is back on `set_whitespace_mode`.

## Summary by CodeRabbit

- **New Features**
  - Open diff views now refresh automatically when files change or Git status updates.
  - Source Control, Timeline, staged, branch, previous-commit, and two-file diffs update in place while preserving scroll position and other view state.
  - Diff results refresh even when file content changes without a timestamp or size change.

- **Documentation**
  - Documented live refresh behavior for open diff views.

- **Chores**
  - Bumped the package version to 0.1.876.


vitali87:
Pushed 88a2b04 for the three findings: the watcher's changed-path list now bypasses the `(mtime, len)` gate for the sides it names (the stamp stays as the idle-poll fallback, and the same-length preserved-mtime case has a test on both paths); `carry_view_from` keeps only the needle and options and the app recomputes the match set against the new rows when a find bar is open; the git worker records that a status refresh arrived, changed or not, and the drain re-runs staged / branch / previous-commit views on every arrival while the HEAD blob stays gated on the root's oid. Tests cover all three, including a re-staged file reaching an open staged view through the real worker drain.

vitali87:

Ran because neither bot reviewed this head (603f825): CodeRabbit is at its OSS review limit and Greptile reviewed only the PR's first head. This is the fourth pass; the earlier passes' two majors (workspace-root vs toplevel keying of the HEAD refresh; the revert force by a re-read `scm_root()`) and the third pass's major (hunk anchor carried across changed rows) are fixed in the commits since. Minors below are left as judgment calls: the per-path `canonicalize` is bounded by real source writes and skipped entirely when the watcher names nothing; the pointer identity is compared, never dereferenced, inside a loop that does not mutate the vector; first-match recompute for a background view is an improvement over dropping the band and the active tab still gets the full recompute.

# CodeRabbit / Greptile fallback review — PR #486 @ `603f825`

## Walkthrough

Every `DiffData` now records a `DiffSource` (`HeadVsWorking{root,rel}`, `FixedLeft`, `TwoFiles`, `GitCommand{root,kind}`, `Static`) plus `(mtime,len)` stamps of its file-backed sides, so `App::refresh_open_diff_views` can sweep every split group's diff tabs and rebuild each in place through `rebuild_diff_view`. Two triggers with two costs: an FS event or poll re-reads only file-backed sides whose stamp moved (or whose path the watcher explicitly named, bypassing the stamp gate), while the git worker's debounced status drain also re-runs `git diff` behind command-backed views and re-reads the HEAD blob for roots whose HEAD oid moved. `carry_view_from` transfers whitespace mode, clamped scroll, pan, the needle, and — only when `self.rows == old.rows` — the hunk anchor, recomputing the active find match against the new rows; the revert-hunk path now routes through the same machinery instead of the deleted `rebuild_open_scm_diff`.

| File | Summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | Version 0.1.870 → 0.1.876. |
| `docs/ARCHITECTURE.md` | Extends the `diff.rs` entry with the #471 live-refresh mechanism. |
| `src/app/git_worker.rs` | Adds `status_arrived` + `take_status_arrival()` so an *unchanged* status still counts as an arrival; one new unit test and a `for_test_with_responses` fixture. |
| `src/app/mod.rs` | New `refresh_open_diff_views` / `tag_open_diff` / free fn `rebuild_diff_view`; call sites in `poll_filesystem_changes`, `drain_fs_events`, `drain_git_responses` (with `heads_moved` keyed on `status.repo_root`), and `confirm_pending_revert_hunk`; deletes `rebuild_open_scm_diff`; tags the five openers. |
| `src/app/tests.rs` | 11 new integration tests covering each source, the HEAD-moved gating, byte facts, watcher bypass, inactive splits, find recompute, the anchor drop, and the background find band. |
| `src/release_notes/0.1.876.md` | New release note for the feature. |
| `src/widgets/diff.rs` | `DiffSource` / `GitDiffKind` enums, `source`/`left_stamp`/`right_stamp` fields, and `carry_view_from` / `stamp_sides` / `sides_moved_on_disk` / `same_content_as`. |
| `src/widgets/editor.rs` | `open_diff` tags `TwoFiles` and stamps at open. |

## Findings

Nothing at blocker or major severity. The three newest fixes verify correct:

- **`heads_moved` keying** — `src/app/mod.rs:6063` records `status.repo_root` (git's `--show-toplevel`) with `ws_root` as fallback, and `rebuild_diff_view` compares `r == root || canon(r) == canon(root)`. This matches what `scm_root()` at `src/app/mod.rs:6620-6627` stamps into the tag, so a subdirectory workspace or a symlinked root now agrees on both sides. The empty-`heads_moved` fast path means no `canonicalize` syscall on the common drain.
- **Revert forced by the view's own root** — `src/app/mod.rs:24140-24155` reads `DiffSource::HeadVsWorking{root,..}` off the live view before falling back to `scm_root()`. `refresh_after_hunk_op()` is non-blocking, so `self.editor.diff` is still the tagged view when it is read on the next line.
- **`nav_anchor` drop** — `src/widgets/diff.rs:1087-1089`. `action_row()` is `nav_anchor.unwrap_or(scroll)` and the sole input to the S/U/R patch builder; `hunk_range_at` does walk forward from a context row to the next change, so a stale index really could have selected a hunk the reader never saw. Since `carry_view_from` runs only after `same_content_as` returned false, and that already requires `rows` equality, the surviving case is precisely "rows identical, byte facts changed", where the index still names the same hunk. Correct and complete.
- **Background find band** — `src/widgets/diff.rs:1096-1105` computes the match from `self.rows`, index-aligned with `display_rows`, so the band lands where the renderer paints. The app-side recompute is correctly gated on `active_changed`, and `std::iter::once(&mut self.editor)` cannot alias `editor_layout.inactive_groups_mut()`.

Remaining items, all minor:

- `src/app/mod.rs:24209-24212` **[minor/performance]** `touched` canonicalizes every watcher-reported path before checking whether any diff view is open. Bounded by real source writes (noise dirs are filtered), but a formatter rewriting hundreds of files costs one syscall each with zero diff tabs open. An early bail when no tab holds a diff, or a lazily built `touched`, would avoid it.
- `src/app/mod.rs:24217` + `:24242` **[minor/style]** identifying the active tab by raw pointer when `EditorTabs::active_index()` exists. Sound as written (compared, never dereferenced, no `Vec` mutation in the loop), but an index comparison reads the same and cannot rot.
- `src/widgets/diff.rs:1096-1100` **[minor/correctness]** the recomputed background match is always `.first()`, so a reader parked on match 5 of 20 in a background split is moved to match 1 on a rebuild. Still an improvement over dropping the band; the nearest match to the old row would preserve the place.
- `src/app/mod.rs:24259` **[minor/performance]** for the active tab, `find_matches` runs twice per rebuild (once in `carry_view_from`, again in `diff_find_apply`). Harmless at typical sizes.
- `src/app/mod.rs:34230` (via `diff_find_apply`) **[minor/correctness]** `scroll_to_match` unconditionally clears `nav_anchor`, so on the active tab with a find bar open the one case `carry_view_from` preserves the anchor for loses it anyway. Narrow.

Repo conventions verified: `Cargo.toml` bumped to 0.1.876 and `src/release_notes/0.1.876.md` added; `python3 scripts/check_doc_ownership.py origin/main HEAD` reports nothing lost or stranded across 5 changed Rust files. No CI/workflow files touched.

## Test coverage

- The anchor-survives half of the `nav_anchor` fix is untested: replacing the `self.rows == old.rows` conjunct with `false` would still pass. A test on the byte-facts fixture (rows identical, `right_no_final_nl` flips) asserting the anchor is unchanged would pin it.
- The revert-hunk success path (the `Ok` arm of `confirm_pending_revert_hunk`, including the new `force_root` extraction) has no test; this was already true of the deleted `rebuild_open_scm_diff`.
- The deletion tombstone's `Static` classification is asserted nowhere.

## Nitpicks

- `src/widgets/diff.rs:1074` — one over-long doc line; re-wrap.
- `src/app/mod.rs:24196` — `refresh_open_diff_views` is `pub` with no callers outside `App`.
- The `&std::collections::BTreeSet::new()` literal is spelled out at three call sites.
- `right_is_file` is duplicated between `stamp_sides` and `sides_moved_on_disk`.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: no correctness defects inside the new lines; the three newest fixes hold up against the code they protect. Only minors and the untested positive branch of the anchor condition.
- **Greptile score (codebase fit) — 4/5**: version bump and release note satisfy CONTRIBUTING, the doc-ownership gate passes, no caller of the deleted function remains, all five openers are tagged and the two intentionally `Static` ones are correctly left alone, and the `scm_root()`/`repo_root` agreement is verified end to end. Held off 5 by the eager `canonicalize` and the uncovered revert-hunk success path.

VERDICT: APPROVE

PR #487, opened 2026-09-06
feat(log): highlight rendered logs with tailspin (#466)

### LINKED ISSUE #466: Highlight log outputs with tailspin
https://github.com/bensadeh/tailspin

vitali87:
Closes #466.

## What changes

- **Rendered logs are highlighted with tailspin.** The `tailspin` crate (7.0, library only, `default-features = false`) is added, and the log view's window refill runs each plain line through it before the ANSI parse. Dates, durations, numbers, UUIDs, IPs, URLs, emails, paths, key=value pairs, quotes and severity keywords come out as ordinary SGR spans through the same palette path a self-coloured log already uses.
- **Keywords included.** The library's default highlighter omits the CLI's built-in keywords, so the highlighter is built explicitly with ERROR/FATAL/PANIC (red bold), WARN (yellow bold), INFO (green), DEBUG/TRACE (magenta), booleans/null (cyan) and HTTP methods (blue) added first, ahead of the standard groups, since the group added first wins on overlap.
- **A line that already carries colour is left alone.** Any line with an escape byte, or one still inside a colour block an earlier line opened, is parsed as written, so a logger's own colours are never overpainted and a red stack trace stays red.
- **Toggle and preference.** "Log: Toggle Highlighting (tailspin)" in the palette and a Settings row flip every open log view in every split group, steer views opened later, and persist as `disable_log_highlight` in `config.json` (stored as a disable flag so older configs mean "on"; allowed in the workspace layer like the other display toggles). Flipping drops the parsed window only; index, selection and find never read colours.
- One `Highlighter` per thread, built once, since building compiles its regex set.

## Tests

- A plain line is coloured, its ERROR keyword is coloured, and its visible text is unchanged; the same line is unstyled with the flag off.
- A self-coloured line equals its own parse.
- A plain line inside an inherited colour block keeps the block colour.
- The process-wide default steers new views only.
- The app toggle flips the open view, the default for later opens, and the status line.

## Verification

- 138 log-related tests pass through the host suite lock, including the five new ones.
- Local reviewer ran once at 2/5; every finding is addressed: crate pinned to the published 7.0 (7.1 exists only on tailspin's main), `Cargo.lock` updated for `--locked`, `apply`'s `Cow` return used directly, keyword highlighters added so the docs are true, the default-flip test race removed by setting the flag explicitly, the inherited-colour case gated on the carried style, and `cargo fmt` applied.

Version 0.1.868 with a release note; SETTINGS.md, KEYBINDINGS.md and ARCHITECTURE.md updated.


vitali87:
Pushed de0970a: `App::new` now seeds `log_view::set_default_highlight` from the saved preference before any log can open (new test `startup_seeds_the_log_view_default_from_the_saved_preference`), and the tests that read or flip the process-wide default take a shared test-only mutex, so a parallel toggle cannot land inside another test's snapshot/open window.

vitali87:

Ran because neither bot reviewed this head (0f16079): CodeRabbit is at its OSS review limit and Greptile reviewed only the PR's first head. This is the fourth pass; the earlier passes' majors (unresolvable crate version, stale lockfile, `Cow` type error, docs promising keyword colouring the default highlighter omits, the workspace key never applied on remerge, a tautological startup test) are fixed in the commits since. The one minor below (a comment mentioning a startup remerge that does not exist) is left as a judgment call: the guard it explains is still correct for the reason the comment's second clause gives.

## Walkthrough

The PR adds the `tailspin` crate (7.0, `default-features = false`) as a library and runs each *plain* line of the rendered log view's parsed window through it before the ANSI parse, so dates, numbers, UUIDs, IPs, URLs, paths, quotes and severity keywords come out as ordinary SGR spans through the existing palette path. A line that carries its own escape bytes, or one still inside a colour block an earlier line opened, is parsed as written, so a logger's own colours are never overpainted. The setting is a per-view flag seeded from a process-wide default, driven by a new `disable_log_highlight` preference, a palette command and a Settings row, and applied live on a workspace-layer settings remerge.

This fourth pass reviews the resolution of round three's major: routing was deliberately left alone (an uncoloured `.log` still opens as an editable text file) and the scope is now stated in the docs, the release note and ARCHITECTURE.

| File | Summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | Version 0.1.876 → 0.1.877; adds `tailspin = { version = "7.0", default-features = false }`; lock records the CLI-free dep set. |
| `src/log_view.rs` | `build_highlighter` (keywords first, foreground-only), a `thread_local` `HIGHLIGHTER`, `MAX_HIGHLIGHT_BYTES` width cap, `highlighted()`, the per-view `highlight` flag + `set_highlight`/`highlight`, the `DEFAULT_HIGHLIGHT` atomic and its test lock, the `ensure` two-part guard, and five new tests. |
| `src/app/mod.rs` | `log_highlight` field, `toggle_log_highlight`, shared `set_log_highlight` fan-out over all split groups, remerge application, and the startup seed with a `#[cfg(test)]` per-thread recording shim. |
| `src/app/tests.rs` | Three tests: the toggle's fan-out and status, the startup seed pinned by a workspace layer in both directions, and a workspace-layer remerge reaching a log parked in another split. |
| `src/prefs.rs` | `disable_log_highlight` (`#[serde(default)]`, stored inverted) and `save_disable_log_highlight` via `prefs_for_update`. |
| `src/config_layers.rs` | Adds `disable_log_highlight` to `WORKSPACE_ALLOWED_KEYS`. |
| `src/widgets/command_palette.rs` | `Command::ToggleLogHighlight` in the enum, `ALL_COMMANDS`, label, keys and id fns. |
| `docs/{ARCHITECTURE,KEYBINDINGS,SETTINGS}.md`, `src/release_notes/0.1.877.md` | Document the mechanism, the honest routing scope, the allowlist key and the release note. |

## Findings

**On round three's major — accepted as resolved.** `Editor::should_open_as_log` requires `looks_like_ansi` over the first 8 KiB, and `only_files_carrying_sgr_sequences_route_to_the_log_view` asserts *"a log without colours is an ordinary editable file"*. The rendered view is read-only, so widening the routing would silently make plain `.log` files uneditable: a product change, not a highlighting change. The scope is now stated in three places, each naming widening as a separate decision. **Not re-raised.**

**Merge with main verified clean.** The diff against main for `src/app/mod.rs` and `src/app/tests.rs` is pure additions on both sides of the conflict; main's `git_worker` and `refresh_open_diff_views` work survives; no conflict markers remain; `check_doc_ownership.py` reports nothing lost or stranded across 6 changed Rust files.

Findings at major or above: **none.**

- `src/app/mod.rs:38525-38528` **[minor/docs]** The comment justifying the `want_highlight != self.log_highlight` guard says "the startup remerge runs before any log is open", but `apply_merged_settings` is reached only from a config-file save and from a workspace re-root, never from `App::new`. The guard is still correct and needed for its second clause (writing the process-wide default under test would race the tests that read it); only the first clause is false.

Verified rather than trusted: the background-padding constraint is real in tailspin's `KeywordFinder`/`render` and every keyword style here sets `fg` and `bold` only; keyword precedence is first-added-wins per byte; the width cap's per-byte allocation claim matches `merge_spans`; every highlighted line ends in a reset so the carried-style guard cannot be corrupted; find/copy read the raw sweep path and cannot diverge from the display; the cache short-circuit cannot fire on a dropped window; highlighting is per visible line, not per window; all six `Command` registration sites are covered; the version bump and single release-note file satisfy CONTRIBUTING; the prefs field follows the `disable_secret_redaction` precedent; `default-features = false` keeps clap/rayon/signal-hook/toml out of the graph.

## Test coverage

- The agreement between the cached (highlighted) text and the uncached sweep path on a highlighted line is not asserted directly; a test comparing `visible_text` with `selection_text` over one highlighted line would pin the invariant the `build_highlighter` doc protects.
- No spot check that `ALL_COMMANDS` contains `ToggleLogHighlight` with id `toggle_log_highlight`; optional by convention.

## Nitpicks

- `docs/ARCHITECTURE.md:36` — stray comma before the parenthetical.
- `src/log_view.rs:73` and `:604` — `--` where the file otherwise uses an em dash.
- The plain-lines test clones two lines in full to compare `.text`; `visible_text(0)` reads the same value.

## Scores

- **CodeRabbit score (diff quality) — 5/5.** No finding's evidence lives inside the changed code: the guards are each correct against tailspin's actual behaviour, the cache short-circuit is provably safe after a clear, the keyword styles honour the documented no-background constraint, and the new tests pin the width cap on both sides.
- **Greptile score (codebase fit) — 4/5.** One minor (a comment asserting a startup remerge that does not exist) plus nits; everything cross-file lands correctly, including a merge with main that is pure additions on both conflicted files.

VERDICT: APPROVE

PR #489, opened 2026-09-06
test(app): budget the probe and task-pane waits, silence the quick-select pane (part of #397)

vitali87:
Part of #397; that issue stays open for the members without a captured failure.

Three members of the flaky-under-load class, each with a captured full-suite failure and each a different shape.

## The probe helper (six end-to-end tests)

`await_terminal_probe` waited a fixed 15 s. A full parallel run captured the probe command echoed on screen and still unanswered at 15 s, which is a starved shell, not a broken one. The helper now waits `spawn_budget(TERMINAL_PROBE_BASE)`: 15 s at the floor, up to twice that under load, and the panic still prints the screen. This covers `terminal_copy_paste_into_other_panes_works_in_split_and_maximised_layout` and `terminal_cmd_c_copies_even_when_search_sidebar_is_open`, both on the issue's list.

## The task-pane tests (three tests)

They waited up to 5000 ms for the pane's shell to return to its prompt and broke out of the loop silently on timeout. Under load the rerun then met a busy pane, opened a second one, and the count assertion reported "3 panes, expected 2" as if reuse were broken, which is the panic the issue records. They now go through `await_spawned(TASK_PANE_PROMPT_BASE)`: 5 s at the floor, and a timeout fails loudly naming what it waited for.

## The quick-select streaming test

It had row slack against a real shell's wrapped prompt. The captured failure showed the pane holding only that prompt and the trailing `x1 x2 x3` with the URL gone: the shell's prompt repaint after the maximize's SIGWINCH rewrote the rows between the feeds. Slack absorbs a prompt that appends, not one that repaints. The pane is now a silent `/bin/sh -c 'sleep 30'`, the same pattern the other silent-pane tests use; the subject is croft's label overlay following streamed content, which needs a grid, not a shell. The slack assertion stays because the scroll still needs the rows.

## Bases

Both bases live in `test_budget::tests` beside `RESTORED_SHELL_BASE`, with a pin test asserting each reproduces the constant it replaced at the floor, so lowering either fails there rather than at a call site.

## Not in this PR

The remaining fixed waits in `src/app/tests.rs` (about forty `waited < N` / `from_secs(15)` deadlines) are the same shape but have no captured failure; converting them blind is what this issue's history warns against. Test-only, no version bump.

## Verification

- The 20 affected tests (probe users, task-pane tests, quick-select, `test_budget`) pass through the host suite lock.
- The quick-select and three task-pane tests pass 15 consecutive runs through the lock.
- Local reviewer ran once; see the PR thread for anything it raised.

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -22,29 +22,37 @@ fn probe_command(probe: &str) -> String {
     format!("printf '%s%s\\n' '{head}' '{tail}'\r")
 }
 
-/// How long the terminal end-to-end tests wait for their probe to come back
-/// through the PTY. Generous on purpose: a real shell has to start, run the
-/// command, and have the reader thread drain the bytes into the grid, and on
-/// a box running the whole suite in parallel that took longer than the three
-/// seconds this used to allow (issue #226).
-const PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(15);
-
 /// Run the probe command in the app's embedded terminal and wait until the
 /// probe is on screen. Panics with the actual screen contents if it never
 /// arrives, so a failure says what the terminal was showing instead of
 /// unwrapping `None` several lines later.
+///
+/// The wait is load-scaled (#397): a real shell has to start, run the
+/// command, and have the reader thread drain the bytes into the grid, and
+/// what stretches that is contention from the rest of the suite, not this
+/// test. The fixed 15s this replaced (#226 had raised it from 3s) is what a
+/// quiet machine still gets at the floor; the captured failure had the probe
+/// command echoed on screen and unanswered at 15s, a starved shell rather
+/// than a broken one. `spawn_budget` rather than `await_spawned` so the
+/// panic can keep showing the screen.
 fn await_terminal_probe(app: &mut App, probe: &str) {
     app.terminal_mut()
         .write_input(probe_command(probe).as_bytes());
+    let base = crate::test_budget::tests::TERMINAL_PROBE_BASE;
+    let budget = crate::test_budget::spawn_budget(base);
     let started = std::time::Instant::now();
-    while started.elapsed() < PROBE_TIMEOUT {
+    while started.elapsed() < budget {
         if app.terminal().visible_text().contains(probe) {
             return;
         }
         std::thread::sleep(std::time::Duration::from_millis(20));
     }
     panic!(
-        "the shell never printed the probe {probe:?} within {PROBE_TIMEOUT:?}; terminal showed:\n{}",
+        "the shell never printed the probe {probe:?} within {budget:?} ({base:?} base x{} total \
+         for calibration and load); if this is a real failure it would also fail at 1x, so \
+         re-run the full suite on the unmodified merge base under the same load before \
+         suspecting your diff. Terminal showed:\n{}",
+        budget.as_millis() / base.as_millis().max(1),
         app.terminal().visible_text()
     );
 }
@@ -6227,13 +6235,14 @@ fn rerunning_a_task_reuses_its_pane_instead_of_stacking_new_ones() {
     app.handle_key(chord).unwrap();
     // Wait for the pane's shell to come back to its prompt so the rerun
     // path sees an idle pane (a busy pane legitimately gets a new one).
-    let started = std::time::Instant::now();
-    while started.elapsed() < std::time::Duration::from_millis(5000) {
-        if app.terminals[app.active_terminal].foreground_is_shell() {
-            break;
-        }
-        std::thread::sleep(std::time::Duration::from_millis(20));
-    }
+    // Load-scaled and LOUD (#397): the fixed 5000ms loop broke out silently
+    // on timeout, so under suite load the next step met a pane still busy
+    // and the assertion after it misreported the timeout as its own failure.
+    crate::test_budget::await_spawned(
+        crate::test_budget::tests::TASK_PANE_PROMPT_BASE,
+        "the task pane's shell to return to its prompt",
+        || app.terminals[app.active_terminal].foreground_is_shell(),
+    );
     let after_first = app.terminals.len();
     app.handle_key(chord).unwrap();
     assert_eq!(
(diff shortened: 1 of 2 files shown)


vitali87:

Ran because neither bot reviewed this head: CodeRabbit is at its OSS review limit ("Review rate limited" is a no-op, not a review) and Greptile has not reviewed this PR.

## Walkthrough

Test-only change converting three fixed wall-clock waits in `src/app/tests.rs` to the repo's load-scaled `test_budget` helpers, and swapping one flaky test's live workspace shell for a silent child. `await_terminal_probe` (used by six end-to-end tests) moves from a fixed 15s to `spawn_budget(TERMINAL_PROBE_BASE)`; three task-pane tests replace a 5000ms loop that `break`-ed silently on timeout with `await_spawned(TASK_PANE_PROMPT_BASE, …)`, which panics loudly; and `quick_select_labels_follow_content_that_streams_below_them` now runs pane 0 as `/bin/sh -c 'sleep 30'` so no shell prompt repaint can rewrite the grid between feeds. Two `pub(crate) const` bases plus a floor-pin test are added inside `src/test_budget.rs`'s `tests` module.

| file | summary |
|---|---|
| `src/app/tests.rs` | Deletes the `PROBE_TIMEOUT` const; `await_terminal_probe` now uses `spawn_budget(TERMINAL_PROBE_BASE)` and reports the scale factor in its panic. Three task-pane tests swap their silent `while`/`break` loop for `await_spawned(TASK_PANE_PROMPT_BASE, …)`. `quick_select_labels_follow_content_that_streams_below_them` replaces `app.terminals[0]` with a silent `/bin/sh -c 'sleep 30'` pane and rewrites the surrounding rationale comments; the `h >= 16` tripwire and both assertions are kept with reworded messages. |
| `src/test_budget.rs` | Adds `TERMINAL_PROBE_BASE` (3750ms) and `TASK_PANE_PROMPT_BASE` (1250ms) beside `RESTORED_SHELL_BASE` inside `pub(crate) mod tests`, plus `the_probe_and_task_pane_bases_keep_their_old_budgets`, pinning each base's floor product against the constant it replaced. |

## Findings

None at blocker, major, or minor severity. Everything I set out to falsify held up:

- **Release-hygiene exemption verified, not assumed.** `.github/workflows/ci.yml:69-71` filters `^src/app/tests\.rs` out of `shipped` by path before anything else runs, and I ran the second filter myself against the true merge base (`571dd9d`): `python3 scripts/ships_nothing.py 571dd9d HEAD src/test_budget.rs` exits 0. Only those two files are in the PR, so `shipped` is empty and the job exits 0 at `ci.yml:100` without ever reaching the version or notes checks. No bump required — this matches the PR body's claim and CONTRIBUTING.md:219-225.
- **Doc gate passes.** `python3 scripts/check_doc_ownership.py 571dd9d HEAD` → "No documentation lost or stranded across 2 changed Rust file(s)."
- **The floor arithmetic is right.** `BASE_CALIBRATION = 2` (`src/test_budget.rs:57`), `MIN_SCALE = 2` (line 37). `3750ms × 4 = 15000ms` and `1250ms × 4 = 5000ms` — both exactly reproduce the constants they replaced, and `the_probe_and_task_pane_bases_keep_their_old_budgets` (line 411) asserts this against the very constants the call sites import.
- **The `await_spawned` closures borrow-check.** All three read `app.terminals[app.active_terminal]` immutably inside an `FnMut` while `app` is otherwise unborrowed at that statement — the pattern already compiling at `src/app/tests.rs:24481` and `:24820` against `RESTORED_SHELL_BASE`.
- **The `crate::test_budget::tests::` path is legal and precedented.** `src/main.rs:78-79` declares `#[cfg(test)] mod test_budget;`, and `mod tests` inside it is `pub(crate)` with `pub(crate)` consts (`src/test_budget.rs:261`). No `dead_code` exposure under `cargo clippy --all-targets -- -D warnings` since both new consts have live `cfg(test)` users.
- **No stale `PROBE_TIMEOUT` reference.** `grep -rn PROBE_TIMEOUT src/` returns only `src/gui_path.rs`, an unrelated const of its own.
- **The silent-pane swap does not disturb the geometry the test depends on.** `PtyTerminal::new_running` (`src/widgets/terminal.rs:1343`) writes its run label into the grid synchronously (`terminal.rs:1727-1732`), not as a widget row, so `last_inner.height` is untouched and the `h >= 16` tripwire still measures the same thing. The header is scrolled away by the `h * 2` newline feed and contains neither `http://` nor `drift.io`. The `[Process exited]` footer only fires on child exit, and `sleep 30` outlives the test.
- **The kept assertions are not weakened by the swap.** With no shell prompt on screen the URL is likely the sole quick-select match, so `assign_labels` (`src/quick_select.rs:185`) hands it the first single-char label deterministically; the assertion is still the label-overlay invariant, not the letter.
- **The panic's scale arithmetic is sound.** `budget.as_millis() / base.as_millis().max(1)` divides 15000 by 3750 exactly at the floor; `.max(1)` guards the zero-base case.
- **The three task-pane tests can actually reach their condition.** `foreground_is_shell` (`src/widgets/terminal.rs:3491`) settles true once the `make build` recipe (`true`, instant) returns. The comment at `src/app/tests.rs:6307-6311` correctly identifies the nastier case in `a_task_pane_from_the_old_workspace_is_not_reused_after_a_re_root`: a timed-out busy pane made `after + 1` hold without ever exercising the cwd guard, a silent false pass.

## Test coverage

No gap that warrants a new test. The only uncovered change is `await_terminal_probe`'s new panic text, a diagnostic on a path only reached when a test is already failing; the repo does not test the equivalent string in `await_spawned` either. The behavioral changes are each covered: the two new bases are pinned at the floor, the ceiling stays pinned by `the_worst_case_stretch_is_unchanged`, and the quick-select test's own `h >= 16` tripwire is the precondition check for its own fix.

One note: the PR body said the probe helper covers "seven end-to-end tests" while the in-tree doc at `src/test_budget.rs:264` says "six". Every call site (`src/app/tests.rs:6159, 7574, 7625, 7687, 7861, 7895`) maps to one of exactly six distinct `#[test]` fns; the shipped documentation is the correct one, and the PR body has been corrected.

## Nitpicks

- `src/app/tests.rs:30419-30429` — the comment block explaining the silent-pane swap runs eleven lines to re-narrate the captured failure that the commit message already tells in full. Within local convention; the same hunk deletes three comparably long blocks, so net density is roughly unchanged.

## Scores

- **CodeRabbit score (diff quality) — 5/5**: no findings against the changed lines. The floor arithmetic is exact in both directions, the pin test is written against the constants the call sites import rather than literals beside it, and the panic-message scale division is guarded.
- **Greptile score (codebase fit) — 5/5**: no findings whose evidence lives outside the diff. The helper choice follows CONTRIBUTING.md:435-455, the const path and the silent-pane idiom match in-repo precedent, both CI gates were run and pass, and the release-hygiene exemption is genuine.

VERDICT: APPROVE

vitali87:
Additional captured evidence from a parallel session on a Linux box: `double_click_in_terminal_word_selects_in_split_and_maximised_layout` failed under full-suite load at the same probe wait this PR budgets: `the shell never printed the probe "croft-dclick-…" within 15s; terminal showed: printf '%s%s\n' 'c' 'roft-dclick-…'`. The shell had echoed the command line but not reached a prompt, so the 15 s expired on shell startup, not on anything in the double-click or selection path; run alone straight afterwards it passed in 2.05 s. That member is therefore covered by the `await_terminal_probe` change here (it is one of the six callers), with no change to its selection assertions.

PR #497, opened 2026-09-06
feat(provenance): undo and redo restore the map; a lens shows who typed each line

vitali87:
## Summary

Part of #349 (provenance blame). Second slice, after #446 shipped the per-buffer map: the map now survives an undo/redo cycle, and there is a lens to look at it.

## What changes

- **Undo/redo carry the map.** The editor's undo `Snapshot` now holds the provenance map alongside the lines, so undoing an edit brings back the attributions of the text it brings back and redoing brings back the edit's own seat. Before this the map was cleared on every restore, deliberately, because keeping a stale map would have credited lines to whoever wrote a different buffer; a reload still clears it, since a reload replaces the text with something the map never described.
- **`Editor: Toggle Provenance`** (Command Palette) turns on a session-only lens: a thin bar in the gutter's spacer lane in a per-seat hue (blue you, purple navigator, orange agent, green peer, grey generated), painted over the git bar while the lens is on, and the seat's name appended to the inline-blame annotation (`Vitali, 1 minute ago • fix: the thing • agent (pane 2)`). A line with no record paints nothing and adds nothing: unknown lines are never guessed. The flag is synced onto the active editor each tick the way the inline-blame preference is.

- Three seats, three colours, correct names on hover: the gutter half and the annotation half are here; hover text is the inline-blame annotation.
- Survives save, reopen, and an undo/redo cycle: undo/redo yes (this PR); save and reopen need the map persisted beside local history, which stays for the next slice.
- Unknown lines are never attributed: unchanged, and the lens keeps it visible.

## Tests

`an_undo_cycle_restores_the_provenance_of_each_step` (red before the snapshot change: the peer's line came back unattributed), `provenance_overlay_paints_a_seat_coloured_bar_per_line`, `provenance_overlay_names_the_seat_in_the_blame_annotation`, `toggle_provenance_command_flips_the_editor_lens`. Related provenance, blame-annotation, gutter and toggle tests pass; clippy with `-D warnings` on all targets is clean.

Version 0.1.897 with `src/release_notes/0.1.897.md` (main is 0.1.894; 0.1.895 and 0.1.896 are held by other open PRs).


vitali87:

CodeRabbit is at its rate limit on this repo and Greptile reviews only a PR's first head, so this pre-push review is the recorded review. It is the second round on the branch (round one was REQUEST_CHANGES: a doc comment captured by an inserted item, the seat label painting over a conflict header's accept actions, and the lens reaching only the active tab; all three fixed before this round). After this round's approve, and before the push, its two coverage items were added as well: a `Seat::hue()` distinctness test and a split-group assertion in the sync test, plus the "per-frame" wording correction. The version bump and release note were added last, against a fresh read of main.

## Walkthrough

The second slice of #349 makes the line-provenance map survive history and gives it two surfaces. `Snapshot` now carries a clone of `Provenance`, so `restore_snapshot` re-installs the map that described the text it brings back (truncated to the restored line count) instead of clearing it — meeting the issue's criterion 2, while a reload still clears deliberately. A new session-only `Editor: Toggle Provenance` command flips `App::provenance_overlay`, which paints a per-seat `▎` bar in the gutter's spacer lane over the git bar, and appends the seat's label to the inline-blame annotation. Round one's two majors are genuinely fixed: the seat-alone arm is now guarded by `!in_conflict`, and the flag is pushed onto every editor in both `sync_focus_flags` loops.

| File | Summary |
|---|---|
| `docs/KEYBINDINGS.md` | One row for `Editor: Toggle Provenance`, stating the inline-blame dependency and the conflict-block behaviour. |
| `src/app/mod.rs` | New `provenance_overlay: bool` App field (default false); pushed to the active editor in `sync_blame` (before the blame-off early return) and to every editor in both `sync_focus_flags` loops; `Cmd::ToggleProvenance` dispatch and `toggle_provenance()`. |
| `src/app/tests.rs` | Two tests: the toggle command flips the active editor's lens and survives the tick sync; the flag reaches a second tab (and, after this round, a split's inactive group) after `sync_focus_flags`. |
| `src/provenance.rs` | `Seat::hue()` — five RGB triples, one per variant — inserted below the complete `label()` item; after this round, a test that the five hues are distinct. |
| `src/widgets/command_palette.rs` | `Command::ToggleProvenance` added to the enum, `ALL_COMMANDS`, title, keybinding hint and `id()`. |
| `src/widgets/editor.rs` | `Snapshot.provenance` + its clone in `snapshot()`; `restore_snapshot` restores and truncates instead of clearing; `current_line_blame_annotation` split into a git half (`committed_blame_annotation`) and a lens-aware wrapper; the gutter overlay block; five tests. |

## Findings

Nothing rises to blocker or major. The four items set out to verify all check out:

- **`self.conflicts` cache freshness at annotation time — correct.** The new `conflict_containing(&self.conflicts, …)` reads the cached field on the same footing as the git half: the sole production caller is the render block, and the same `Widget::render` pass calls `self.conflicts()` before the row loop, so the cache is fresh by the time the annotation is built.
- **The multi-tab test exercises a second editor.** `push_editor` appends and activates the new tab, so the toggle reaches only it through `DerefMut`; the first tab's flag can only be changed by `sync_focus_flags`. Both directions have teeth.
- **No remaining `Snapshot { .. }` or `Editor { .. }` literal.** One `Snapshot` constructor in `editor.rs`, updated; construction of `Editor` goes through `Editor::new()`, updated.
- **`hue()`'s "one distinct hue per kind" holds in both themes.** Reproducing `luma`/`light_fallback` in a standalone probe: all five stay distinct and legible on white.

## Test coverage

- `Seat::hue()` had no unit test and the render test compared only two variants. Added after this round: `every_seat_kind_has_its_own_hue`.
- The inactive-group loop in `sync_focus_flags` was not exercised. Added after this round: the sync test now splits the editor and asserts the lens across the active group and the inactive one.
- The `restore_snapshot` truncate is still untested (acknowledged follow-up): a test that records seats past the restored line count and asserts they are gone, with a contrast assertion that an in-range seat survives.

## Nitpicks

- The field doc said "per-frame flag sync" while `sync_focus_flags` runs at its event call sites, not per frame; corrected after this round to name the function.
- The KEYBINDINGS row sits under the Source Control section beside the inline-blame row it depends on; an `Editor:` command might read better under the editor section. Left as-is.
- The two adjacent gutter blocks order their guards differently. Left as-is.
- `sync_blame`'s doc describes blame fetching only while also pushing non-blame editor flags (as `auto_close_pairs` already did). Left as-is.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: no correctness defect inside the changed lines (the match is exhaustive, the truncate is correctly ordered after the empty-buffer push, the conflict guard is right); deductions for the untested `hue()` and the field comment's wording, both addressed after this round.
- **Greptile score (codebase fit) — 4/5**: clean against everything outside the diff (the cached-`conflicts` read matches the git half's freshness contract, the two `sync_focus_flags` loops mirror `show_bracket_colors`/`show_indent_guides`, diff tabs return early so the lens is inert there, reload paths clear the undo stacks so no stale map can be restored onto reloaded text, no `ALL_COMMANDS` gate is broken, and the doc-ownership gate passes); deductions for the untested inactive-group loop, addressed after this round, and the KEYBINDINGS placement.

VERDICT: APPROVE

---
Merging on this verdict once CI on the head is green.

PR #498, opened 2026-09-06
feat(explorer): a lane's root row wears its branch and its agent (part of #348)

vitali87:
## Summary

Part of #348 (worktree lanes), after #449 (the worktree operations) and #496 (the pane wiring). This is the Explorer half: a lane's root row now shows what is happening in it.

- **The badge.** After the folder name, dim: the lane's branch and the agent seated in its pane with the pane pill's glyph, `repo-fix-login  agent/fix-login · ◆ claude ●`. Before the agent is seated the configured agent's bare name shows (`agent/fix-login · claude`); a plain-shell lane shows the branch alone.
- **Where it comes from.** `FileTree::root_badges` is a map keyed by root path, painted only on depth-0 rows and clipped like the rest of the row. The app fills it in `sync_lane_root_badges` at the top of render from `lane_panes` and each pane's `agent()`, swapping the map only when its contents change; with several panes on one lane, the one with an agent seated decides. Paths match because both the tree's roots and the recorded lane paths are canonical.

## Tests

- `widgets::file_tree::a_root_badge_is_painted_after_the_root_name_and_only_there`: the badge follows the lane root's name, appears on that row alone, and goes when the map clears.
- `a_lanes_root_row_wears_its_branch_and_its_agent`: the three badge shapes, then the app-side sync on a real lane (keyed by the canonical path, the primary has none, a closed lane's badge goes with it).

The lane and root-badge filters pass locally; clippy with `-D warnings` and the doc-ownership gate are clean.

Still open on #348 after this: grouping the review queue by worktree, and "Agent: Merge Lane".

Version 0.1.899 with `src/release_notes/0.1.899.md` (re-read against main at merge time).

Part of #348


vitali87:

Both bots were unavailable on this PR (Greptile: out of credits for the billing period; CodeRabbit: "Review rate limited"), so this review was run locally. Round 1 (head e500e6b) returned APPROVE at 4/5 4/5 with two minors (a glyph sniff deciding the multi-pane tie-break, and `DarkGray` which the theme does not map on light grounds) and three test gaps; those are fixed in 44548a8. This is the round-2 result on head 44548a8.

---

## 1. Walkthrough

A worktree lane's Explorer root row now carries a dim trailing badge — its branch plus the agent seated in its pane (`agent/fix-login · ◆ claude ●`), the configured agent's bare name before the sampler seats it, or the branch alone for a plain-shell lane. `FileTree` gains `root_badges: Arc<HashMap<PathBuf, String>>`, painted only on depth-0 rows and cleared by `set_root` alongside `ignored`/`agent_touched`; `App::sync_lane_root_badges` rebuilds it from `lane_panes` at the top of every `render`, resolving several panes on one lane by an explicit rank (seated 2, configured 1, plain 0), and swaps the `Arc` only when the contents differ. Path keying holds: `open_lane_pane` canonicalizes `LaneRecord.path` and `add_workspace_folder` canonicalizes before `tree.add_root`, so the create, `restore_workspace_folders` and `restore_terminal_session` paths all produce the same key.

| file | summary |
|---|---|
| `Cargo.toml`, `Cargo.lock` | 0.1.898 → 0.1.899 |
| `src/release_notes/0.1.899.md` | new, one `feature:` line |
| `src/widgets/file_tree.rs` | `root_badges` field + `HashMap` import; cleared in `set_root`; depth-0 paint in the dir branch of `render` using `ui(Color::Gray)`; two render tests |
| `src/app/mod.rs` | `sync_lane_root_badges()` at the top of `render`; the rank-based `sync_lane_root_badges`; free fn `lane_root_badge` |
| `src/app/tests.rs` | `a_lanes_root_row_wears_its_branch_and_its_agent`, `a_lanes_badge_follows_the_seat_and_a_seated_pane_outranks_a_plain_one` |
| `docs/ARCHITECTURE.md` | `git.rs` entry gains a clause on the Explorer badge |
| `docs/LAYOUT.md`, `docs/KEYBINDINGS.md` | the badge described on the Agent-lanes bullet and the `Cmd+K Shift+L` row |

## 2. Findings

None.

Both findings from the previous round are fixed, and I verified each against the working tree rather than the diff:

- `src/app/mod.rs:25419-25429` — the glyph sniff is gone. `rank` is derived from `(seated, lane.agent)` directly, and `best.get(&key).is_some_and(|(have, _)| *have >= rank)` skips only a strictly-not-better pane. `>=` (not `>`) keeps first-wins among equal ranks, matching the old seated-pane behaviour and staying stable under the `BTreeMap`'s uid order. No formatted string is read back for state, so the `agents.json` row name that could previously spoof the guard (`src/agents.rs:115`, arbitrary user text) is now inert.
- `src/widgets/file_tree.rs:1534` — now `self.theme.ui(Color::Gray)`, which `Theme::ui` maps to `Color::Rgb(0x61, 0x61, 0x61)` on a light ground (`src/theme.rs:339`). The added comment names the reason the old constant was wrong. This is now the repo's established dim idiom (16 `ui(Color::Gray)` call sites, zero remaining `ui(Color::DarkGray)`).

## 3. Test coverage

No gaps I can identify. The three I raised are all closed, and each new test actually fails if its guard is removed:

- `a_root_badge_skips_the_same_path_where_it_appears_as_a_child` (`file_tree.rs:2096`) puts `second/lib` on screen twice — `add_root` appends it as a depth-0 section while `two_root_fixture`'s expanded `second` root already renders it as a depth-1 child (verified `add_root` sets `expanded: true` and calls `load_children`, `file_tree.rs:176-183`). Asserting the badge count is exactly 1 makes `node.depth == 0` at `file_tree.rs:1527` load-bearing in a way the original test did not.
- `a_lanes_badge_follows_the_seat_and_a_seated_pane_outranks_a_plain_one` (`tests.rs:43195`) walks configured → seated → back to configured, and runs the two-pane tie in both orders. The ordering is real rather than assumed: `uid` is a monotonic global `AtomicU64::fetch_add` (`widgets/terminal.rs:2180`), so `assert!(app.terminals[idx].uid() < second_uid)` holds by construction and the `BTreeMap` genuinely walks the seated pane first in one case and second in the other.
- The test's agent table is correctly isolated. `AgentTable::from_json` starts from `builtin()` and *replaces* a same-named row rather than appending (`agents.rs:228-243`), so `{"name":"claude","launch":"true"}` overrides the real launch line — the lane records `agent: Some("claude")` (the `seated` flag in `open_lane_pane` needs a launch line) without spawning anything. The comment's claim about not starting the real thing is accurate.

## 4. Nitpicks

- `src/app/mod.rs:25431-25432` The `best` map is now collected into a second `HashMap` purely to drop the rank before the equality compare. Correct and cheap at these sizes, but the compare could run against `best` directly if the tree's field were `(u8, String)`-free by construction — not worth changing; noting only that the extra pass exists.
- `src/widgets/file_tree.rs:1686` The sticky band still rebuilds root-row spans by hand and omits the badge, so a lane root pinned by sticky scroll loses its branch/agent. The team's stated reason — the band drops the other decorations too — checks out against the code (it re-derives only chevron/icon/name and applies neither the ignored dim's full treatment nor the `#345` dot), so this is consistent rather than an oversight.
- `docs/ARCHITECTURE.md:175` The `file_tree.rs` entry still has no pointer to `root_badges`; it is documented on the `git.rs` (#348) entry. The stated precedent holds — I confirmed the `#345` agent dot is likewise absent from the widget entry and lives on `agent_lane.rs:85`.
- `src/app/mod.rs:25381` vs `src/app/mod.rs:45064` The case flip for a capitalised custom `agents.json` row name (`LaneRecord.agent` is lowercased, `AgentLane.name` is the row's name as written) remains. Invisible for the built-ins, which are lowercase (`agents.rs:555`, `620`), and explicitly deferred by the team.

## 5. Scores

- **CodeRabbit score (diff quality) — 5/5.** No findings in the changed lines: the tie-break now reads state instead of rendered text, the painter uses the mapped grey, the doc comment's allocation claim is accurate, and both new tests fail if the behaviour they name regresses.
- **Greptile score (codebase fit) — 5/5.** No findings outside the diff either. `set_root` now clears `root_badges` beside its sibling absolute-path caches; the dim colour matches the 16-site `ui(Color::Gray)` convention; path keying verified consistent across `add_workspace_folder`, `restore_workspace_folders` and `restore_terminal_session`; `close_terminal_at` already prunes `lane_panes`, so no orphaned-terminal case exists; version bump, release note, and doc-block placement all satisfy CONTRIBUTING, and I ran `scripts/check_doc_ownership.py origin/main HEAD` at this head — clean across all 3 changed Rust files.

VERDICT: APPROVE

PR #499, opened 2026-09-06
feat(provenance): the map survives a restart, recorded beside local history

vitali87:
## Summary

Part of #349 (provenance blame). Third slice, after #446 (the map) and #497 (undo/redo and the lens): the map now survives a restart.

## What changes

- **Recorded beside local history.** Every save already records a byte-exact snapshot of the file under the history root. The save now also writes `<millis>.seats` beside `<millis>.snap`: the tab's provenance map as JSON, through the same staging-and-rename as the snapshot, so it is never listed half-written. A map with nothing attributed writes nothing. A snapshot that is superseded inside the merge window or pruned past the cap takes its sidecar with it.
- **Restored only onto the text it described.** Once per buffer generation, when a tab is clean and knows nothing yet, croft reads the newest snapshot; the seats beside it come back only if that snapshot is identical to the file on disk AND the buffer itself holds that text. Any difference (an edit since, a save croft never saw, a file with no history, a sidecar that does not parse) leaves every line unknown. The buffer check is what stops a clean-but-stale tab, one side of a split after the other side saved, from inheriting the saving buffer's attributions. A map is never applied to text it did not describe, so the issue's "unknown lines are never attributed" holds across the restart boundary too.
- The map is read off the saved tab itself at save time, on the UI thread, at every save site (Cmd+S, the auto-save sweep of both groups, a deferred write after formatting, the hex, sheet and forced saves), never looked up by path afterwards: a split can hold the same file in a second buffer with a different map. The restore is synced onto every tab of every group, the way the lens flag reaches every pane. A viewer tab (hex, sheet, image, SVG, log, archive, PDF, media and document previews) records no map: every viewer opener now clears it the way the text opener does, since the viewer's placeholder line was written by no seat. The predicate is the save path's own `has_non_text_view`, so a markdown preview, whose lines are still the file's own text, keeps and records its map.
- The once-per-buffer marker lives on the editor and is keyed on its `edit_seq`, which every opener bumps: a reload, a revert, a disk-conflict reload, Reopen as Text and a re-decode are all reconsidered by construction, and two tabs on one file (a split) are two buffers.
- **"Local History: Restore Snapshot"** carries the restored snapshot's own seats onto the reverted buffer and onto the snapshot it records, so a restore inside the merge window no longer replaces a persisted map with an empty one; a snapshot that recorded no seats restores every line unknown.

- Survives save, reopen, and an undo/redo cycle: complete with this PR (undo/redo landed in #497).
- Three seats, three colours, correct names on hover: landed in #497.
- Unknown lines are never attributed: unchanged, and enforced at restore by matching the snapshot against both the file on disk and the buffer itself.

Still open on #349 after this: the Source Control "group by seat" diff view.

## Tests

Red first: `seats_are_recorded_beside_their_snapshot_and_read_back` (history), `provenance_survives_a_restart_when_the_file_is_unchanged` (a second app over the same history root gets the navigator's and the agent's lines back, the untouched line stays unknown), with `provenance_is_not_restored_onto_text_the_snapshot_never_saw` as the control (a file edited outside croft after the save opens with every line unknown).

The restore side is pinned by three more, each red against the first cut: `provenance_comes_back_after_every_same_path_swap` (a revert and a same-path reopen each clear the map and get it back on the next tick, with nothing reset by hand: the marker is keyed on the buffer's `edit_seq`, which every opener bumps), `a_dirty_tab_is_reconsidered_for_provenance_once_it_is_clean` (a sync that lands on a dirty buffer does not burn the path), and `restoring_a_snapshot_carries_its_seats_onto_the_restore` ("Local History: Restore Snapshot" puts the restored snapshot's own seats onto the buffer and onto the snapshot it records, so a restore inside the merge window no longer replaces a persisted map with an empty one).

Four more came out of later review rounds, each red first: `a_split_of_a_synced_file_gets_the_map_in_the_new_group` and `provenance_is_synced_onto_the_inactive_group_too` (two tabs on one file are two buffers, and the lens paints in every pane), `an_inactive_group_save_records_the_map_of_that_buffer` (a save records the map of the buffer it wrote, not one found by path), `a_viewer_swap_drops_the_map_and_the_text_reopen_gets_it_back` and `a_markdown_preview_save_keeps_the_map` (the viewer predicate is the save path's own), and `a_stale_tab_is_not_credited_with_the_map_of_the_saving_tab` (verified revert-sensitive: disabling the buffer check fails exactly that test), with `the_map_of_an_emptied_file_survives_a_restart` and `the_map_of_a_utf16_file_survives_a_restart` pinning the comparison against how a file is actually loaded. Related provenance and history tests pass; clippy with `-D warnings` on all targets is clean.

Version 0.1.900 with `src/release_notes/0.1.900.md` (main is 0.1.899).

## Summary by CodeRabbit

* **New Features**
  * Per-line authorship information now persists alongside local-history snapshots.
  * Reopening an unchanged file restores its authorship information across restarts.
  * Snapshot restores preserve authorship when the restored content matches exactly.
  * Files changed since their snapshot reopen with authorship marked as unknown, preventing stale attribution.

* **Bug Fixes**
  * Prevented viewer content and mismatched or dirty files from inheriting incorrect authorship information.

* **Documentation**
  * Updated architecture documentation to describe authorship persistence and restoration behavior.


vitali87:

CodeRabbit is rate-limited on this repo, so this branch went through seven rounds of local Opus 5 review before the push. Rounds one to six each returned REQUEST_CHANGES; every finding was fixed with a failing test first. This is the round-seven verdict on the reviewed head, plus a note of the two minors it raised, which are fixed in the commit above it.

---
# Review: `feat/provenance-persisted` (round 7)

**Reviewed head:** `eb8d3d5` (the two minors below are fixed in `d0ceec8`, and `c43b5f0` adds the version bump).

## 1. Walkthrough

The branch persists the per-line provenance map as a `<millis>.seats` JSON sidecar beside each local-history `<millis>.snap`, written on save with the saved buffer's own map and restored onto clean, unattributed text buffers when the newest snapshot is byte-identical to disk *and* the buffer itself holds that text. Round six's major — a clean-but-stale split tab being credited with the saving tab's map — is fixed by `Editor::holds_exactly`, which decodes the snapshot with the BOM's encoding (else the buffer's) and compares `split_into_lines` against `self.lines`.

**Verification of the round-six fix.** Checked against every case:

- **CRLF, no-trailing-newline, trailing-blank-line** — all round-trip correctly. `split_into_lines` normalizes `\r\n`→`\n` before `.lines()`, matching `open` exactly.
- **BOM'd / UTF-16 / non-UTF-8** — the encoding precedence (`for_bom` else `self.encoding`) mirrors `open` and `reopen_with_encoding` exactly.
- **No false-positive path exists.** A `true` return requires the snapshot to decode, under this buffer's own decoding rules, to exactly this buffer's lines.
- **Ordering of the cheap guards is right.** Marker → `entries_in` → sidecar read → `metadata` size compare → two full reads → `holds_exactly`. The decode is last, behind a proven byte-identical match, and the per-buffer marker caps it at once per buffer generation.
- **No other path assigns a stored map without the check.** All 14 `.provenance = ` sites enumerated: eleven are `Provenance::new()` clears, one is the undo/redo snapshot (in-memory, correct), and only two assign persisted data — `sync_provenance_for` (gated) and `restore_history_snapshot` (safe by construction).

**Branch-wide re-check.** Sidecar lifecycle is complete: removal on merge-window supersession at both timestamps, on prune, and overwrite-in-place on the unchanged-content path. `entries_in` filters on `SNAP_EXT` so sidecars never pollute the listing. The shared staging shape with an atomic counter means the snap and seats writes cannot collide. All nine viewer openers clear the map. Every save site passes `provenance_to_record()`. The group walk covers all tabs in all groups.

## 2. Findings

- **`src/widgets/editor.rs` (`holds_exactly`)** **[minor/correctness]** returns `false` for an empty file, so an empty buffer never gets its map back. `open` pushes a sentinel line, leaving `lines == [""]`, but `holds_exactly` compares against the bare `split_into_lines("")`, which is `[]`. Reachable: attribute line 0, backspace to empty, save. The effect is fail-closed — a *missed* restore, never a wrong credit.

- **`src/widgets/editor.rs` (`holds_exactly`)** **[minor/correctness]** the `!had_errors` conjunct makes it stricter than `open`, so a file with undecodable bytes never gets its map back even when the buffer genuinely holds that text. `open` discards `had_errors` and builds `lines` from the replacement-character text. Also fail-closed.

## 3. Test coverage

Coverage is strong — 13 new app tests plus 4 history tests, and the ones read carry explicit positive controls that prove a careless implementation would have had something to apply. Confirmed with a working grep control that **no test touches `provenance_synced_seq` directly** (zero hits in tests against 4 in production).

Gaps, both matching the findings above: an empty file's map never returns; and no test exercises the encoding branch, so a regression dropping `for_bom` would stay green.

## 4. Nitpicks

- The restore assigns the map after `revert_to_disk()` without re-checking the buffer, which is safe by construction but leaves a narrow window if another process rewrites the file between the write and the reload.
- `seats_path` takes an already-resolved `dir` while its siblings take `root_dir` and resolve internally.

## 5. Scores

- **CodeRabbit score (diff quality) — 4/5.** Two minors inside the changed lines, both fail-closed; no majors, and the new tests carry real positive controls.
- **Greptile score (codebase fit) — 4/5.** No broken callers: all 14 provenance-assignment sites, all nine viewer openers, every save site, and the group walk check out against the surrounding code. Docs and comment density match the repo's conventions.

VERDICT: APPROVE

---

**Both minors and the restore nitpick are fixed in `d0ceec8`**, with a test for the emptied file and one for a UTF-16 file whose snapshot has to be decoded through its BOM. The full suite selection for this branch is 39 tests, green, with clippy `-D warnings` clean on all targets.

vitali87:
All three findings were real and are fixed in `bc40031`, each verified against the code first.

- **Sidecar before publish** (`src/history.rs`): the superseded same-millisecond sidecar is now removed *before* `write_snapshot` publishes the replacement, so an interruption in between leaves a snapshot with no seats rather than one wearing the seats of text it does not hold.
- **Marker before the sidecar read** (`sync_provenance_for`): the marker is now set only once `seats_for` has found a map. A save records its snapshot and then its sidecar from a worker thread, and a tick landing in that window used to conclude the buffer had no map and burn the marker for the session.
- **Map bound to its bytes** (`record_history_snapshot`): the recorder takes the bytes the map describes and records the map only while those bytes are still on disk; otherwise the snapshot is recorded alone and every line reads unknown. The bytes come from the save site rather than a lookup here, because a search by path finds the focused tab, which need not be the tab that saved; a snapshot restore passes `None`, since its seats describe bytes it wrote itself.

The third is pinned by `the_map_is_not_recorded_against_another_save_s_bytes`, with a positive control in the same test and verified revert-sensitive: making the binding inert fails exactly that test. Suite selection for this branch is 40 tests, green, clippy `-D warnings` clean, and the branch is merged up to main at 0.1.902.

PR #500, opened 2026-09-06
feat(agents): group the review queue by the worktree each file is in (part of #348)

vitali87:
## Summary

Part of #348 (worktree lanes), the last of its slices after #449 (worktree operations), #496 (the pane wiring) and #498 (the Explorer badge).

"Agents: Show Changed Files" now reads as one section per workspace root, which is what makes the queue usable once lanes exist: with an agent working in a lane and you working in the main checkout, a flat list of basenames cannot say which `mod.rs` it means.

- **Grouping.** `AgentLedger::lane_by_root` groups an agent's unreviewed files by the root that owns them, longest-prefix like `WorkspaceRoots::owning_root`, in root order with the queue's existing recency inside each group. A file under no root groups under "outside the workspace" rather than being dropped.
- **Naming.** A lane's group is named by its **branch** (the directory is `repo-fix-login`, the branch is what the agent is on), any other root by its display label.
- **Single-root sessions are untouched.** With one group the row keeps the pre-grouping flat shape, `claude: 2 to review (b.rs, a.rs)`: the label would name the only place the files could be and the group count would merely repeat the total. Grouping appears exactly when there is something to tell apart.
- **Truncation says so.** A group lists at most three names and ends in `…` when there are more, so a count is never set against a silently shortened list.

The grouping is pure and takes the roots rather than reaching for them, so ordering and the unowned-file case are testable without a workspace.

## Tests

- `agent_lane::a_lane_groups_its_files_by_the_root_that_owns_them`: root order, recency within a group, unowned last, and a group that disappears once its files are reviewed.
- `the_review_queue_names_the_lane_each_agents_files_are_in`: the lane named by its branch, the primary by its folder, order, and an unowned file named rather than dropped.
- `a_single_root_review_queue_reads_as_it_always_did`: the flat shape byte for byte, the `…` on a truncated sample, and an emptied lane with no dangling separator.

The ledger and queue filters pass locally (14 tests); clippy with `-D warnings` and the doc-ownership gate are clean.

This closes out the lane work on #348 except "Agent: Merge Lane".

Version 0.1.901 with `src/release_notes/0.1.901.md` (re-read against main at merge time).

Part of #348


vitali87:

Both bots were unavailable on this PR (Greptile: out of credits for the billing period; CodeRabbit: "Review rate limited"), so this review was run locally. Round 1 (head `ea3bac0`) returned REQUEST_CHANGES with one major — a grouped row long enough for the one-row status bar's middle-elision to eat its group labels, the one thing grouping adds — and two minors. Those are fixed in `8f7a363`. This is the round-2 result on head `8f7a363`; the docs-only finding it raised was then fixed in `9a4fd4e`, along with a test for the branch-collision fallback.

---

## 1. Walkthrough

`Cmd::ShowAgentLane`'s inline row-building is extracted into `App::agent_lane_rows` and gains grouping: `AgentLedger::lane_by_root` partitions an agent's unreviewed files by the workspace root that owns them (longest-prefix, mirroring `WorkspaceRoots::owning_root`), and `agent_lane_rows` names each group by the lane's branch or the root's `root_display_labels` entry. A grouped row carries labels and counts only, sized to survive the one-row status bar's middle-elision; a one-group row keeps the pre-grouping flat shape with its three-name sample and `…`. Version 0.1.901 with a matching release note.

| file | summary |
|---|---|
| `Cargo.toml` / `Cargo.lock` | 0.1.899 → 0.1.901 (0.1.900 held by peer PR #499) |
| `src/agent_lane.rs` | new `AgentLedger::lane_by_root`; test covering root order, intra-group recency, unowned-last, group emptying, and nested roots |
| `src/app/mod.rs` | new `App::agent_lane_rows`; grouped rows are label+count, with a branch-collision fallback to the disambiguated root label; `Cmd::ShowAgentLane` arm reduced to one call |
| `src/app/tests.rs` | four tests: multi-root/lane naming, single-root flat shape, a 120x40 render test pinning the row against the status bar, and the branch-collision fallback |
| `docs/KEYBINDINGS.md` | new palette row describing the grouped format, label tiers, and the unowned group |
| `docs/ARCHITECTURE.md` / `docs/LAYOUT.md` | one clause each on the grouping wiring |
| `src/release_notes/0.1.901.md` | new, one `feature:` line |

**Verification of the fixes since `ea3bac0`** — all four land, and I verified each rather than taking the summary's word for it:

- **The status-bar major is fixed and genuinely regression-tested.** `agent_lane_rows` now emits `format!("{label} {}", files.len())`. I simulated `elide_middle` against the new test's own fixture: the new row is 47 chars against roughly 57 available at 120 cols after the right cluster, so it lands whole; the previous format was 95 chars and elides to `claude: 8 to revie…nt/fix-login 4: l3.rs, l2.rs, l1.rs, …`, which contains **neither** `agent/fix-login 4` nor `repo 4`. So `a_grouped_review_queue_row_survives_the_status_bar` fails on the old format for the right reason — it is a real regression test, not a tautology.
- **The branch-collision minor is fixed correctly.** The branch label is guarded on that branch occurring exactly once among the groups being rendered, so a branch shared by two groups sends both to the root label. Single occurrence keeps the branch.
- **The nested-roots gap is closed.** The ledger test asserts `/w/repo/crates/foo/src/lib.rs` groups under the deeper root — component count 4 beats 2 — and that group order still follows `roots` order.
- **The single-root shape is untouched.** `a_single_root_review_queue_reads_as_it_always_did` is unmodified by the fix commit, so the flat row and its `…` remain pinned byte-for-byte.

Both new items sit *after* a complete item (CONTRIBUTING.md's doc-ownership rule), and version + release note satisfy "Every shipped change is a release".

## 2. Findings

- **`docs/KEYBINDINGS.md`** **[minor/docs]** — **The row contradicted itself: it carried both the obsolete format example and the new one.** The clause read "a worktree lane's group is named by its branch (`agent/fix-login 2: other.rs, mod.rs`)" — the label-count-plus-sample shape — while the next sentence correctly said a grouped row carries labels and counts rather than file names. The first example was unproducible: the code emits `{label} {count}` with no `: ` and no sample, and the test asserts `!row.contains("mod.rs")` for exactly that reason. A reader hit the stale example first.
  **Fixed in `9a4fd4e`:** the parenthetical now reads `agent/fix-login 4`.

## 3. Test coverage

- **The branch-collision fallback had no test.** It was the newest logic in the PR and the only untested behavior it introduced: the uniqueness guard could have been inverted and every test would still have passed.
  **Fixed in `9a4fd4e`:** `two_lanes_sharing_a_branch_are_named_by_their_folders_instead` inserts two lane records with distinct paths and the same branch, asserts both groups fall back to their folder labels, then removes one and asserts the now-unique branch names its lane again.
- **`may_be_incomplete` combined with grouping is still untested.** The watcher-overflow suffix appends ` · at least: …` to the joined rows, so a grouped session nests `;`, `—` and `·` in one string with no coverage. Left as is; the suffix is orthogonal to the grouping and unchanged by this PR.
- **Two agents each with groups is still untested.** Both grouping tests use one agent, so the ` · ` join between two grouped rows — the longest string the feature can produce — is never rendered. Left as is; each row is independently bounded, which is what the render test pins.

## 4. Nitpicks

- `agent_lane_rows` materialises `branches` as a `Vec<Option<String>>` and rescans it per group, O(groups²) string comparisons. Fine at realistic root counts.
- `lane_by_root`'s `.unwrap_or(usize::MAX)` in the sort key is dead: `owner()` only ever yields a reference drawn from `roots`. Harmless — `sort_by_key` is stable, so even a collision with the `None` key could not reorder or duplicate a group.
- `roots` and `labels` are built on every call even when the one-group early return leaves `labels` unused. Negligible on a palette command.
- The multi-root test's `row.find(a) < row.find(b)` ordering assertion would pass silently if the first label appeared earlier for an unrelated reason; splitting on `"; "` would be exact.

## 5. Scores

- **CodeRabbit score (diff quality) — 4/5.** The status-bar major and the branch-collision minor are both fixed; what remains inside the changed code is a handful of nits (the O(groups²) uniqueness rescan, the dead `unwrap_or`, the eager `labels`).
- **Greptile score (codebase fit) — 4/5.** The elision contract is now respected and pinned by a render test built in the same shape as the existing `status_transient_keeps_its_tail_when_longer_than_the_bar`; `root_display_labels`' uniqueness guarantee is no longer bypassed; the unowned group is documented as the defensive path the ledger's only feeder makes it. Version, release note, doc-ownership ordering, and the `·`/`—` separator nesting all match house convention.

VERDICT: APPROVE

Issue #186, opened 2026-08-17
Flaky: busy-cd re-root test's preconditions accept foreground_is_shell's fallback answers as resolved samples


vitali87:
## What happens

`change_workspace_root_skips_cd_when_terminal_runs_a_foreground_app` still flakes (~1 in 5 standalone) after #169's `sleep 60` hardening: the re-root takes the seed branch and the assert on the "terminal busy" status fails.

## Why

Both preconditions poll `foreground_is_shell()`, whose answers include policy fallbacks, not just resolved samples:

1. The "shell at its prompt" wait can exit instantly through the startup carve-out (`_ => true` while the spawn has not yet reported a shell pid), before rc files have run. `sleep 60` is then written as startup type-ahead.
2. The "command is foreground" wait then reads an rc-startup child as the busy command.
3. By re-root time the rc finished and the shell sits at its prompt about to read the type-ahead - `foreground_is_shell` is genuinely true, the cd seeds, and the test's premise never existed.

## Fix direction

A strict `#[cfg(test)]` accessor that returns only RESOLVED samples (`Some(fg == shell)`, `None` otherwise), with the test gating each precondition on `Some(true)` / `Some(false)` - the carve-outs stay in the production path where they belong.

PR #501, opened 2026-09-06
test(terminal): scale the two shell-paint waits with load (part of #397)

vitali87:
## Summary

Part of #397, taking the two members the issue's latest analysis names. Both wait on a real shell to paint into the grid, on budgets that do not stretch under load:

- `cmd_clicking_a_non_web_hyperlink_refuses_instead_of_opening` passed a bare `Duration::from_millis(1000)` to `await_spawned` for "the shell to paint the linked cell". That is 4s at the floor, against 15s for `TERMINAL_PROBE_BASE`, which waits for the same thing. The wait had already been raised 5s → 8s once, and the 1000ms base handed back half of that.
- `change_workspace_root_skips_cd_when_terminal_runs_a_foreground_app` used a hand-rolled `loop` with `assert!(waited < 8000)`, so it did not scale at all, and its failure said only "marker output never arrived" — equally true of a starved shell and a broken one.

Both now share `SHELL_PAINT_BASE` (2s), which reproduces the 8000ms each had at the floor and gives up to twice that under load, with the floor assertion its three siblings already have. `RESTORED_SHELL_BASE`'s doc had explicitly deferred this reconciliation as "a question about that test's operation rather than about these"; the owner's control run on unmodified main answers it. Painting a cell is the same operation those waits perform, so it takes the same value rather than a guessed fourth number, and stays a separate constant so that neither operation's cost can move the other silently.

No test's observable behaviour changes; only what it waits.

## No version bump

Every hunk is inside `#[cfg(test)] mod tests` or `src/app/tests.rs`, so the shipped binary is byte-identical: `scripts/ships_nothing.py origin/main HEAD src/test_budget.rs` exits 0. This matches #489, the directly analogous merged PR on this issue, which changed the same two files with no bump and no release note. A note here would tell users their build's headline change is that two tests stopped flaking.

## Tests

`test_budget::the_probe_and_task_pane_bases_keep_their_old_budgets` gains the floor assertion for the new base; the assertion is not vacuous, since a 1999ms base yields 7996ms and fails it. Both converted call sites and the full `test_budget::` module pass locally (14 tests); clippy with `-D warnings` and the doc-ownership gate are clean.

The issue stays open: 31 unscaled `waited < N` loops remain in `src/app/tests.rs`, the residual of this class.

Part of #397


vitali87:

Both bots were unavailable on this PR (Greptile: out of credits for the billing period; CodeRabbit: "Review rate limited"), so this review was run locally. Round 1 (head `a1163aa`) returned REQUEST_CHANGES over a load-bearing doc claim that turned out to be false; that is fixed in `28fdb50`. This is the round-2 result on head `28fdb50`; the two minors it raised were then fixed in `8f3b392`.

---

## 1. Walkthrough

The change puts three waits that depend on a real shell reaching a state onto a single load-scaled base, `test_budget::tests::SHELL_PAINT_BASE` (2s → 8s at the floor via `BASE_CALIBRATION * MIN_SCALE == 4`). Two were the members #397's analysis named: the hyperlink test's bare `Duration::from_millis(1000)` handed to `await_spawned`, and the re-root test's hand-rolled `loop` with `assert!(waited < 8000)` that did not scale and whose failure named nothing. The follow-up commit adds the third — the re-root test's second wait, a fixed 4000ms poll of `foreground_resolved_for_test` that did not assert on timeout at all — and rewrites two paragraphs of the new constant's doc that made unsupported claims. The existing floor-assertion test gains a `paint >= 8000ms` clause. No shipped code changes.

| File | Summary |
| --- | --- |
| `src/test_budget.rs` | Adds `SHELL_PAINT_BASE = Duration::from_secs(2)` inside `#[cfg(test)] pub(crate) mod tests`; extends `the_probe_and_task_pane_bases_keep_their_old_budgets` with a third floor assertion. Doc justifies separation from `TERMINAL_PROBE_BASE` by the differing constants each replaced, and records `RESTORED_SHELL_BASE`'s deferral as stale rather than answered. |
| `src/app/tests.rs` | Hyperlink test: `from_millis(1000)` → `SHELL_PAINT_BASE`. Re-root test: the `waited < 8000` marker loop → `await_spawned` over `app.terminal().grid_lines()`; the `waited < 4000` foreground loop → `await_spawned` with the sample captured into an outer `foreground` from inside the closure. |

**Both prior majors are fixed, and I verified the fixes rather than taking the summary's word.** The false provenance paragraph is gone: the doc now says the 2s is settled by both call sites having replaced the same 8000ms, and explicitly records the deferral as stale because the counterexample it names no longer waits on anything. I re-confirmed that independently — `a_terminal_link_binding_is_not_refused_for_an_editor_side_reason` drives the grid with `feed_bytes_for_test`, and a grep over its body for `await_spawned|waited|from_millis|from_secs` returns nothing. The `TERMINAL_PROBE_BASE` sentence now gives the real reason (15s vs 8000ms replaced, one number cannot pin both floors) instead of undercutting itself.

**The new conversion is mechanically sound.** `await_spawned` evaluates `ready()` *before* its first sleep, so the immediate pre-loop sample the old code took is preserved — no first-iteration behaviour change. `foreground_resolved_for_test` is a `tcgetpgrp` comparison holding no lock and blocking on nothing, so the closure is cheap and cannot deadlock or busy-spin. Capturing into the outer `Option` from inside the closure preserves the retained-sample invariant the `#186` comment protects, and it is exactly the shape `wait_for_conflicted_entry` already uses.

**Release gate re-verified at the new head.** The diff is exactly the two test files. The CI workflow strips `^(src/app/tests\.rs|tests/)` before `ships_nothing.py` runs, and the script exits 0 on `src/test_budget.rs`. `Cargo.toml` is `0.1.902` on both sides and the diff against main for `Cargo.toml` and `src/release_notes/` is empty. Matches merged precedent #489.

## 2. Findings

- **`src/test_budget.rs` (floor assertion message)** **[minor/tests]** — The message read *"the shell-paint base must reproduce the 8000ms **its call sites** replaced at the floor"*, but the constant has three call sites and the third replaced **4000ms**, not 8000ms — `git log -S 'waited < 4000 && foreground'` puts that loop's introduction at `eab97b7` at 4000ms, unchanged until this PR. So the second re-root wait is a deliberate **raise** from 4s to 8s at the floor, not a reproduction. The raise is defensible (the old loop did not assert on timeout, so its 4000ms was never a deadline the test could fail on), and the assertion still passes. But the doc immediately above states the house rule as *"each base reproduces its own at the floor"*, and this call site quietly did not fit it.
  **Fixed in `8f3b392`:** the message now names the largest replaced constant and says the third site is deliberately raised; the doc gained a paragraph recording the same, so a later `git blame` does not surprise the next author.

- **`src/app/tests.rs` (precondition assert)** **[minor/tests]** — The `assert_eq!(foreground, Some(false), "precondition: …")` is unreachable on failure now that `await_spawned` panics on timeout rather than returning. Not a defect — the same pattern is established at `wait_for_conflicted_entry`, which acknowledges it with `.expect("await_spawned returns only once the entry is present")` — but as written it read like a live check, and a reader could later relax the closure's predicate assuming the assert still guarded it.
  **Fixed in `8f3b392`:** the message now says what it actually pins, with a comment naming the sibling pattern; the closure also takes the immutable accessor, matching the marker wait above it.

## 3. Test coverage

- **Call-site drift is still unpinned, and now across three sites.** Reverting any one call site to a literal leaves every test in `test_budget::tests` green, because the floor assertion tests the constant, not its use. Left out deliberately: the hole is shared by all four bases and a source-level `include_str!` tripwire is crude, so it belongs to the module rather than this PR. Worth its own issue — this PR takes `SHELL_PAINT_BASE` from zero to three call sites, making it the most-used of the four.
- **The rewritten predicates are not verified as discriminating.** The marker closure's `!l.contains("echo")` clause is what makes the quote-split trick work, and nothing fails if it is dropped. Likewise the foreground closure would still compile and usually pass if `Some(false)` were relaxed to `.is_some()`, defeating the `#186` reasoning. One assertion after each wait would close both.
- **Stale `RESTORED_SHELL_BASE` doc** (*"the two restored-shell waits"* against six actual call sites, four of them git-worker waits). Correctly left alone as pre-existing and outside this PR's blast radius; flagged so it is not lost.

## 4. Nitpicks

- The second wait now carries two comment paragraphs — the original `#186` retained-sample one and the new load-scaling one — which overlap on why the sample is captured rather than re-read.
- `SHELL_PAINT_BASE`'s doc is the longest of the four bases. The content is accurate; the module is comment-dense by house convention.

## 5. Scores

- **CodeRabbit score (diff quality) — 4/5.** Both majors from the previous round are genuinely fixed in the changed code. The new conversion is correct: pre-poll ordering preserved, closure cheap and lock-free, retained-sample invariant kept via the established `wait_for_conflicted_entry` shape. The two remaining minors were documentation-accuracy issues about the new code, both since fixed.
- **Greptile score (codebase fit) — 4/5.** The stale-deferral finding required reading a test outside the diff and is now resolved accurately in-tree, which I re-confirmed independently. The release-gate exemption holds at the new head by the same two-stage path check plus `ships_nothing.py` exit 0, matching precedent #489. The one minor whose evidence lived outside the diff was the 4000ms provenance of the third call site, found via `git log -S`.

The change is a real improvement to three real flakes, and each round addressed the substantive feedback rather than arguing with it.

VERDICT: APPROVE

PR #502, opened 2026-09-06
feat(provenance): group a Source Control diff by seat

### LINKED ISSUE #349: Provenance blame: colour the gutter by who typed each line (you, the navigator, an agent pane)

vitali87:
## Summary

Closes #349 (provenance blame). The fourth and last slice, after #446 (the map), #497 (undo/redo and the gutter lens) and #499 (persistence beside local history): item 4 of the issue's design, "a Source Control diff view option 'group by seat' so a commit can be reviewed as 'the agent's part' and 'my part'".

## What changes

- **`Diff: Toggle Group by Seat`** (Command Palette) marks who wrote the lines a change **adds**: a thin bar in the diff's right gutter in that seat's hue, and a per-seat tally in the diff header (`• by seat: you 3 • agent (pane 2) 1 • unrecorded 2`). The status line says the same on toggle.
- **Only the lines the change adds.** Added and replaced rows are marked; an unchanged line's seat describes work this change did not do. A line with no record is counted as `unrecorded` and painted nothing, never guessed. A change that only removes lines reads `no added lines` rather than leaving the label empty.
- **Only where the join is sound.** Offered on the Source Control HEAD-vs-working view alone, whose right side *is* the working tree and whose row indices are therefore real 0-based file lines, the same key the provenance map uses. A diff parsed from raw `git diff` text numbers its rows by position in that text, so the same join would credit lines at random; the lens refuses it and says why.
- **Only onto the text it describes.** The map is loaded fresh when the lens comes on, and only when the newest history snapshot still holds both the bytes on disk *and* the lines the view is showing. A same-length rewrite inside one filesystem timestamp tick moves the file without moving the stamp that triggers a rebuild, so the view can be showing older lines than the snapshot describes; that case reads as unrecorded rather than crediting lines the map never saw.
- **Survives a rebuild.** When a view is re-read because its file moved (#471), the lens stays on rather than going dark mid-review, and its map is reloaded against the new rows. The map itself is deliberately not carried: it names line indices the rebuilt rows have renumbered.
- The tally counts the real rows, the basis staging uses, so the ignore-whitespace lens cannot drop a line the change actually wrote; the gutter bar keys on the same rows, so the header and the bars always agree.

- Three seats, three colours, correct names on hover: landed in #497.
- Survives save, reopen, and an undo/redo cycle: #497 and #499.
- Unknown lines are never attributed: unchanged, and enforced here too, at both the tally and the paint.

With this the issue's four design items are all shipped.

## Tests

Red first throughout. `added_lines_are_tallied_by_seat`, `added_lines_with_no_record_are_their_own_bucket`, `a_replaced_line_counts_as_added_for_its_seat`, `the_seat_summary_reads_as_counts_per_seat`, `the_seat_summary_of_an_unrecorded_change_says_unrecorded`, `the_seat_summary_of_a_deletion_only_change_says_so`, `the_lens_paints_a_seat_bar_on_added_lines_only` (whose fixture carries an unchanged line *with* a seat, so dropping the added-rows-only guard fails it) and `the_bars_and_the_tally_agree_under_the_whitespace_lens`.

At the app level: `the_group_by_seat_lens_reports_who_wrote_the_added_lines`, `the_group_by_seat_lens_refuses_a_diff_that_is_not_the_working_tree`, `the_group_by_seat_lens_refuses_a_map_for_lines_the_view_does_not_show`, `the_group_by_seat_lens_handles_a_lone_carriage_return` (the diff opener and the editor's loader split a lone `\r` differently, and the binding comparison has to match the opener), and `the_group_by_seat_lens_survives_a_rebuild_of_the_view`, which drives the real `refresh_open_diff_views` over a git repository fixture so deleting the production reload fails it.

Every fix on this branch was verified revert-sensitive. Related provenance, history and diff tests pass; clippy with `-D warnings` on all targets is clean.

Version 0.1.903 with `src/release_notes/0.1.903.md` (main is 0.1.902).

## Summary by CodeRabbit

* **New Features**
  * Added a Source Control option to group working-tree diffs by seat.
  * Added author-colored gutter markers for added and replaced lines.
  * Added per-seat change totals in diff headers, including unrecorded lines.
  * Added command palette access and documentation for toggling the new view.
  * The grouping view remains available as diffs refresh and files move.
* **Bug Fixes**
  * Prevented seat grouping from being enabled for unsupported diff types.
  * Improved handling of line matching and carriage returns.


vitali87:

CodeRabbit is rate-limited on this repo, so this branch went through three rounds of local Opus 5 review before the push. Rounds one and two returned REQUEST_CHANGES (two majors each); every finding was fixed with a failing test first and verified revert-sensitive. This is the round-three verdict, on commit `a26dd05`; the two minors it raises are fixed in `0d75813` above it.

---

# Review: `feat/provenance-seat-groups` (round 3)

## 1. Walkthrough

Adds the final slice of #349: a per-view "group by seat" lens on the Source Control HEAD-vs-working diff, which colours the right gutter of each line the change *adds* in its author's seat hue and puts a per-seat tally in the diff header, so a review reads as "the agent's part" and "my part". The lens is gated to `left_is_git_head` diffs (the only view whose right side is the working tree and whose `right` row index is a real file line), loads the persisted provenance sidecar only when the newest snapshot still matches both the bytes on disk *and* the lines the view is showing, and survives a view rebuild by reloading the map against the new rows.

### Round-two fixes — verified

1. **`split_into_lines` / lone `\r`** — **fixed.** `split_into_lines` is private again. The check now uses `text.lines().eq(right_lines.iter().map(String::as_str))`. I checked **both** builders: `open_head_diff_with_text` does `right_text.lines().map(str::to_string).collect()`, and `rebuild_diff_view`'s closure is identical. Equivalent for every case: trailing newline, no trailing newline, empty file, only-newlines, CRLF. The strict-UTF-8 decode matches the opener's `read_to_string` too.
2. **Gutter bar vs header tally** — **fixed.** The renderer takes `seat_row` from `diff.rows`, the same basis `added_lines_by_seat` iterates. I verified the index-alignment claim: `normalised_rows` is a strict 1:1 `map`, so `display_rows()` is always exactly the same length as `rows`, and it is written in only five places. The unified-deletion view has no seat-bar code and cannot reach this path. `row_idx` is unconditionally in range, so **the `unwrap_or(row)` is provably dead** — it cannot silently mix bases.
3. **`reload_seats_for` gates on `left_is_git_head`** — **fixed.**
4. **Rebuild test drives the real loop** — **fixed and it has teeth.** It uses `repo_with_seed`/`rewrite_later`, tags a real `HeadVsWorking` source, and calls the real `refresh_open_diff_views`. If `carry_view_from` stopped carrying `group_by_seat`, both the flag assert and the summary assert would fail. No `#[cfg(test)]` wrapper remains.

## 2. Findings

- **`src/widgets/editor.rs:1379`** **[minor/style]** The seat bar bypasses `theme.ui()`, unlike every other colour in `render_diff` and unlike the sibling provenance lens in the editor pane. Of the five seat hues, only `Seat::Me`'s has a light-theme mapping — so on a light theme the "you" bar renders washed-out light blue here while the identical bar in the editor renders the darker mapped blue: the same seat, two colours, on screen at once.

- **`src/widgets/diff.rs:1239`** **[minor/correctness]** `seat_summary()` returns the empty string when the diff adds nothing, producing a dangling label in both the header (`• by seat:  `) and the status line (`Diff: group by seat • `). Reachable from the normal Source Control flow: click a `Modified` file whose working copy only removed lines. The other "nothing to say" case is already handled deliberately — unattributed lines are *named* rather than omitted — so the empty case is the one gap in that stance.

### Deliberately not reported

- **The bar on a whitespace-lens `Equal` row is correct.** Under an active ignore-whitespace lens a re-indented line takes the equal background yet wears a seat bar from the real `Replaced` row. Judged honestly, this is the right trade and reads coherently: the header already says `• ignoring whitespace: leading`, so a bar on a visually-unchanged row parses as "this line *did* change, whitespace-only, and X wrote it". The alternative — the header counting a line the gutter does not mark — is the worse defect. The indices agree: `normalised_rows` preserves both `left` and `right` when it rewrites `Replaced` to `Equal`.

## 3. Test coverage

- The empty `seat_summary()` on a deletion-only diff — the gap behind finding 2.
- The `same_content_as` early-continue path in the refresh: correct (the map still describes the same lines) but unpinned.
- The light theme: no test renders the seat bar under one, which is why finding 1 is invisible to the suite.

## 4. Nitpicks

- The `diff.seats = Provenance::new()` reset before the reload reads slightly better as `default()`, matching the `unwrap_or_default()` used for the same purpose in `reload_seats_for`.
- `seats.truncate(right_lines.len())` is documented as provably dead; keeping it is defensible as a guard and the comment is honest about it.
- `self_history_root` is an unusual local name for a clone of `self.history_root`.

## 5. Scores

- **CodeRabbit score (diff quality) — 4/5**: both findings are minors living inside the new lines; no correctness defect in the new logic, and the seven new unit tests plus five app tests cover the tally, the bars, the gates, and the rebuild.
- **Greptile score (codebase fit) — 4/5**: one minor whose evidence is cross-file (the seat bar diverges from `render_diff`'s themed-colour convention and from the sibling provenance overlay); the palette wiring is complete across all four sites, the `left_is_git_head` gate matches the single production caller that sets it, and the docs row exceeds the existing convention.

Both round-two majors are fully fixed and the `unwrap_or` fallback flagged as a risk is provably unreachable rather than a silent base-mix.

VERDICT: APPROVE

---

**Both minors are fixed in `0d75813`**, with a test for the deletion-only summary. The suite selection for this branch is 34 tests, green, clippy `-D warnings` clean on all targets.

vitali87:
Fixed in `3ba5ffd`. You are right and the test was weaker than its own comment claimed: it described the same-length rewrite the `(mtime, len)` stamp gate cannot detect, but `OTHER` is one byte longer than `mine`, so the gate could see it and the case the test exists for went unexercised. The replacement is now `else`, genuinely the same length, and the test still passes for the right reason: the view keeps its pre-rewrite lines and the newer map is refused.

PR #503, opened 2026-09-06
test(terminal): put 17 more shell waits on the shared load-scaled base (part of #397)

vitali87:
## Summary

Part of #397, the first batch of the residual #501 named: 30 hand-rolled `waited < N` loops still in `src/app/tests.rs`.

This converts the 20 that already assert on timeout. Each waits on a real shell to produce something into a pane — a shell-integration decoration mark, an inline image, a broadcast line reaching every pane, a progress report, an OSC 7 host — which is the operation `SHELL_PAINT_BASE` already covers. Each keeps the budget it had at the floor and now stretches under load instead of failing on a busy box.

**Scope, deliberately narrow.** All 20 already asserted on timeout, so this is a conversion and not a change of teeth. The loops that instead `break` silently, and the ones doing work inside the body (pumping `drain_terminal_bells`, or asserting on intermediate state), are left for a later batch: each of those is a judgement about that test's operation rather than a rewrite, and mixing them in would hide the mechanical change among the interesting ones.

**Five raises, recorded as a rule.** Fifteen sites replaced 8000ms exactly; five replaced 4000ms and so are raises. `SHELL_PAINT_BASE`'s doc now states the rule — every call site reproduces its 8000ms, the rest are raises from 4000ms — rather than naming them, since a list needs an edit per batch and that is exactly the staleness the rule replaces. The doc also records that a wait whose old bound *exceeded* 8000ms would need its own base, since the floor assertion checks the constant rather than its call sites. A raise cannot introduce a flake; the assertion guards the direction that can.

**Predicates.** The rewrite was mechanical, then hand-polished, so the risk is a negation that silently weakens a wait: a test that still passes but no longer waits for what it did. Each closure is the exact negation of the loop condition it replaces, including the two compound conditions where de Morgan was applied by hand (`!(a || !b)` → `!a && b`), which still refuse to return true on a half-arrived condition, and two `count() < 2` conditions that became `count() >= 2`. Each description names what is awaited rather than restating the old failure text, since `await_spawned` renders it as "waiting for {what}", and names the sentinel it waits on where four sites would otherwise read identically.

## Tests

All 20 converted tests pass, as do the 12 `test_budget::` tests. Clippy with `-D warnings`, the doc-ownership gate and `ships_nothing.py` are clean.

## No version bump

Test-only: the diff touches `src/app/tests.rs` and the `#[cfg(test)]` module of `src/test_budget.rs`, so the shipped binary is unchanged and `scripts/ships_nothing.py` exits 0. Same as #501 and #489 before it.

10 `waited < N` bounds remain (8 at 8000, 2 at 4000), so the issue stays open.

Part of #397

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -2972,12 +2972,11 @@ fn terminal_decoration_dot_click_is_not_hijacked_by_the_sidebar_splitter() {
         tmp.path(),
     )
     .unwrap();
-    let mut waited = 0u32;
-    while app.terminals[0].command_decorations().is_empty() {
-        assert!(waited < 8000, "no decoration mark arrived");
-        std::thread::sleep(std::time::Duration::from_millis(40));
-        waited += 40;
-    }
+    crate::test_budget::await_spawned(
+        crate::test_budget::tests::SHELL_PAINT_BASE,
+        "the shell to emit a decoration mark",
+        || !app.terminals[0].command_decorations().is_empty(),
+    );
     let backend = ratatui::backend::TestBackend::new(120, 40);
     let mut term = ratatui::Terminal::new(backend).unwrap();
     term.draw(|frame| app.render(frame)).unwrap();
@@ -3643,18 +3642,18 @@ fn cmd_opt_up_parks_the_previous_prompt_mark_at_the_viewport_top() {
         tmp.path(),
     )
     .unwrap();
-    let mut waited = 0u32;
-    while app.terminals[0].prompt_lines().is_empty()
-        || !app.terminals[0]
-            .grid_lines()
-            .0
-            .iter()
-            .any(|l| l.contains("filler-39"))
-    {
-        assert!(waited < 4000, "mark/filler output never arrived");
-        std::thread::sleep(std::time::Duration::from_millis(20));
-        waited += 20;
-    }
+    crate::test_budget::await_spawned(
+        crate::test_budget::tests::SHELL_PAINT_BASE,
+        "the prompt mark and filler output",
+        || {
+            !app.terminals[0].prompt_lines().is_empty()
+                && app.terminals[0]
+                    .grid_lines()
+                    .0
+                    .iter()
+                    .any(|l| l.contains("filler-39"))
+        },
+    );
     app.focus_pane(Pane::Terminal);
     assert_eq!(
         app.terminal().viewport_top_line(),
@@ -3697,19 +3696,19 @@ fn ctrl_shift_space_quick_select_labels_matches_and_a_label_commits() {
     // sentinels, so wait until the SHA shows up on its own echoed row too
     // (a second occurrence); otherwise quick-select can open on a
     // header-only viewport and the hint set is timing-dependent.
-    let mut waited = 0u32;
-    while app.terminals[0]
-        .grid_lines()
-        .0
-        .iter()
-        .filter(|l| l.contains("de40bdf12ab34"))
-        .count()
-        < 2
-    {
-        assert!(waited < 4000, "output never arrived");
-        std::thread::sleep(std::time::Duration::from_millis(20));
-        waited += 20;
-    }
+    crate::test_budget::await_spawned(
+        crate::test_budget::tests::SHELL_PAINT_BASE,
+        "both copies of the pasted hash",
+        || {
+            app.terminals[0]
+                .grid_lines()
+                .0
+                .iter()
+                .filter(|l| l.contains("de40bdf12ab34"))
+                .count()
+                >= 2
+        },
+    );
     app.focus_pane(Pane::Terminal);
     assert!(app.terminal_quick_select.is_none(), "starts off");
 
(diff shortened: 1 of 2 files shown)


vitali87:

Both bots were unavailable on this PR (Greptile: out of credits for the billing period; CodeRabbit: "Review rate limited"), so this review was run locally. Round 1 (head `9853580`) raised a soft edge: three pure asserting loops had been deferred without matching the stated criteria. Those are now converted, taking the batch from 17 to 20. This is the round-2 result on head `bcfeac5`; its two findings were then fixed in `1bb1503`.

---

## Walkthrough

Twenty hand-rolled `while … { assert!(waited < N); sleep; waited += N }` loops in `src/app/tests.rs` are replaced by `test_budget::await_spawned(SHELL_PAINT_BASE, …)`, so waits on a real shell painting into a pane stretch under load rather than failing on a busy box (#397). Fifteen replaced an 8000ms bound exactly; five replaced 4000ms and are deliberate raises. `SHELL_PAINT_BASE`'s doc drops both its stale opening enumeration and its per-batch raise list in favour of a rule, and records that the floor assertion pins the constant rather than its call sites.

| file | summary |
|---|---|
| `src/app/tests.rs` | 20 `waited < N` loops → `await_spawned(SHELL_PAINT_BASE, "…", closure)`. Closures are exact negations of the loop conditions (two by de Morgan, two by `count() < 2` → `count() >= 2`). Four formerly identical `"the pane output"` descriptions now name their sentinels, and the progress one names its value. SHELL_PAINT_BASE call sites go 3 → 23. |
| `src/test_budget.rs` | `SHELL_PAINT_BASE`'s opening line now names the class of thing awaited; the raise paragraph states the rule without listing members; a new paragraph records that the floor assertion checks the constant, so a wait whose old bound exceeded 8000ms would need its own base. Assertion message reworded to match. Both hunks inside `#[cfg(test)] pub(crate) mod tests`. |

## Findings

I re-verified all 20 negations against the merge base and found 0 mismatches. The two `count()` rewrites are exact: `!(count < 2)` ≡ `count >= 2` on `usize`, with no off-by-one and no weakening — `>= 2` still refuses to fire on the header-only viewport (count 1) that the comment above the call site says the wait exists to exclude. The two de Morgan rewrites remain correct. Every removed line across the full diff is a loop condition or scaffolding, so no converted body dropped work.

**All three points from the previous round are genuinely fixed.** The exclusion criteria now match the code: I read all nine residual loops and each hits a stated exclusion — `drain_terminal_bells()` in body (three), `drain_fs_events()` in body (one), an intermediate `assert_eq!` before `break` (three), silent `break`/fall-through on timeout (two). No pure asserting loop remains, so the previous soft edge is closed rather than papered over. The residual tally is right and the earlier off-by-one is sidestepped by counting *bounds* rather than loops: `grep -n "waited <"` returns exactly 10 hits, 8 at 8000 and 2 at 4000, across 9 distinct loops. The doc's opening enumeration and raise list are both gone.

- **`src/app/tests.rs` (two `count() >= 2` descriptions)** **[minor/correctness]** Both said **"pasted"**, and nothing in either test pastes anything. `await_spawned` renders these as `waiting for both copies of the pasted hash`, which sends a reader hunting for a clipboard path that does not exist. The scripts are plain `echo`s, and the *reason* two copies exist is the run-label header: `PtyTerminal::spawn_with_preamble` writes `format!("\x1b[2m▶ {label}\x1b[22m\r\n")`, which contains the raw command text, so copy one is the header and copy two is the shell's own echoed row. The comment directly above the first call site says exactly this. The word also buried the thing that makes the predicate non-obvious — why `>= 2` rather than `>= 1`.
  **Fixed in `1bb1503`:** both now read "the hash/URL on its own echoed row as well as in the run-label header".

- **`src/test_budget.rs` (the new floor-assertion note)** **[nit/docs]** The note said a wait whose old bound exceeded 8000ms "would need its own base rather than this one — none exists today". True of `SHELL_PAINT_BASE`'s call sites but not of the wider tree: `src/widgets/terminal.rs` still carries a `waited < 20_000`. Out of scope for #397's `src/app/tests.rs` work and a different module, so the claim was defensible, but a reader taking it as repo-wide would be surprised.
  **Fixed in `1bb1503`:** scoped to "no such call site exists among these".

## Test coverage

- **The floor assertion still cannot see its call sites.** It asserts a property of the constant alone, so nothing mechanically catches a future batch converting a >8000ms loop onto this base and silently cutting it. This round now *documents* the gap, which is the right response — a source-scraping test would be worse than the note. Recorded as a known, deliberate limit rather than a defect.
- **No coverage regression from the conversion.** All 20 already asserted on timeout and `await_spawned` panics on timeout, so teeth are unchanged. The five raises can only lengthen a wait, which cannot turn a failing test green: each predicate is the same condition, just given more time to become true.

## Nitpicks

- The sentinel-naming fix works, but embedding escaped quotes renders as `waiting for the pane output "ccc3"`; `the pane output line ccc3` would avoid the escaping noise.
- `RESTORED_SHELL_BASE`'s doc still says "The base the two restored-shell waits use" against 6 call sites. Pre-existing and untouched by this PR: it is the identical staleness just fixed one constant down, and fixing it would leave the module consistent.

**Release gate:** correctly no version bump and no release note. CONTRIBUTING exempts `src/app/tests.rs` by path and "a `.rs` file whose diff touches nothing outside a `#[cfg(test)]` module"; both `src/test_budget.rs` hunks fall inside the module, and `ships_nothing.py`'s shape requirement is met.

## Scores

- **CodeRabbit score (diff quality) — 4/5**: all 20 negations verified correct including both `count() >= 2` rewrites, and no body work dropped; the only finding inside the changed code was two descriptions that misdescribed the mechanism as a paste, since fixed.
- **Greptile score (codebase fit) — 5/5**: no findings whose evidence lives outside the diff. The residual loops all match the stated criteria, the raise count and bound tally check out against the tree, `PtyTerminal::spawn_with_preamble` confirms the run-label header behaviour the `>= 2` predicates depend on, and the release-gate exemption is verified against both CONTRIBUTING and `ships_nothing.py`.

VERDICT: APPROVE

Issue #493, opened 2026-09-06
Editor::open runs unbounded work on the frame loop, now reachable from any pane via croft view


vitali87:
`Editor::open` does work of unbounded duration on the frame loop, and `croft view` (#362) makes that reachable by any process running in any pane rather than only by a human clicking the Explorer.

The path: `App::apply_view_request` calls `open_pinned`, which reaches `Editor::open` (`src/widgets/editor.rs`). For a PDF that is `open_pdf` at `:4446`, which calls `crate::pdf::rasterize_page`, which runs `Command::new("pdftoppm").output()` (`src/pdf.rs:369`) synchronously **with no timeout**. `open_archive` and `open_doc_preview` walk their containers on the same thread.

So `croft view big.pdf` from a build script parks the whole editor for as long as the render takes, and a `pdftoppm` that hangs (a malformed document, a stalled filesystem) parks it indefinitely. There is no size cap on that branch either: `MAX_FILE_BYTES` is applied after the extension dispatch has already returned for PDFs.

This is pre-existing, but the exposure is new. The FIFO refusal already in `apply_view_request` draws exactly this distinction:

> Neither is reachable through the Explorer without a deliberate human gesture, and this is reachable by any process in any pane.

The FIFO and `/dev/zero` cases were bounded by checking the file's KIND. That does not help here, because a regular PDF is a legitimate request; what is unbounded is the work, not the type.

**Not fixed in #452**, deliberately: moving the open off the frame loop changes how every viewer reports failure, which is a larger change than the channel that exposed it. #452 documents the caveat at the call site instead.

Options worth weighing:

1. A timeout on the `pdftoppm` subprocess specifically, which is the one unbounded external call, and the cheapest real bound.
2. A size cap on the request side of `croft view`, rejecting a routed kind above its viewer's practical limit before the open. Cheap, but it only covers the IPC entry point and leaves the Explorer click unbounded.
3. Move `Editor::open`'s expensive branches to the existing background-work pattern and reply `Ok` on dispatch. The complete fix, and the one that touches the most code.

vitali87:
claimed by M1 croft-2 (session croft-2): taking option 1 first, a bounded pdftoppm run with a clear error when it overruns, as its own PR; option 3 stays open for a follow-up.

vitali87:
claimed by M1 croft-2

Survey first, since one option is already partly shipped:

[...]

PR #504, opened 2026-09-06
fix(archive): bound a zip listing before the parse, and say why on refusal

vitali87:
## Summary

Part of #493. The issue lists three options; the survey below found most of it already handled, so this fixes the one genuine gap: a zip listing runs unbounded work on the frame loop.

## What the survey found

The issue's premise is partly stale, and it is worth recording why this change is narrower than the issue reads:

- **Option 1, a `pdftoppm` timeout, already shipped in #495.** `PDF_RENDER_BUDGET` bounds the subprocess, kills the process group on expiry and settles stderr, so the indefinite-hang case is closed.
- **The "cap applies after the dispatch" claim is true of the ordering but is not the defect it sounds like.** Every extension-routed kind bounds itself with a limit appropriate to it: images at `MAX_IMAGE_BYTES`, SVG at `MAX_SVG_BYTES`, documents at `MAX_DOC_BYTES`, tar at `TAR_LIST_CAP` with a decoded-byte cap on top for `.tar.gz`. The general cap sitting later is deliberate.
- **The zip branch had no bound at all**, and that is what this fixes.

## The defect

A zip's entry count is not bounded above by its file size. Measured at 87 bytes per member, a 33 MB file, well under the editor's 50 MB cap, declares 400k members and takes **1.65s to list on the frame loop**.

The first cut guarded the wrong side of that cost. `ZipArchive::new` parses the whole central directory before `len()` exists, allocating a name, extra field, comment and record per member; measured, that constructor is **71% of the total** (1.17s of the 1.65s). A guard on the entry count arrives too late to prevent most of what it is for.

## What changes

- **`ZIP_LIST_CAP` (8 MB), checked against the file size before `File::open`.** Entry count is bounded *below* by file size, so this is the only measure available before the parse, and it is what the tar branch already does. 8 MB admits ~97k members at the measured rate, ~400ms; the true worst case is ~178k, ~800ms, since the constructor reads only the central directory. For scale, a large library jar holds 38k members in 3.2 MB, so ordinary archives pass with room, and a test at 38k members pins that.
- **A refused archive still opens as hex, and says why.** The fall-through a corrupt archive already takes, but a browser silently replaced by a hex dump reads as a bug rather than as a policy. All four routes into `open_archive` (by extension, and by sniffed Zip, Gzip and Tar magic) carry the reason through a new `route_note`.
- **The reason is keyed to the file it describes, and cleared where it would go stale.** An open can arm the note and then fail before reaching any viewer, and "Reopen as Hex" reaches the hex viewer without passing through `open` at all. Keying stops a note explaining a *different* file; the two places that reach no fresh status line clear it so it cannot go stale against its *own* file, which is what the sibling `pdf_restore_page` already does for the same reason.

## Tests

Eight, each verified red before green, and five verified revert-sensitive individually: the cap itself, the `>` boundary at exactly the cap, an ordinary 38k-member archive still listing, the hex fall-through with its reason, the extensionless zip route, the `.tar.gz` decoded-cap route, the raw `.tar` extension route, and a stale note not explaining its own file later.

**One thing I could not prove, stated plainly.** The `Err(_)` arms distinguish a parse failure (no reason shown) from a policy refusal (reason shown). That distinction is real in the code and is why `is_list_cap_refusal` exists, but I wrote two tests for it and both stayed green with all four guards widened to catch every error. I could not explain the mechanism and deleted them rather than ship coverage I cannot demonstrate. The behaviour is unpinned; a reviewer who sees how to pin it should.

Also corrected along the way: I claimed the raw tar cap was unreachable through the editor. It is not — `kind_from_ext` runs above the size guard, so a large `.tar` does reach it. Documenting a live path as dead is the dangerous direction, and there is now a test on that route.

Version 0.1.904 with `src/release_notes/0.1.904.md` (main is 0.1.903).

## Summary by CodeRabbit

- **New Features**
  - Added safeguards that prevent oversized ZIP archive listings from freezing the editor.
  - Oversized or otherwise declined archive listings now open in the hex viewer with a status message explaining the reason.
  - Added bounded listing behavior for supported archive formats.

- **Documentation**
  - Documented archive listing limits and related behavior.

- **Chores**
  - Updated the package version to 0.1.904.


vitali87:

CodeRabbit is rate-limited on this repo, so this branch went through five rounds of local Opus 5 review before the push. Every round returned REQUEST_CHANGES and every finding was fixed with a failing test first. The round-five verdict is below, on commit `a8d2ace`; its finding is fixed in `7e78e1f`.

---

# Review: `fix/viewer-size-cap` (round 5)

## 1. Walkthrough

`zip::ZipArchive::new` parses the entire central directory before any entry count exists, so a zip under the editor's 50 MB cap can declare hundreds of thousands of members and park the frame loop for seconds. A new `ZIP_LIST_CAP` (8 MB) gates the zip branch on *file size* before the constructor runs, mirroring the existing tar gate. Because a size refusal is policy rather than a parse failure, the reason is carried to the hex fallback through a path-keyed `route_note`, and all four routes that reach `open_archive` share one predicate owned beside the messages it matches.

**Round four's raw-tar correction is confirmed correct.** I traced both routes independently. Extension route (live): `kind_from_ext` runs at `editor.rs:3922`, the `MAX_FILE_BYTES` guard at `:3980`, so a 100 MB+1 `.tar` reaches `list`, trips `TAR_LIST_CAP`, arms the note, and lands in hex with the reason. Magic route (genuinely dead for raw tar): the magic match is reached only after a read gated by `MAX_FILE_BYTES`, and 50 MB < 100 MB, so `Magic::Tar` can never reach that gate. The scoped comment is now accurate.

**On the deleted corrupt-zip test — I can explain the mechanism.** The natural fixture (`PK\x03\x04` + printable garbage) is **not binary**: no NUL, non-text ratio 0, so `is_binary` returns false. Under a widened guard the note *is* armed, but the file then falls past the hex fallback into the **text path**, which sets its own status and never reads `route_note`. `open_hex` is never called. The probe read `None` there because `open_hex` never ran. The assertion passed on the text status line, not the hex one — the test's *subject* was never produced. Right to delete it.

## 2. Findings

- **`src/widgets/editor.rs:4797-4812`** **[major/correctness]** — `open_hex`'s same-path refresh early-returns without consuming `route_note`, so a cap refusal can be re-armed and then leak into a later status line **for the same path**. Keying guarantees a surviving note never explains a *different* file; it does not guarantee a surviving note is still *true* about its own. Verified sequence: a 20 MB `.zip` opens as hex with the reason (note consumed); the file changes on disk, the FS sweep reloads, the note is re-armed and the refresh early-returns leaving it armed; the user shrinks the file to a valid 1 MB zip, which lists fine; "Reopen as Hex" then reports *"zip too large to list (20971520 bytes)"* for a file that is no longer too large. The field's doc comment argues against the sibling `pdf_restore_page`'s unconditional clear, but that sibling is the one that would have caught this. Fix: clear on the two paths that reach no fresh status line — the refresh, and a successful listing, which positively disproves any armed refusal for that path.

- **`src/widgets/editor.rs:2345-2347`** **[minor/docs]** — the doc comment's justification is wider than its mechanism. "Does not depend on finding every path" is true for the cross-file case and false for the stale-same-file case, and it is what invites skipping the clears.

## 3. Test coverage

The stale-same-path leak is untested and no existing test can reach it — every test opens each fixture once into a fresh `Editor`, so the note is always consumed or abandoned on the first open.

Everything else is well covered: the gate itself, the `>` boundary, the negative control that the guard cannot be satisfied by refusing everything (the 38k fixture computes to ~3.2 MB, a 5.1 MB margin), and all four routes.

## 4. Nitpicks

- The new test block strands the module's `use std::io::Write as _;` in the middle of the file.
- The tar.gz fixture relies on `Drop` to finalize rather than propagating errors like its sibling does.
- The refusal message reports the source size but not the cap it exceeded.

## 5. Scores

- **CodeRabbit score (diff quality) — 3/5.** One major inside the changed code: the new field's lifecycle is incomplete, so a refusal reason can outlive its condition. The gate, the four converted arms, the predicate and the boundary tests are correct, and the fixture math checks out.
- **Greptile score (codebase fit) — 3/5.** The same major's decisive evidence is outside the diff: the leak is reachable only through `reload_from_disk` and the `Cmd::ReopenAsHex` dispatch, and the repo's own `pdf_restore_page` precedent already solves this class with an unconditional clear.

VERDICT: REQUEST_CHANGES

---

**Fixed in `7e78e1f`**, with a test verified red first: it fails on the stale reason quoting the old byte count, and passes with both clears. The doc comment no longer argues against the precedent it should have followed.

The parse-versus-policy coverage the reviewer suggested is **not** included: I wrote it twice and it stayed green with all four guards widened, so it proved nothing. That behaviour ships unpinned and the pull request says so.

vitali87:
Thanks, but I am keeping the unconditional clear here, and the reasoning is worth recording because the previous review round arrived at the opposite conclusion first.

The refresh branch (`src/widgets/editor.rs:4804-4822`) writes **no** status line on any of its exits: it early-returns on pending edits, and otherwise refreshes the view and returns after updating only `disk_stamp`/`disk_conflict`. So clearing the note there removes nothing the reader would have seen. The status from the open that first showed the file already reads `Opened … in the hex viewer: Archive open failed: zip too large to list (…)`, and a refresh leaves it standing.

The alternative you describe — keeping a note the current `open` armed and clearing only older ones — is what this branch shipped two commits ago, and it is the defect `7e78e1f` fixes. A note that survives the refresh outlives the condition that armed it: the FS sweep reloads an over-cap archive, the user shrinks the file so it lists fine, and `Reopen as Hex` then quotes the old refusal with the old byte count. That sequence is pinned by `a_stale_refusal_reason_does_not_explain_the_same_file_later`, verified red before the fix.

So the trade is between a reason that is not re-displayed on a refresh that displays nothing, and a reason that can be wrong. I took the first. If the refresh ever does write a status line, the clear should move after it rather than become conditional.

PR #505, opened 2026-09-06
docs(remote): explain the declined background update

vitali87:
## Why

A user hit `Background croft update failed; staying on current version` in the status bar of a remote session and asked why. Nothing in the docs covered it, and the message is the hardest of the three update-failure messages to diagnose:

- It is the only one of the three (`Staged`, `SelfInstall`, `Watch`) that names **no log path** — the other two print theirs.
- The reason lands in the **launching** machine's `~/.cache/croft/install.log`, while the user is looking at a *remote* session on another host. So the natural place to look is the wrong one.
- The wording reads like a crash, but it is a deliberate refusal: with the fast cross-build path unavailable and a session already attached, the only route left is a from-source compile on the box the user is typing into, so `install_only_streaming_over`'s `present` gate declines it rather than loading their live session.

## What

One new `###` subsection in `docs/LINUX.md`, under the existing `## Remote: croft remote <host>` section: what the message means, why it is a refusal rather than a failure, where the reason is actually written, the prerequisites to install, and that a **reconnect** is required afterwards (the declined update is not retried on the running session).

Docs-only — no shipped change, so the release-hygiene gate's docs exemption applies.

## Verification

Every technical claim was checked against the source rather than inferred:

| Claim | Source |
| --- | --- |
| Refusal needs `present` **and** no cross-build | `src/remote.rs:376-387` |
| `clear_remote_updating` → marker vanishes → `Failed` → this string | `src/remote.rs:388-391`, `src/update_watch.rs:378-389`, `src/app/mod.rs:26208` |
| Log path resolves against the **local** `HOME` | `src/remote.rs:429-434` |
| `setup-cross` installs zig + cargo-zigbuild + targets, queries from the checkout | `src/cli.rs:1361-1386`, `1448-1474` |
| Not retried; reconnect required | `spawn_background_install` called once, `src/remote.rs:822` |
| Prerequisites named are the ones probed | `src/remote.rs:698`, `1838-1839` |

Both quoted log lines are verbatim against their `format!` strings.

Two corrections were made during review. An earlier draft called a toolchain bump "the usual cause" — inferred from a code comment, not evidence — and it now names the three prerequisites the code actually probes. The draft also called `setup-cross` "one command", but it prints a plan and blocks on `[y/N]`, and anything but `y` prints `Aborted.` and **exits 0**; following the doc unattended would have been a silent no-op, so the prompt, `--yes`, and "check the output not the exit code" are documented.

### DIFF
--- a/docs/LINUX.md
+++ b/docs/LINUX.md
@@ -55,20 +55,11 @@ The **launching** machine needs `rsync` on its `PATH` — croft uses it to sync 
 
 Only the very first connect (no croft on the box yet) waits for an install. Every later connect attaches to the installed croft immediately: when your local source is newer, the cross-build and ship run in the background while you work, and the running session offers `F9` to reload into the new binary once it lands. Background self-updates use a dedicated throttled SSH lane so install bytes never queue ahead of live keystrokes, keeping input latency at zero even while a newer binary streams in.
 
-#### "Background croft update failed; staying on current version"
+### "Background croft update failed; staying on current version"
 
-A background update declines itself, rather than failing, when you are
-already attached to the remote croft and the fast cross-build path is not
-available. Both halves have to be true, and the outcome is deliberate: the
-only route left is a `cargo install` on the box itself, and starting an
-unrequested compile on the machine you are actively typing into would spend
-its cores and disk on a build nobody asked for. So croft declines, clears
-the `updating` marker, and leaves you on the running binary. Nothing is
-broken and nothing is half-installed; the update simply did not happen.
+A background update declines itself, rather than failing, when you are already attached to the remote croft and the fast cross-build path is not available. Both halves have to be true, and the outcome is deliberate: the only route left is a from-source compile on the box itself, and starting an unrequested build on the machine you are actively typing into would spend its cores and disk on work nobody asked for. So croft declines, clears the `updating` marker, and leaves you on the running binary. Nothing is broken and nothing is half-installed; the update simply did not happen.
 
-The status line is deliberately short, and the reason is written to the
-install log on the **launching** machine — the Mac or laptop you connected
-*from*, not the server named in the message:
+The status line is deliberately short, and the reason is written to the install log on the **launching** machine — the Mac or laptop you connected *from*, not the server named in the message:
 
 ```bash
 tail ~/.cache/croft/install.log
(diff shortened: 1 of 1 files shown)


vitali87:

CodeRabbit's check on this PR reports `pass` while annotated **"Review limit reached"** — the free OSS quota for this repo is exhausted, so no review actually ran and `reviews[]` is empty. This local Opus 5 review fills that gap.

**Verdict: REQUEST_CHANGES** (2 major, 2 minor). **All four are fixed in `4622849`**, pushed after the review; details at the bottom.

---

### Walkthrough

Adds a troubleshooting subsection to `docs/LINUX.md` explaining the status-bar string "Background croft update failed; staying on current version": that it is a deliberate refusal rather than a crash, where the reason is logged, and how `croft setup-cross` fixes it. Every substantive claim was checked against the source and holds — the refusal gate, the marker/watcher path, the local log location, `setup-cross` behaviour, the no-retry-so-reconnect rule, and the dialog-still-up fallback prompt. The defect was in the two quoted log lines, which readers are told to search for and which did not match the code.

| file | summary |
| --- | --- |
| `docs/LINUX.md` | New `### "Background croft update failed…"` subsection between the background-update paragraph and `### Surviving sleep and network drops`. |

Verified correct: the `present && cross-build unavailable` gate (`src/remote.rs:378-387`, doc-comment `:325-331`); `clear_remote_updating` on the declined path (`:388-391`, `:2810-2817`) feeding `UpdateEvent::Failed` when the marker vanishes without the stamp advancing (`src/update_watch.rs:378-389`) to the exact string at `src/app/mod.rs:26209`; `install_log_path()` resolving against the local `HOME` (`src/remote.rs:429-434`); `setup_cross` installing zig + cargo-zigbuild + both musl targets, `[y/N]`, `Aborted.` with `return Ok(())` (`src/cli.rs:1361-1387`), idempotent skips (`:1396`, `:1428`, `:1463`), both rustup calls `.current_dir(checkout)` (`:1450-1470`); `spawn_background_install` called exactly once at `src/remote.rs:822` with a hardcoded `&mut |_| false` (`:857`), so a decline is never retried; `confirm_fallback` reaching the dialog when `present == false` (`src/install_session.rs:64-81`); prerequisites matching `cross_compile_unavailable_reason`'s probes (`src/remote.rs:1832-1849`) plus `rust_target_installed` (`:2241-2254`).

### Findings

- `docs/LINUX.md` (log-quote block) **[major/docs]** The first quoted line is not what croft writes. `src/remote.rs:697-700` composes `reason` as ``rustup target `{triple}` missing (run `rustup target add {triple}` once to enable the fast path)`` then logs `Local cross-build skipped: {reason}`. The doc dropped the parenthetical — which names the one-command fix. The section tells readers to search the log, so a truncated quote makes a literal search fail and hides the remedy the log itself offers.
- `docs/LINUX.md` (same block) **[major/docs]** `Update NOT installed: cross-build unavailable (...)` substitutes a literal `(...)` for the content the reader is sent to find. `src/remote.rs:380-382` writes the reason plus `Run \`croft setup-cross\` on this machine, then reconnect — updates then ship a prebuilt binary in seconds.` The elided `{reason}` is the diagnostic value; the elided sentence is the fix.
- `docs/LINUX.md` (`tail ~/.cache/croft/install.log`) **[minor/docs]** Every line in that file is timestamp-prefixed — `spawn_log_tee` writes `writeln!(f, "[{secs}] {line}")` (`src/remote.rs:419`) — but the quoted block shows none, so a reader comparing sees a mismatch on every line.
- `docs/LINUX.md` ("logged on the launching machine") **[minor/docs]** `install.log` is truncated per session, not appended: `File::create` at `src/remote.rs:411`. A user who reconnects before reading it — which this section instructs — destroys the evidence.

### Test coverage

None. Docs-only; the described behaviours are already covered (`src/widgets/connect_dialog.rs:681` pins the fallback prompt, `src/update_watch.rs` tests cover marker-vanished-without-stamp, `src/remote.rs:3935-3937` pins the session check before the compile).

### Nitpicks

- "Two things that catch people out." — headless fragment; the file uses full sentences.
- "the musl rustup targets" — plural is right (`CROSS_TARGETS`, `src/cli.rs:1358`) but the quoted line named only x86_64, reading as a contradiction.

Release gate: **exempt**. `.github/workflows/ci.yml:70-72` filters shipped files with `grep -E '^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)'`; `docs/LINUX.md` matches none.

### Scores

- **CodeRabbit score (diff quality) — 2/5** — two majors: both quoted log lines differed from the `format!` strings readers are told to search for.
- **Greptile score (codebase fit) — 4/5** — heading depth, line-wrapping style and voice match the file; two minors on log-file mechanics.

**VERDICT: REQUEST_CHANGES**

---

### Fixes applied in `4622849`

Both quoted lines now match their `format!` strings verbatim, including the `[timestamp]` prefix (confirmed against a real `install.log` on this machine). Added "Read it **before** reconnecting: each connect truncates the file." Both musl targets are named, and the fragment is a full sentence.

PR #507, opened 2026-09-06
test(terminal): put the last nine shell waits on the load-scaled base

vitali87:
## Summary

Part of #397, and its last batch: **no hand-rolled `waited < N` loop is left in `src/app/tests.rs`.** All nine move to `test_budget::await_spawned` on the shared `SHELL_PAINT_BASE`, so they stretch under load instead of failing on a busy box. Call sites go 24 → 33.

Three shapes, all deferred by #503 for reasons that do not survive contact with `await_spawned`'s signature:

- **Pump in the loop body** (four). It takes `FnMut`, so `drain_terminal_bells()` and `drain_fs_events()` move into the closure. #503 said work in the body blocked conversion; it does not.
- **Length check then equality assert before `break`** (three). The length becomes the predicate and the assert moves after the wait, where it reads a settled buffer: each sink is written by `dd bs=1 count=N`, so once `len() == N` the content is final.
- **Silent `break`** (two). Both were already safe — a standalone assert followed each — so the teeth are unchanged and only the failure message moves.

## A wrong turn on this branch, recorded rather than buried

An earlier revision left the fs-events loop unconverted behind a comment claiming a conversion "drains and checks in the same instant" and fails at 12s. **Both claims were wrong.** `await_spawned` sleeps 20ms *between* `ready()` calls, so from the second poll a converted closure gets the same drain → sleep → drain sequence the loop gave it. The 12s came from a loaded box taking the ×3 scale — the panic string said `x3 for load` in the same sentence I misread — not from an event that never arrives. Converted, the test passes in ~0.47s over five runs.

The local reviewer caught it by converting the loop and running it, which is the only thing that could have caught it: the false comment was self-consistent and cited a real measurement.

## Predicates

Each closure is the exact negation of the condition it replaced, including two de Morgan cases (`while waited < 4000 && foreground_is_shell()` → `|| !foreground_is_shell()`) and two `count() < 2` → `count() >= 2`. Descriptions name what is awaited, since `await_spawned` renders them as "waiting for {what}".

Poll frequency changes for four sites that slept 40ms, since `await_spawned` polls at 20ms. That changes cadence, not budget, and cannot turn a passing test red.

## Tests

All nine converted tests pass by exact name (9 passed, 0 failed, 1.28s); `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean. Full suite: 4249 passed, 2 failed — both confirmed pre-existing on `origin/main` (a missing local cross-target toolchain and a sandbox socket bind).

## No version bump

Test-only: the diff is `src/app/tests.rs` alone, which CI filters out by path before `ships_nothing.py` runs. Same as #489, #501 and #503.

Part of #397

## Summary by CodeRabbit

- **Tests**
  - Improved automated test reliability by standardizing asynchronous wait handling.
  - Preserved existing validation checks for process status, foreground groups, captured output, history, and sink data.
  - No user-facing behavior changes.

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -4107,17 +4107,14 @@ fn trigger_hits_surface_in_the_status_bar_via_the_drain_tick() {
     .unwrap();
     app.drain_terminal_bells();
     app.terminals[0].write_input(b"\n");
-    let mut waited = 0u32;
-    while !app.status.starts_with("Trigger in") {
-        app.drain_terminal_bells();
-        assert!(
-            waited < 8000,
-            "trigger never surfaced, status: {}",
-            app.status
-        );
-        std::thread::sleep(std::time::Duration::from_millis(40));
-        waited += 40;
-    }
+    crate::test_budget::await_spawned(
+        crate::test_budget::tests::SHELL_PAINT_BASE,
+        "the trigger to surface in the status line",
+        || {
+            app.drain_terminal_bells();
+            app.status.starts_with("Trigger in")
+        },
+    );
     assert!(
         app.status.contains("build: OK"),
         "interpolated message expected, status: {}",
@@ -9475,13 +9472,16 @@ fn change_workspace_root_seeds_cd_while_shell_startup_owns_the_tty() {
     )
     .unwrap();
 
-    let mut waited = 0u32;
-    while waited < 4000 && app.terminal_mut().foreground_is_shell() {
-        std::thread::sleep(std::time::Duration::from_millis(20));
-        waited += 20;
-    }
+    crate::test_budget::await_spawned(
+        crate::test_budget::tests::SHELL_PAINT_BASE,
+        "the startup child to take the foreground group from the shell",
+        || !app.terminal().foreground_is_shell(),
+    );
+    // await_spawned panics on timeout, so this pins the state the rest of
+    // the test reads rather than guarding the wait, the same shape as
+    // wait_for_conflicted_entry.
     assert!(
-        !app.terminal_mut().foreground_is_shell(),
+        !app.terminal().foreground_is_shell(),
         "precondition: the startup child must own the foreground group"
     );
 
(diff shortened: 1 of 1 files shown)


vitali87:

Both bots are silent on this PR (zero entries in `reviews[]`, zero comments), so this review was run locally by the `greptile-local` agent. It reviewed the tip at the time, `c8afab2`.

**Confidence Score: 4/5** — one incorrect comment on the changed surface; the code itself correct and the tests passing.

### The finding, and it was right

**[P1] The "NOT converted" comment stated a measurement that does not reproduce, and its mechanism argument was false** — `src/app/tests.rs:40181-40196`

That revision left the fs-events loop unconverted behind a comment claiming a conversion "drains and checks in the same instant" and fails at 12s. The reviewer converted it and ran it five times: `ok` at 0.46–0.52s each.

Both claims were wrong, and the reviewer's reasoning is correct at `src/test_budget.rs:220`:

- `await_spawned` sleeps 20ms *between* `ready()` calls, so from the second poll a converted closure gets exactly the drain → sleep-20ms → drain sequence the hand-rolled loop gave it. The two are equivalent.
- The 12s was the *budget*, not evidence of starvation: `2s × BASE_CALIBRATION(2) × scale(3)` = 12s. The panic string interpolated `x3 for load` in the same sentence I misread. The box was loaded by my own concurrent runs.

I reproduced the reviewer's result independently before accepting it — five runs at 0.50 / 0.48 / 0.46 / 0.46 / 0.48s. **Fixed by converting the loop and deleting the comment**, so this PR now converts all nine and closes #397's class at zero rather than eight-plus-a-caveat.

### Verified, not defects

- All nine closures are exact negations, including the two de Morgan cases and two `count() < 2` → `count() >= 2`.
- The three moved `assert_eq!`s cannot read a partial buffer: each sink is written by `dd bs=1 count=N`, so once `len() == N` the content is final. Verified via a FIFO.
- The four `drain_*` closures are sound — the pumps are the event loop's own per-tick calls, non-blocking, and were called every iteration by the base loops too.
- Budget arithmetic reproduces: 8s at the floor, sites replacing 8000ms reproduce exactly, the two replacing 4000ms are deliberate raises.
- Release gate correct: `src/app/tests.rs` is filtered by path before `ships_nothing.py` runs. No bump, no release note — matching #489, #501, #503.

### Corrections to the brief I gave the reviewer

- `MAX_SCALE` is **4**, not 8 (`src/test_budget.rs:34`). The 8 is `BASE_CALIBRATION * MAX_SCALE`, the total stretch. My error, caught by the reviewer.

### Verification run

Nine named tests by `--exact`: 9 passed / 0 failed. Full suite on the branch: 4249 passed / 2 failed / 11 ignored — both failures confirmed pre-existing on `origin/main` (a missing local cross-target toolchain, and a sandbox socket bind). `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

VERDICT: APPROVE (after the P1 was fixed)

Issue #318, opened 2026-08-27
Compounds: honour preLaunchTask, and decide on presentation/stopAll


vitali87:
`launch.json` compounds carry five properties: `name`, `configurations`,
`preLaunchTask`, `stopAll`, `presentation`. croft's `parse_compounds` reads the
first two and discards the rest.

That was harmless while no compound could launch. #311 makes a **one-member**
compound launch its single member — one session, which croft has supported since
#250 — and launching via the member alone runs neither the compound's own
`preLaunchTask` nor respects its `presentation`.

#311 therefore refuses a one-member compound that actually asks for one of those
keys, rather than launching it and silently dropping them. That is the safe
behaviour, but it is a placeholder: the user is told croft cannot honour the key,
not that it ever will.

### What is still missing

**`preLaunchTask` is the tractable one.** The machinery already exists a few
lines away — `launch_debug_config` resolves a *member's* `preLaunchTask` against
the task registry, runs it in its terminal pane, and gates the launch on exit 0.
A compound's own task wants exactly that treatment before its member starts.
Honouring it would turn the refusal into a launch.

**`presentation`** (`hidden`, `group`, `order`) affects picker ordering and
visibility. croft's picker has no grouping concept yet, so this is a larger
change than it looks.

**`stopAll`** governs whether ending one session ends the others. It is
meaningless for a single session and is deliberately *not* part of the
one-member refusal gate — at either value, including `false`, which is VS Code's
own default. It becomes real only alongside the multi-session work in #310.

### Why this is not #310

#310 is the multi-session limitation, and a one-member compound needs one
session. Citing #310 for a key-based refusal tells the user something false
about their own `launch.json` — the same error #311 exists to correct on the
adjacent path. The refusal should name this issue instead.

### Note on the multi-member path

A compound with several members reports #310 and never mentions a dropped
`preLaunchTask`. The precedence is right (the session count is the harder
blocker) but the user cannot see both obstacles at once: they split the compound
to work around #310 and then meet the key refusal. Worth surfacing both.

vitali87:
## `preLaunchTask` has shipped — this issue is now only `presentation` and `stopAll`

Checked against `origin/main` before picking this up, and the tractable half is already done. Recording it so the next session does not re-derive the same thing.

**Shipped** in `f21dcef` ("feat(debug): run a compound's own preLaunchTask (#318)"), on main now:

[...]

Issue #149, opened 2026-08-15
Multi-root Phase 2a: per-root git workers; Source Control follows the active file's repo


vitali87:
Multi-root workspaces Phase 2, first slice, on the folder-set gestures (#147). Git state is still singular: one `GitWorker` rooted at the primary, one shared `GitStatus`, and every consumer (status line, gutter baselines, blame, Source Control, the ignored set) reads it — so a secondary root's repository is invisible: no branch pill when its files are focused, no gutter marks, no panel entries, and its files never dim as git-ignored.

Slice 2a:

- One `GitWorker` per workspace root, keyed by root, each discovering its own repo toplevel (#139's discovery already rides the status poll). Responses stay root-scoped by construction — each worker's drain lands in its root's slot.
- Per-file consumers resolve through the OWNING root's worker: gutter baselines, inline blame, the timeline, and `scm_root` (which becomes owning-root-aware for panel operations on the active repo).
- The status line's branch pill and the Source Control panel follow the repository of the root owning the focused file, falling back to the primary — the panel still shows ONE repo at a time in this slice; VS Code's all-repos sectioned view is the next slice.
- The Explorer's ignored dimming unions every root's ignored set (each is cwd-scoped to its root already).

With one folder, behavior is byte-identical: one worker, the same status, the same panel.

PR #508, opened 2026-09-06
docs: simplify the language, drop the changelogs and issue refs

vitali87:
Three passes over every doc except `CLAUDE.md`, which is deliberately excluded — it is the instruction file agents are bound by, and rewording its rules risks changing what they require.

## 1. Plain language

Shorter sentences, less throat-clearing, the same information in fewer words. 12 of the 13 files shrank.

## 2. No release narrative

`MULTIPLAYER.md` was substantially a changelog: `PR 3a ... SHIPPED in 0.1.627`, headings titled by release, prose narrating how 0.1.634 became 0.1.635. A design doc says how the thing works; git already records when it changed. The "Phased PR plan" section is gone entirely.

**Kept:** `## Field test (0.1.630, real SSH, Linux musl)`. That version names which binary a real-hardware test actually ran on — evidence, not history.

## 3. No bare issue refs

111 of them sent the reader to a browser to learn something the sentence should have said. Each is now the fact it stood for: `Multi-root (#149): one GitWorker per root` became a statement about multi-root support.

**Kept:** four pointing at genuinely open work (#310, #318, #363 — verified OPEN via `gh`), written as links, plus `rust-lang/cargo#8684` cited as upstream evidence.

## 4. ARCHITECTURE.md is restructured, not trimmed

Its module tree had grown descriptions of up to **17,000 characters on one line inside a code fence** (`terminal.rs`), with 36 entries holding 93,442 of the tree's 152,468 characters. That is not a navigation index any more.

The tree is now short labels; the design reasons moved into 73 `###` prose sections under `## Module notes`, where they can be read — the HUP-trapping shell that left a zombie per closed pane, the `.env` destroyed while status reported the tree clean, the NTP step-back that must be clamped rather than dropped.

A 16-line block about file encoding was also rescued: it sat in the tree indented as a continuation with **no owning module row**, and would otherwise have been dropped silently. It is now `### File encoding`.

## On the word count

**65,671 → 66,465 words. This PR does not make the docs shorter.**

Twelve files shrank; `ARCHITECTURE.md` went 22,993 → 25,406 and swallowed the gains. The reason is worth stating plainly: the first version of the restructure looked like a 2,000-word saving, but review showed part of that "saving" was **deletion** — 32 modules had reasoning cut with no prose section to receive it, ~19,500 characters including the `ZIP_LIST_CAP` arithmetic and why retiring a port cancelled a live SSH tunnel. Restoring those reasons as readable prose costs more than the compressed one-line form did.

The docs are better. They are not shorter. Accepted deliberately rather than by cutting rationale.

## Verification

`scripts/check_arch_content.py` is added as a gate for the failure mode found in review. The obvious check is the one that fails silently: verifying module **rows** and identifiers survive stays green while the reasoning *after* each row is cut away. The gate instead diffs each module's tree-description length against the base and demands a `###` section wherever it shrank by more than 400 characters.

It went through the same trap once itself — it keyed on filename, and nine basenames repeat in this tree (`registry.rs` ×4, `mod.rs` ×3), so 12 rows carrying 8,364 characters were unmonitored. It now keys on the full tree path, and that fix immediately surfaced two more real losses (`collab_agent.rs`, `pair/local.rs`), both since restored.

Controls, all re-run against the final script:

| Injected regression | Result |
| --- | --- |
| `src/remote.rs` gutted 5395 → 161, section deleted (duplicated basename) | caught, full path named |
| `src/widgets/remote.rs` row deleted | caught as a dropped row |
| `archive.rs` gutted 1037 → 20, section deleted | caught |
| Nothing changed (identical files) | passes, exit 0 |

Also verified: all 152 module filenames present, every code identifier per module reappearing in its prose, no duplicate headings, no skipped heading levels, fences balanced, `KEYBINDINGS.md` at 457 table rows before and after with an identical first-column set, and no `CONTRIBUTING.md` rule weakened.

The release gate needs no version bump here: the `shipped` filter selects only `src/`, `assets/`, `build.rs` and the Cargo files, and this branch touches none of them.


vitali87 (commented):
LGTM

PR #511, opened 2026-09-06
docs: restore the module reasoning and the content gate that main is missing

vitali87:
#508 was merged at `0bc0cfc`, the commit *before* its review fixes. The fallback review on that PR returned REQUEST_CHANGES with three majors; I fixed all three plus a fourth in `cc525ed` and pushed `6d19bb5`, but GitHub's PR head lagged and never updated, so the merge took the reviewed-and-rejected state.

`main` today is the version that scored **3/5**. This PR is `cc525ed`, cherry-picked onto current main. The re-review of that exact commit scored **5/5 APPROVE**, verified with positive controls rather than inspection.

## What main is missing

| | on `main` now | with this PR |
|---|---|---|
| `### ` module sections | 73 | 100 |
| `SHRINK_THRESHOLD` | 400 | 120 |
| Gate wired into CI | no | yes |
| Unit test for the gate | none | 9 cases |

## The three majors

**27 modules lost their reasoning with nowhere to put it.** ~19,500 characters deleted, not relocated: `widgets/output.rs`'s `dropdown_scroll` and "overflow channels used to be unreachable by any gesture"; `plot.rs`'s reason the sixel probe cannot be used there ("needs a raw tty on stdin, which is the data pipe here"); `media.rs`'s header-parser list; `sqlite_view.rs`'s "no system dependency"; `session.rs`'s phantom-row fact.

**The threshold was tuned to pass.** I set 400 on the argument that both error directions are cheap. The largest unguarded shrink was 353 — the gate sat just above the cluster of real losses. Now 120, taken from the measured distribution, with the mistake recorded in the docstring rather than quietly corrected. The reviewer sanity-checked it independently: the nearest unsectioned loss below 120 is 20 characters away, and everything above it has a section.

**The gate ran nowhere.** `grep` for `check_arch_content` returned only its own docstring — no ci.yml step, no test. A guard nothing executes cannot fail, which is the exact shape it was written to catch. It is now a step in the `docs` job on the same merge-base contract as `check_doc_ownership.py`.

## A fourth, found by checking rather than assuming

The reviewer listed `session.rs`'s phantom-row fact as lost. Rather than assume the restore covered it, I grepped: it had not. `session.rs` and `dap/session.rs` both exist, and one `### session.rs` heading satisfied the gate for **both** — so the shorter module kept a lost fact while the gate reported success. That is the same collapse-on-basename bug in its third form (first a threshold above the losses, then a dict keyed on filename, then heading-matching).

`section_keys` now accepts any unambiguous path suffix and refuses a bare filename two rows share, which forced four headings to be qualified. The reviewer reproduced this: renaming `### src/session.rs` back to a bare `### session.rs` makes the new code flag **both** rows where the old code flagged **neither**.

## Verification

`scripts/tests/test_check_arch_content.py`, 9 cases: path-keying, the heading-anchored fence, the threshold either side, a dropped row, and the same-basename regression. Full script suite **125 tests, OK** (`/tmp/cgr-suite-logs/20260906T234618Z-82224.log` — the summary is in the log, not the terminal).

The reviewer additionally ran four positive controls against the real document, and measured every row rather than only the ones named: of 101 shrunken rows, the only three without sections are −100, −48 and one fully carried by `### testing/`.

Docs plus `scripts/` — outside ci.yml's `shipped` filter, so no version bump.

## Summary by CodeRabbit

* **Documentation**
  * Expanded architecture documentation with clearer source references and descriptions of key modules and capabilities.
  * Clarified terminal problem-entry behavior in matcher documentation.
  * Updated Windows guidance for using WezTerm with the real UI.

* **Quality Improvements**
  * Added automated checks and tests to detect unintended reductions or omissions in architecture documentation.
  * Improved validation for modules with similar filenames and documentation layout changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,6 +175,16 @@ jobs:
           set -euo pipefail
           base="$(git merge-base "origin/$GITHUB_BASE_REF" HEAD)"
           python3 scripts/check_doc_ownership.py "$base" HEAD
+      # ARCHITECTURE.md's tree carries the reasons behind each module, and a
+      # rewrite can cut a line's rationale away while the row and its
+      # identifiers survive - which is what a row-level or identifier-level
+      # check sees as a pass. This one compares per-module description
+      # length and asks for a `### <module>` section wherever one shrank.
+      - name: No module may lose its reasoning
+        run: |
+          set -euo pipefail
+          base="$(git merge-base "origin/$GITHUB_BASE_REF" HEAD)"
+          python3 scripts/check_arch_content.py "$base"
 
   test:
     name: clippy + tests
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -291,7 +291,7 @@ A tree-sitter highlighted editor with a full write path, mouse-drag selection, n
 
 **Breadcrumbs and sticky scroll.** `EditorTabs` renders both. App pushes `breadcrumbs` and `sticky_lines` each frame from the outline scope chain, and `breadcrumb_target_at` / `sticky_line_at` map clicks to jumps. Crumbs are budgeted and advanced in display CELLS via `set_stringn`, so a CJK segment is neither overpainted by the next crumb nor mis-hit. The sticky band floats over content but stops above the caret's PAINTED row, read back out of the frame's own row layout so a collapsed fold or comment box between the top and the caret cannot make the band overshoot. A caret under the band cannot be painted at all, so `caret_floor_row` keeps the scroll from parking it there in the first place — dragging it to the first row the band does not cover, instead of to `scroll`, which had suppressed the band for the whole of a wheel scroll.
 
-### remote.rs
+### src/remote.rs
 
 Remote (SSH) target metadata and launch dispatch, plus the ssh-pane re-root offer.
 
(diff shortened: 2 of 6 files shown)


vitali87:
Both flagged items addressed — one taken, one declined with a reason.

**Title check — taken.** "merged without" was a dangling clause. Retitled to "docs: restore the module reasoning and the content gate that main is missing", which names what the diff does rather than what happened to a previous PR.

**Docstring coverage — declined, with reasoning.** The 31.25% is measured across `scripts/check_arch_content.py` and `scripts/tests/test_check_arch_content.py`. The gate script's own functions are documented, including a module docstring explaining why the check exists and why the obvious version of it fails silently. The uncovered functions are the nine test methods, and their names are the specification:

```
test_keys_on_path_so_repeated_basenames_stay_distinct
test_only_the_gutted_one_of_two_same_named_rows_is_reported
test_gutted_row_with_no_section_is_reported
test_just_under_the_threshold_is_not
```

A docstring restating those would add characters without adding information. Where a test needed something its name could not carry, it has one — six of the nine do, and the class-level docstrings explain the two blind spots the file exists to pin. That matches the convention in `scripts/tests/`, where `test_ships_nothing.py` documents the module and the non-obvious cases rather than every method.

I would rather leave the coverage number low than pad it, since padding is the failure mode the metric is trying to prevent.

Worth noting for the record: this PR exists because #508 was merged at a stale head. Its review fixes were pushed as `6d19bb5`, GitHub's PR object kept reporting `0bc0cfc`, and the merge took the older commit — so `main` today carries the version a review scored 3/5, with the reasoning gate present in `scripts/` but wired into no workflow. That last part is the reason this is worth landing promptly: a guard nothing runs reads as coverage while providing none.

Issue #473, opened 2026-09-05
No leaking of text in terminal is allowed


vitali87:
<img width="836" height="231" alt="Image" src="https://github.com/user-attachments/assets/9cdfe5ac-19f8-4300-b5c4-7781c09cbd75" />

vitali87:
claimed by M1 croft-2 (session croft-2): #475 covered the resize half (repaint parsed into the old grid); taking the remaining leak, where cells of a cleared full-screen child stay painted, next. Investigating the terminal widget's cell-painting path first.

vitali87:
Progress note from the investigation so far.

- #475 fixed one half: a full-screen child's repaint after a resize was parsed into the old grid. That is merged (0.1.870).
- The painter is not the remaining leak: `PtyTerminal::render` writes every cell of the viewport into the frame buffer each paint (`src/widgets/terminal.rs`, the `for x in 0..cols` loop ends in `set_symbol` + `set_style` for every cell), so nothing from a previous frame can survive in croft's own buffer. The fragments in the screenshot are therefore in the alacritty grid: cells the child wrote and never erased.
- The fragments read like remnants of Claude Code's box ("aude co", "chat", "g-g sa"), scattered at the right-hand side. Claude Code does not use the alternate screen; its renderer erases its previous output by counting the lines it wrote. After a width change alacritty reflows the primary screen, the old output occupies a different number of rows, and a line-count erase misses the difference. That leaves exactly this kind of residue in any terminal, and croft resizes the grid every time the pane's inner width changes (split drags, side-bar auto-hide, the seam strip), so it will hit it more often than a window that is resized once.

To confirm and fix this I need the steps behind the screenshot: what was running in the pane (Claude Code?), and what happened just before the text was left behind (a resize, a split, the side bar collapsing or expanding, scrolling?). If it is the reflow case, the fix is on croft's side: either suppress reflow for a pane whose child is line-erasing (alacritty's `reflow` config), or debounce grid resizes so a transient width during a layout change never reaches the grid.

vitali87:
Not obvious where this happened

PR #512, opened 2026-09-07
fix(terminal): drop last_inner with last_area when a frame skips a pane

### LINKED ISSUE #510: fix croft overflow (picture attached)
Image

vitali87:
Closes #510.

## The mechanism

`last_area` and `last_inner` are assigned on **adjacent lines** by the same function (`terminal.rs:4474-4475`). After that they are treated completely differently: `last_area` is cleared in thirteen places, including for a pane the frame declines to paint (`mod.rs:15440`); `last_inner` had exactly one production write and **zero** clears. One of a pair is defended and its twin is not.

That matters because `last_inner` anchors the terminal image overlay, which is emitted *after* `terminal.draw()` with absolute cursor positioning and is bounded by nothing else:

```rust
// mod.rs:48343-48357
let _ = write!(out, "\x1b[{};{}H", layout.cell_y + 1, layout.cell_x + 1);
let _ = out.write_all(osc.as_bytes());
```

`update_terminal_image_overlay` validates the payload against that *same* stale rect (`mod.rs:41805`), so every bounds check passes and the write lands wherever the old origin points — over a sibling pane, the sidebar, or the status bar. Being a raw write rather than a buffer index, an out-of-range origin cannot panic.

## Why the corruption survives a repaint

This is what separates #510 from the stale-cell class (#473, closed). ratatui's diff has no record of the cells these raw writes touched — the code says so at `mod.rs:48307` — so it believes they still hold the old content and never repaints them. A foreign character landing mid-row therefore **stays embedded inside a word** (`not re2ponding`, `nfs ser6er`) instead of clearing on the next frame.

Credit where it belongs: the reporter's observation that digits landed *inside* words is what discriminated between the candidate mechanisms. A screenshot alone cannot separate a paint-path bug from grid corruption, but that specific artefact can.

## The trigger is maximize, not a plain split

Non-active panes get `Rect::default()` (`mod.rs:15398`), so the render loop skips them (`:15439`) and clears `last_area` while `last_inner` keeps the pre-maximize split column. Collapse alone does not do it — `sync_focus_flags` moves focus off a collapsed pane, so it stops being the active terminal the overlay consults.

## Verification

Red/green, both halves shown by name against a real `test result:` line:

| run | result |
| --- | --- |
| with the fix | `test a_hidden_pane_drops_last_inner_not_just_last_area ... ok` — `1 passed; 0 failed; 4277 filtered out` |
| fix line removed, nothing else | `FAILED` — `panicked at src/app/tests.rs:18579`, on the assertion it names |

The test maximizes a split and asserts the hidden pane's `last_inner` is zeroed, mirroring the existing assertion that already pins `last_area` for the same panes.

## Scope: this fixes the carrier, not the class

The same emit-without-clip shape repeats in that block for the **editor, markdown, minimap, welcome and activity-bar** overlays (`mod.rs:48332/48365/48381/48408`, `:5934`). I fixed the one the reported symptom points at. Filing the remaining four is worth doing separately, so the next report of this class does not send someone back to code that is already fixed.

Version 0.1.906 rather than 0.1.905: another session holds 905 committed but unpushed for the docx container cap, so 904+1 would have collided at merge with work `origin` cannot see.


vitali87:

CodeRabbit's check here reports `pass` while annotated **"Review limit reached"** — its quota is per-PR, and this PR's window has not opened, so no review ran and `reviews[]` is empty. This local Opus 5 review fills the gap, since #510 is the repo's only `priority: high` issue.

**Verdict: REQUEST_CHANGES** (1 major, 2 minor). **All fixed in `a608af2`**, plus the two tests it asked for.

---

### Walkthrough

The mechanism holds. `last_area` and `last_inner` are assigned on adjacent lines (`terminal.rs:4474-4475`); on `origin/main` `PtyTerminal::last_inner` is assigned there and cleared in **zero** production sites (the one `last_inner = Rect::default()` on main is `self.tree.last_inner`, a different widget). `update_terminal_image_overlay` anchors `cell_x`/`cell_y` on it, and the payload is emitted post-`draw()` with absolute CUP at `mod.rs:48349`, unclipped. The fix is correct and minimal; the findings are about claims made *around* it.

### Findings

- **`mod.rs:41811-41816` [major/correctness-of-claim]** — The comment asserts the `inner.width < 6 || inner.height < 3` floors "already reject" a pane the last frame did not paint, presenting them as a sufficient second line of defence. True only for a **zeroed** rect. They are magnitude checks, not freshness checks: a pane painted at a real size that then stops being painted keeps a full-size `last_inner` and passes both. That *was* the bug; it is now false only because this PR zeroes the rect. The comment inverts the dependency, so a future reader deleting the clear would conclude the floors still cover them — the precise failure mode #510 was.
- **commit message / `tests.rs:18541` [minor/accuracy]** — "cleared in thirteen places" verifies as 13, but only **6** are `PtyTerminal` panes (15440, 15632, 15649, 15666, 15682, 15707); the other 7 are `open_editors`, `outline`, `timeline`, `dependencies`, `tree`, `commit_graph`, `osk`. The count conflates widgets and overstates the asymmetry it is being used to argue.
- **`mod.rs:15456` [minor/completeness]** — Worth recording why the other five terminal-pane `last_area` clears need no matching `last_inner` clear. Verified: they fire when the bottom panel shows Problems/Output/Ports/Captures or is hidden, and `update_terminal_image_overlay` returns at `41796-41802` on exactly `!self.show_terminal || self.bottom_panel_tab != BottomPanelTab::Terminal`. Unreachable for this consumer — by a guard 26k lines away.

### Test coverage

The test is sound and non-vacuous: `tests.rs:18570-18573` asserts `last_inner.width > 0` after the first draw and before maximizing, so the final assertion cannot pass by the rect having always been zero. Two gaps:

- It sets `terminal_pane_maximized` directly rather than calling `toggle_terminal_pane_maximize()`, so it pins the render loop but not the gesture.
- No test covers the **collapse** arm of the same condition (`t.collapsed && !maximized`), the other route into the identical clear.

### Verified, not defects

- **Completeness.** Swept every mutation of `self.active_terminal` and every route a pane can stop being painted. `update_terminal_image_overlay` reads only `self.terminals[self.active_terminal]` and is called at `mod.rs:16177` **inside `App::render`**, ~700 lines after the pane loop — so every route that repoints `active_terminal` (`reveal_terminal_pane`, rail/label clicks, the menu toggle, `move_terminal`, the index shift in `close_terminal_at`, and `undo_close_terminal` re-inserting a parked pane with its pre-close rect) is closed by that ordering. **No route leaves `last_inner` stale for this consumer.**
- **No regression in the other readers.** `record_active_screen`, `toggle_session_recording` and the Shift+PageUp page size all read the **active** terminal — never zeroed here — and the first two are zero-guarded anyway.
- **Release gate.** `Cargo.toml`, `Cargo.lock` and `src/release_notes/0.1.906.md` consistent; `release_notes.rs` makes a missing note a build error. No collision with main (0.1.904), and the commit explains the 905 skip.
- **Scope claim — accurate on the editor, wrong on markdown.** `update_editor_image_overlay` takes `editor_area: Rect` as a parameter and never reads `Editor::last_inner`, so it is safe as claimed. But `update_markdown_image_overlay` anchors on **`md.last_area`**, not `Editor::last_inner` — the latter is read only for `viewport_rows`, a height bound. #513 has been corrected.

### Scores

- **CodeRabbit score (diff quality) — 3/5** — one major: a comment that inverts the dependency it documents, in the exact place the next reader would consult.
- **Greptile score (codebase fit) — 4/5** — mechanism and completeness verified against every route; two minors on an overstated count and an unrecorded reason.

**VERDICT: REQUEST_CHANGES**

---

### Fixes applied in `a608af2`

Comment reworded to state the direction of the dependency and that deleting the clear breaks it. Count corrected to six. The other five clears' exemption recorded at the clear site. Two tests added — one driving `toggle_terminal_pane_maximize`, one covering the collapse arm — for 3 passing, each named in the output against a real `test result:` line.

Issue #468, opened 2026-09-01
left terminal uncollapse fails


vitali87:
<img width="1156" height="313" alt="Image" src="https://github.com/user-attachments/assets/161bf37a-8f46-482c-86ec-58f788dc7755" />

This works fine for right-side-opened terminals but the default terminal that is on the left side which fold also on the elft side,c annot be recovered after collapse.

vitali87:
Investigated the residual left by #478 (which said "Part of", not "Closes"). I have not found the mechanism and would rather ask than keep guessing — two theories have died against the tree already.

**What is already fixed.** #478 exempted a collapsed pane's strip from the *sidebar seam*, which runs down the side of the panel. That covered the left pane with the side bar on the left, and the right pane with it on the right.

**What I checked and ruled out:**

1. *The editor/panel splitter.* Same shape of bug on the perpendicular edge: its grab zone is two rows, and I expected the strip's expand chevron to sit in it. It does not — the chevron renders at y=20, the splitter at y=19, so the zone (rows 18-19) is one row clear. A test asserting the overlap as a precondition failed on that precondition, which is what disproved it.
2. *An all-strips row.* Both routes are guarded: `toggle_terminal_collapse` refuses to fold the last expanded pane, and `ensure_a_terminal_pane_is_expanded` covers the close route.
3. *Persistence.* `collapsed` is not written to prefs, so a fold cannot survive a restart.
4. *Mouse routing.* Left-click on the strip, right-click "Expand Terminal", the header `‹` button and `Cmd+K ]` (restore all) all reach `toggle_terminal_collapse` and all look correct.

**One real inconsistency, though probably not this bug.** `toggle_terminal` (`src/app/mod.rs:12612`) clears `terminal_maximized` when Ctrl+J hides the panel, on the stated rationale that Ctrl+J is the user's "show my normal layout again" key — but it does not clear `collapsed`. So a folded pane survives a hide/show cycle. The rationale covers the collapsed case as much as the maximized one, so this reads as an oversight; happy to fix it separately if you agree it should follow the same rule.

**What would help.** Could you say what the state was when you hit it — how many panes were open, which one you folded, whether the side bar was showing and on which side, and whether `Cmd+K ]` brought it back? The screenshot shows the strips but not the sequence. If `Cmd+K ]` did recover it, the bug is the advertised click gesture being dead in some layout I have not reproduced; if even that failed, it is something else entirely and that would narrow it fast.

[...]

PR #514, opened 2026-09-07
test(terminal): cover the leftmost pane's collapse strip

vitali87:
Part of #468. **Does not fix it** — the probe came back green, which narrows the search rather than closing it.

## Why this test exists

#468 reports the leftmost pane specifically: "works fine for right-side-opened terminals but the default terminal that is on the left side, which fold also on the left side, cannot be recovered after collapse."

Every strip test in the suite used index 1 or 2. `terminal_strip_rects[0]` was read by **no test at all** — the shape in which a bug survives a fully green suite, and it matched the reported asymmetry exactly.

## Result: green

```
test app::tests::the_leftmost_panes_strip_offers_expand_terminal ... ok
test app::tests::the_leftmost_pane_comes_back_from_its_strip ... ok
test result: ok. 2 passed; 0 failed; 4277 filtered out
```

Both halves of the gesture on pane 0: left-click unfolds, right-click offers **Expand Terminal**. So index 0 is not the mechanism, and the layout loop's identical treatment of every index (`mod.rs:15459`) holds in practice as well as by reading.

Three gates on the run, so the green means something: a `test result:` line exists, the filter matched something, and both tests appear by name. `0 passed; 0 failed` would have been reported inconclusive rather than passing.

## Worth keeping regardless

The asymmetry the report describes now has coverage, and the next person to look at #468 can start from "not index 0" instead of re-deriving it.

## One theory killed on the way

While fixing #510 I found `PtyTerminal::last_inner` is assigned in one place and cleared nowhere, so a pane that stops being painted keeps a stale rect — and a collapse cycle is exactly that shape. It is **not** this bug: the strip click resolves through `terminal_strip_rects` (`mod.rs:37653`) and dispatches straight to `toggle_terminal_collapse`, while `last_inner` is read only by pane-body hit-test helpers. Nothing on the path from a strip click to an unfold reads it.

That is the third dead theory on #468, and the first to die on evidence rather than on a failed precondition. Recorded on the issue so nobody spends a round re-deriving it.

## What is still needed

The repro, now a narrower question: side bar left or right, how many panes, and whether anything was resized or maximized between the collapse and the attempt to recover. The three untested conditions in priority order are on the issue.

Test-only: `src/app/tests.rs` is excluded by ci.yml's `shipped` filter, so no version bump.


vitali87:

CodeRabbit reports `pass` here while annotated **"Review limit reached"**, and `reviews[]` is empty — its quota is per-PR and this PR's window has not opened, so no review ran. Reviewed by hand instead; the change is 63 added lines in one test file.

### Walkthrough

Adds two tests covering `terminal_strip_rects[0]`, which no test read. Both pass, so this documents that index 0 is **not** the mechanism behind #468 rather than fixing it.

| File | Change |
|---|---|
| `src/app/tests.rs` | `the_leftmost_pane_comes_back_from_its_strip`, `the_leftmost_panes_strip_offers_expand_terminal` |

### Findings

**None.** Test-only; no production code touched, so no behaviour can regress from it.

### What I checked

- **The preconditions are real, not decorative.** Both tests assert `strip.width == 1` after the draw, so a run where the strip was never painted fails on the precondition rather than passing vacuously on an unfoldable pane. The first also asserts `app.terminals[0].collapsed` before clicking, so "it unfolded" cannot be satisfied by a pane that was never folded.
- **They drive the production path.** `app.handle_mouse(...)` at the strip's real coordinates, not `toggle_terminal_collapse` directly — so they exercise the hit-test ordering the click depends on (`mod.rs:37653`, strips tested before the splitter and the pane body). A test calling the toggle would have passed while the click path was broken, which is the failure this issue is about.
- **The green is not vacuous.** Three gates on the run: a `test result:` line exists, the filter matched something, and **both tests appear by name** — `0 passed; 0 failed` would have been reported inconclusive. Result: `2 passed; 0 failed; 4277 filtered out`.
- **Release gate: exempt.** `src/app/tests.rs` is excluded by ci.yml's `shipped` filter (`grep -vE '^(src/app/tests\.rs|tests/)'`), so no version bump. Verified by running the workflow's own filter against the diff.
- **Base is current.** Merged `origin/main` after #511 landed; the diff against main is now exactly the 63 added lines, and `scripts/check_arch_content.py` still passes post-merge.

### Test coverage

The tests ARE the coverage. Worth noting what they do not reach, since that is where #468 now lives: the probe is a 3-pane split with the default side bar and no resize. Untested and named on the issue in priority order — side bar on the right, two panes rather than three, and a resize between the fold and the unfold.

### Nitpicks

None.

### Scores

- **CodeRabbit score (diff quality) — 5/5** — no findings; preconditions guard against a vacuous pass and the tests drive the real entry point.
- **Greptile score (codebase fit) — 5/5** — matches the file's existing strip-test shape (`app_with_terminal_panes`, `handle_mouse` at `strip.x, strip.y + 1`), and fills a genuine coverage hole rather than duplicating an index already covered.

**VERDICT: APPROVE**

PR #515, opened 2026-09-07
fix(docx): cap the document preview before the central-directory parse

### LINKED ISSUE #506: docx/odt admits 50 MB into the same central-directory parse the zip lister caps at 8 MB
Found while assessing #493 option 3. Separate from that issue and not a reason to close it.

`docx::to_markdown` (`src/docx.rs:80`) gates on file size **before** `ZipArchive::new`, so it does *not* have the defect #504 fixed in the archive lister. The bound is the problem, not its placement:

```rust
// src/docx.rs:14
pub const MAX_DOC_BYTES: u64 = 50 * 1024 * 1024;
```

```rust
// src/archive.rs:45
pub const ZIP_LIST_CAP: u64 = 8 * 1024 * 1024;
```

Two paths into the same `zip::ZipArchive::new` central-directory parse, admitting sizes 6.25x apart. #504 set the 8 MB figure precisely because that parse allocates per member before the entry count is knowable, so it cannot be bounded from inside.

### Reachability

`.docx`/`.odt` dispatch at `src/widgets/editor.rs:3921`, on the `Editor::open` cascade — the same cascade `croft view` (#362) reaches. So this is triggerable by any process in any pane, which is the exposure argument #493 was filed on.

### Measurement, and its caveat

Crafted zips of empty long-named members (the layout that maximises central-directory size per byte), parsed with **Python's** `zipfile`:

| file size | members | parse |
|---|---|---|
| 8 MB (the archive cap) | 39,568 | 182-192 ms |
| 50 MB (the docx cap) | 247,305 | 1205-1696 ms |

6.25x the bytes gives ~6.3x the members and ~7-9x the time.

**This is indicative, not conclusive.** Python's parser is not the `zip` crate's, and the absolute numbers will differ. What transfers is the shape: cost scales with member count, member count scales with file size, and the docx path admits six times as much. I did not run the Rust measurement because the host suite lock was held by two other sessions; happy to produce it before any fix lands if you want the real figure first.

Also worth saying: a *real* 50 MB docx is mostly media members and would have a far smaller entry count than this crafted layout. The concern is the adversarial case, which is the one `croft view` makes reachable.

### Suggested fix

Give the doc preview its own pre-parse cap in the same shape as `ZIP_LIST_CAP`, sized to what a document preview actually needs, and let an over-cap file fall through to the archive route (which then applies its own 8 MB gate) or to hex. The existing 50 MB `MAX_DOC_BYTES` stays as the decompressed-member bound it already is — the two are different limits and the fix should not conflate them.

I can take this if it's wanted.

vitali87:
## Summary

Closes #506. `docx::to_markdown` gated on `MAX_DOC_BYTES` (50 MB) before `ZipArchive::new`. The placement was right and the bound was wrong: that constant limits how much a **member may decompress to** once the archive is open, while the constructor's cost is the **central-directory parse** it performs to open it at all — which the archive lister caps at 8 MB (`ZIP_LIST_CAP`, #504). Two routes into one parse, six times apart.

`MAX_DOC_LISTING_BYTES` is that pre-parse bound, **aliased to `ZIP_LIST_CAP`** so the two cannot drift. `MAX_DOC_BYTES` is unchanged.

## The refusal has to be named on this path

`archive::kind_from_ext` claims `.zip`/`.jar`/`.whl` and **not** `.docx`/`.odt`, so an over-cap document never reaches the archive route and cannot pick up its cap refusal on the way past. It would fall to the hex viewer with nothing said — the silent dump #504 added `route_note` to prevent. The cascade now sets that note itself, with text `archive::is_list_cap_refusal` recognises, so it lands in the existing policy branch rather than a parallel mechanism.

An earlier revision of this branch removed that message on the false premise that the archive route would supply it. A peer caught it. The comment at `editor.rs` that stated that premise is corrected here, since it is what sold two of us the wrong model.

## The cap is a crude proxy, and the doc says so

File size does not track parse cost. Measured:

| shape | size | members | parse | verdict |
|---|---|---|---|---|
| crafted, empty members | 8.39 MB | 97,799 | 384 ms | **ADMITTED** |
| document of 600 KB photos | 12.3 MB | 21 | 0.19 ms | **REFUSED** |

The most expensive thing it admits costs ~2000x the cheapest thing it blocks. Media is bytes without members; an attack is members without bytes — the two are anti-correlated with size, so the cap orders backwards the very cases it exists to separate.

The benign threshold is a **range, not a number**: holding the cap fixed and varying only media size moves it 27x (165 members of 50 KB images, 26 of 330 KB, 6 of 2 MB photographs). The doc says not to read a member count out of those rows and tune against it — any such figure describes the fixture it came from.

## Why it is not raised, given all that

The local review recommended raising the value toward `MAX_DOC_BYTES` since size buys little protection. I disagreed, and recorded why in the doc: `ZIP_LIST_CAP`'s own comment accepts **~800 ms** as its bounded worst case at this value, so the 384 ms this admits is *inside* the budget #504 chose deliberately for the same constructor. Raising this route would take the same parse several times past that bound on one of its two routes and break the alias that keeps them honest.

Bounding the real cost needs a member-count limit, which needs a partial central-directory reader — `ZipArchive::len()` exists but only on a *constructed* archive, so asking how many members there are costs exactly what we are avoiding. That is a design change and belongs in its own issue, not round eleven of this one.

## This is a trade, not an improvement

A large image-heavy document that opened under the old 50 MB gate is now refused. The release note says so plainly. The justification is the asymmetry of the two failures: **a refused document is recoverable — it opens as hex and says why — and a frozen frame loop is not.**

## Tests

- the cap refuses before the parse, and an under-cap document still opens (so the refusal is the size, not the fixture)
- the fixture asserts **both** bounds — past the new cap *and* under `MAX_DOC_BYTES` — so only the new gate can be what refuses it
- `kind_from_ext` returns `None` for a `.docx` **with a `.zip` control in the same run**, so the premise is asserted alongside proof the call works
- end to end: opening an over-cap document leaves a status line containing `hex viewer:` **and** satisfying `is_list_cap_refusal`, so a reword on either side fails the test rather than silently decoupling them
- `const _: () = assert!(MAX_DOC_LISTING_BYTES < MAX_DOC_BYTES)` — a compile-time check, verified load-bearing by inverting the constant and watching the build fail

Verified at `5252a63`: `CLIPPY_EXIT=0`, `TEST_EXIT=0` (5 passed, named individually), `FMT_EXIT=0`.

0.1.905 with a release note; `ships_nothing.py` exits 1 (it ships).

Closes #506


vitali87:

Both bots are silent on this PR (zero entries in `reviews[]`, zero comments), so this review was run locally by the `greptile-local` agent against `1728d2d`. It returned **Confidence Score 2/5** with two P1s. Both are fixed in `5252a63`.

### P1 — an unrelated file was committed

`CLAUDE.md.bak-priority`, a 40-line local instructions backup, was swept into the commit by `git add -A`. `.gitignore` covered `/CLAUDE.md` but not its backups, so it would have shipped to every clone.

**Fixed:** removed from the commit, and `.gitignore` now covers `/CLAUDE.md.*` so the next `add -A` cannot repeat it. I verified the staged file list explicitly before re-committing rather than trusting `add -A` a second time.

### P1 — the cap's selectivity is inverted

The reviewer measured what I had only reasoned about:

| shape | size | members | parse | verdict |
|---|---|---|---|---|
| crafted, empty members | 8.39 MB | 97,799 | 384 ms | ADMITTED |
| document of 600 KB photos | 12.3 MB | 21 | 0.19 ms | REFUSED |

The most expensive thing the cap admits costs ~2000x the cheapest thing it blocks, and my doc comment claimed it "bounds the central-directory parse". That was the intent, not the effect.

**Fixed, but not as recommended.** The reviewer suggested raising the value toward `MAX_DOC_BYTES` since size buys little protection. I disagreed and recorded the reason in the doc: `ZIP_LIST_CAP`'s own comment accepts **~800 ms** as its bounded worst case at 8 MB, so the 384 ms this admits is *inside* the budget #504 chose deliberately for the same constructor. Raising this route would take the same parse several times past that bound and break the alias that keeps the two routes honest.

What changed instead: the doc now describes the cap as a crude proxy, carries both measurement directions, records that the benign threshold is a **range spanning 27x** rather than a number (so nobody tunes it against their own fixture), names the residual explicitly (~98k members still costs ~400 ms), and states that this is a **trade** — an image-heavy document that opened under 50 MB is now refused. The release note warns users of that regression rather than hiding it.

### Verified, not defects

- `archive::kind_from_ext` returns `None` for `.docx`/`.odt` — confirmed by reading the function and by the test's own `.zip` control. The `route_note` is genuinely required, not redundant; the earlier revision that removed it was wrong.
- `route_note` is consumed with a path-equality guard and cleared on success, so a stale note cannot explain the wrong file.
- The `const` assertion is load-bearing: the reviewer inverted `ZIP_LIST_CAP` to 80 MB and the build failed with `E0080`. The equality assertion it replaced was a true tautology — it still compiled at the inverted value.
- The fixture's two bounds hold. `set_len(CAP+1)` appends NUL padding after the EOCD, and Rust's `zip` 7.2 reader still opens it (unlike Python's `zipfile`), so with the gate reverted the parse would succeed and the test would fail for the reason it claims.
- Release gate: `ships_nothing.py` exits 1 (it ships); 0.1.904→0.1.905 with a matching note.

### Noted for follow-up

`archive::list`'s zip branch has the same size-as-proxy weakness. Now that the constants are aliased, a proper fix — a member-count limit via a partial central-directory reader — should cover both routes. Worth its own issue.

### Verification at `5252a63`

`CLIPPY_EXIT=0`, `TEST_EXIT=0` (5 tests passed, each named individually in the output), `FMT_EXIT=0`, tree confirmed by the runner's own control.

VERDICT: APPROVE (after both P1s fixed)

Issue #316, opened 2026-08-26
Port the sibling repo's required scored-review gate: CLEAN is near-zero evidence a review happened


vitali87:
## The measurement

Taken from the actual board on 2026-08-26:

| | |
|---|---|
| open PRs reporting `mergeStateStatus: CLEAN` | **5** |
| of those, carrying an **unaddressed REQUEST_CHANGES** at their exact head SHA | **3** (#297, #309, #311) |

**A 60% false-permission rate.** Merging on the aggregate would have landed all three over live objections.

## Why CLEAN says so little here

`CLEAN` reports one thing: no branch-protection rule is currently blocking the merge. croft has **no rule requiring a review**, so the signal never consults review state at all. It is not weak evidence that a review happened — it is close to zero evidence, because the question it answers does not mention reviews.

The same shape appears one level down. `gh pr checks` on #297 showed:

```
CodeRabbit    pass    Review rate limited
```

A green row meaning **no review ran**. And the rate-limit notice on that PR is 10,076 characters — *larger* than both genuine reviews (3,245 and 4,626) — so "longest bot comment" selects the non-review. Length is not weak evidence here, it is inverted.

## The general form

Requiring the positive artifact is necessary but **not sufficient** — the artifact must also be *about the thing you named*:

| signal | real artifact? | but about… |
|---|---|---|
| stale review | yes | the wrong commit |
| `test result: ok`, 0 tests selected | yes | no test at all |
| `mergeStateStatus: CLEAN` | yes | branch rules, not review |
| rate-limited bot `pass` | yes | the check having run |

The second half is invisible to the first, which is why it keeps being rediscovered.

## Proposal

Port the equivalent of the sibling `code-graph-rag` repo's required `Greptile 5/5 Gate` job: poll until a scored review names the **exact head SHA**, and fail on a stale, missing, or lower-scored one. That converts a habit into a mechanism, so the merge is blocked rather than merely discouraged.

`CONTRIBUTING.md`/`CLAUDE.md` already describe this discipline in prose and note croft has no such gate today. The measurement above is the argument for closing that: the rules are being followed by attention, and attention is what a hurried session skips.

**This is a repo-policy change (branch protection + a required job), so it needs the owner's decision rather than being actioned from a session.**

vitali87:
[...]

PR #516, opened 2026-09-07
fix(welcome): drop the ambient dotted wave from the release card

### LINKED ISSUE #469: Better remove the effect in "IN THIS RELEASE" welcome screen
Image

vitali87:
## Summary

Closes #469. The welcome screen's "IN THIS RELEASE" card loses its ambient dotted wave.

## The effect was correctly built — that is not the problem

`paint_card_wave` found each row's **last occupied column** and painted only past it, specifically so a dot could never land between two words of a release note. The doc said so and the code did it. That contract held.

What it could not address is scale. On a wide terminal the blank region to the right of three short notes is most of the card, so a field sized as "sparse texture" becomes a large block of dots. In the screenshot on the issue the dotted region occupies more of the card than the notes it was meant to frame. A correct implementation of the wrong amount.

## What was removed

The call site in `src/app/mod.rs`, `paint_card_wave`, its five tests, and five tuning constants. Clippy with `-D warnings` then flagged five test helpers — `dots_at`, `luma`, `LUMA_BRANCH`, `blank`, `area` — as never-used, which left `mod wave_tests` containing zero tests, so that module went too. ~304 lines from `src/gradient.rs`.

The function was **not** left behind an unused path. Its tuning tests would have stayed green while nothing rendered them, which reads as coverage of live behaviour.

`src/release_notes/0.1.860.md` still announces the wave and is untouched. `build.rs` reads exactly one note file per version, so a 0.1.907 binary never displays 0.1.860's text — and `src/release_notes.rs` states the rule directly: "a version's file describes that version alone rather than accumulating a changelog." Amending it would make it falsely describe what that release shipped.

## Also here

`.gitignore` gains `/CLAUDE.md.*`. It covered `/CLAUDE.md` but not local backups of it, and a `git add -A` swept one into an unrelated commit on another branch earlier tonight.

## Verification

Local review returned **5/5, no findings**, and checked the deletion empirically rather than by grep: rendering `render_welcome` at 120x30 gives 22 middot cells on base and **0** on this branch, so the effect is genuinely gone rather than merely compiled out. It also confirmed `paint_gradient_box` and every palette constant survive intact with all 14 call sites compiling, and that `inner_x`/`inner_y`/`inner_w_actual` remain live in the surrounding card code.

At `3c38b7e`: `CLIPPY_EXIT=0`, `TEST_EXIT=0` (gradient filter 8 passed, welcome filter 19 passed), `FMT_EXIT=0`.

Full suite on this host: 4256 passed, 5 failed — all five environmental and pre-existing (two reproduce on unmodified `origin/main`: a missing `aarch64-unknown-linux-musl` rustup target and a sandbox socket bind; three are load-scaled timeouts that pass in isolation). None touch changed code.

0.1.907 with a release note; `ships_nothing.py` exits 1.

Closes #469


vitali87:

Both bots are silent on this PR, so this review was run locally by the `greptile-local` agent against `3c38b7e`.

**Confidence Score: 5/5 — no findings.**

### What it verified rather than assumed

- **The deletion is effective, not just compiled out.** A probe rendering `render_welcome` at 120x30 in throwaway worktrees: base `PROBE_MIDDOT_COUNT=22`, branch `PROBE_MIDDOT_COUNT=0`. That distinguishes "the effect is gone" from "the code no longer references it", which a grep cannot.
- **Nothing else used the removed code.** Zero hits for `paint_card_wave`, `WAVE_`, `wave_tests`, `LUMA_BRANCH`, `dots_at` outside the removed block. `blank`/`area`/`luma` existed only inside `mod wave_tests`, which was the file's sole `#[cfg(test)]` module, so no other test module could have depended on them.
- **No over-deletion.** `paint_gradient_box`, `lerp_rgb`, `rgb_color` and every palette constant survive; the diff touches only lines after the end of `paint_gradient_box`. Its 14 call sites still compile and its two tests still pass.
- **The card still renders.** `inner_x`, `inner_y`, `inner_w_actual` remain live below the removed block; `box_rect.height` was read only inside `wave_area`.

### On leaving `0.1.860.md` alone

I asked whether a shipped release note should be amended when its feature is later removed. The reviewer settled it from the codebase rather than taste: `build.rs:bake_release_notes` reads exactly one file, `src/release_notes/{CARGO_PKG_VERSION}.md`, so a 0.1.907 binary never displays 0.1.860's text. `src/release_notes.rs:13` states the rule — "a version's file describes that version alone rather than accumulating a changelog" — and every commit that ever edited a note edited the version then in `Cargo.toml`, never a shipped one. Amending it would make it falsely describe what 0.1.860 shipped.

### Full-suite note

The reviewer corrected something in my own report: my `32 passed` was a **filtered** run, not the full suite. The full suite on this host is 4256 passed / 5 failed, and all five are environmental — two reproduce on unmodified `origin/main` (a missing `aarch64-unknown-linux-musl` rustup target; a sandbox socket bind), three are load-scaled timeouts that pass in isolation and whose own panic text says "if this is a real failure it would also fail at 1x". None touch changed code.

Worth stating plainly since I reported the narrower number first: a filtered green and a full-suite green are different claims.

VERDICT: APPROVE

vitali87:
**Rebased onto `ef878e9` (main after #515 merged). Head moved `3c38b7e` → `eb03b20`.**

Flagging this explicitly because the 5/5 review above names `3c38b7e`, and a review that names a commit the merge will not take is exactly the failure #316 records — #508 merged at the commit *before* its review fixes for that reason.

What changed in the rebase, and why the review still applies:

- **`.gitignore` dropped out of the diff.** Both this branch and #515 carried the same `/CLAUDE.md.*` addition; #515 landed it first, so the hunk is now redundant and git resolved it away. The fix is on main — verified — so nothing is lost.
- **The version conflict resolved to 0.1.907, unchanged.** Main is 0.1.905 after #515; 906 is held by #512 and 908 by #452, so 907 remains the free slot this branch already claimed.
- **No source file changed.** `src/app/mod.rs`, `src/gradient.rs` and `src/release_notes/0.1.907.md` are byte-identical to the reviewed commit; the diff is 5 files instead of 6 purely from the dropped `.gitignore` hunk.

So every finding-bearing surface the reviewer examined is unchanged, and the empirical check it ran — 22 middot cells on base, 0 on the branch — is unaffected by a version-line resolution.

Re-verified on the rebased tree rather than assuming the earlier run carried over:

```
TREE: fix/469-remove-welcome-wave @ eb03b20
CLIPPY_EXIT=0    (Compiling croft-software v0.1.907, finished in 15.70s)
TEST_EXIT=0      32 passed; 0 failed; 4243 filtered out
FMT_EXIT=0
```

Pushed with `--force-with-lease=fix/469-remove-welcome-wave:3c38b7e`, so the push would have been refused had anything other than my own reviewed commit been at the remote head.

PR #518, opened 2026-09-07
fix(overlays): refuse a post-draw overlay that does not fit the screen

### LINKED ISSUE #513: Post-draw image overlays write with absolute CUP and no clipping to the owning widget
#510 fixed one carrier of a class, and the other writers in the same block are unfixed. Filing so the next report of this shape does not send someone back to code that is already fixed.

### The shape

`src/app/mod.rs:48325-48420` is a block of post-draw flushes that run *after* `terminal.draw()` has emitted ratatui's diff. Each parks the cursor with absolute CUP and dumps a payload:

```rust
let _ = write!(out, "\x1b[{};{}H", layout.cell_y + 1, layout.cell_x + 1);
let _ = out.write_all(osc.as_bytes());
```

Nothing clips the payload to the owning widget's rect at emit time. Two properties make a wrong origin dangerous rather than merely wrong:

1. It is a raw write, not a buffer index, so an out-of-range origin cannot panic — ratatui's `Buffer` would have.
2. ratatui's diff has no record of the cells it touched (`mod.rs:48307` says so), so nothing repairs them on the next frame. The corruption persists instead of clearing on repaint.

The bound is therefore entirely the rect the layout was computed from. In #510 that rect was `PtyTerminal::last_inner`, which was assigned in one place and cleared nowhere, so a pane the frame skipped kept its pre-maximize column.

### What is still exposed

| writer | anchor | stored rect? |
| --- | --- | --- |
| terminal image (`:48343`) | `PtyTerminal::last_inner` | **fixed in #512** |
| editor image (`:48332`) | `editor_area` **parameter** (`:41621`, callers at `:15257-15278`) | no — recomputed per frame |
| markdown image (`:48365`) | `md.last_area` **and** `self.editor.last_inner` (`:41954-41955`) | **yes** |
| minimap (`:48381`) | `MinimapLayout` | unaudited |
| welcome (`:48408`) | — | unaudited |
| activity-bar icons (`:5934`) | `sidebar_areas`, with a `positions_moved` check | partial |

The editor path is the reassuring one: it takes its rect as a parameter, so it cannot outlive the layout that produced it. That is the shape the others should have.

**The markdown path is the concrete concern.** `update_markdown_image_overlay` reads `Editor::last_inner`, and that field has the same lifecycle defect the terminal one had: assigned at `editor.rs:10700` and `:10729`, cleared nowhere in production (`editor.rs:18010` is a test; `mod.rs:10119` is the file tree, a different widget). I have not established a route by which the editor stops painting while the markdown overlay still fires, so this is an unproven exposure rather than a second live bug — but it is the same missing clear on the same kind of field feeding the same unclipped writer.

### Suggested shape of a fix

Two options, and the second is better:

1. Clear each stored rect wherever the frame declines to paint its widget — what #512 does for the terminal. Correct, but it is one more invariant to remember at every new skip site.
2. Clip at the emitter. The flush block knows the payload's origin and size; a widget rect passed alongside would let the block refuse a payload that falls outside it, which makes the whole class unreachable regardless of what any cached rect says. The editor path already demonstrates the parameter-passing half of this.

Option 2 is the one that ends the class. Option 1 is what each individual report will otherwise cost.

### Not urgent

Nothing here is a reported user-visible failure today. #510 was, and it is fixed. This exists so that the next one is diagnosed from the class rather than rediscovered from the symptom.

vitali87:
## Summary

Closes #513. The post-draw flush block writes overlays with absolute CUP straight to stdout, **after** ratatui has flushed its diff. Two properties make a wrong origin dangerous rather than merely wrong:

1. It is a raw write, not a buffer index, so an out-of-range origin **cannot panic** the way `Buffer` would.
2. ratatui holds no record of the cells it touched, so **nothing repairs them** on the next frame. The corruption persists instead of clearing on repaint — which is what #510 looked like from the outside.

#512 fixed one carrier by clearing the stored rect it read. Every other writer in the block keeps its own rect with its own lifecycle, so that shape needs the clear remembered at every future site a frame can decline to paint a widget — and this class exists precisely because someone remembered it for `last_area` and not for its twin on the adjacent line.

`overlay_fits_on_screen` checks the payload against `last_frame_area` at **emit** time instead. All six writers in the block go through it: the two editor split slots, the terminal pane, the markdown preview, the minimap, the welcome logo, and the activity-bar icons.

## What this does NOT do

Stated in the doc, because the bound is easy to oversell: **it removes off-screen writes only.**

A stale rect that still lands somewhere on screen passes untouched — and that is the case #510 actually exhibited, an image painted over a sibling pane, the sidebar and the status bar, all mid-screen. So this neither subsumes #512 nor closes #513's class. It removes the unbounded half, where a wild origin can write anywhere in the terminal. Making it exact needs each payload to carry its owning rect, which is the wider change #513 describes.

## Details worth knowing

- **A clipped payload retries.** The `mark_*_displayed()` calls sit inside the guarded blocks and `payload()` is not gated on `displayed`, so it is re-offered each frame rather than dropped for good. This was the risk worth being wrong about — a fix that trades a visual glitch for a permanently missing image would be worse than the bug.
- **The activity-bar bound is 1x1 and knowingly narrower** than its 2–3 cell count badges: that writer's helper discards each block's `Rect` and returns only an origin. Not reachable today, since the badges anchor to the left-hand bar. Recorded at the call site rather than left to be rediscovered.

## Tests

- `an_overlay_past_the_screen_edge_is_refused_before_it_is_written` — both edges, a saturating wild origin, and zero-sized payloads
- `no_overlay_is_emittable_before_the_first_frame` — the window before `last_frame_area` is set
- `every_post_draw_overlay_writer_checks_the_screen_bound` — counts guarded call sites, so a **seventh** writer added later without one fails rather than silently reopening the class

**Red control:** neutering the bound to `true` makes the first test fail (exit 101), so it tests the guard rather than passing incidentally. The second still passes under that mutation — it covers the zero-area early return, which the control does not touch.

The call-site test exists because the first revision of this change guarded **four of six** writers; the local review caught the minimap and welcome ones still emitting unguarded, both named in #513's own exposure table.

## Verification

At `e6fd02c`: `CLIPPY_EXIT=0`, `TEST_EXIT=0` (3 tests, each named in the output), `FMT_EXIT=0`. `check_doc_ownership.py` clean, `ships_nothing.py` exits 1 (it ships).

0.1.910 with a release note — main is 0.1.909 after #512.

Closes #513

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.909"
+version = "0.1.910"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.909"
+version = "0.1.910"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -5931,6 +5931,22 @@ impl App {
         let cursor_on = self.cursor_should_be_visible();
         let _ = write!(out, "\x1b[?25l\x1b[s");
         for ((x, y), seq) in &overlays {
+            // Same screen bound as the image writers (#513): a raw
+            // absolute-CUP write that ratatui cannot repair, so the origin
+            // is checked before it is used.
+            //
+            // 1x1 is NARROWER than some payloads here, and knowingly so.
+            // `pending_activity_image_overlays` discards each block's
+            // `Rect` and returns only its origin, and a count badge is 2-3
+            // cells wide (`count_badge_cells_w`), so a badge whose left
+            // edge is on screen and right edge is not would pass this.
+            // Not reachable today -- the badges anchor to the left-hand
+            // activity bar -- and making it exact means returning the
+            // `Rect` from that function, which is a wider change than this
+            // guard. Recorded rather than left to be rediscovered.
+            if !self.overlay_fits_on_screen(*x, *y, 1, 1) {
+                continue;
+            }
             let _ = write!(out, "\x1b[{};{}H", y + 1, x + 1); // 1-based
             let _ = out.write_all(seq.as_bytes());
         }
@@ -5941,6 +5957,52 @@ impl App {
         let _ = out.flush();
         self.overlays.activity.mark_emitted();
         self.overlays.activity.store_positions(positions);
+    }
+
+    /// Whether a post-draw overlay at this cell rect may be emitted (#513).
+    ///
+    /// The flush block below writes with absolute CUP straight to stdout,
+    /// after ratatui's diff. Two properties make a wrong origin dangerous
+    /// rather than merely wrong: it is a raw write, not a buffer index, so
+    /// an out-of-range origin cannot panic the way `Buffer` would; and
+    /// ratatui has no record of the cells it touched, so nothing repairs
+    /// them on the next frame. The corruption persists instead of clearing
+    /// on repaint, which is what #510 looked like from the outside.
+    ///
+    /// So the payload is checked against the screen at EMIT time rather
+    /// than trusting whichever cached rect produced the layout. Every
+    /// writer in the block goes through it, which is the part #512's shape
+    /// could not give: that fixed the terminal carrier by clearing its
+    /// stored rect, but each other writer keeps its own rect with its own
+    /// lifecycle, so every future site a frame can decline to paint is one
+    /// more place to remember the clear.
+    ///
+    /// What this does NOT do, stated plainly because the bound is easy to
+    /// oversell: it removes OFF-SCREEN writes only. It is a screen bound,
+    /// not the owning widget's rect, so an overlay anchored on a stale rect
+    /// that still lands somewhere on screen passes it untouched -- and that
+    /// is the case #510 actually exhibited, a terminal image painted over a
+    /// sibling pane, the sidebar and the status bar, all of them mid-screen
+    /// (#512 records the landing sites). So this does not subsume #512 and
+    /// does not close #513's class; it removes the unbounded half, where a
+    /// wild origin can write anywhere in the terminal.
+    ///
+    /// The bound is the screen because the emitter does not know which
+    /// widget a payload belongs to. Making it exact needs each payload to
+    /// carry its owning rect, which is the wider change #513 describes.
+    fn overlay_fits_on_screen(&self, cell_x: u16, cell_y: u16, cell_w: u16, cell_h: u16) -> bool {
+        let area = self.last_frame_area;
+        if area.width == 0 || area.height == 0 {
+            // No frame has been drawn yet, so there is no screen to land on.
+            return false;
+        }
+        // A zero-sized payload has nothing to place; treat it as unemittable
+        // rather than trivially in-bounds, so a cleared rect cannot slip a
+        // bare cursor park through.
+        if cell_w == 0 || cell_h == 0 {
+            return false;
+        }
+        cell_x.saturating_add(cell_w) <= area.width && cell_y.saturating_add(cell_h) <= area.height
     }
 
     /// Reset the blink phase so the caret is solidly visible for the next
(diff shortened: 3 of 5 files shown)


vitali87:

Both bots are silent on this PR, so this review was run locally by the `greptile-local` agent against `bb29d60`. It returned **Confidence Score 3/5** with one P1 and two P2s. All three are addressed in `e6fd02c`.

### P1 — two of six writers were left unguarded

I guarded four writers. The flush block has six. `minimap_image_payload()` (`mod.rs:48428`) and `welcome_image_emit_payload()` (`:48455`) still emitted raw CUP with no bound — and both are named as exposed in #513's own table, so this was not a subtle miss.

The reviewer proved it reachable rather than asserting it: on an 80x24 `TestBackend`, both return `Some` for a rect at (200,200) that `overlay_fits_on_screen` refuses, and the block then writes CUP to row 201.

**Fixed:** all six writers now go through the guard, and `every_post_draw_overlay_writer_checks_the_screen_bound` counts the call sites so a seventh added later without one fails the suite.

### P2 — the doc overclaimed, and this is the correction that matters

My comment said the guard "ends the class". The reviewer showed a screen bound **would not have refused #510 itself**: PR #512 records that overlay landing "over a sibling pane, the sidebar, or the status bar", and a stale pre-maximize `last_inner.x` is a mid-screen column, so it passes this check.

So the residual I had written as an aside is in fact *the case the motivating bug exhibited*.

**Fixed:** the doc now says the bound removes off-screen writes only, that it neither subsumes #512 nor closes #513's class, and that making it exact needs each payload to carry its owning rect.

### P2 — the activity-bar bound is narrower than its payloads

`pending_activity_image_overlays` discards each block's `Rect` and returns only an origin, but count badges are 2–3 cells wide (`count_badge_cells_w`). A badge with its left edge on screen and right edge off would pass a 1x1 check. The reviewer could not construct a reachable case — the badges anchor to the left-hand bar — so this is imprecision, not a demonstrated bug.

**Addressed by documenting it at the call site.** Fixing it properly means returning the `Rect` from that function, a wider change than this guard warrants.

### Verified, not defects

- **`mark_*_displayed()` is inside every guarded body**, and `payload()` is not gated on `displayed`, so a clipped payload is re-offered each frame and retries. No permanent drop — this was the claim worth checking independently.
- **`last_frame_area` is current at emit time**: assigned as the first statement of `App::render`, and the flush block sits after an unconditional `terminal.draw(|f| app.render(f))?` in the same iteration.
- **Activity-bar move detection is intact**: the `positions_moved` early return precedes the loop, and `positions` is built before the guard, so a skipped icon does not perturb the stored strip.
- **Doc ownership and release gate** both clean; 0.1.910 verified as the free slot independently.

### Noted for a separate audit

Five more raw-CUP writers exist outside #513's table — `flush_run_debug_icon_overlay`, `flush_problems_badge_overlay`, `flush_no_repo_hero_overlay`, `flush_ssh_empty_state_overlay`, `flush_tooltip_over_image`. They take single cells or self-clamped rects, so they are a different audit rather than this fix's scope.

### Verification at `e6fd02c`

`CLIPPY_EXIT=0`, `TEST_EXIT=0` (3 tests, each named), `FMT_EXIT=0`, doc-ownership clean, `ships_nothing.py` exit 1.

VERDICT: APPROVE (after the P1 and both P2s addressed)

vitali87:
**Rebased onto `d7cc9d4` (main after #512 and #517). Head moved `e6fd02c` → `69e91c8`.**

Flagging explicitly because the review above names `e6fd02c`, and a review that names a commit the merge will not take is the #316 failure mode — #508 merged at a pre-fix commit for a cousin of that reason.

What changed: **the version conflict only.** Main went to 0.1.909 when #512 landed, so `Cargo.toml`/`Cargo.lock` re-resolved to 0.1.910. I re-enumerated every remote branch's claim before keeping it — main 0.1.909, `feat/croft-view` 0.1.908, nothing else on 0.1.9xx — so 910 is still free and still mine.

**No source file changed.** `src/app/mod.rs`, `src/app/tests.rs` and `src/release_notes/0.1.910.md` are byte-identical to the reviewed commit, so every finding-bearing surface the reviewer examined is unchanged, including the six guarded call sites and the corrected doc.

Re-verified on the rebased tree rather than assuming the earlier run carried over:

```
TREE: fix/513-clip-overlays-at-emitter @ 69e91c8
CLIPPY_EXIT=0    (Compiling croft-software v0.1.910, finished in 13.72s)
TEST_EXIT=0      3 passed; 0 failed; 4280 filtered out
                 every_post_draw_overlay_writer_checks_the_screen_bound ... ok
                 an_overlay_past_the_screen_edge_is_refused_before_it_is_written ... ok
                 no_overlay_is_emittable_before_the_first_frame ... ok
FMT_EXIT=0
```

Pushed with `--force-with-lease=fix/513-clip-overlays-at-emitter:e6fd02c`, so the push would have been refused had anything other than my own reviewed commit been at the remote head.

Note the filtered count moved 4275 → 4280: main gained five tests from #512 and #517. The three named tests are unaffected.

PR #519, opened 2026-09-07
fix(pdf): bound the page-count probes on the frame-loop open path

### LINKED ISSUE #493: Editor::open runs unbounded work on the frame loop, now reachable from any pane via croft view

vitali87:
Part of #493 — **does not close it**, see the bottom.

## The premise was half stale, and the code said which half

`run_pdftoppm` has been bounded since it was written: own process group, polled, group killed at `PDF_RENDER_BUDGET`. So the issue's headline — unbounded `pdftoppm` on the frame loop — is fixed.

But `pdf.rs:283` carried this, and it was accurate:

```rust
// Unbounded (#493): `pdfinfo` and `mdls` below each run with `.output()` on
// the same frame-loop open path as the render, just before it. The
// follow-up that moves the open off the frame loop covers them; until then
// the pdftoppm runner is the shape to reuse.
```

Both reach `.output()` on the same `Editor::open` path, immediately *before* the render that is bounded. A document that makes `pdfinfo` hang parks the editor exactly as the original report describes — one call earlier than the issue names. `croft view` makes it reachable from any pane.

## The fix reuses an existing shape

croft already had this pattern in `dap/install.rs:142`, so this is a third caller rather than a third implementation: piped stdout, read on a thread, `recv_timeout`, kill and reap.

The reader is deliberately **never joined**, for the reason `run_pdftoppm` already records for its own stderr drain — a descendant the child left behind keeps the pipe's write end open, and a join would wait on that descendant past the budget, on the success path too.

`PDF_INFO_BUDGET` is **2s** against the render's 10s, and the gap carries the argument: these parse a header rather than rasterise, so one still running after two seconds is wedged, not slow. Overrunning costs only the page count — the viewer degrades to the single-page path rather than failing the open.

## The test hangs a real child

A build passing proves the happy path, which is exactly the path that never needed bounding. So:

```rust
let out = run_bounded_stdout(Path::new("sleep"), &[Path::new("30")], Duration::from_millis(150));
assert!(out.is_none());
assert!(elapsed < Duration::from_secs(5), "the deadline must fire rather than wait out the child");
```

**The elapsed assertion is the load-bearing half.** Without it the test would still pass with `recv_timeout` deleted entirely, since `sleep 30` returns `None` eventually either way — that would be a test of "something happened", not of the deadline. A paired presence case pins that a child finishing in time still yields its stdout, so the refusal cannot pass vacuously.

Verified with all three gates: a real `test result:` line, the filter matching something, and both new tests named in the output (`named=11`).

## Why this does not close #493

The structural half remains: moving `Editor::open` off the frame loop, which also covers `open_archive` and `open_doc_preview` walking their containers on the same thread. **No timeout addresses that** — it is a different size of change and a scope decision. Bounding these two calls is the "until then" the code comment describes, not a substitute for it.

Version 0.1.911: main is at 0.1.909, PR #518 holds 0.1.910, PR #452 holds 0.1.908. Enumerated across open PRs rather than taking main+1, since a claim on an unpushed or unmerged branch is invisible to `origin/main`.

## Summary by CodeRabbit

- **Bug Fixes**
  - PDF page-count checks now have a time limit, preventing the editor from freezing when document metadata tools become unresponsive.
  - Documents with unreadable page counts now open on their first page instead of hanging.

- **Release**
  - Updated the package version to 0.1.911.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.909"
+version = "0.1.911"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.909"
+version = "0.1.911"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/pdf.rs
+++ b/src/pdf.rs
@@ -303,7 +303,12 @@ pub fn detect_page_count(pdf: &Path) -> Option<u32> {
 
 fn page_count_via_pdfinfo(pdf: &Path) -> Option<u32> {
     let program = which("pdfinfo")?;
-    let out = run_bounded_stdout(&program, &[pdf], PDF_INFO_BUDGET)?;
+    let (status, out) = run_bounded_stdout(&program, &[pdf.as_os_str()], PDF_INFO_BUDGET)?;
+    // The guard the unbounded version had: stdout from a failed run is not a
+    // page count, and every other subprocess in this file checks the same.
+    if !status.success() {
+        return None;
+    }
     parse_pdfinfo_pages(&out)
 }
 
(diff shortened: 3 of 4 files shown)


vitali87:

CodeRabbit's check reports `pass` while annotated **"Review limit reached"** — its quota is per-PR and this one's window has not opened, so no review ran and `reviews[]` is empty. Two rounds of local Opus 5 review below; **all findings fixed in `472dbbd0`**.

---

### Round 1 — `fee066b5` — REQUEST_CHANGES (3 major, 3 minor, 1 nit)

**Major 1 — the dropped exit-status guard.** Both probes had `if !out.status.success() { return None; }` on `origin/main`; the rewrite removed it while every other subprocess in `pdf.rs` still checks (`grep -n "status.success"` → 73, 527, 624 — the two probe sites gone). Measured, not reasoned: `mdls -raw -name kMDItemNumberOfPages <non-pdf>` prints `(null)` and exits **0**; on a missing file it exits **1** with the error on a stderr this now discards. Latent today because both fail `parse::<u32>()` — but the guard was what kept that true.

**Major 2 — duplicate of `dap/install.rs:142`.** Identical statement-for-statement at that head, the `&str`/`&Path` signature difference forcing `Path::new("-raw")` for arguments that are not paths.

**Major 3 — a fixed 150ms handed to a `recv_timeout`** is the case `CONTRIBUTING.md:435-456` names verbatim, and `pdf.rs`'s own test module has the `budget()` helper twelve lines above the new test, used by its neighbour. CI runs `RUST_TEST_THREADS: 4`. This is the class #397 was filed for.

**Minors:** child leaked when `stdout.take()` returns `None`; `read_to_string` lost the `from_utf8_lossy` recovery the old code had, and the read was uncapped; the process-group divergence from `run_pdftoppm` was undocumented while citing its reasoning. **Nit:** `a_quick_probe` used a bare `echo`, so it never reached the success-path kill/reap.

### Round 2 — `5838d650` — one new major

**Portability regression introduced by the Major 3 fix.** `budget()` is `#[cfg(unix)]`; the three new spawning tests were not gated, so the test build breaks on Windows with `cannot find function budget in this scope`. All nine pre-existing spawning tests in the module carry the gate; these were the only ungated items calling a gated helper. Invisible locally and in CI — both are unix, and the ship path builds only the two musl triples.

**Major 2 withdrawn, on evidence.** Fixing Major 1 falsified the premise: `run_bounded_stdout` now returns the status, caps at 1 MiB and recovers lossily, while `dap/install.rs`'s caller validates via `Path::new(path).is_file()` — so an interactive rc that prints a usable path while exiting non-zero **must still succeed** there. Merging would impose a rejection that caller specifically must not have.

---

### Fixes in `472dbbd0`

| Finding | Fix |
| --- | --- |
| M1 status guard | Returns `(ExitStatus, String)`; both probes gate on it; test uses output that **genuinely parses** from a failed run |
| M3 fixed timing | `budget(150)` with the ceiling derived from it, matching the neighbour |
| New major | `#[cfg(unix)]` on all three spawning tests; the pure invariant test left ungated |
| Child leak | Reaped explicitly instead of returning through `?` |
| Lossy read | `take(MAX_PROBE_STDOUT).read_to_end()` + `from_utf8_lossy` |
| Process group | Documented, with why neither probe forks |
| Success-path reap | `sh -c 'sleep 0.05; echo …'` reaches it; a bare `echo` never did |

**Coverage gap closed after I got the framing wrong.** I declined direct probe tests because they would no-op where `pdfinfo` is absent — the false green this session has been about. The reviewer pointed out the gap was narrower: the **`mdls` argv**, coverable with no subprocess at all. `mdls_argv` is now a pure function asserted element by element, so a reordering or misspelt attribute name fails on any machine.

Also pinned: `PDF_INFO_BUDGET < PDF_RENDER_BUDGET`, which the doc comments argue at length and nothing else would have noticed being raised.

**Final scores: CodeRabbit 4/5, Greptile 4/5.** Verified locally at head: clippy clean under `-D warnings`, 29 pdf tests pass, all five new ones named in the output.

PR #520, opened 2026-09-07
feat(terminal): draw the focused pane's border in heavy glyphs

### LINKED ISSUE #470: Make the selected terminal border twice as thick for more obvious visual cue

vitali87:
## Summary

Closes #470. The focused pane's box is drawn in heavy line-drawing glyphs (`┏━┓`) against the unfocused panes' light ones (`┌─┐`), so which pane has the keyboard reads from **weight** and not only from the border's colour.

**Weight rather than a second ring.** The issue asks for "twice as thick", and a terminal cell is the smallest unit there is — so a literally doubled border means a second box one cell inside the first, costing the shell a row and a column on every side of every pane. Heavy glyphs give the emphasis for free, which is what iTerm2 and WezTerm do for the same reason. The title's stated goal is "a more obvious visual cue"; the thickness was the proposed means.

## Three collisions, none obvious from the diff

**The OSC 9;4 progress gauge.** It paints its fill over the bottom border, and read as fill only because the border was `─` and the fill `━`. A heavy border made both `━`, so the gauge showed **100% on the focused pane** — the one most likely to be running the build being measured. The fill glyph is now picked against the border's weight: `═` on heavy, `━` on light, each the same contrast trick one step up. `═` appears nowhere else in the codebase, so the "cleared gauge" assertion cannot be satisfied by an intact border.

**The Black theme.** There `paint_gradient_box` *replaces* the block's border with its own glyphs, so a `BorderType::Thick` underneath is painted straight over — the change would have appeared to work on every theme except the one it was drawn over. That function has 28 call sites, so it keeps its light weight and gains a `paint_gradient_box_heavy` sibling that only the focused terminal pane uses. All 27 non-terminal callers are unaffected.

**A guard I degraded while refactoring.** Adding the weight parameter left `paint_gradient_box_weighted` with a weaker bounds check than the original — it dropped the lower-bound tests and the buffer-origin offsets, so the heavy path could panic on a rect the light path safely declines. Both entry points now share the original guard.

Corners stay rounded in both weights: there is no heavy arc glyph, and the arcs are the house look every caller of the gradient box shares. `Block::inner()` is unaffected by border type, so the PTY does not resize on focus change.

## Tests

`the_gradient_box_comes_in_two_weights` pins all four edges in both weights and asserts the corners match. Four rather than two because the first revision replaced one horizontal edge twice and left the other light — a test reading only the top would have missed its own inverse.

The gauge test's two hard-coded `'━'` literals now derive from focus state. The local review checked this was not a weakening by reverting only the fix and re-running the *updated* test: it still fails (`left: 0, right: 23`), so it still catches the exact bug it was edited around.

## Verification

At `eb680db`: clippy `-D warnings` exit 0, `cargo fmt --check` exit 0, doc-ownership clean, `ships_nothing.py` exit 1.

Unfiltered suite: 4268 passed / 5 failed — 3 pass in isolation (load flakes), and 2 fail identically on base `a970d5f` (a missing `aarch64-unknown-linux-musl` rustup target; a socket-path-length `EINVAL` under the long scratchpad prefix). Zero failures attributable to this branch.

0.1.912 with a release note — main is 0.1.911 after #493's PDF fix.

Closes #470

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.911"
+version = "0.1.912"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.911"
+version = "0.1.912"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -5876,6 +5876,49 @@ fn render_welcome_does_not_panic_in_default_80x25_with_many_notes() {
     term.draw(|f| app.render_welcome(f, area)).unwrap();
 }
 
+/// The focused pane's box is drawn in HEAVY glyphs and an unfocused one
+/// in light, so which pane owns the keyboard reads from weight and not
+/// only from colour (#470). Both weights keep the rounded corners: there
+/// is no heavy arc glyph, and the corners are the house look every other
+/// caller of the gradient box shares.
+#[test]
+fn the_gradient_box_comes_in_two_weights() {
+    use ratatui::buffer::Buffer;
+    let area = ratatui::layout::Rect {
+        x: 0,
+        y: 0,
+        width: 8,
+        height: 4,
+    };
+    let sym = |b: &Buffer, x: u16, y: u16| b[(x, y)].symbol().to_string();
+
+    let mut light = Buffer::empty(area);
+    crate::gradient::paint_gradient_box(&mut light, area);
+    let mut heavy = Buffer::empty(area);
+    crate::gradient::paint_gradient_box_heavy(&mut heavy, area);
+
+    // All four edges, not just two: the bug this was written for replaced
+    // one horizontal edge and left the other light, so a test that reads
+    // only the top would miss its own inverse.
+    for (x, y, what) in [(3u16, 0u16, "top"), (3, 3, "bottom")] {
+        assert_eq!(sym(&light, x, y), "\u{2500}", "light {what} edge");
+        assert_eq!(sym(&heavy, x, y), "\u{2501}", "heavy {what} edge");
+    }
+    for (x, y, what) in [(0u16, 1u16, "left"), (7, 1, "right")] {
+        assert_eq!(sym(&light, x, y), "\u{2502}", "light {what} edge");
+        assert_eq!(sym(&heavy, x, y), "\u{2503}", "heavy {what} edge");
+    }
+    // The corners are shared, which is what keeps the two boxes the same
+    // shape rather than two different visual languages.
+    for (x, y) in [(0, 0), (7, 0), (0, 3), (7, 3)] {
+        assert_eq!(
+            sym(&light, x, y),
+            sym(&heavy, x, y),
+            "corner ({x},{y}) is the same arc in both weights"
+        );
+    }
+}
+
 #[test]
 fn paint_gradient_box_draws_rounded_corners_with_corner_colours() {
     let rect = Rect {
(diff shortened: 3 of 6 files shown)


vitali87:

Both bots are silent, so this was reviewed locally by the `greptile-local` agent across two rounds.

**Round 1 (`caf3c16`): 2/5** — three P1s, two P2s.
**Round 2 (`eb680db`): 5/5, no findings.** Every prior issue fixed, and each fix confirmed by **mutation testing** rather than by reading the diff.

### What round 1 caught

**P1 — the heavy border erased the OSC 9;4 progress gauge.** The gauge paints `━` over the bottom border and reads as fill only because the border is `─`. A heavy border made both `━`, so it showed 100% on the focused pane. I reproduced it independently: 50 heavy cells counted where 23 were expected. Fixed by picking the fill glyph against the border's weight.

**P1 — my verification method.** I reported "86 passed" for a change to terminal border rendering. The binary has 4,284 tests and my filter never reached the one that breaks. This was the most valuable finding of the two rounds: I had spent the session catching filter-shaped false greens in other runners and then produced one myself at the final step. I now run the unfiltered suite.

**P1 — version collision.** `origin/fix/493-bound-pdfinfo` already held 0.1.911. Moved to 0.1.912 and renamed the note.

**P2 — a guard I degraded.** `paint_gradient_box_weighted` got a weaker bounds check than the function it was extracted from, so the heavy path could panic where the light path safely returns. Both entry points now share the original.

**P2 — my test checked two of four edges**, so it would have missed its own inverse. Now all four.

### What round 2 verified, by mutation

- **The gauge test update is not a weakening.** The decisive check: revert only the fix, run the *updated* test — it still fails (`left: 0, right: 23`). A test that still catches the bug it was edited around has kept its teeth.
- **Contrast holds in all four states** (focused/unfocused × gradient/plain), with 23-cell fills rendered and read out of the buffer.
- **The light path is byte-identical to base** — the `horiz`/`vert` indirection resolves to the same literals at `heavy: false`; 27 of 28 call sites are unaffected.
- **The weight test catches its own inverse**: mutating the bottom edge to stay light fails it; mutating the right edge fails it.
- **The heavy path now declines the rect that panicked it** — buffer at (10,5) 40×20, rect at (2,1) 6×4, both entry points return leaving byte-identical buffers.
- **`Block::inner()` is unchanged by border type**, so no PTY resize on focus change — the regression that would have been worse than the bug.

### Suite

4268 passed / 5 failed unfiltered. 3 pass in isolation (load flakes); 2 fail identically on base `a970d5f` (missing rustup cross-target; socket-path `EINVAL`). Zero attributable to this branch.

VERDICT: APPROVE

vitali87:
**Rebased onto `origin/main` after #493's PDF fix landed at 0.1.911. Head `eb680db` → `391fc36`.**

Flagging because the 5/5 review above names `eb680db`, and a review naming a commit the merge will not take is the #316 failure mode.

**Version conflict only.** Main took 0.1.911, so `Cargo.toml`/`Cargo.lock` re-resolved to 0.1.912. Re-enumerated every remote branch's claim before keeping it: only main holds 911, nothing holds 912. **No source file changed** — `src/widgets/terminal.rs`, `src/gradient.rs` and `src/app/tests.rs` are byte-identical to the reviewed commit, so every finding-bearing surface is unchanged.

Re-verified on the rebased tree with the **unfiltered** suite, which is the check that mattered in round 1:

```
TREE: fix/470-thick-focus-border @ 391fc36
CLIPPY_EXIT=0
TEST_EXIT: 4271 passed / 7 failed
FMT_EXIT=0
```

All 7 triaged rather than waved off. Five pass in isolation (1.04s, load contention): `a_restored_lane_pane_spawns_in_its_worktree…`, `change_workspace_root_skips_cd…`, `closing_a_lane_whose_pane_is_the_last_one…`, `the_explorer_dots_follow_the_agent_ledger`, `terminal_session_restores_pane_layout…`. The remaining two are the known environmental pair that fail identically on base — a missing `aarch64-unknown-linux-musl` rustup target, and a socket-path `EINVAL` under the long scratchpad prefix. **Zero attributable to this branch**, and `osc_9_4_progress_paints_a_border_gauge_and_pill_percent` — the regression round 1 caught — passes.

Pushed with `--force-with-lease=fix/470-thick-focus-border:eb680db`.

PR #521, opened 2026-09-07
feat(terminal): release Ctrl+B to the app in a focused pane

### LINKED ISSUE #304: Should a reserved chord be releasable to a full-screen terminal app (e.g. Claude Code's Ctrl+B)?

vitali87:
## Summary

Closes #304, which asked whether a reserved chord should be releasable to a full-screen terminal app, and whether a **general opt-out** was the right shape or whether the existing **F-key precedent** already answered it.

The answer is no opt-out. Claude Code uses `Ctrl+B` to background a running command and has no other route to that gesture; croft's Toggle Primary Side Bar has three — `Cmd+B`, the command palette, the View menu. That asymmetry is the argument: **releasing costs a chord reachable three other ways, holding costs a chord reachable no other way.**

## Conditional, not global

The chord falls through to the PTY only while the terminal pane is focused and the panel is on the Terminal tab, mirroring `terminal_owns_fkeys` directly above it. Everywhere else `Ctrl+B` still toggles the side bar.

That makes it a rule about **when the chord is unambiguous**, not a user preference — which is why it needs no setting. A preference would make every future report of this chord start with "what's your config?", and would let someone strand themselves by releasing a chord that is croft's only way back.

Only the `Ctrl` form. `Cmd+B` is untouched, which is what keeps the side bar reachable from the terminal, and it matches the prior art the docs already cite: iTerm2 and Ghostty reserve the Cmd form and pass Ctrl through. `Cmd+F` stays reserved for the opposite reason, now stated beside it — it is croft's only escape hatch *out* of a full-screen app.

## Docs

Two rows in `KEYBINDINGS.md` needed it, and the second is the one worth noting: the measured-chords table at line 508 listed `Ctrl+B` among chords "the shell never sees" and closed with *"`Ctrl+B` does not, which matters for Claude Code, where it backgrounds a running command"* — verbatim the behaviour this reverses. The local review caught it by rerunning that row's own stated methodology (driving every letter through the real key path and watching which reach the PTY), which is a better check than reading.

## Tests

`ctrl_b_belongs_to_the_app_in_a_focused_terminal` asserts the side bar does not move, `Cmd+B` still toggles it, **and the byte arrives** (`0x02` in `written_bytes_for_test`). `ctrl_b_still_toggles_the_side_bar_outside_the_terminal` pins the boundary.

The byte assertion exists because the review proved the original pair could not fail on a swallowed chord: injecting a guard that declines the toggle and then eats the key left all three tests green. I reproduced that mutation after strengthening the test — it now fails under it, as it should.

Two mutation controls run:
- neutering the guard → the release test fails, the boundary test stays green
- swallowing the chord → the release test fails on the missing byte

The reviewer independently confirmed `0x02` reaches the pane **including in the alternate screen**, which is the real Claude Code case and something my tests do not cover.

## Verification

At `0baf9ef`: clippy `-D warnings` exit 0, `cargo fmt --check` exit 0, doc-ownership clean, `ships_nothing.py` exit 1.

Unfiltered suite: 4271 passed / 9 failed. Seven pass in isolation (load contention — four sessions share this host tonight), including `widgets::terminal::tests::the_reader_thread_records_shell_output_for_rewind`, which I checked specifically because it is in the widget this change touches. The remaining two fail identically on base (a missing `aarch64-unknown-linux-musl` rustup target; a socket-path `EINVAL`).

0.1.913 with a release note.

Closes #304


vitali87:

Both bots are silent, so this was reviewed locally by the `greptile-local` agent against `e7509ca`. **Confidence Score 3/5**, one P1 and one P2. Both fixed in `0baf9ef`.

### P1 — a doc line contradicted the change, in the same table

`docs/KEYBINDINGS.md:508` lists chords "**Claimed by croft, whatever pane has focus** — the shell never sees it" and included `Ctrl+B`, closing with *"`Ctrl+B` does not, which matters for Claude Code, where it backgrounds a running command"*. That is verbatim the behaviour this PR reverses, six rows above the row I did update.

The reviewer found it by **rerunning that row's own stated methodology** — sweeping `'a'..='z'` with CONTROL through the real key path and recording which reach the PTY. Measured result: croft claims `J P Q S V`; the doc claimed `B F J P Q S V`.

**Fixed:** `B` dropped from the list and the trailing clause rewritten. (`F` also differs, but identically on `origin/main` — pre-existing and out of scope, noted below.)

### P2 — my tests could not fail on a swallowed chord

Both new tests asserted only that `show_tree` did not move. A guard that declines the toggle and then eats the key satisfies that perfectly while delivering nothing — which is the entire point of the change.

The reviewer proved it rather than asserting it: injecting `if terminal_owns_ctrl_b && is_sidebar_toggle_key(key) { return Ok(()); }` left all three `ctrl_b_` tests green.

**Fixed:** the release test now asserts `0x02` reaches `written_bytes_for_test()`. I reproduced the swallow mutation afterwards to confirm the strengthened test catches it — it fails under the exact mutation that previously passed.

### Verified, not defects

- **The bytes genuinely arrive.** Exactly one byte (`0x02`) is written, **including in the alternate screen** — the real Claude Code case, which my own tests do not cover.
- **`key.modifiers == KeyModifiers::CONTROL` (exact) is right.** Ctrl+Alt+B and Cmd+Opt+B still toggle the secondary side bar, Ctrl+Shift+B stays inert, Ctrl+Cmd+B still toggles the primary. It composes with `is_sidebar_toggle_key` (which rejects ALT/SHIFT and requires CONTROL-or-SUPER) rather than duplicating or contradicting it, so `Cmd+B` is never released.
- **Nothing claims `Ctrl+B` first.** Every modal and picker returns before the guard, so with a modal open the chord goes to the modal — same as the F-key precedent.
- **The `Cmd+F` contrast is accurate.** The side bar genuinely has three routes (`toggle_side_bar()` at three call sites); `open_terminal_find` has exactly one caller, so `Cmd+F` genuinely is the only route to terminal find.
- **Uppercase `Ctrl+Shift+B`** behaves consistently, since `is_sidebar_toggle_key` uses `eq_ignore_ascii_case`.
- **0.1.913 is unclaimed** across every remote branch.

`KEYBINDINGS.md:508` also lists `Ctrl+F` as croft-claimed, but `is_terminal_find_key` requires SUPER on macOS, so `Ctrl+F` reaches the PTY there. Identical text on `origin/main` — a separate correction.

VERDICT: APPROVE (after both findings fixed)

PR #522, opened 2026-09-07
test(terminal): sweep every collapse-recovery path (part of #468)

vitali87:
## Summary

Part of #468. **Does not fix it** — the sweep comes back green, which is itself the finding.

## Why this exists

Four theories about #468 have died against the tree:

1. The editor/panel splitter's grab zone — the chevron sits a row clear of it.
2. `toggle_terminal` clearing `terminal_maximized` but not `collapsed` — a real inconsistency, but recoverable with `Cmd+K ]`.
3. Index 0 being special — disproven by #514, now permanent coverage.
4. The last-expanded guard — it makes the state *unreachable* rather than unrecoverable.

Each died **by inspection**, and that is a weaker claim than it sounds. "I could not construct the path" and "the paths do not exist" are different statements, and only the second is worth acting on. This enumerates instead of arguing.

## What it covers

36 combinations: **2 and 3 panes** × **both side bar positions** × **nothing / a resize / a maximize cycle** × **every pane index**. Each folds a pane, applies the perturbation, and asserts the pane comes back from its one-column strip.

The perturbations are not arbitrary. #475 fixed a resize-related repaint bug, so resize is the one input already known to leave stale geometry behind; maximize bypasses the pane-constraint function entirely.

All 36 recover.

## What that means for #468

The reported failure needs a condition outside this space — and the space is now written down rather than argued about. If a future change breaks any of these paths, this test names the case directly, which is worth having whether or not it ever helps with the original report.

The single most useful thing on that issue remains one observation from the reporter: **did `Ctrl+L`, or resizing the window by a column, bring the pane back?** That splits a paint-path bug from state corruption in one answer.

## Verification

Test-only (`src/app/tests.rs` alone), so no version bump — CI strips that path before `ships_nothing.py` runs, matching #503, #507 and #514.

At `493980a`: clippy `-D warnings` exit 0, `cargo fmt --check` exit 0, doc-ownership clean, the new test passes in 6.36s.

Part of #468

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage verifying that collapsed panes remain stable across supported layouts, resizing, and maximize actions.
  * Verified that clicking a collapsed pane’s strip consistently restores the pane.

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -42324,6 +42324,127 @@ fn maximize_ignores_the_collapse_flags_and_gives_them_back_on_exit() {
     );
 }
 
+/// Closing the last EXPANDED pane leaves a panel of nothing but strips,
+/// and something must bring one back (part of #468).
+///
+/// `toggle_terminal_collapse` refuses to fold the last expanded pane,
+/// but closing reaches that state from the other side, which is a route
+/// the fold guard cannot see. `ensure_a_terminal_pane_is_expanded` is
+/// the answer and has exactly one caller; this pins that the caller is
+/// on the path that needs it.
+#[test]
+fn closing_the_last_expanded_pane_leaves_one_expanded_behind() {
+    let (_tmp, mut app, mut term) = app_with_terminal_panes(2);
+    term.draw(|f| app.render(f)).unwrap();
+
+    app.toggle_terminal_collapse(1);
+    assert!(app.terminals[1].collapsed, "premise: pane 1 is folded");
+    assert!(
+        !app.terminals[0].collapsed,
+        "premise: pane 0 is the only expanded one"
+    );
+
+    // Close the expanded one. The fold guard never fires -- nothing is
+    // being folded -- so the panel would be all strips without the
+    // close path's own guard.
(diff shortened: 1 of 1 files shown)


vitali87:
Fixed in `8d624d4` — CodeRabbit's finding was right, and it is the exact failure mode this test exists to refuse.

The `continue` skipped any case where the perturbation had cleared `collapsed`. A maximize cycle that unfolds panes would have made those cases pass **without ever clicking a strip**, so the sweep would report 36 paths verified while walking far fewer. A test that silently narrows its own coverage is worse than one that does not claim the coverage at all.

Replaced with an assertion that the fold survives the perturbation. It passes, which also answers the question the skip was hiding: **no case was actually skipping** — all 36 genuinely walk the strip click. The `continue` was not masking anything today; it made that unverifiable, and would have masked it silently after any future change to the perturbation behaviour.

Thanks for catching it. This is the first bot review on any PR this session — Greptile has been out of credits and CodeRabbit rate-limited since the start, so every merge tonight was gated by a locally-run Opus review instead. Good to have the independent gate back.

PR #523, opened 2026-09-07
feat(git): record a lane's base branch and range the commit graph

vitali87:
Part of #348 — the two structural pieces design item 4 needs. **Not the command itself**, and deliberately so; see the bottom.

## What item 4 actually asks for

> **Agent: Merge Lane** opens the branch's **diff** against the base as a commit-graph selection and hands off to the existing merge editor for conflicts.

I had been reading that as a merge and asking which root it should run in. It says *diff*. The command name misled me, which is worth recording because it's the same shape as the stale comment that misled two sessions on #506 tonight — a label disagreeing with the text beside it.

Neither half of "the branch's diff against the base" was expressible.

## `branch_history_range`

`branch_history` ran `git log --first-parent HEAD` with no range, so `base..lane` — the commits a lane added and none of the base's — could not be asked for.

Refactored rather than copied: `branch_history` is now a one-line call into the ranged version with `"HEAD"`, so there's one body and one place for the `--first-parent` reasoning to live.

## `WorktreeLane.base`

`worktree add -b` takes no start-point (`git.rs:1169`), so a lane forks from whatever HEAD happened to be. Git records that **only in the reflog** — which is local, prunable at `gc.reflogExpire`, and absent from a fresh clone. A lane created on a laptop has no base to recover on a desktop. `merge-base` can't help either: it needs the candidate, which is the thing being asked for.

So the base is knowable at creation and unrecoverable afterwards. `add_worktree_lane` now reads HEAD **before** creating the worktree, while it's still the branch being forked from. `Option<String>` because a detached HEAD has no branch name and must not be guessed at.

## The test

Names a **non-default** branch — `release-42`, not `main` — so an implementation that passed a hardcoded default through would fail rather than look plausible. It also pins the two shapes a careless version takes:

- `plan` leaves `base` as `None`, since it's pure and reads no git
- the base is never the lane's *own* branch

## Verification

`cargo check`, `clippy -D warnings`, and the lane tests all pass, with the new test required **by name** in the runner.

One test failed on the way — `a_new_lane_opens_a_named_pane`, which polls a real shell. Probed rather than assumed: **3/3 alone on this tree, and green without the diff**, so it's the load-sensitive class rather than this change. The second half is the control; three green runs alone would only show it *can* pass, not that a signature change I made was innocent.

Merged 52 commits of main and re-ran everything, since a build that passed before that merge proves nothing about after it.

## Why the command isn't here

The remaining work is a palette entry, a dispatch arm, and the hand-off to `open_merge_editor_for` — all mechanical now that the base and the range exist. What isn't mechanical is whether item 4 is *only* the diff viewer it describes, or a diff viewer plus an actual merge. If it's the latter, the direction question matters: `git merge <lane>` in the primary brings the lane's work home, in the lane it refreshes the lane from base — and `close_worktree_lane` deliberately acts on `active_scm_root`, with a comment recording that targeting the primary destroyed a lane in testing.

These two pieces are needed either way, so they land now rather than waiting on that.

## Summary by CodeRabbit

* **New Features**
  * Worktree lanes now remember the branch they were created from.
  * Branch history can be retrieved for custom revision ranges, supporting comparisons between a lane and its base branch.
  * Detached worktrees continue to work without base-branch metadata.

* **Documentation**
  * Added release notes for version 0.1.916, including current limitations around displaying lane-base information.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.915"
+version = "0.1.916"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.915"
+version = "0.1.916"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -25502,7 +25502,7 @@ impl App {
     /// and the Explorer would show the lane inside the root it was cut from.
     fn create_worktree_lane(&mut self, name: &str) {
         let repo = self.workspace_root().to_path_buf();
-        let Some(lane) = crate::git::WorktreeLane::plan(&repo, name) else {
+        let Some(mut lane) = crate::git::WorktreeLane::plan(&repo, name) else {
             // Refused rather than defaulted: a name with nothing usable in
             // it would otherwise become the branch `agent/`.
             self.status = format!("{name:?} has no letters or digits to name a lane after");
@@ -25512,7 +25512,7 @@ impl App {
             self.status = format!("{} already exists — pick another name", lane.path.display());
             return;
         }
-        match crate::git::add_worktree_lane(&repo, &lane) {
+        match crate::git::add_worktree_lane(&repo, &mut lane) {
             Ok(()) => {
                 self.add_workspace_folder(lane.path.clone());
                 self.open_lane_pane(&lane);
--- a/src/git.rs
+++ b/src/git.rs
@@ -1008,6 +1008,18 @@ pub struct WorktreeLane {
     pub path: PathBuf,
     /// The branch created for the lane.
     pub branch: String,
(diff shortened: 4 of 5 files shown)


vitali87:
Fixed in `dcdcfed`. The finding was correct and the subtlety is worth recording, because the naive fix would have broken the feature.

The **read** genuinely must precede `worktree add`: HEAD is the fork point only until the worktree exists, which is the whole reason this field is recorded at creation rather than derived later. But the **assignment** did too, and those are separable — I had conflated them. Now:

```rust
let base = current_branch(repo);
run_git_mut(repo, &["worktree", "add", "-b", &lane.branch, &path])?;
lane.base = base;
```

Moving the *read* after the command would have compiled, passed the existing tests, and silently recorded the wrong branch — the lane's own, since `worktree add -b` checks it out. So "move the assignment" and "move the call" look alike here and only one is right.

Added `a_failed_lane_creation_leaves_the_base_unset`, driven through the real duplicate-name refusal rather than a mocked failure, so it fails the way a caller sees it. It asserts the first lane's base as a precondition, so the `None` cannot pass vacuously.

Red/green verified rather than assumed: with the ordering reverted the test fails on `left: Some("release-9")` / `right: None`. A test written in response to a review finding is easy to write so it passes either way — asserting only that a failed creation returns `Err` would have been green with the bug fully present.

PR #524, opened 2026-09-07
feat(debug): honour a compound's presentation.hidden

vitali87:
## Summary

Part of #318. A `launch.json` compound asking for `presentation` was refused wholesale, so one asking **only** to be hidden was declined for a capability it never requested.

The key splits, and the issue's "larger change than it looks" applies to only half of it:

- **`presentation.hidden`** is a filter on the picker's rows — croft can do this today.
- **`presentation.group` / `presentation.order`** want an ordering the picker has no concept of, and still refuse.

Testing the **sub-keys** rather than the parent is what turns a blanket refusal into a launch for the common case.

## Two surfaces, not one

The first revision filtered only the picker. `refresh_run_debug` also counts compounds for the Run and Debug config row, and its own comment states the invariant: *"the picker lists them, so counting only configurations makes the row vanish."* Counting what the picker drops advertises an entry that isn't there — a workspace whose only compound is hidden would render `⚙ 1 debug entry ▾` over a picker offering nothing. Both now use the same filter, with a test asserting they agree.

## Index stability

Row ids are `compound:{i}` indexing `debug_compounds`, so the filter runs **after** `enumerate`. Filtering before renumbers every row past a hidden one, and selecting "Third" would launch whatever slid into its slot. A red control confirms the test catches exactly that: moving the filter ahead of `enumerate` fails on the "Third keeps index 2" assertion.

## Malformed shapes still refuse

`{"hidden": "true"}` — a quoted boolean, the likely typo here — must not read as "asks for nothing". On main it refused and the user found out; silently launching it unhidden would be a regression dressed as a feature. A `presentation` that isn't an object, and a `hidden` that isn't a boolean, both record as unsupported.

## Keys are named with their parent

The refusal says `presentation.order`, not bare `order`. `order` is a generic word, and "remove that key" would then read as deleting the whole `presentation` block — taking a working `hidden` with it, which is the outcome this split exists to avoid.

## Also

The issue body is **stale** in one respect I verified: it calls `preLaunchTask` "the tractable one, still missing", but it shipped — parsed in `configs.rs`, chained via `launch_after_compound_task`, gated on exit 0. Only `presentation` remained.

`docs/KEYBINDINGS.md` and `docs/LAYOUT.md` both still said a compound setting `presentation` is refused; both now name the sub-keys and document that `hidden` keeps a compound out of the picker.

## Tests

Three existing tests asserted the coarse `["presentation"]` contract and now expect the qualified path. Each fixture was checked to use `{"order": N}`, so the refusal path is still exercised — including the one serving as a **positive control** against six vacuous `is_empty` assertions, whose comment was updated since `hidden` can no longer serve that role.

## Verification

At `a7c31f6`: clippy `-D warnings` exit 0, `cargo fmt --check` exit 0, doc-ownership clean, `ships_nothing.py` exit 1.

Unfiltered suite: **4330 passed / 3 failed** — one load-sensitive shell-probe timeout and the two known environmental failures that fail identically on main (a missing `aarch64-unknown-linux-musl` rustup target; a socket-path `EINVAL`).

0.1.917 with a release note — main is 0.1.915, and 0.1.916 is held by `feat/348-merge-lane`.

Part of #318

## Summary by CodeRabbit

- **New Features**
  - Added configurable project-wide TypeScript checking with background status reporting.
  - Added reviewed-baseline and reviewed-diff views for agent files, plus worktree commit diffs.
  - Ctrl-clicking valid file references in logs now opens those files.
  - Hidden debug compounds are excluded from picker results and counts while remaining compounds retain stable identifiers.
  - Added support for installing target-specific archives with `cargo binstall`.

- **Bug Fixes**
  - Symbol tabs now close when the active file changes.

- **Documentation**
  - Clarified debug compound behavior, including hidden, grouped, and ordered compounds.


vitali87:

Greptile is out of credits; this was reviewed locally by the `greptile-local` agent against `002dd1b`. **Confidence Score 3/5** — one P1 and three further findings. All fixed.

### P1 — the change stopped at the picker

`refresh_run_debug` counts compounds for the Run and Debug config row, unfiltered. I filtered the picker and not the count, so a workspace whose only entry is a hidden compound would render `⚙ 1 debug entry ▾` over a picker offering nothing.

The reviewer noted this inverts the invariant an existing test states in its own comment — *"the row counts what the picker lists"*. **Fixed:** both surfaces use the same filter, with `the_config_row_does_not_count_a_hidden_compound` asserting they agree.

### P2 — a malformed `presentation` silently lost its refusal

`{"hidden": "true"}` — a quoted boolean, the likeliest typo here — read as "asks for nothing" through `as_bool()`, so it launched listed and unhidden with no signal. On main it refused and the user found out.

**Fixed:** a `presentation` that is present but not an object, and a `hidden` that is present but not a boolean, both record as unsupported. New coverage in `presentation_hidden_is_honoured_while_group_and_order_are_not`.

### P2 — the refusal named a bare sub-key

The message read *"sets order, which croft does not honour yet — remove that key"*. `order` is a generic word, and "remove that key" then reads as deleting the whole `presentation` block — taking a working `hidden` with it, which is exactly what this split exists to prevent.

**Fixed:** keys are qualified paths (`presentation.group`, `presentation.order`), so the user can find them in the file.

This also corrected an assertion of my own. I had asserted the message must *not* contain `"presentation"`, reasoning the parent overstates the refusal. The reviewer agreed with the principle but showed the fix overshot — the assertion was a proxy for the intent, and it broke when the implementation improved in the direction the assertion wanted. It now strips the qualified path and checks the parent never appears alone.

### P2 — docs still described the old behaviour

`docs/KEYBINDINGS.md` and `docs/LAYOUT.md` both said a compound setting `presentation` is refused, and KEYBINDINGS claimed compounds are listed "with their members" without qualification. **Fixed:** both name the sub-keys and document that `hidden` keeps a compound out of the picker.

### Verified, not defects

- **Index stability holds.** Filtering after `enumerate` keeps `compound:{i}` naming the same element; the reviewer confirmed that filtering before would emit `compound:1`, which is the hidden compound.
- **`hidden` is honoured on every route.** `open_debug_config_picker` is the only builder of `compound:` rows and the only `ListPurpose::DebugConfig` picker; `selected_debug_config` stores a resolved member config's name, never a compound, so no saved selection or F5 re-launch reaches a hidden one. This was the "is it half-done?" question I asked, and the picker/row-count gap above was the answer.
- **The three changed tests are legitimate updates**, not weakened — each fixture moved to `{"order": N}`, keeping `unsupported_keys` non-empty and the refusal branch exercised.

### Verification at `fc5c218`

Rebased onto main twice during review as it moved 0.1.915 → 0.1.918; now at **0.1.920** (0.1.919 is held by `feat/348-lane-diff`).

Unfiltered suite: **4334 passed / 6 failed** — four load-sensitive tests that pass in isolation, and the two known environmental ones that fail identically on main. Clippy `-D warnings` clean, `cargo fmt --check` clean, doc-ownership clean, `ships_nothing.py` exit 1.

VERDICT: APPROVE (after all four findings fixed)

vitali87:
**Rebased onto main (0.1.924). Head `9b39a63` → `a0fb76f`, version 0.1.923 → 0.1.926.**

Fifth re-bump on this PR (914 → 917 → 920 → 923 → 926), each one a peer landing while review findings were being fixed. 925 is croft-2's; 924 is main.

Two changes since the last review, both from peer findings rather than my own checking:

**The config-row test was drop-only.** Every assertion said "nothing is there" — `config_count == 0`, no compound rows — and an empty `launch.json`, a broken parser and a working filter all produce exactly that. Added the positive leg: the *same* compound with `"hidden": false` must produce a count of 2 and appear in the picker, so absence now means something because presence is proven in the same test. Mutation-verified: deleting the filter fails it, and I asserted the anchor text was present before mutating, since a mutation that silently fails to apply reads as one the test survived.

**The fixture's discriminating property is now asserted, not commented.** The counts only mean "the compound was excluded" while exactly one configuration is declared. That was a comment; a later edit adding a second config would have made 1 and 2 wrong for a reason unrelated to hiding, and the obvious repair — bumping the numbers — would leave a test that no longer measures what it claims. It now asserts `discover_configs(...).len() == 1` up front.

The general form, which is worth more than either fix: **mutation testing answers "does this test detect the implementation being gone", not "does it test what it says it tests".** A fixture can be wrong in a way that makes both the right and wrong implementations agree.

### Verification at `a0fb76f`

Cleanest run of the session: **4346 passed / 2 failed** unfiltered, and both failures are the known environmental pair that fails identically on main (a missing `aarch64-unknown-linux-musl` rustup target; a socket-path `EINVAL`). No load flakes — the host was quiet.

Clippy `-D warnings` zero warnings, `cargo fmt --check` clean, doc-ownership clean, `ships_nothing.py` exit 1.

PR #525, opened 2026-09-07
feat(fleet): saved host groups, refused from workspace layers

vitali87:
Part of #363 — **criterion 3 only**. Criteria 1 and 2 both describe *tiles*, and there is no fleet widget; that is a view design, not plumbing.

## The trust question was already answered

I was going to ask whether a repo may name SSH hosts. It can't, and croft had already decided: `prefs.rs` has `remote_offer_excluded_hosts: Vec<String>` — a host-name list, **absent from `WORKSPACE_ALLOWED_KEYS`**.

`fleet_groups` follows it exactly, and the reason is sharper here: a fleet run executes arbitrary text on every host in the set, so a checked-out repo defining a group could add a production box to a set the user then broadcasts to. `config_layers.rs` states the rule — workspace layers are repo-controlled input, and keys that change trust posture are refused with a visible warning.

I extended the **existing** refusal test rather than writing a parallel one, so `fleet_groups` is now covered by the same assertion that proves a clone cannot grant `mcp_consented`.

## What the expansion preserves

`parse_request_with_groups` keeps the unknown-host check on **every expanded name**. A group is shorthand for hosts croft already knows, never a way to name a machine absent from `~/.ssh/config`.

Three tests pin what a careless version gets wrong:

| test | what it catches |
| --- | --- |
| `a_saved_group_expands_to_its_hosts_and_only_those` | fixture gives 2 of 3 known hosts, so expanding to the whole fleet — the shape `*` already has — fails |
| `a_group_cannot_smuggle_an_unknown_host_past_the_check` | an unknown member refuses the **whole** request; paired presence case proves the refusal isn't a broken fixture |
| `a_group_name_takes_precedence_over_a_host_of_the_same_name` | stated rather than left to whichever branch runs first |

`BTreeMap` not `HashMap`: a fleet comparison is read across tiles, and a group whose order changed between runs would move them.

## Three defects, all found by building

1. **`App` has no `prefs` field** — prefs load into dedicated fields.
2. **The live re-merge path needed it too.** Found while fixing #1. Without it, `fleet_groups` would work at startup and then silently stop tracking `config.json` — a group edit would take effect only after a restart, and nothing would fail. That is the one I'm most glad of; it fails quietly.
3. **Dead code.** Keeping `parse_request` as a wrapper left it with no production caller once I rewired the real one, and `-D warnings` implies `-D dead-code`. Removed rather than `#[allow]`ed — a wrapper alive only through a test is the shape that rots. The tests now call the shipping path.

## Verification

63 group tests pass with all three new ones required **by name**; the trust test passes; clippy clean under `-D warnings`.

Verification was killed three times for host memory — no verdict either way, not failures — so the runner is now staged with tests before clippy, letting clippy reuse the test harness instead of building `--all-targets` from cold twice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.915"
+version = "0.1.917"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.917"
+version = "0.1.918"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.917"
+version = "0.1.918"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.915"
+version = "0.1.917"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -86,7 +86,7 @@ allowlist — appearance and editor/terminal behavior:
 `disable_inline_values`, `disable_bracket_colors`, `disable_indent_guides`,
 `disable_inlay_hints`, `copy_on_select`, `disable_secret_redaction`, `disable_log_highlight`, `explorer_views`.
 
-Everything else — `disabled_extensions`, `mcp_consented`, `disable_remote_offer`, `remote_offer_excluded_hosts`, `lane_agent`,
+Everything else — `disabled_extensions`, `mcp_consented`, `disable_remote_offer`, `remote_offer_excluded_hosts`, `fleet_groups`, `lane_agent`,
 `mcp_tool_fingerprints`, `host_accents`, `notifications`, and any future key not explicitly
 allowlisted — is ignored from workspace layers with a visible warning.
 Extending the allowlist is a deliberate review decision, not a default.
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2841,6 +2841,9 @@ pub struct App {
     remote_offer_disabled: bool,
     /// `remote_offer_excluded_hosts` from prefs (#364).
     remote_offer_excluded: Vec<String>,
+    /// `fleet_groups` from prefs (#363): named host sets a fleet run can
+    /// expand, so a fleet is named once rather than retyped per run.
+    fleet_groups: std::collections::BTreeMap<String, Vec<String>>,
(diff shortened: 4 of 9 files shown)


vitali87:

CodeRabbit's check here reports `pass` while annotated **"Review limit reached"** — its quota is per-PR (#523 got a real review minutes earlier), so no review ran on this one and `reviews[]` is empty.

**Verdict: REQUEST_CHANGES** — 1 blocker, 3 majors, 1 minor. **All fixed in `cc0e7f8`.**

---

### Blocker — a live version collision the release gate cannot catch

PR #524 also sets `version = "0.1.917"` and also adds `src/release_notes/0.1.917.md`. `origin/main` is at 0.1.915; #523 holds 0.1.916.

The gate compares each head against **its own merge base**, so both PRs pass legitimately and only the second to merge conflicts — on `Cargo.toml`, `Cargo.lock` *and* the notes filename. I had enumerated open PRs before bumping; #524 appeared afterwards. **Fixed: 0.1.918.**

### Major — the release note would render wrong

The note began `change:`. `release_notes.rs:87-100` matches only `feature:` and `fix:` case-insensitively; anything else falls to the `_` arm, which classifies it `Fix` **and keeps the whole line as the summary** — so the welcome panel would have shown a fix bullet reading literally "change: name a set of SSH hosts…". The CI gate only checks the file is non-empty, so it ships silently. **Fixed: `feature:`**, verified against the parser.

### Major — the doc claim was false

I wrote that the doc comment "moved to the surviving function intact". It didn't — the two blocks were **concatenated**, leaving two summary sentences with no blank line, so rustdoc's short description was the *old* function's.

The sharp part: `check_doc_ownership.py` passes because it detects docs that are **lost**, not two that are **fused**. Its green was not evidence. **Fixed:** one summary, then the syntax/refusal prose, then the group rules.

### Major — an empty group silently narrowed the fleet

With `{"web": []}`, the spec `web,db1: cmd` returned `Some([db1])`. The user believes they addressed a fleet and addressed one box — the same harm the unknown-host check exists to prevent, and every other refusal in that function is loud. **Fixed: refuses**, with a test covering the alone and mixed forms plus a paired presence case.

### Minor — `docs/SETTINGS.md` omission

The settings reference enumerates refused keys by name, and every sibling host-name key is listed. `fleet_groups` was absent, covered only by the trailing catch-all. **Fixed.**

### Test coverage, addressed

- **A user layer could not be shown to work.** The refusal test would have passed even if `fleet_groups` never deserialized at all — an inert feature. It now carries a user-layer group asserted to survive, so refusal is proven rather than assumed.
- **Self-reference and mutual reference** are now pinned. The doc promises one-level expansion; a later move to recursion would otherwise break that silently.
- **Empty groups** covered in both forms.

### Verified clean

Case-insensitive matching is consistent with `remote.rs`'s ssh host handling. Group-beats-host precedence is defensible, documented and tested — a collision is user-authored and both directions of silent shadowing are worse. Removing `parse_request` was clean: no surviving callers. `apply_merged_settings` genuinely carries the field, so a `config.json` edit tracks live.

### Scores

**CodeRabbit 1/5** at review time — one blocker plus three majors. **Greptile 3/5** — the trust boundary follows `remote_offer_excluded_hosts` exactly and the security property held under every expansion path traced, but the note prefix, the SETTINGS omission and a stale rationale each departed from conventions the repo keeps consistently.

**After `cc0e7f8`:** 65 group tests pass with all five named, trust test passes, clippy clean under `-D warnings`.

PR #526, opened 2026-09-07
feat(agent): diff a worktree lane against the branch it was cut from

### LINKED ISSUE #348: Worktree lanes: one chord creates a git worktree, adds it as a workspace root, and starts an agent in it

vitali87:
**Closes #348**, design item 4 — the last item on that issue.

## The spec says diff, not merge

> **Agent: Merge Lane** opens the branch's **diff** against the base as a commit-graph selection and hands off to the existing merge editor for conflicts.

I spent several rounds asking which root a *merge* should run in. The command's name says merge; the sentence beside it describes a viewer. Reading the label over the text is the same failure as the stale comment that misled two sessions on #506 earlier.

## The tag, not new machinery

I first reported that the COMMITS channel "cannot carry the answer" and offered three designs. That was wrong, and worth correcting explicitly: `graph_fetch_inflight`, `graph_refetch_queued`, and a drain that already discards replies for a since-left root **all exist**. What was missing was the tag's *resolution* — `PathBuf` alone, so "history for root X" and "lane diff for root X" were indistinguishable.

`(PathBuf, Option<String>, rows)` and the existing discard rule does the rest. No modality to own, no new state machine.

## `LaneRecord` carries the base

Caught while wiring: #523 records the base in memory, but without persisting it the diff breaks after any restart. `worktree add -b` leaves no start-point and the reflog that records one is local and prunable — `LaneRecord` is the only copy that outlives the session that made the lane.

## Guards

Uses `active_scm_root` like `close_worktree_lane` beside it, not the primary: a lane is never index 0, and a sibling command that silently switched roots is how that one destroyed a lane in testing. A lane with no recorded base names the lane rather than showing an empty graph that reads as "no commits".

## Tests

The stale-reply test now covers the new dimension — a reply for the **right root but the wrong view** must also be dropped. That is the entire point of widening the tag, and it would fail if someone later compared only the root.

The `terminal_session` round-trip got a real base (`Some("main")`) rather than `None`, so it proves the field survives rather than that absence survives.

## Four build failures, each a real integration point

| Failure | What it was |
| --- | --- |
| 7 stale `LaneRecord` constructors | the new field, unhandled across the tree |
| `id()` match | a designed tripwire — its doc says a new variant must fail to compile until it declares an id |
| old tuple in a test | the very test covering stale-reply discard |
| keybinding match | third of three exhaustive matches on `Command` |

I found the last two by searching for **every** arm on the enum rather than fixing compiler errors one at a time — four rounds of one-at-a-time was three too many.

## One failure probed, not assumed

`a_restored_lane_pane_spawns_in_its_worktree` polls a real shell and failed once. **3/3 alone on this tree, and green without the diff** — the load-sensitive class. The second half is what makes it evidence: three green runs alone would show only that it *can* pass, and I had modified the struct that test restores.

Version 0.1.919: main is at 0.1.918, and PR #524 holds a now-stale 0.1.917.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.918"
+version = "0.1.919"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.918"
+version = "0.1.919"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -6430,6 +6430,12 @@ impl App {
     fn seed_active_scm(&mut self, active: PathBuf) {
         self.active_scm_root = active.clone();
         self.default_branch_label = None;
+        // A lane diff belongs to the root it was asked for. Carried into
+        // another repo the range names a branch that does not exist there,
+        // and `branch_history_range` answers an unresolvable range with an
+        // EMPTY vec rather than an error - so the panel would show a repo
+        // with commits as having none, with nothing on screen saying why.
+        self.graph_revspec = None;
         let w = self.git_worker_for_root_mut(&active);
         w.bypass_debounce();
         w.request_status_and_changes();
@@ -25918,7 +25924,7 @@ impl App {
         let lane = self.active_scm_root.clone();
         if lane == self.roots.primary() {
             self.status =
-                String::from("The primary folder is not a lane - open a file in the lane first");
+                String::from("The primary folder is not a lane — open a file in the lane first");
             return;
         }
         let Some(record) = self
@@ -25967,6 +25973,10 @@ impl App {
                 String::from("The primary folder is not a lane — open a file in the lane first");
             return;
         }
+        // The diff outlives the lane otherwise: `remove_workspace_folder`
+        // re-seeds the primary, and its COMMITS panel is then stuck on a
+        // range whose right side no longer resolves.
+        self.graph_revspec = None;
         if let Some(why) = crate::git::lane_removal_block(&lane) {
             self.status = format!("Keeping the lane: {why}");
             return;
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -25904,13 +25904,6 @@ impl App {
         (closed, kept)
     }
 
-    /// Remove the active root's worktree lane, refusing when it is dirty
-    /// (#348).
-    ///
-    /// The refusal is the point. `git worktree remove` DISCARDS uncommitted
-    /// work, so this is the one place in the lane flow where being wrong
-    /// destroys something rather than annoying someone — and the check fails
-    /// closed, treating a tree git cannot report on as unsafe.
     /// Point the COMMITS graph at `base..lane` for the active lane (#348).
     ///
     /// Uses `active_scm_root` like `close_worktree_lane` beside it, rather
@@ -25952,6 +25945,13 @@ impl App {
         self.status = format!("COMMITS: {branch} against {base}");
     }
 
+    /// Remove the active root's worktree lane, refusing when it is dirty
+    /// (#348).
+    ///
+    /// The refusal is the point. `git worktree remove` DISCARDS uncommitted
+    /// work, so this is the one place in the lane flow where being wrong
+    /// destroys something rather than annoying someone — and the check fails
+    /// closed, treating a tree git cannot report on as unsafe.
     fn close_worktree_lane(&mut self) {
         // The ACTIVE root, not the primary. `workspace_root()` is
         // `roots.primary()`, and a lane is appended by `add_workspace_folder`
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2362,6 +2362,17 @@ type PendingTestDebug = (
     PathBuf,
 );
 
+/// A COMMITS graph reply: the root and revspec it answers, and the rows.
+/// Tagged rather than bare so a reply for a since-left root or a superseded
+/// view is discarded on drain (#348).
+type GraphReply = (
+    PathBuf,
+    Option<String>,
+    Vec<crate::widgets::commit_graph::GraphRow>,
+);
+type GraphReplyRx = std::sync::mpsc::Receiver<GraphReply>;
+type GraphReplyTx = std::sync::mpsc::Sender<GraphReply>;
+
 pub struct App {
     pub tree: FileTree,
     pub search: SearchPanel,
@@ -2373,8 +2384,16 @@ pub struct App {
     /// Background `git log --topo-order` + lane layout for the COMMITS
     /// section, tagged by the root it describes so a stale reply after a
     /// Make Root is ignored on drain.
-    graph_rx: std::sync::mpsc::Receiver<(PathBuf, Vec<crate::widgets::commit_graph::GraphRow>)>,
-    graph_tx: std::sync::mpsc::Sender<(PathBuf, Vec<crate::widgets::commit_graph::GraphRow>)>,
+    graph_rx: GraphReplyRx,
+    graph_tx: GraphReplyTx,
+    /// What the COMMITS panel is currently showing: `None` for the root's
+    /// own history, `Some("base..lane")` for a worktree lane's diff (#348).
+    ///
+    /// The reply tag carries this too, so a history refresh landing while a
+    /// lane diff is up is discarded rather than replacing it - the panel
+    /// already discarded replies for a since-left ROOT, and this is the same
+    /// rule at the resolution the lane diff needs.
+    graph_revspec: Option<String>,
     /// A graph fetch is running; suppresses duplicate spawns when a burst of
     /// StatusAndChanges responses lands. Cleared when the reply drains.
     graph_fetch_inflight: bool,
@@ -4440,6 +4459,7 @@ impl App {
             source_control,
             commit_graph: crate::widgets::commit_graph::CommitGraphPanel::new(),
             graph_rx,
+            graph_revspec: None,
             graph_tx,
             graph_fetch_inflight: false,
             graph_refetch_queued: false,
@@ -8727,9 +8747,9 @@ impl App {
             }
         }
         // Drain any COMMITS graph replies; ignore one for a since-left root.
-        while let Ok((root, rows)) = self.graph_rx.try_recv() {
+        while let Ok((root, revspec, rows)) = self.graph_rx.try_recv() {
             self.graph_fetch_inflight = false;
-            if root == self.active_scm_root {
+            if root == self.active_scm_root && revspec == self.graph_revspec {
                 self.commit_graph.set_rows(rows);
                 changed = true;
             }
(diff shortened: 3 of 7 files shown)


vitali87:

CodeRabbit's check on this PR reports `pass` but its comment body says "Review limit reached" — the check did not block, but no review happened. This is that review, at head `df850d0a483a0219ff62700fd8b633e02fcc24b9`.

---

## Walkthrough

Adds an "Agent: Diff Lane Against Its Base" palette command that sets `App::graph_revspec` to `base..lane` and refetches the COMMITS graph through the pre-existing `git::branch_history_range`. Persists the lane's base branch on `LaneRecord` so the command survives a restart (`WorktreeLane::base` already existed on `main`; only the serialized record is new). Widens the graph reply tag from `(root, rows)` to `(root, revspec, rows)` so a background history refresh cannot overwrite a lane diff.

| file | changes |
| --- | --- |
| `src/app/mod.rs` | `GraphReply`/`GraphReplyTx`/`GraphReplyRx` aliases; new `graph_revspec` field; producer and drain carry/compare the revspec; new `diff_worktree_lane`; `Cmd::DiffWorktreeLane` dispatch; `base` propagated into `LaneRecord` at lane creation |
| `src/terminal_session.rs` | `LaneRecord.base: Option<String>`, `#[serde(default, skip_serializing_if)]`; round-trip test now pins a real base |
| `src/widgets/command_palette.rs` | new `DiffWorktreeLane` variant + title, empty keybinding hint, id `diff_worktree_lane` |
| `src/app/tests.rs` | reply-tag arity fixes at four `LaneRecord` literals and the graph-drain test; one new injected reply |
| `Cargo.toml`, `Cargo.lock`, `src/release_notes/0.1.919.md` | 0.1.918 → 0.1.919 with notes |

## Findings

- `src/app/mod.rs:6430-6440` **[major/correctness]** `seed_active_scm` switches `active_scm_root` to a different repo and calls `refresh_commit_graph()`, but never clears `graph_revspec`. Once a lane diff is up, the revspec is sticky across every root switch. Reproducer: run the command in lane `croft-fix-login` (`graph_revspec = Some("main..agent/fix-login")`), then click another repository in the SCM overview (`src/app/mod.rs:38362` → `seed_active_scm`). The worker at `src/app/mod.rs:19050` runs `branch_history_range(&other_root, "main..agent/fix-login", 400)`; the ref does not exist in that repo, `run` fails, and `branch_history_range` (`src/git.rs:1251`, empty-on-failure by design) returns `Vec::new()`. The reply tag matches on both root and revspec, so `set_rows(Vec::new())` lands and the COMMITS panel shows an empty graph for a repo with commits, with no status explaining it. Same path via focus-follow (`src/app/mod.rs:6317`). Fix — reset in `seed_active_scm`, where the root identity that made the revspec meaningful is being discarded:
  ```rust
  self.active_scm_root = active.clone();
  self.graph_revspec = None; // the lane diff belonged to the root we just left
  ```
- `src/app/mod.rs:25904-25943` **[major/correctness]** Nothing ever returns `graph_revspec` to `None`, so there is no way back to the root's own history short of restarting croft. The doc claims "returning to history is the ordinary refresh path", but every refresh path re-reads `self.graph_revspec` at `src/app/mod.rs:19046` and reproduces the diff: HEAD-moved (`src/app/mod.rs:6371`), `seed_active_scm` (`6439`), and the SCM-panel refresh (`19026`). Verified by grepping every write to `graph_revspec` at `df850d0a` — the only one is `25943`. Worse after `Cmd::CloseWorktreeLane`: `close_worktree_lane` (`src/app/mod.rs:25947`) removes the worktree and the branch's lane dir, `remove_workspace_folder` re-seeds the primary (`src/app/mod.rs:40964`), and the primary's COMMITS panel is then permanently stuck on `main..agent/fix-login` — a range whose right side may no longer resolve — showing an empty graph. Fix: either make the command a toggle, or clear on lane close. Minimum, in `close_worktree_lane`'s `Ok(())` arm:
  ```rust
  self.graph_revspec = None;
  ```
- `src/app/tests.rs:10540-10551` **[major/tests]** The new injected reply asserts nothing — the test passes identically with the revspec check deleted. `app.graph_revspec` is `None` here and the injected reply carries `Some("a..b")` with `Vec::new()` rows. The only assertion afterwards is the `while !app.commit_graph.is_loaded()` poll, and `set_rows` sets `loaded = true` unconditionally (`src/widgets/commit_graph.rs:209-210`), including for an empty vec. So with the tag check removed the reply would be *accepted* and satisfy the loop sooner, not fail it. This is the repo's own "absence assertion needs a paired presence" trap: it needs a positive value to pin. Fix — put a real revspec on the app, send a mismatched reply carrying distinguishable rows, and assert the panel did not take them:
  ```rust
  app.graph_revspec = Some(String::from("main..lane"));
  app.commit_graph.set_rows(layout_graph(vec![commit("aaa")]));
  app.graph_tx.send((app.active_scm_root.clone(), None, layout_graph(vec![commit("bbb")]))).unwrap();
  app.sync_explorer_panels();
  assert_eq!(app.commit_graph.commit_at(0).map(|c| c.hash.clone()), Some(String::from("aaa")),
      "a history reply must not replace the lane diff");
  ```

## Test coverage

- `diff_worktree_lane` has zero tests — no test names it, and no test sets `graph_revspec` (verified by grep over `src/app/tests.rs` and `tests/` at `df850d0a`). All four of its branches are untested: primary-root refusal, no-lane-open refusal, `base: None` refusal, and the success path setting `graph_revspec` and the status. Write a unit test building an `App` with a `lane_panes` entry whose `path` equals `active_scm_root` and `base: Some("main")`, calling `app.diff_worktree_lane()`, and asserting `app.graph_revspec == Some("main..agent/fix".into())` plus the status text; and a companion with `base: None` asserting `graph_revspec` stays `None`. No process spawn, so no `#[cfg(unix)]` gate needed.
- The two major bugs above are each one test: switch roots after a diff and assert `graph_revspec` is `None`; close the lane after a diff and assert the same.
- `LaneRecord.base` round-trips through serde (`src/terminal_session.rs:199`) but no test covers restore-then-diff — that a lane restored from `~/.croft` session state can still be diffed. The `base: None` literals added at `src/app/tests.rs:45097/45108/45479/45597` all pin the absent case only.

## Nitpicks

- `src/widgets/command_palette.rs:209` — `DiffWorktreeLane` is the only variant in its block with no doc comment; `NewWorktreeLane` (206) and `CloseWorktreeLane` (208) both carry one with the `(#348)` tag. Add `/// Point the COMMITS graph at what the current lane added (#348).`
- `src/widgets/command_palette.rs:975` — id is `"diff_worktree_lane"` while its two siblings are `"agent_new_worktree_lane"` / `"agent_close_worktree_lane"`. `Command::id`'s own doc (`:806-809`) says the string is the `keybindings.json` contract and "must never drift once shipped", so this is the last moment it is free to fix: `"agent_diff_worktree_lane"`.
- `src/app/mod.rs:25921` — the refusal message uses an ASCII hyphen where the identical message in `close_worktree_lane` at `:25964` uses an em dash. Same sentence, two spellings.

## Scores

**CodeRabbit score (diff quality) — 2/5**: one major inside the diff (the vacuous test at `tests.rs:10540`) plus three nits.

**Greptile score (codebase fit) — 2/5**: two majors whose evidence is outside the diff — `seed_active_scm` and `close_worktree_lane` both leave `graph_revspec` set, and `branch_history_range`'s empty-on-failure contract turns that into a silently blank panel.

VERDICT: REQUEST_CHANGES

vitali87:

CodeRabbit is still rate-limited at this head (its check reports `pass` over a "Review limit reached" body), so this is the review of record. Re-verified at `3d0e4cc415a5c0172f6188539a3b3b2a190394f2` after the round-1 fixes.

---

## Verification of the three majors

- **`seed_active_scm` (`src/app/mod.rs:6433-6438`) — fixed, correctly placed.** `self.graph_revspec = None;` sits after the `active_scm_root` assignment and before `refresh_commit_graph()` at `:6445`, so the refetch this function performs already runs unranged. Both entry paths traced originally — the SCM-overview repository click (`:38368`) and the focus-follow switch (`:6317`) — go through here, as does the `remove_workspace_folder` fallback (`:40970`). The comment names the empty-vec mechanism accurately.

- **`close_worktree_lane` (`src/app/mod.rs:25976-25979`) — fixed; the placement is defensible, with one caveat below.**

- **`a_history_reply_does_not_replace_the_lane_diff` (`src/app/tests.rs:10547-10636`) — fixed, and it is now a real test.** The vacuous injection is gone from the re-fire test (`:10728-10735` now sends only the since-left-root reply, which the original test's root tag was always about). The new test pins a positive value: the panel holds `aaa…`, the `None`-tagged reply carrying `bbb…` must leave `aaa…` on screen, and the third leg with `Some("main..lane")` carrying `ccc…` must land. Deleting the `revspec == self.graph_revspec` clause at `:8756` now fails leg two — `bbb…` would replace `aaa…`. The third leg rules out a dead channel. The `commit_at(area.y + 2)` after a real draw is right: `commit_at` is a hit-test over `last_area`, which is zero-sized until rendered.

## On clearing before the refusal

The author moved the clear ahead of the dirty-tree refusal rather than into the `Ok(())` arm suggested in round 1. That is the better placement, for a reason the original suggestion missed: `remove_worktree_lane` can fail *after* `lane_removal_block` passed (a locked worktree, called out at `:25984-25989`), and the `Err` arm returns with the revspec intact. Clearing early covers that arm too. Keep it where it is.

One consequence worth naming:

- `src/app/mod.rs:25979` **[nit/correctness]** On the refusal path, `graph_revspec` is set to `None` and the function returns at `:25982` without calling `refresh_commit_graph()` — the refresh sites at this head are `:6371`, `:6445`, `:8766`, `:19032`, `:25950`, none reachable from the refusal. So between the refused close and the next refresh, the panel *displays* diff rows while the state says it is showing history. Not a correctness bug: the rows on screen are real commits from the lane, the lane still exists, `active_scm_root` is unchanged, and the next HEAD move or root switch resolves it correctly. But state and pixels disagree for a window, which is the class of thing the revspec tag exists to prevent. Either refresh on that path, or say so in the comment.

## New findings

None. Re-read the two new coverage tests: `diffing_a_lane_refuses_when_there_is_no_base_to_diff_against` (`:10450`) exercises all three refusals and pins the success path at `Some("main..agent/fix")`, which is the positive control making the three `assert_eq!(app.graph_revspec, None)` non-vacuous. `switching_roots_drops_the_lane_diff` (`:10525`) sets a revspec, calls `seed_active_scm`, asserts `None`; `git_worker_for_root_mut` (`:6963-6968`) falls back to `self.git` for an unregistered root, so the unregistered `other` dir is safe. Neither spawns a process, so no `#[cfg(unix)]` gate is needed.

Round-1 nits all applied: em dash at `:25973` matches its sibling, doc comment on the palette variant, id renamed to `agent_diff_worktree_lane` — the right call given `Command::id`'s own doc says the string is a `keybindings.json` contract that must never drift once shipped, and this was the last moment it was free.

## Release gate

`Cargo.toml` is 0.1.919, `origin/main` is 0.1.918, and `src/release_notes/0.1.919.md` exists with a valid `feature:` line.

## Scores

**CodeRabbit score (diff quality) — 4/5**: no majors remain inside the diff; one nit on the stale-rows window at `mod.rs:25979`.

**Greptile score (codebase fit) — 5/5**: both cross-file majors are closed, and the clear covers the `Err` arm the original suggestion would have missed.

VERDICT: APPROVE

PR #527, opened 2026-09-07
feat(agent): diff a lane file against the snapshot you reviewed

vitali87:
Criterion 2 of #345: "Diff for a row compares against the last-reviewed snapshot, verified after two rounds of edits."

Marking a file reviewed recorded only a content hash. A hash answers "changed?" but not "changed how?", so a lane row could show that an agent had touched a file again without being able to show what it did since you looked.

## What changed

`LaneFile` gains `reviewed_millis`: the local-history snapshot holding the content that became the baseline. `Agents: Diff This File Since Review` opens the working file against that snapshot rather than against HEAD.

The anchor is recorded from the same bytes the baseline hash is taken from. That matters more than it first looks: the history store is written only by croft's **own** saves — the agent-write path that creates lane rows records nothing — so adopting whatever snapshot happened to be newest would anchor the review to an unrelated earlier save and then label that diff "(reviewed)". `record_in` also dedupes identical content and replaces inside the merge window, so the store is read back rather than assuming the timestamp names a file.

## Refusals

A row with no usable baseline says which reason, rather than falling back to a HEAD diff:

- never reviewed
- reviewed before a snapshot existed
- the snapshot has since been pruned — the store keeps 50 per file against a 1s autosave, so this is a live case, not only a legacy one

A diff against a baseline other than the one claimed is worse than no diff, because nothing on screen tells you which one you got.

Files over `MAX_HASH_BYTES` keep `read_baseline`'s stamp path rather than being slurped and copied into the history store on every review. They review with no anchor, and the diff says so.

## Review findings fixed before this was opened

- **Stale restore (data loss).** Opening this diff left "Restore Snapshot" armed at whatever file the TIMELINE last showed, so restoring would have overwritten a file the user was not looking at. Disarmed on entry — refusals included, since a refused diff leaves the previous one on screen — and armed to this snapshot on success. The repo already had a test for this shape on the timeline path; this adds the matching one here.
- **Split read.** The whole-lane path took the hash and the anchor from two separate reads, so a write landing between them left the baseline hash describing content the anchor does not hold. One closure now returns both from one read.

## Verification

`cargo check` clean, `clippy -D warnings` clean, 15 tests pass.

Both new behaviours were mutation-tested rather than trusted: removing the restore disarm fails `the_review_diff_disarms_another_files_restore`, and reverting the anchor to "newest snapshot" fails `reviewing_anchors_to_the_content_it_baselined_not_an_older_save`. Each mutation was applied to an asserted anchor and the file restored afterwards, so a mutation that silently failed to apply cannot read as a survived one.

Part of #345. Criteria 1 (attribution while working) and 3 (badge counts and tree decorations) are already on main; this is criterion 2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.918"
+version = "0.1.921"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.918"
+version = "0.1.921"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/agent_lane.rs
+++ b/src/agent_lane.rs
@@ -30,6 +30,13 @@ pub struct LaneFile {
     /// Content hash when the user last marked it reviewed; `None` until the
     /// first review, which is what makes a newly-touched file unreviewed.
     pub reviewed_hash: Option<u64>,
+    /// Milliseconds of the local-history snapshot holding the content that
+    /// became the baseline, so the row can be DIFFED against what the user
+    /// approved rather than only compared to it (#345). `None` when the
+    /// snapshot could not be taken — the store is best-effort, and a row
+    /// with no anchor refuses the diff rather than showing one against a
+    /// baseline it cannot name.
+    pub reviewed_millis: Option<u64>,
     /// Content hash at the most recent attributed write.
     pub current_hash: u64,
     /// How many attributed writes have landed since the last review.
@@ -200,6 +207,7 @@ impl AgentLedger {
                         LaneFile {
                             path: path.to_path_buf(),
                             reviewed_hash: None,
+                            reviewed_millis: None,
                             current_hash: hash,
                             writes_since_review: 1,
                             last_write_seq: seq,
@@ -322,7 +330,18 @@ impl AgentLedger {
     /// hash now, which may differ from the last attributed write if the
     /// user edited it themselves in between — reviewing means "I have seen
     /// what is on disk", not "I have seen what the agent last wrote".
-    pub fn mark_reviewed(&mut self, agent: &str, path: &Path, current: u64) -> bool {
+    ///
+    /// `snapshot_millis` anchors WHICH local-history snapshot holds that
+    /// content, so the row can later be diffed against it. `None` is a
+    /// legitimate outcome, not an error: the store is best-effort, and a row
+    /// without an anchor is still reviewed — it just cannot show a diff.
+    pub fn mark_reviewed(
+        &mut self,
+        agent: &str,
+        path: &Path,
+        current: u64,
+        snapshot_millis: Option<u64>,
+    ) -> bool {
         let Some(lane) = self.lanes.get_mut(agent) else {
             return false;
         };
@@ -330,6 +349,7 @@ impl AgentLedger {
             return false;
         };
         entry.reviewed_hash = Some(current);
+        entry.reviewed_millis = snapshot_millis;
         entry.current_hash = current;
         entry.writes_since_review = 0;
         self.settle_if_empty();
@@ -347,6 +367,7 @@ impl AgentLedger {
         &mut self,
         agent: &str,
         read: impl Fn(&Path) -> Baseline,
+        snapshot: impl Fn(&Path) -> Option<u64>,
     ) -> (usize, usize) {
         let Some(lane) = self.lanes.get_mut(agent) else {
             return (0, 0);
@@ -360,6 +381,10 @@ impl AgentLedger {
             match read(&entry.path) {
                 Baseline::Hash(current) => {
                     entry.reviewed_hash = Some(current);
+                    // Per row, not once for the lane: each file has its own
+                    // snapshot, and one that failed to record must not
+                    // inherit a sibling's anchor.
+                    entry.reviewed_millis = snapshot(&entry.path);
                     entry.current_hash = current;
                     entry.writes_since_review = 0;
                     cleared += 1;
@@ -467,7 +492,7 @@ mod tests {
 
         // Reviewing one clears just that row, and it leaves the lane's
         // ordering with the unreviewed file first.
-        assert!(led.mark_reviewed("claude", &p("/w/src/a.rs"), 1));
+        assert!(led.mark_reviewed("claude", &p("/w/src/a.rs"), 1, None));
         assert_eq!(led.unreviewed_count("claude"), 1);
         assert!(!led.is_unreviewed(&p("/w/src/a.rs")));
         let rows = led.lane("claude");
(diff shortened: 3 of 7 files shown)


vitali87:
## 🤖 Fallback review (round 2) — greptile-local, Claude Opus 5

CodeRabbit is rate-limited on this PR (its check reports `pass` over a "Review limit reached" body), so this is the review of record. Round 1 ran at `2c058a5` and returned 2/5 with a P0 and a P1; this is the re-verification at `8d5bd02`.

---

### Greptile-local Summary

Adds `LaneFile.reviewed_millis` plus `App::diff_agent_lane_row` and a palette command to diff an agent-touched file against the snapshot that was its review baseline. Both findings from the previous round are fixed correctly, and the two follow-up concerns raised with the fix (one read per row on the whole-lane path, and the restored large-file guard) check out. Ready to merge.

### Confidence Score: 5/5

Both prior findings fixed at the mechanism level with non-vacuous tests; no new defects found.

### Findings

None.

### Verified fixes from the previous round

1. **[was P0] Stale "Restore Snapshot"** — `src/app/mod.rs:31437`, `31489`
   `self.history_restore = None;` is now the first statement of `diff_agent_lane_row`, ahead of all five `return` paths (row-not-in-lane, no anchor, snapshot pruned, unreadable snapshot, and the `open_head_diff_with_text` error). On success it arms `Some((path, millis, content))` at 31489, after the diff opens, so restore puts back the version that was reviewed. This is strictly safer than `open_timeline_diff`, which disarms but arms only for the local-history case.
   New test `the_review_diff_disarms_another_files_restore` (`tests.rs:12601`) exercises the refusal path — the dangerous one, since a refusal leaves the previous diff on screen — and asserts both `history_restore.is_none()` and that `a.txt` is byte-unchanged after `restore_history_snapshot()`. Removing the disarm fails it.

2. **[was P1] Anchor did not name the reviewed content** — `src/app/mod.rs:31409`
   `newest_snapshot_millis` is gone (confirmed: zero references at this head, so no dead code). `record_review_baseline` records a snapshot of the bytes being hashed and then finds the entry whose content equals those bytes, rather than assuming `millis` names it. Reading the store back is required, not defensive: `history.rs:141-143` returns early on dedupe without writing at `millis`, and `146-160` writes at `millis` while deleting the previous newest inside the merge window — so the passed-in millis genuinely need not name a file. `mark_agent_file_reviewed_inner` reads once and derives both hash and anchor from that read.
   New test `reviewing_anchors_to_the_content_it_baselined_not_an_older_save` (`tests.rs:12641`) plants a decoy snapshot at `millis=1000` holding an unrelated earlier save, then asserts the anchor's *content* equals the agent's bytes, with `assert_ne!(millis, 1000)` as the paired positive control. Under the old `entries_in(...).first()` both assertions fail.

### Checked, not a defect

- **One read per row on the whole-lane path:** confirmed. `mark_lane_reviewed` now takes a single `impl Fn(&Path) -> (Baseline, Option<u64>)` (`agent_lane.rs:370`), and the closure at `mod.rs:31552` performs exactly one `std::fs::read` feeding both `content_hash` and `record_review_baseline`. All three `Baseline` outcomes are still produced and routed: `Hash` from `Ok`, `Gone` from `NotFound`, `Unreadable` from any other error. The ledger arms are unchanged, and the per-row comment correctly notes a failed snapshot must not inherit a sibling's anchor.
- **Large-file guard on both paths:** correct. Both `mark_agent_file_reviewed_inner` (`31513`) and the lane closure (`31555`) check `metadata().len() > MAX_HASH_BYTES` before reading, delegate to `read_baseline` (which stamps from metadata), and pass `None` for the anchor — so a 4 MiB+ file is never copied into the history store, and reviews with no anchor. All three `Baseline` outcomes are handled on the single-file guarded branch too.
- **Hash domains stay comparable:** the normal path now calls `content_hash` directly instead of going through `read_baseline`, but for sub-threshold files `read_baseline` computes the same `content_hash` over the same bytes, and over-threshold files go through `read_baseline` on both sides. So `reviewed_hash` remains comparable with the `current_hash` that `record_write` stored, and unreviewed-detection is unaffected.
- **Call-site sweep:** all 6 `mark_lane_reviewed` sites converted to the tuple form (5 tests + 1 production); no stragglers.
- **Test coverage gap closed:** `diff_agent_lane_row` had zero test references in the previous round and now has two (`tests.rs:12623`, `12681`), covering a refusal and the success path.
- **Refusal paths:** still all three reachable and distinguishable; `reviewed_hash`/`reviewed_millis` are written together in `mark_reviewed`, so "reviewed before a snapshot existed" stays distinct from "never reviewed".
- **Encoding, `entries_in` ordering, doc ownership, version/release notes (0.1.921 + `src/release_notes/0.1.921.md`, `feature:` prefixed), and the 6 palette/dispatch wiring sites:** all re-confirmed unchanged and correct at this head.

- `view_staged_diff_source_control` / `view_default_branch_diff_source_control` / `view_previous_commit_diff_source_control` (`mod.rs:24394, 24434, 24467`) do not clear `history_restore`, despite the field doc saying "Cleared when a non-snapshot diff opens". They open git diffs rather than snapshot diffs, so the stale-restore risk is lower than the case fixed here, but the invariant is still enforced in only two of five diff-opening paths.
- If `stamp_hash` returns `None` (no mtime available), `read_baseline` falls through and slurps a large file to hash it. Pre-existing; it only hashes and never snapshots, so the "never copy 4 MiB+ into the store" goal still holds.

### Verification run

- tests: not run — reviewed by reading only, so the host stayed free for the author's gate run. Fix verification is anchored to lines read at `8d5bd02`.
- The author's reported results (check clean, clippy clean under `-D warnings`, 15 tests passing, both fixes mutation-verified) are consistent with what was read but are the author's evidence, not the reviewer's.

vitali87:
Merged `origin/main` in after #526 landed (GitHub flipped this to `DIRTY` immediately after reporting 8/8 green — the version-collision shape).

Resolution:
- **Cargo.toml**: version line only. 0.1.921 stands, one above main's new 0.1.919.
- **Cargo.lock**: regenerated rather than hand-merged, then verified its `croft-software` version matches the manifest. Taking main's lock wholesale would have shipped a lock reading 0.1.919 against a manifest reading 0.1.921.
- **Source files auto-merged.** Verified both features survived rather than assuming: 2 `graph_revspec` clears from #526 and 2 `history_restore` disarms from this PR, both at their expected counts. The `<<<<<<<` hits remaining in the tree are string literals inside croft's own conflict-parser tests, none at line start.
- Merge ancestry asserted: two parents, `origin/main` is a genuine ancestor.

The merged tree is code neither branch tested, so the gates were re-run on it rather than carried over: `cargo check` clean, `clippy -D warnings` clean, 18 tests pass with all five named tests **from both features** confirmed present.

The 5/5 review above was performed at `8d5bd02`, before this merge. The merge adds no source changes of its own — only the version resolution and main's `src/release_notes/0.1.919.md` arriving with #526.

vitali87:
Correction to my previous comment: "the merge adds no source changes of its own" was imprecise. `git diff 8d5bd02..HEAD` does show 315 changed source lines — those are **#526's changes arriving from main**, not drift in this PR's work.

Verified precisely, since the 5/5 review was performed at `8d5bd02` and it matters whether that review still covers what is being merged:

- `src/agent_lane.rs` (this PR's core) is byte-identical to its reviewed state — empty diff.
- Of the changes to `src/app/mod.rs`, 14 lines carry #526 identifiers (`graph_revspec`, `diff_worktree_lane`, `GraphReply`).
- **Zero** lines touch this PR's identifiers (`record_review_baseline`, `reviewed_millis`, `diff_agent_lane_row`, `history_restore`).

So the review at `8d5bd02` still describes the code under review; what the merge brought in is the already-reviewed-and-merged #526. The interaction between the two is what the re-run gates on the merged tree cover — 18 tests, all five named tests from both features.

Issue #376, opened 2026-08-29
Prebuilt release binaries and cargo binstall support


vitali87:
## Motivation

#261 covers prebuilt binaries for the *remote* install path. Local users get the same ten-minute compile. Publishing a release matrix and declaring `binstall` metadata makes `cargo binstall croft-software` a five-second install and feeds the Homebrew formula.

## Proposed design

1. A `release.yml` job on tags builds `aarch64-apple-darwin`, `x86_64-apple-darwin`, `x86_64-unknown-linux-musl`, `aarch64-unknown-linux-musl` (the musl targets are what #261 needs too; share the cross setup), strips, packages `tar.gz` with a `SHA256SUMS`, and attaches them to the GitHub Release the existing tag flow creates (#86).
2. `[package.metadata.binstall]` in `Cargo.toml` with the URL template and archive layout, so `cargo binstall croft-software` resolves without any config.
3. Signed checksums (minisign or sigstore cosign, decided in the PR); the remote-install fast path verifies them.
4. `croft --version` reports `release` provenance for these builds (#237).

## Acceptance criteria

- [ ] `cargo binstall croft-software` installs the tagged version on macOS and Linux without compiling.
- [ ] Every release has four archives and a checksum file; a corrupted archive fails verification.
- [ ] The remote install fast path (#261) consumes the same artifacts.

## Related

- Tracking epic: #385

vitali87:
claimed by M1 croft-2

vitali87:
Starting on this. Two things in the proposed design need a decision from you first, because building either as written would contradict something already shipped.

## 1. Item 4 contradicts #282, which is guarded by a test

The design says `croft --version` should report `release` provenance. `src/cli.rs` has a test that forbids exactly that:

```rust
/// #282: `--version` is a plain `x.y.z`. The regression this guards is a
/// well-meaning one — re-adding provenance "so bug reports carry it" is
/// exactly how the hash got into the common path the first time.
fn version_is_a_bare_semver_and_provenance_lives_behind_build_info()
```

It asserts `--version` renders `croft x.y.z` and contains no `(`. Provenance did not disappear — it moved to `--build-info` (`VERBOSE_VERSION`, carrying `CROFT_GIT_HASH` and `CROFT_BUILD_TIME`).

[...]

Issue #375, opened 2026-08-29
Homebrew tap and bottles: brew install croft


vitali87:
## Motivation

`cargo install` compiles for ten minutes on a laptop and needs a Rust toolchain first. Most people who hear about a terminal tool try `brew install <name>` and move on if it does not exist. A tap with bottles is the single largest funnel fix available.

## Proposed design

1. A `vitali87/homebrew-croft` tap with a formula that installs the prebuilt release artifact (from the GitHub Releases matrix issue) for macOS arm64/x86_64 and Linux x86_64/aarch64, falling back to `cargo install --locked` when no bottle matches.
2. The release workflow bumps the formula automatically on tag (a job that opens a PR against the tap with the new version and SHA256s), so the tap never lags a release.
3. `croft --version` and the F9 self-update path recognise a brew-installed binary and say "upgrade with `brew upgrade croft`" instead of reinstalling over Homebrew's cellar (the provenance work in #237/#242 provides the hooks).
4. Document in README under Install, above `cargo install`.

## Acceptance criteria

- [ ] `brew install vitali87/croft/croft` works on a clean macOS with no Rust toolchain.
- [ ] A new tag updates the formula within the release run.
- [ ] F9 on a brew install shows the upgrade hint and does not overwrite the cellar.

## Related

- Tracking epic: #385
- Depends on: #376

PR #528, opened 2026-09-07
feat(release): prebuilt binaries and cargo binstall support

vitali87:
Part of #376. `cargo install croft-software` compiles for ten minutes and needs a Rust toolchain first; this publishes four archives per tag so `cargo binstall croft-software` is a download, and so the Homebrew formula (#375) has something to point at.

## What ships

* **`.github/workflows/release.yml`** — tag-triggered (`v*`) matrix over `aarch64-apple-darwin`, `x86_64-apple-darwin`, `x86_64-unknown-linux-musl`, `aarch64-unknown-linux-musl`. Packages one `tar.gz` per target, computes per-target sums on the runner that built the artifact, then a publish job collects them, builds `SHA256SUMS`, verifies, and creates the release.
* **`[package.metadata.binstall]`** in Cargo.toml.
* **`scripts/check_binstall_contract.py`** — a CI gate comparing the two, because nothing else does.
* **9 tests** for that gate: one control plus eight proven failure modes.

## Two decisions carried over from ci.yml

The musl recipe is deliberately the one `ship-path build` already proved, including its two non-obvious fixes — `rustup target add` runs from the checkout so `rust-toolchain.toml` resolves the channel, and `cargo-zigbuild` needs `pip install --break-system-packages ziglang` on current runner images. The Apple targets build on a macOS runner, since zigbuild does not cross to them without an SDK.

The profile is `release`, **not** the `remote-fast` profile ci.yml builds. remote-fast drops LTO for build speed because a user is waiting on it during `croft <host>`; a published artifact is built once and run many times.

## Why the gate exists

`pkg-url` and `bin-dir` are resolved by binstall at **install** time. Drift between them and the workflow is invisible to every other CI job and lands on users as a 404 or a "corrupt download" — the worst place to find it. Nothing else in the repo compares the two files.

## Review: 3/5, two P1s, both fixed

All three findings were the same shape, and it is the one CLAUDE.md documents: **a check that cannot fail is indistinguishable from a check that passed.**

1. **The transit-corruption check could not fail.** The publish job ran `sha256sum croft-*.tar.gz > SHA256SUMS` and then `sha256sum -c` on what it had just generated — the downloaded bytes checked against themselves. It passes on a corrupted archive. The comment above it claimed the sums were verified "on a different machine from the one that produced them"; they were not, and the per-target sums that *are* independent evidence were uploaded and never read. Now built from those.
2. **The gate missed the drift its own docstring promised.** It claimed to check the directory *inside* the tarball and only ever compared archive *names*. Flattening the archive while keeping its name passed the gate and broke every install — the reviewer confirmed this against real `cargo-binstall` 1.23.0, which reported `bin file .../croft-aarch64-apple-darwin/croft not found`. Now parses the `tar -czf ... -C <dir> <member>` line.
3. **The extension guard disarmed itself.** `if "tar -czf" in workflow and suffix != ".tar.gz"` only fires while `tar -czf` is present, so switching to `zip` removed the substring that armed it. Now unconditional.

Two tests cover the tar-layout drifts, both verified **red** against the previous checker and green against this one.

Worth stating plainly: I wrote a gate to catch install-time drift and put three unfalsifiable assertions inside it. My own teeth-proof caught a fourth instance before review — I proved the *tests* could fail without proving each individual *guard* could, and that gap is exactly where these three lived.

## Verified

* `cargo binstall croft-software` resolves, extracts and installs a runnable `croft` — reviewer tested end to end with binstall 1.23.0 against the exact tar layout this produces.
* `python3 -m unittest discover -s scripts/tests` → 143 passed.
* `check_binstall_contract.py`, `check_doc_ownership.py`, `check_arch_content.py` all exit 0.

## Not in this PR, raised on the issue

* **Item 4 (`--version` provenance)** contradicts #282, which has a test forbidding it whose comment names the regression as "well-meaning". Provenance lives behind `--build-info`; adding a release marker there is the version that does not reopen #282. Needs a decision.
* **Item 3 (signed checksums)** needs a minisign-vs-cosign choice. `SHA256SUMS` ships now; signing can follow without blocking the install path.
* **There is no existing tag/release flow** on `main` — `ci.yml` triggers only on push-to-main and PRs. So this workflow *creates* the release rather than attaching to one. If releases are cut by hand today, say so and I will match that process instead.

## Summary by CodeRabbit

- **New Features**
  - Install prebuilt `croft` binaries with `cargo binstall croft-software` on supported macOS and Linux platforms.
  - Tagged releases now provide platform-specific archives and SHA256 checksums for download verification.
  - Manual release builds are available without publishing a release.

- **Documentation**
  - Updated installation guidance to prioritize prebuilt binaries while retaining source-build instructions.
  - Added release notes covering faster installation and artifact verification.

- **Bug Fixes**
  - Added automated checks to detect inconsistencies between published binaries and installation metadata.


vitali87:
All four addressed. Each was right, and the fourth is the third instance of one defect class in this PR — worth stating plainly rather than fixing quietly.

**Workflow permissions.** Taken. `contents: write` at the top level gave the four *build* jobs the ability to rewrite the repo, when only `publish` needs it and only to create a release. Now `contents: read` at workflow level, `contents: write` scoped to the publish job, and `persist-credentials: false` on both checkouts so the token is not left in `.git/config` for any later step to reach.

**The README checksum command was wrong.** Taken, and it would have failed for every user who followed it. `SHA256SUMS` covers all four archives, so `sha256sum -c SHA256SUMS` checks three files that are not on disk and exits non-zero. Now:

```bash
grep "croft-aarch64-apple-darwin.tar.gz" SHA256SUMS | sha256sum -c -
```

with the `shasum -a 256` variant alongside.

**`cargo binstall` prerequisite.** Taken — the README told people to run a command that does not exist on a stock toolchain. Added the install line and a link.

**The checker discarded the tar output path.** Taken, and this is the finding worth dwelling on. My regex captured the output path into a group and then never looked at it:

```python
tar = re.search(r'tar -czf\s+"?\$?\{?[^"\s]*"?\s+-C\s+(\S+)\s+"?([^"\s]+)"?', workflow)
member = tar.group(2)   # group(1) was the -C directory; the output was never captured at all
```

So `tar -czf "dist/$name.zip"` writes a file binstall never requests, while every name-based check still agrees — the archive *name* is built from `$name` either way, only the extension moved. Now the output path is captured and compared against what `pkg-url` expects, with the regression test you asked for (`test_a_mismatched_output_extension_is_caught`), verified red against the previous checker.

---

**The pattern across this PR's reviews.** Three reviewers, three different unfalsifiable checks, all in code written specifically to make install-time drift falsifiable:

* a checksum step that verified downloaded bytes against themselves
* a contract gate that promised in its docstring to check the tarball's internal layout and only compared file names
* this one, capturing evidence and discarding it

My own teeth-proof caught a fourth before any review. The gap in my method was proving the *tests* could fail without proving each individual *guard* could — a test suite that goes red when you disable the whole gate says nothing about whether each assertion inside it can fail on its own.

10 contract tests now, each pinned to a specific drift and each verified red against the code that preceded it.

PR #529, opened 2026-09-07
feat(log): Ctrl+click a path:line printed in a rendered log

### LINKED ISSUE #257: Render ANSI colors when opening log files in the editor

vitali87:
A `path:line` printed in a rendered ANSI log is now Ctrl+clickable, the same as one printed in a terminal pane.

This is the last unmet acceptance criterion of #257. The rest of that issue shipped across #293, #393, #402, #490 and #188 — the issue body still reads as a greenfield proposal, so I walked the criteria individually and commented the delta there.

## Why it did not work

The plumbing was all present and simply never connected. `log_view::visible_text`'s own doc comment names "`path:line` scanning" among its intended consumers:

> The escape-free text of `idx`, for find, selection, copy, and `path:line` scanning — every consumer that must see what the user sees rather than the raw bytes.

Nothing ever called it for that. A modifier click in a rendered log started a selection instead, because `begin_log_selection` is dispatched ahead of the editor's Ctrl handling and returns early — so the gesture never reached the code that would have followed the reference.

## What changed

`log_file_click` scans the **escape-free** text, not the raw bytes. That distinction is the substance of the change rather than a detail: once SGR sequences are stripped, a reference's column on screen is not its column in the file, and a path can sit inside an escape entirely. `log_cell_at` already maps a screen cell to a column in that text, and `cell_map::CellMap` was already doing the same job for copy and selection.

Dispatched ahead of `begin_log_selection` and guarded on CONTROL, so a plain click still starts a selection drag.

Relative paths resolve against the workspace root. A log has no shell cwd to consult, unlike a terminal pane — `terminal_file_click` asks the pane's live shell, which has no analogue here. The click only lands when the resolved path is a real file, which also filters lookalikes like `host.com:443`.

## Verification

`cargo check` clean, `clippy -D warnings` clean, tests pass.

The test asserts Ctrl+click follows the reference and lands on the right line, that `host.com:443` does NOT navigate, and that a plain click at the same cell still begins a selection drag.

Both halves are mutation-verified: removing the dispatch fails it, and dropping the CONTROL guard fails it too. The second matters because a feature that works but swallows ordinary clicks would otherwise pass.

### The fixture had to be made to discriminate

The first version of this test *named* escape-free mapping and could not fail on it. The escapes shifted the offset by 9 characters while the tempdir path is ~50, so the visible column still landed inside the raw path token and a raw-bytes scan returned the same `FileRef`. Both mutations still killed it — it detected a missing dispatch fine — so mutation testing alone gave a clean result to a test that did not check its own headline claim.

Two further fixtures looked correct and were not: a longer escape run keeps the column inside the token (the run pushes the token right by the same amount), and a short `/tmp/w/` model path worked in arithmetic but failed at column 136 on real tempdirs.

What works is putting the escape run BETWEEN two references and scaling it past the path length, so the second reference's visible column falls inside an escape in the raw bytes — where a raw-bytes scan finds nothing at all. The test now asserts its own fixture discriminates, with a message blaming the fixture rather than the code, so a later edit that quietly destroys that property fails loudly instead of passing.

Two smaller corrections the failing runs surfaced: `has_selection()` is false after a plain press by design (`begin_log_selection` sets anchor == head, and it requires them to differ), so the control asserts `dragging`; and `host.com:443` matches the `path:line` pattern with only `is_file()` rejecting it, which nothing asserted — now covered.

Closes #257.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.921"
+version = "0.1.924"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.921"
+version = "0.1.924"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -29286,6 +29286,51 @@ impl App {
         true
     }
 
+    /// Ctrl+click a `path:line` printed in a RENDERED log (#257).
+    ///
+    /// Scanned against the escape-free text, not the raw bytes: the
+    /// reference's column in the file is not its column on screen once SGR
+    /// sequences are stripped, and a path can sit inside an escape entirely.
+    /// `log_cell_at` already maps a screen cell to that text's column.
+    ///
+    /// Dispatched ahead of `begin_log_selection`, which runs before the
+    /// editor's own Ctrl handling and would otherwise swallow the gesture
+    /// into a selection drag. Relative paths resolve against the workspace
+    /// root - a log has no shell cwd to consult, unlike a terminal pane.
+    fn log_file_click(&mut self, col: u16, row: u16) -> bool {
+        let scroll = self.editor.scroll;
+        let Some(log) = self.editor.log.as_ref() else {
+            return false;
+        };
+        if !rect_contains(log.last_body, col, row) {
+            return false;
+        }
+        let (line, column) = log_cell_at(log, scroll, col, row);
+        let Some(text) = log.visible_text(line).map(str::to_string) else {
+            return false;
+        };
+        let Some(fr) = crate::file_ref::file_ref_at(&text, column) else {
+            return false;
+        };
+        let raw = std::path::PathBuf::from(&fr.path);
+        let abs = if let Some(rest) = fr.path.strip_prefix("~/") {
+            match std::env::var_os("HOME") {
+                Some(h) => std::path::PathBuf::from(h).join(rest),
+                None => return false,
+            }
+        } else if raw.is_absolute() {
+            raw
+        } else {
+            self.tree.root.join(raw)
+        };
+        // Only when it resolves to a real file, which is also what filters
+        // lookalikes like `host.com:443`.
+        if !abs.is_file() {
+            return false;
+        }
+        self.open_resolved_file_ref(&abs, fr.line, fr.column)
+    }
+
     /// Extend a live log drag and finish it on release.
     fn update_log_selection(&mut self, m: MouseEvent) -> bool {
         let scroll = self.editor.scroll;
@@ -37935,6 +37980,13 @@ impl App {
                 // the rendered view starts a drag-selection over what the
                 // user can SEE, not the source buffer beneath it.
                 if self.begin_preview_selection(m.column, m.row) {
+                    return;
+                }
+                // Before the selection: a Ctrl+click on a printed
+                // `path:line` follows it rather than starting a drag (#257).
+                if m.modifiers.contains(KeyModifiers::CONTROL)
+                    && self.log_file_click(m.column, m.row)
+                {
                     return;
                 }
                 if self.begin_log_selection(m.column, m.row) {
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -2885,26 +2885,40 @@ fn the_swallow_guard_reads_the_clicked_terminal_not_the_active_one() {
 /// A `path:line` printed in a RENDERED ANSI log is Ctrl+clickable, the
 /// same as one printed in a terminal pane (#257).
 ///
-/// The rendered view strips the escapes, so the reference is scanned
-/// against the text the user can SEE - the raw bytes would put the column
-/// in the wrong place, or hide the path inside an escape entirely.
+/// The reference is scanned against the text the user SEES. The fixture is
+/// built so that matters: the second path sits after a 36-character SGR run
+/// with no printed width, so its visible column falls INSIDE an escape
+/// sequence in the raw bytes. A raw-bytes scan finds no reference at that
+/// column at all, which is what makes this test turn on the mapping rather
+/// than merely coexist with it.
 ///
 /// Ctrl is what separates it from a selection drag: `begin_log_selection`
-/// runs before the editor's Ctrl handling, so without this the modifier
-/// click just starts a selection.
+/// runs before the editor's Ctrl handling, so without the guard the modifier
+/// click would just start a selection.
 #[test]
 fn a_file_reference_in_a_rendered_log_is_ctrl_clickable() {
     let tmp = tempfile::tempdir().unwrap();
-    // The file the log points AT.
-    let target = tmp.path().join("src.rs");
-    std::fs::write(&target, "one\ntwo\nthree\n").unwrap();
+    let first = tmp.path().join("first.rs");
+    let second = tmp.path().join("second.rs");
+    std::fs::write(&first, "a\nb\nc\n").unwrap();
+    std::fs::write(&second, "one\ntwo\nthree\nfour\nfive\nsix\nseven\n").unwrap();
 
-    // A log line with SGR escapes around the reference, so the escape-free
-    // column mapping is exercised rather than bypassed.
+    // An SGR run with zero printed width, placed BETWEEN the two references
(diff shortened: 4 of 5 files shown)


vitali87:
## 🤖 Fallback review (round 2) — greptile-local, Claude Opus 5

CodeRabbit is rate-limited repo-wide right now (its check reports `pass` over a "Review limit reached" body), so this is the review of record. Round 1 ran at `9e97d42` and returned 4/5 with two P2s; this is the re-verification at `1ba7062`.

---

### Greptile-local Summary

Adds `App::log_file_click`, dispatched from the `Down(Left)` arm ahead of `begin_log_selection` and gated on CONTROL, so a `path:line` printed in a rendered ANSI log opens like one in a terminal pane. Both review findings are fixed; the fix is test-only, so the previously verified implementation is unchanged. The reworked fixture genuinely discriminates the escape-free mapping from a raw-bytes scan, and its self-check is non-vacuous. Ready to merge.

### Confidence Score: 5/5

No P0/P1/P2 findings; the changed path is covered by a test that fails if either the dispatch, the CONTROL guard, or the escape-free mapping is removed.

### Findings

None.

### Re-verification of the two fixed findings

**Finding 1 — the author's reasoning is correct, and my round-1 suggestion was wrong.** Lengthening the SGR run in place does not discriminate: the run shifts the raw token right by the same amount it shifts the visible column, so the column stays inside the token. Confirmed with my own earlier example (36-char run, 50-char path): column 49, token spans [47, 95) — still inside, exactly as computed. Placing the run *between* two references is the right construction, because only then does the second reference's raw offset advance while its visible offset does not.

The geometry holds, and not by luck. Simulating `PATH_LINE_RE`'s char-index spans against the committed fixture:

| tmpdir | visible col | raw char there | raw-bytes scan | escape-free scan |
|---|---|---|---|---|
| macOS real (46ch) | 110 | `;` | `None` | finds `second.rs` |
| macOS long (56ch) | 132 | inside SGR | `None` | finds `second.rs` |
| Linux `/tmp` (15ch) | 48 | inside SGR | `None` | finds `second.rs` |
| `/tmp/w` (6ch) | 30 | inside SGR | `None` | finds `second.rs` |
| CI long (55ch) | 140 | inside SGR | `None` | finds `second.rs` |

Swept every tmpdir length 2..200: zero failures. The margin is wide rather than marginal — on a real macOS tempdir the click column sits 53 chars past the end of the first raw token and 142 chars before the start of the second. The `len/10 + 4` scaling is sound because each repetition adds 21 raw chars for 0 visible chars, so the divergence grows ~2.1x faster than the path length it must outrun. The author's `/tmp/w` regression is covered by the sweep.

**The fixture guard is correctly stated and can fire.** Testing the committed `is_none_or(|fr| fr.path != second)` against deliberately weakened fixtures:

- escape run removed entirely → guard **fails** (raw scan finds `second.rs`)
- shortened to 1 repetition → guard **fails**
- shortened to 2 repetitions → guard **fails**
- 3+ repetitions → guard passes
- as shipped (9 reps on macOS) → guard passes

So it is a real check, not an assertion that can never fail, and it catches precisely the "later edit shrinks the escape run" regression it was written for. Blaming the fixture rather than the code in the message is the right call.

**Finding 2 — the lookalike is the right case to have picked, and the other three are not blocking.** `host.com:443` does match `PATH_LINE_RE` (capture `('host.com', '443', None)`), so it reaches the `is_file()` guard, and the paired `dragging` assertion distinguishes "declined and fell through" from "eaten". The `editor.open(&log)` before it resets `dragging` to false, so the assertion starts clean rather than inheriting the previous drag.

The three omitted arms are nice-to-have:
- `path:12:5` — `fr.column` is passed straight to `open_resolved_file_ref`, shared with `terminal_file_click` and already covered by its tests.
- `~/` and relative — both resolve through logic structurally identical to `resolve_pane_path`, which has its own coverage; and both fail closed via `is_file()`, so a bug there declines rather than opening the wrong file.

None of them can produce a wrong-file open, which is the only outcome that would warrant blocking.

### Checked, not a defect

- `src/app/mod.rs`, `Cargo.toml`, and `src/release_notes/0.1.924.md` are byte-identical to `9e97d42` — the fix is test-only, so round 1's verification of dispatch ordering, column mapping, path resolution, and borrow scoping still stands unchanged.
- Backend widened 120→200: the click column is 110–140 on real tempdirs against ~125–155 cells of visible text, so it fits with headroom. The escapes have no printed width, so the ~314–386 byte raw line is irrelevant to the cell budget.
- `raw.lines().next()` is exact — the fixture writes `\n` only, no `\r`.
- `second.rs` is written with 7 lines, so the `:7` reference resolves and `cursor_row == 6` is in range.
- `first.rs` is never clicked; it exists solely to place a token before the escape run, which is what creates the byte/column divergence. Correct as written.
- The `visible` closure re-reads through `app.editor.log` after each re-open, so it cannot serve text from a stale view.

- None.

### Latent fragility (not blocking, no action needed)

- The 200-cell backend stops fitting the click column once the tempdir path reaches 91 characters (real values: macOS ~46, Linux ~15, CI ~55 — roughly 2x headroom). If `TMPDIR` were ever that long the test fails loudly on the `find`/click rather than passing silently, so it cannot become a false green.

### Verification run

- tests: not run by the reviewer (host suite lock; `cargo` blocked by hook). Gates reported clean by the author on this head — check, clippy under `-D warnings`, tests passing, and both mutations still killed.
- Fixture geometry, guard firing behaviour, and cell-width headroom independently re-derived by simulating `PATH_LINE_RE` and the SGR stripper outside the repo.

Issue #119, opened 2026-08-13
Task problem matchers: build errors from task panes into PROBLEMS


vitali87:
## Summary

Tasks (Ctrl/Cmd+Shift+B and "Tasks: Run Task") run in a named terminal pane, but their compile errors never reach the PROBLEMS panel — VS Code's problem matchers are why its Problems view stays trustworthy during builds. croft's PROBLEMS panel is LSP-only today.

## Design (croft-native)

- Ride the existing reader-thread line scanning (the TriggerScanner already strips escape sequences and matches completed lines per pane): a problem matcher is a regex with named captures (`file`, `line`, `col`, `severity`, `message`) applied to the TASK pane's lines while a task runs.
- Built-in matchers as data, like [[test_runners]]: rustc/cargo (`error[E0308]: ... --> src/x.rs:12:5`), tsc, eslint-stylish, python tracebacks, generic `file:line:col: severity: message`.
- Hits land in PROBLEMS under a "build" source beside the LSP diagnostics (badged separately), navigable exactly like LSP entries; cleared when the same task re-runs.
- Paths resolve against the task's cwd; unresolvable paths keep the row but don't navigate.

## Non-goals (v1)

Custom user-defined matchers in tasks.json (the built-ins cover the shipped runners; the data-driven registry leaves the door open); multi-line matcher state machines beyond the built-ins' needs.

PR #530, opened 2026-09-07
feat(problems): check the whole project and put findings in PROBLEMS

vitali87:
Issue #256's third acceptance criterion: "Problems: Check Whole Project" populates PROBLEMS from `tsc --noEmit` and replaces results on rerun.

Criteria 1, 2 and 4 shipped in #296 — I verified that against main rather than trusting the issue body, which still reads as unstarted. Criterion 5 (a file-count threshold so the sweep does not auto-enable on giant monorepos) is deliberately **not** here; it is separable and gets its own PR.

## What it does

`tsc --noEmit` where a `tsconfig.json` sits at the workspace root, fed through the existing matcher pipeline — which already parses both of tsc's output shapes, so no new parsing.

The sweep gets **its own producer id** rather than borrowing a pane's. Diagnostics are keyed by producer so a re-run replaces that producer's previous contribution; sharing an id would make a pane's next command wipe the sweep's rows and the sweep wipe the pane's. Pane uids are allocated from 1 upward by an `AtomicU64`, so `u64::MAX` cannot collide.

It runs off the main loop on the same channel-and-drain pattern as history snapshots, with a guard against stacking compilers.

## Three defects found in review, all verified against real `tsc`

**A checker that failed to run reported "no problems."** Both tsc matchers require a `file(line,col)` prefix, so a global error like `error TS5023: Unknown compiler option` matches nothing and the row count is zero — which is exactly what a clean project looks like. Ignoring the exit status is correct for diagnostics (tsc exits non-zero precisely *when it finds them*), but that left nothing to separate the two cases. Zero rows **and** a non-zero exit is now an invocation failure, named with the reason from stderr, falling back to stdout.

**`npx tsc` was actively dangerous.** On a project without TypeScript installed, npx does not fail — npm's registry `tsc` is an unrelated Unix terminal-control tool, which it downloads and runs against your source. Combined with the above, the user was told their project was clean. Now `node_modules/.bin/tsc` first, `npx --no-install` second, so a missing compiler is an error rather than a download.

**The status counted the wrong thing.** `problems.total_count()` is the panel's projection: it mixes in every language server's diagnostics and every pane's build output, and the Open Files scope filters it — so a sweep whose findings were all in unopened files reported "no problems", which is the one thing this feature exists to do. It now counts its own producer's contribution.

## Verification

`cargo check` clean, `clippy -D warnings` clean, 5 tests pass.

Four mutations, each killed: borrowing a real pane id, removing the replace loop, ignoring the exit status again, and reverting to the panel's count. The test helper drives the real channel and `drain_project_check` rather than setting the status itself, so it cannot pass while the drain is broken.

The helper uses `ExitStatusExt::from_raw`, so it and its two tests are `#[cfg(unix)]`-gated — an ungated spawning helper breaks the Windows test build without failing anything locally or in CI.

The version bump is its own final commit, so a re-bump is one amend rather than a rewrite of the commits carrying the work.

Part of #256.


vitali87:
## 🤖 Fallback review (round 4) — greptile-local, Claude Opus 5

CodeRabbit is rate-limited repo-wide (its check reports `pass` over a "Review limit reached" body), so this is the review of record. Rounds 1–3 ran at `8708712` (2/5, three P1s) and `aae693a` (3/5, two P1s); this is the re-verification at `ac20ff1`.

---

### Greptile-local Summary

Both round-3 findings are fixed and I verified each against the real captured bytes rather than the fixture. `first_meaningful_line` now yields a useful reason on the actual `npx --no-install tsc` output, and moving the count to scan time gives the right answer in both producer orderings. The P2 redraw is fixed. No findings remain; the `install_build_diags` follow-up is correctly left out of this PR.

### Confidence Score: 5/5

No P0/P1/P2 findings; both fixes are pinned by tests that fail in opposite directions under the old code, and the conventions check out.

### Findings

None.

### Answers to the author's questions

- **The fixture now matches what I captured.** Real `npx --no-install tsc --noEmit` with no `node_modules`: stderr is **0 bytes**, stdout opens with an empty line then ANSI-coloured banner lines. The fixture reproduces all three load-bearing properties — leading blank line, ESC on the text line, empty stderr. It is a shortened banner (the real output has an extra all-whitespace banner line before the text), but that difference does not weaken the test: I ran the implementation's logic over the **real captured bytes** and it produces `The project checker failed: This is not the tsc command you are looking for`, with no escape characters and a non-empty reason. The whitespace-only banner line strips to empty and is skipped by the same `find` that skips the leading blank.
- **The counting scan matches the branch `apply_build_scan` takes.** Confirmed, not assumed. `task_matcher_by_pane` has exactly one insert site (`src/app/mod.rs:24481`, inside `assign_task_matcher`), and both of its callers (`:24448`, `:24457`) pass a `PtyTerminal::uid()`, allocated from `AtomicU64::new(1)` (`src/widgets/terminal.rs:2219-2222`). `PROJECT_CHECK_ID` is `u64::MAX` and is never a terminal uid, so the `get(&pane)` at `src/app/mod.rs:8326` is always `None` for the sweep and the `matchers.scan_batch(output, cmd)` arm is always taken — the same call the count makes. The other early return, `watch_published_panes` (`:8320`), is likewise only ever inserted for terminal panes (`:31143`), so it cannot make the count and the install disagree either.
- **Both orderings are pinned correctly, and non-vacuously.** Leg 1 (sweep, then pane on the same file, then sweep) asserts `1 problem` where the old store-sum read **2**. Leg 2 (pane first, then sweep reports `shared.ts` + `only.ts`) asserts `2 problems` where the old store-sum read **1**, because `install_build_diags` never listed `shared.ts` in the sweep's `touched`. The two legs fail in opposite directions under the old implementation, so neither alone would have caught it.
- **Not filtering on `source` was the right call.** `src/build_matchers.rs:105` hardcodes `source: String::from("tsc")` for both matcher shapes, so a pane running tsc is indistinguishable from the sweep by that field. Counting at scan time is the right fix.
- **The `install_build_diags` follow-up should stay out of this PR.** The defect is that the second producer to reach a shared file never records it in its own `touched`, so it cannot later clear its row — that is `install_build_diags`'s own contract, reachable from any two producers, and the correct fix (per-producer row ownership) touches the #119/#252 path this PR does not otherwise modify. This PR makes the overlap *routine* rather than *possible*, which raises the follow-up's priority but does not make it this PR's bug: the symptom is a stale row after a producer stops reporting, not a wrong count, and the count is now correct in both orderings.

### Checked, not a defect

- `project_check_rows` cannot go stale: single writer (`ingest_project_check`), single reader (`project_check_count`), and the drain always ingests before reading.
- `first_meaningful_line` uses `remote_connect::strip_ansi`, which consumes CSI sequences by scanning to a `0x40..=0x7e` final byte — correct for the `\x1b[41m` / `\x1b[0m` SGR codes npx emits.
- Rows-found still ignores exit status, so genuine diagnostics (real `tsc@5` exits **2**) stay out of the failure branch; `tsc` bad-flag output still yields `The project checker failed: error TS5023: Unknown compiler option '--nosuchflag'.`
- `Err` arm now returns `true`, so a spawn failure flags `changed` and repaints.
- The extra `scan_batch` in `ingest_project_check` is a second pass over the same output, on a path that already spawned a compiler — no meaningful cost.
- `project_check_running` still cannot wedge: `Command::output()` returns `io::Result` rather than panicking, `App` owns the sender for its lifetime, and the drain clears the flag before matching.
- Drain placement unchanged and still unconditional in the main loop via `sync_explorer_panels`.
- Version 0.1.927 with the bump as the final commit (`ac20ff1`), `Cargo.lock` consistent, `src/release_notes/0.1.927.md` present; main still 0.1.924.
- `#[cfg(unix)]` gating consistent across all three tests and the helper.

- `install_build_diags` records a path in `touched` only when the entry was empty, so the second producer to report a file cannot later clear its own rows from it. Follow-up, as agreed above.

### Verification run

- tests: not run by the reviewer (host suite lock; `cargo` hook-blocked). Gates reported by the author on this head: check clean, clippy clean under `-D warnings`, 6 tests pass, both round-3 fixes mutation-verified.
- Round-3 fixes verified by replaying `first_meaningful_line` and `strip_ansi` over the **real** `npx --no-install tsc --noEmit` stdout/stderr bytes captured earlier, and by modelling both producer orderings against the `TSC_PAREN` matcher to confirm each test leg fails under the previous implementation.

vitali87:
Merged `origin/main` in — #528 landed while round 4 was running, moving main to 0.1.925 and this branch 6 commits behind.

Version 0.1.927 still clears main, so no re-bump was needed. Only `Cargo.toml` and `Cargo.lock` conflicted (the version line); the lock was regenerated and verified against the manifest rather than hand-merged.

Verified both sides survived rather than assuming: this PR's 29 `project_check` references, and #528's `release.yml` and binstall metadata. Merge ancestry asserted — two parents, `origin/main` a genuine ancestor.

The merged tree is code neither branch tested, so the gates were re-run on it rather than carried over: `cargo check` clean, `clippy -D warnings` clean, 6 tests pass with all four named tests confirmed present.

The 5/5 review above was at `ac20ff1`. This merge adds no source changes of its own — `git diff ac20ff1..02c1558 -- 'src/**/*.rs'` is empty, so the review still describes the code being merged.

PR #531, opened 2026-09-07
feat(problems): cap the whole-project check's automatic runs by size

### LINKED ISSUE #256: Project-wide diagnostics: whole-workspace errors in PROBLEMS, not just open files

vitali87:
Issue #256's fifth and final acceptance criterion: a 100k-file root does not auto-enable the whole-project check, and the setting overrides.

Criterion 3 (the sweep itself) shipped in #530; criteria 1, 2 and 4 in #296. **With this, every acceptance criterion on #256 is met.**

## What it does

`workspace_exceeds(root, limit)` counts files with `ignore::WalkBuilder` and **stops at the threshold** — a cap that counts a 100k-file tree to conclude the tree is too large has already paid the cost it exists to avoid.

Ignored and vendored directories don't count. `node_modules` alone puts most JS projects over any sensible cap, and disabling the sweep for JS projects would disable it for the projects it's *for*. Reuses the file finder's `is_noise_dir`.

`problems_project_scope` is auto/on/off, cycled from a new palette entry. An unrecognised value reads as `auto`, never `on` — a typo must not start a compile on a huge root, which is the exact failure the cap prevents.

An **explicit** sweep always runs; the cap governs *unasked* runs, and the user is told once when automatic ones won't start.

## Review findings, all fixed

**The walk was on the UI thread.** The toggle called the cap inline to describe its effect. My early-stop reasoning was half right: it bounds the cost *above* the cap and does nothing below it, so a 60k-file repo got walked to completion on a keypress. This repo runs whole-workspace walks on a background thread (`mod.rs:19604`, and both `file_finder` and `search` do it). The count now rides the sweep's existing worker.

**That left the cap with no caller.** Rather than an `#[allow]` or a wrapper — I tried the wrapper and deleted it, since it just moved the dead code up a level — it now has the consumer the issue actually specifies: *"otherwise stays off with a one-time hint"*, computed on that same worker and drained beside the sweep's result.

**Mode strings were matched raw** while the sibling `problems_scope` pref normalises, so a hand-edited `"Off"` silently meant `auto` — a size-dependent maybe where the user asked for never. One normaliser now serves both the predicate and the cycle.

## Verification

`cargo check` clean, `clippy -D warnings` clean, 7 tests.

Three mutations killed: unknown-setting fallback flipped to `on`, noise-dir filter removed, normaliser reverted to raw matching.

**A fourth mutation survived and found a real gap.** Flipping the cap's `>` to `>=` passed the whole suite — the boundary was untested, so an off-by-one could have shipped behind seven green tests. The doc comment claimed "exactly `limit` does not exceed it" and nothing checked it. Now pinned on both sides (12 files must not exceed 12; must exceed 11), and the mutation dies.

Closes #256.

## Summary by CodeRabbit

- **New Features**
  - Added controls for whole-project problem checks: **Auto**, **Always**, or **Never**.
  - In Auto mode, checks no longer start automatically in workspaces exceeding 100,000 files.
  - Displays a one-time notice when the workspace is too large for automatic checking.
  - Excludes vendored directories such as `node_modules` from workspace-size limits.
  - The selected behavior is saved across sessions.

- **Release**
  - Updated the package version to 0.1.929.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.927"
+version = "0.1.929"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.927"
+version = "0.1.929"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -8251,6 +8251,13 @@ impl App {
         let _ = self.project_check_tx.send((cwd.to_path_buf(), Ok(out)));
     }
 
+    /// Whether the session's one-shot size hint has been spent (#256).
+    /// Test-only reader; the field is private.
+    #[cfg(all(test, unix))]
+    pub(crate) fn project_check_hinted_for_test(&self) -> bool {
+        self.project_check_hinted
+    }
+
     /// Queue a size hint on the REAL channel (#256). Test-only.
     ///
     /// Does not drain: the point is to have it in flight when the result
@@ -9072,6 +9079,11 @@ impl App {
         // fast-failure path (over-cap root, no compiler installed) is
         // exactly where that happened.
         let hinted_now = self.project_hint_rx.try_recv().is_ok();
+        if hinted_now {
+            // Here, not at the spawn: only a hint that was actually WARRANTED
+            // spends the one-shot.
+            self.project_check_hinted = true;
+        }
         // A whole-project check finished: its rows and its verdict land
         // together, so the status cannot describe a different run's result.
         if self.drain_project_check() {
@@ -34953,8 +34965,13 @@ impl App {
                         let root_for_thread = root.clone();
                         let mode = self.problems_project_scope.clone();
                         let hint_tx = self.project_hint_tx.clone();
+                        // NOT set here: this fires for every explicit check,
+                        // so a run under `on`, `off`, or an under-cap `auto`
+                        // would consume the one-shot and a later oversized
+                        // `auto` run would stay silent forever. The flag is
+                        // set where the hint is RECEIVED, which is the only
+                        // point that knows one was warranted.
                         let hinted = self.project_check_hinted;
-                        self.project_check_hinted = true;
                         std::thread::spawn(move || {
                             // The cap's one-time hint (#256 criterion 5).
                             // Computed HERE, on the thread already spawned
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2570,6 +2570,9 @@ pub struct App {
     /// How many rows the last sweep produced (#256), counted before its
     /// items join a store shared with every other producer.
     project_check_rows: usize,
+    /// Whether the sweep may run unasked (#256): `auto`, `on` or `off`.
+    /// Read at startup; an unrecognised value behaves as `auto`.
+    problems_project_scope: String,
     /// Replies from the background test-source locator (the workspace grep is
     /// too slow for the render thread): the root the search ran under, the
     /// test name, and the hit. A reply from before a re-root is dropped.
@@ -4521,6 +4524,11 @@ impl App {
             project_check_tx,
             project_check_running: false,
             project_check_rows: 0,
+            problems_project_scope: if loaded_prefs.problems_project_scope.is_empty() {
+                String::from("auto")
+            } else {
+                loaded_prefs.problems_project_scope.clone()
+            },
             test_jump_rx,
             test_jump_tx,
             timeline_fetched: None,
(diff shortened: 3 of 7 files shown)


vitali87:
## 🤖 Fallback review (round 3) — greptile-local, Claude Opus 5

CodeRabbit is rate-limited repo-wide (its check reports `pass` over a "Review limit reached" body), so this is the review of record. Round 1 ran at `bc3eae7` (3/5, two P1s), round 2 at `4d4d91e` (4/5, one P2); this is the final verification at `4835d5f`.

---

### Greptile-local Summary

The P2 is fixed correctly. `hinted_now` captures the receive, `drain_project_check()` sets the outcome, then the hint is appended — so the result leads and the one-shot survives the same-tick race that previously erased it. The misleading comment was replaced with an accurate one that names the failure mode. The new test drives the real path, and both helpers are unambiguous. No remaining findings; the branch is ready to merge.

### Confidence Score: 5/5

All three rounds' findings are fixed and verified by reading; the change is covered by tests that exercise the shipped path, and docs, release note and help text agree with the code.

### Findings

None.

### Checked, not a defect

- **P2 fixed.** `src/app/mod.rs:9074-9087`: the `try_recv` result is held in `hinted_now`, `drain_project_check()` runs, then `push_str`. The result no longer overwrites the hint because the hint is written after it. The old comment claimed deprioritisation while the code erased; the new one names the destruction, the one-shot consequence and the fast-failure path that triggered it.
- **The test exercises the real path.** `the_size_hint_is_not_erased_by_a_result_in_the_same_tick` (`src/app/tests.rs:38203-38223`) puts both messages in flight and calls `app.sync_explorer_panels()`, which is the only production caller (`mod.rs:49392`). The hint logic lives in that function, so `finish_project_check_for_test` would have bypassed it entirely — the author's first version would have been green for the wrong reason, and the second helper was necessary, not incidental.
- **The two helpers are not a trap.** Both doc blocks name the distinction in their first two lines, and each names the *other*, so a reader arriving at either is pointed at the choice. More importantly the failure mode is not silent: using `finish_project_check_for_test` in the race test drains before `sync_explorer_panels` runs, so the hint is still queued when the append fires and the "no problems" assertion fails on a status that leads with the hint. A wrong-helper mistake fails rather than passes. The names also carry the difference (`finish` = drains, `queue`/`send` = does not).
- **No truncation or width concern.** `self.status` is pushed as one `Span` into `Line::from(spans)` and rendered by `Paragraph::new(...)` at `mod.rs:16572` with no `.wrap()`, so ratatui clips at `status_rect` width. No panic, no reflow, no layout break — the tail is simply not drawn. This is also not new: `"The project checker failed: {why}"` (`:8217`) already interpolates an arbitrary compiler line into the same field, so unbounded status length is the existing contract, and the appended hint is a bounded ~110 chars on top of it. On a narrow terminal after a long failure reason the hint's tail clips, but its first words ("too large for the whole-project check") come first in the appended text, so the meaning survives.
- **The re-root race is now less reachable, not more.** Appending narrows it. Previously the hint could be applied on any tick after a re-root, with its own status write. Now it only ever modifies a status that `drain_project_check` just set, and that drain already discriminates: it receives `(cwd, done)` and `ingest_project_check` resolves paths against the *sending* run's `cwd`. The remaining window — hint and result both from the old root, drained after a re-root — produces a hint that correctly describes the run whose result it is attached to. Coupling the hint to the result fixed the staleness as a side effect of fixing the erasure.
- **Both orderings read coherently.** When the hint lands alone (slow checker), status is `"Checking the whole project..."`, giving `"Checking the whole project... - too large for the whole-project check to run on its own; set it to always..."`.
- `#[cfg(all(test, unix))]` gating on both new helpers matches the sibling's, and the new test carries `#[cfg(unix)]`; `ExitStatusExt::from_raw` is the reason.
- Doc ownership on both new helpers: each block sits on the function it describes, and the sibling's doc at `:8263` is intact.
- Earlier rounds re-confirmed on this head: no `workspace_exceeds` call on the UI thread; `project_scope_mode` normalises for both the predicate and the cycle; hint gated to `auto` and once per session; `seen > limit` boundary pinned both sides.
- Version 0.1.929 with `src/release_notes/0.1.929.md`, bump is the final commit, above main's 0.1.927.

- None.

### Verification run

- tests: not run by the reviewer — builds and suites are forbidden on this host; verification was by reading `git show 4835d5f:...`. Reported gates on this head: doc-ownership clean across 4 files, `cargo check` clean, clippy clean under `-D warnings`, 8 tests pass, five mutations killed (noise filter, unknown-setting fallback, raw matching, `>` boundary, hint erasure).

vitali87:
Re: CodeRabbit's finding on `project_check_hinted` — fixed in `65e16f7`, and it was a real one-shot leak.

The flag was set at the spawn, so **any** explicit check spent it — including under `on`, `off`, or an under-cap `auto` — and a later oversized `auto` run then stayed silent forever, which is the one case the hint exists for.

It is now set where the hint is **received**, in the drain, since that is the only point that knows a hint was warranted.

`a_check_that_warrants_no_hint_does_not_spend_it` pins both halves: a run producing no hint leaves the flag unspent, and a later run that does warrant one still gets it. Mutation-verified — restoring the eager set at the spawn fails that test.

Worth noting for the record: the local greptile review examined this flag's placement and cleared it, reasoning correctly that it sits inside the spawn arm after the no-checker and already-running guards. That is true, and it misses what CodeRabbit caught — "a check started" and "a hint is warranted" are different questions, and only the second should spend a one-shot. Two reviewers, two different classes of defect; neither was redundant here.

Thread resolved.

PR #534, opened 2026-09-07
fix(editor): a symbol tab does not follow you into another file

vitali87:
A live bug on main, found while sizing #369's remaining work.

`symbol_tab` holds **byte offsets into one buffer**, and nothing cleared it except a straddling edit. Navigate away and it stayed attached — so `follow_symbol_tab_edit` applied the *new* file's edits against the *old* file's offsets, walking the range somewhere meaningless instead of reporting `Gone`.

## The fix

Cleared in `open_at`, beside the `hover_popup` / `hover_diagnostic` resets that are there for exactly the same reason: per-file state does not travel. That is the common path for all ten navigation call sites, so one clear covers them rather than ten.

**Only on a file change.** Reopening the tab's own file — a go-to-definition landing in the same file, a Back jump — never left the buffer the range describes, and clearing then would close the tab on ordinary navigation.

## The over-correction survived my first test

Worth recording, because the test looked correct.

Mutating the guard to clear **unconditionally** — which fixes the reported bug and silently closes the tab on every same-file jump — **passed both tests**, including the paired positive written as a control.

The positive seeded `symbol_tab` *after* the navigation, so the guard never executed on it. It exercised `follow_symbol_tab_edit`'s arithmetic, not the clear: two independent tests, one merely shaped like a control. It now seeds **before** the navigation, so the guard runs and must decline to fire, and the mutation dies.

A pair only pairs if both halves reach the same branch. Asserting the opposite outcome is not the same as exercising the same guard.

## A second wrong assumption, corrected

My first paired case asserted an insertion at offset 0 against a range starting at 0 would *shift* it to `(4, 31)`. It returned `(0, 31)`, and the code is right: `src/symbol_range.rs:86` deliberately treats a zero-width insertion at `start` as **inside**, because otherwise typing at the top of a function pushes the tab off it. The range now starts at 11 so there is a genuine "above" region, with a comment naming why 0 does not work.

## Verification

`cargo check` clean, `clippy -D warnings` clean, 2 tests.

Both mutations killed: removing the clear entirely, and making it unconditional.

## Scope

**Part of #369** — this closes nothing. All three acceptance criteria need the rendering layer, which does not exist: `grep` for `symbol_tab` returns 11 non-test references and none render anything. See the issue for the current-state breakdown.

## Summary by CodeRabbit

- **Bug Fixes**
  - Symbol tabs now close when navigating to a different file or switching editor groups.
  - Symbol tabs remain open and continue tracking correctly when navigating within the same file or making edits.

- **Release**
  - Updated the application to version 0.1.930.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.929"
+version = "0.1.930"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.929"
+version = "0.1.930"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -6181,7 +6181,11 @@ impl App {
     /// but `open_preview` resolves through `find_tab_matching`, which
     /// canonicalises, so it selects the ALREADY-OPEN tab and `editor.path`
     /// keeps its original spelling. The two sides therefore cannot diverge
-    /// here, and a canonicalising branch was unreachable: three attempts to
+    /// FOR THE SAME FILE - they diverge freely when the file really changes,
+    /// which is the point - because the tab's path is a copy of
+    /// `editor.path` and `find_tab_matching` canonicalises before selecting.
+    /// The field itself is not canonical. A canonicalising branch was
+    /// therefore unreachable: three attempts to
     /// write a test that entered it all failed, and a mutation deleting it
     /// survived, which is what proved it dead rather than merely untested.
     fn drop_symbol_tab_if_file_changed(&mut self) {
@@ -14738,6 +14742,12 @@ impl App {
         if self.editor.focused {
             self.poke_cursor();
         }
+        // Group focus changes the active FILE without opening anything and
+        // without a path (#369): `focus_editor_group` and
+        // `move_active_editor` swap another group into `self.editor`, so
+        // neither the open-based sync nor a path-argument guard sees them.
+        // Every focus change lands here, so this is the one place that does.
+        self.drop_symbol_tab_if_file_changed();
     }
 
     fn sync_focus_flags(&mut self) {
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -6158,6 +6158,44 @@ impl App {
     fn sync_open_file_poll_mtime(&mut self) {
         let paths = self.open_tab_paths();
         self.fs_watch.sync_open_file_mtime(&paths);
+        self.drop_symbol_tab_if_file_changed();
+    }
+
+    /// Close a symbol tab once the editor is showing a different file (#369).
+    ///
+    /// The range is BYTE offsets into ONE buffer, so it cannot survive the
+    /// move: `follow_symbol_tab_edit` would apply the new file's edits
+    /// against the old file's offsets and walk the range somewhere
+    /// meaningless rather than reporting `Gone`.
+    ///
+    /// Keyed on the ACTIVE path after the change, not on a path passed in.
+    /// `open_at` looked like the chokepoint and is not - search hits, a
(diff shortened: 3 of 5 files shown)


vitali87:
## 🤖 Fallback review (round 3) — greptile-local, Claude Opus 5

CodeRabbit is rate-limited repo-wide (its check reports `pass` over a "Review limit reached" body), so this is the review of record. Round 1 ran at `2057809` (3/5, two P1s), round 2 at `e64a299` (3/5, one new P1, one finding withdrawn); this is the final verification at `9551b96`, plus the P2 fix in `439c93f`.

---

### Greptile-local Summary

Clears `App::symbol_tab` when the active file changes, guarding a byte-offset range that would otherwise be applied against a different buffer. Over three heads the guard moved from `open_at` (which covered ten of nineteen navigation sites) to `sync_open_file_poll_mtime` plus `focus_pane`, which between them cover every way the active file changes. Both P1s I raised are fixed and verified; my symlink finding was withdrawn as wrong. Ready to merge.

### Confidence Score: 4/5

Correct and now completely covered; the only open item was the group test's conditional branching, addressed in `439c93f`.

### Findings

1. **[P2] The group test branches on a layout that is fully determined** — `src/app/tests.rs`

   `split_editor` → `split_editor_dir(Horizontal, true)` puts the existing group at DFS index 0 and the new active group at index 1 — asserted directly in `split_active_makes_a_two_leaf_horizontal_split_with_the_new_group_active_after` (`src/app/editor_layout.rs:888-911`). So left holds `a.rs`, right holds `b.rs`, and `focus_editor_group(true)` targets `a.rs`.

   The `active == a` branch is therefore the one that always runs, and the `else` — carrying the assertion the test is named for — is dead. The clear is reached only via a nested `focus_editor_group(false)` wrapped in an `if`. An `if` guarding an assertion passes when the condition is false, so a future layout change would leave the test green having asserted nothing about clearing.

   The mutation result (removing the guard from `focus_pane` fails the run) confirms the test reaches the clear today; it does not show the branching is needed.

   **Fixed in `439c93f`:** both directions asserted unconditionally, with the layout pinned as an assertion rather than described in a comment.

### Checked, not a defect

- **`focus_pane` is the right home for the guard.** It is the single exit for every focus change, and `focus_editor_group`, `move_active_editor` and `split_editor_dir` all end in it. Placement after `sync_focus_flags` and `poke_cursor` is correct — the group swap into `self.editor` has already happened, so `editor.path` reads the post-change file.
- **Not too much work for 119 call sites.** `drop_symbol_tab_if_file_changed` opens with `let Some(...) = self.symbol_tab.as_ref() else { return; }`, so sites with no symbol tab pay one `Option` discriminant check. Calling it from `focus_editor_group` and `move_active_editor` separately would repeat the enumerate-the-callers mistake the `open_at` head already made once.
- Terminal and sidebar focus calls are harmless: they do not change `self.editor`, so the paths compare equal and the tab survives.
- **The withdrawn symlink finding stays withdrawn.** `symbol_tab` is written in exactly one production place, from `self.editor.path.clone()`, and `find_tab_matching` canonicalises before `select(idx)`, which touches no path field. The surviving mutation on the deleted branch is the right evidence. My original citations were accurate but described paths with *two independent producers*; this comparison has one.
- The narrowed doc comment now states what actually holds — cannot diverge *for the same file*, and the field is not canonical.
- Duplicate same-file assertion removed; that test still reaches the guard on both the negative and the paired positive.
- Version 0.1.930 in `Cargo.toml` and `Cargo.lock`, `src/release_notes/0.1.930.md` present and updated to match the widened coverage, bump is the final commit.
- No `#[cfg(unix)]` needed — neither test spawns a process.

- `follow_symbol_tab_edit` has one production caller, `apply_collab_spans`, so a symbol tab tracks remote collab edits only, not the user's own typing. This narrows the live impact of the bug fixed here without changing that the stale tab was wrong.
- `rename_open_path` rewrites `editor.path` without notifying `symbol_tab`; the tab is cleared on the next focus or sync, which is defensible for a byte-range view after a rename.

### Verification run

- tests: not run by the reviewer (build and suite prohibited on this host). Reported gates on this head: doc-ownership clean, `cargo check` clean, clippy clean under `-D warnings`, 3 tests, six mutations run across the branch — five killed, and one that **survived**, which is what proved the canonicalising branch dead rather than untested.

Issue #461, opened 2026-09-01
Release gate treats a test-only change inside a shipped file as shipped


vitali87:
The `version bump + release notes` job states its own intent in the comment
above it: "Docs / CI / test-only changes ship nothing and are exempt." It then
encodes "test-only" as a path list:

```
shipped="$(printf '%s\n' "$changed" \
  | grep -E '^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)' \
  | grep -vE '^(src/app/tests\.rs|tests/)' || true)"
```

A change confined to a `#[cfg(test)] mod tests` block inside a shipped file is
test-only in fact, and the release binary it produces is byte-identical, but it
does not match either exempt path. The gate then demands a version bump and a
`src/release_notes/<version>.md` for a change that the welcome panel has
nothing true to say about: the notes file is a user-facing highlight list, so
the only way to satisfy the gate is to describe a change that does not exist in
the binary.

Live instance: #459 changes one assertion inside `mod tests` in
`src/markdown.rs` and nothing else. Every other check is green and the job
fails with:

```
Shipped files changed:
  src/markdown.rs
This PR changes shipped files but the version is still 0.1.857.
This PR changes shipped files but does not write src/release_notes/0.1.857.md.
```

Croft keeps most unit tests beside the code they cover, so this is the ordinary
shape of a test-only fix here, not an edge case. `src/app/tests.rs` is exempt
only because it happens to be a whole file.

## Suggested shape

Filter the shipped list once more: a `.rs` file whose diff touches nothing
outside a `#[cfg(test)]` module ships nothing. Everything else keeps the
current behaviour, and anything the parse is unsure about counts as shipped, so
the failure direction stays "asks for a bump it did not need" rather than
"waives one it needed".

Worth pinning with tests either way, since a gate that cannot fail and a gate
that waives too much look identical from the outside.

PR #546, opened 2026-09-09
fix(lsp): make a broken handshake fail the tests instead of skipping

### LINKED ISSUE #545: LSP handshake tests report ok when initialize fails
The three `initialize_and_shutdown_*` tests in `src/lsp/client.rs` treat a FAILED handshake the same way they treat a missing server: they print a `SKIPPED:` line to stderr and return, so the test passes either way.

```rust
fn run_initialize_shutdown(config: ServerConfig) {
    ...
    if let Err(e) = result {
        eprintln!("SKIPPED: {display_name} initialize/shutdown failed: {e}");
    }
}
```

Two different states collapse into one green row:

- the server is not installed on this machine (a legitimate skip), and
- the server IS installed and the handshake is BROKEN (a regression these tests exist to catch).

Observed while changing the `initialize` path: running the compiled test binary in an environment where the rustup shim could not resolve `rust-analyzer`, the harness printed `SKIPPED: rust-analyzer initialize/shutdown failed: lsp initialize` and the run still reported `test result: ok. 1 passed`. Under `cargo test` the stderr line is captured, so nothing at all is visible: a broken handshake and a working one produce identical output.

The tests are only meaningful when the server is present, so the skip itself is right; what is wrong is that a real failure takes the same exit. Distinguishing them is enough:

- keep the skip when the binary is absent (`server_on_path` already answers that), and
- `panic!` rather than `eprintln!` when the server IS present and the handshake fails.

That way "not installed here" stays a skip and "installed and broken" is a failure, which is the difference the suite currently cannot express.

vitali87:
Closes #545.

The three `initialize_and_shutdown_*` tests gave two very different states the same exit: a `SKIPPED:` line on stderr and a green row. "No server on this machine" and "the server is here and the handshake is BROKEN" were indistinguishable, so a regression in the handshake was invisible to the tests written to catch it. Under `cargo test` the line is captured, so there was nothing at all to see.

## Classifying the error does not work

The obvious fix is to skip on the environmental error variants and fail on the protocol ones. That was the first attempt here, and the re-break falsified it: feeding rust-analyzer malformed `initialize` params makes it EXIT rather than reply, which arrives as `ServiceStopped` - byte for byte what a rustup shim whose toolchain lacks the component produces. The mutation still passed as a skip.

After the fact the two states are the same error. So the machine question has to be settled before the handshake starts.

## The probe

`server_runs_here` spawns the configured command, waits 750ms, and checks it is still alive. A real LSP server blocks reading stdin; a shim that cannot resolve its binary prints and exits. That is exactly the difference `server_on_path` cannot see, because the shim IS a file that exists and is executable.

Once that answers yes, any handshake failure panics.

## Verification

Against a live rust-analyzer, in both directions:

- restored tree: `ok. 1 passed`
- same mutation as above: `FAILED` - `rust-analyzer runs on this machine, but the handshake failed: lsp initialize: service stopped`
- restored again: `ok. 1 passed`

And both skip paths still skip: binary absent (`not on PATH`), and shim present but unresolvable (`is on PATH but does not run on this machine`).

Full suite 4374 passed, clippy clean, fmt clean.

## No version bump

The diff is confined to `#[cfg(test)] mod tests`, so `scripts/ships_nothing.py` reports "ships nothing" for it and the release gate requires no bump or notes (#461). Confirmed by running the gate's own check against this branch.

## Summary by CodeRabbit

- **Tests**
  - Improved language server integration testing by verifying that discovered servers are installed, remain running, and accept input before initialization.
  - Unavailable or non-runnable servers are now skipped, while initialization failures for runnable servers are reported as test failures for more reliable validation.

### DIFF
--- a/src/lsp/client.rs
+++ b/src/lsp/client.rs
@@ -1646,16 +1646,18 @@ mod tests {
     }
 
     fn run_initialize_shutdown(config: ServerConfig) {
-        if !server_runs_here(&config) {
+        // The tempdir comes first because the probe has to run in the SAME
+        // working directory the real spawn will use; see `server_runs_here`.
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let root = tmp.path().canonicalize().expect("canonicalize tempdir");
+        if let Err(reason) = server_runs_here(&config, &root) {
             eprintln!(
-                "SKIPPED: {} is on PATH but does not run on this machine",
-                config.name
+                "SKIPPED: {} does not run on this machine: {reason}",
+                config.command
             );
             return;
         }
         let rt = LspRuntime::new().expect("runtime");
-        let tmp = tempfile::tempdir().expect("tempdir");
-        let root = tmp.path().canonicalize().expect("canonicalize tempdir");
         let expected_name = config.name;
         let display_name = config.name;
 
@@ -1687,7 +1689,7 @@ mod tests {
     }
 
     /// Whether this machine can actually RUN the server, asked BEFORE the
-    /// handshake (#545).
+    /// handshake (#545). `Err` carries the reason, so the skip can say it.
     ///
     /// Classifying the handshake ERROR instead does not work, and the
     /// re-break proved it: feeding rust-analyzer malformed `initialize`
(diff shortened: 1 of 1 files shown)


vitali87:
@coderabbitai review

vitali87:

CodeRabbit reported "Review limit reached" on this PR and then declined to re-review the same commit ("does not re-review already reviewed commits"), so its green check carries no review behind it. This is the standing fallback, run three times across the PR's heads.

**Round 1** (`0a3a6ec`): REQUEST_CHANGES — two majors. The probe inherited the test process's cwd while `LspClient::spawn` uses `current_dir(workspace_root)`, and for a rustup shim the cwd selects the toolchain, so the probe could answer "runnable" where the real spawn could not run the server at all, turning a machine fact into a panic blaming croft (reproduced locally). And the fixed 750 ms sleep is the shape `CONTRIBUTING.md` forbids by name, with `test_budget::spawn_budget` as the sanctioned helper at 101 call sites. Plus two minors and a nit.

**Round 2** (`81b7994`): APPROVE (4/5, 4/5) — all five prior findings verified RESOLVED against the working tree; one minor and three nits left. The reviewer re-broke the new `should_panic` test itself to confirm it genuinely fails on `main`.

**Round 3** (`08bd2d8`), the current head, in full below.

---

### Walkthrough

The delta is a single-file, test-only follow-up in `src/lsp/client.rs` that hardens the `server_runs_here` probe added by 81b7994. It hoists `kill`/`wait` above the `match` so the spawned server is reaped on every path including the `panic!` arm, moves the probe's stderr drain onto a detached thread behind a `recv_timeout(spawn_budget(200ms))` so a forking launcher's grandchild cannot hang the read forever, and replaces the stub test's hard-coded `sleep 5` with a value derived from `spawn_budget(200ms) * 3`. No production code changes.

| Check | Result |
|---|---|
| `cargo fmt --check` | exit 0 |
| `ships_nothing.py <merge-base> HEAD src/lsp/client.rs` (#461 exemption) | exit 0 |
| `check_doc_ownership.py <merge-base> HEAD` | exit 0 |
| `cargo test --bin croft -- the_probe_separates a_server_that_runs_but_never_answers` | `ok. 2 passed; 0 failed` in 3.85s (both named tests present in the run, so not a vacuous empty-filter pass) |

### Findings

No blockers and no majors. Each hazard raised in the brief checks out, verified empirically rather than by reading:

- **Reader thread cannot leak unboundedly, and cannot panic.** Dropping `rx` on the timeout path while the thread is mid-`read_to_string` is safe: `tx.send` returns `Err(SendError)` on a disconnected channel rather than panicking, and the result is discarded. Confirmed directly (`send-after-drop is_err=true`).
- **`recv_timeout` fallback is correct.** Both `Err(Timeout)` and `Err(Disconnected)` fall to the catch-all, as does `Ok` of whitespace-only stderr. Every arm yields a reason containing `before the handshake`, which the probe test asserts on.
- **Reading stderr after `wait()` is sound.** `Child::wait` closes only `stdin`; `stdout`/`stderr` are untouched, so buffered bytes survive reaping. `Child::kill` only errors with `InvalidInput` after a successful `wait`, and no `wait` has happened at that point.
- **`as_secs_f32().ceil()` is correct at every reachable scale and never rounds to 0.** Enumerating the whole domain (`BASE_CALIBRATION = 2`, `load_scale` clamped to `[2, 4]`): `scale=2 → sleep 3`, `scale=3 → sleep 4`, `scale=4 → sleep 5`, ratio ≥ 3x in every case. Rust's `Display` for floats emits `3`, not `3.0`, so the argument is POSIX-valid and does not rely on GNU `sleep`'s float extension.
- **The two `spawn_budget` calls cannot disagree.** `load_scale()` memoises in a process-wide `static SCALE: OnceLock<u32>`, so every call in one test binary returns the identical value no matter how loaded the machine becomes mid-run. The derived stub lifetime exceeds the probe budget by a fixed 3x, not a hoped-for one. The max-scale case reproduces exactly the old hard-coded `sleep 5`, so this is a strict generalisation of the previous constant.

### Previous-finding verification

1. **Unbounded `read_to_string` on probe stderr — RESOLVED.** Read now runs on a spawned thread behind `recv_timeout(spawn_budget(200ms))`; worst case bounded at one extra budget, and the timeout path still produces a usable reason.
2. **`panic!` arm leaked the spawned server — RESOLVED.** `try_wait` is captured first and `kill`/`wait` run unconditionally *before* the `match` whose `Err` arm panics. Unwinding can no longer skip the reap.
3. **Stub `sleep 5` coupled by an undeclared constant — RESOLVED.** Raising `BASE_CALIBRATION`, `MAX_SCALE`, or the base now moves both sides together.

**On the dismissed nit** (the probe spends its whole budget even when the child exits at once): accepted, and the repo backs it more strongly than the author argued. `CONTRIBUTING.md` sanctions `spawn_budget(base)` by name for "a wait that hands its deadline to something else (a `recv_timeout`, a probe's own timeout)" — both new uses are exactly those shapes. Beyond style, polling would not buy what the nit assumed: the probe establishes that the child *stayed* alive, a negative property no amount of polling can confirm early.

### Test coverage (delta only)

Both touched tests exercise the changed code and pass. The probe test covers all three outcomes — alive (`cat`), exited (`true`, which drives the new threaded-stderr path to a prompt EOF), and unspawnable. The stub test covers the derived-lifetime construction end to end: a miscomputed `sleep` argument would surface as "did not panic" rather than a false green.

The one untested path is the `recv_timeout` *timeout* branch, which needs a launcher that forks and exits while a grandchild retains the stderr write end. No configured server does this; its failure mode is a less informative message, not a wrong result.

### Nitpicks

- `sh -c` does fork here, which makes the new defence better-founded than its comment claims: killing `sh -c "sleep 47"` leaves `sleep 47` reparented to PID 1 (dash does not exec a single command). It never reaches the stderr path, because the stub is alive when probed. Worth a word in the comment, since the in-repo example is more concrete than a hypothetical.
- Orphaned `sleep` grandchildren outlive the stub test by 3-5s, for the reason above. Pre-existing on `main` (`sleep 5` behaved identically) and self-limiting; the delta marginally improves it by shortening the sleep at the common floor scale.

### Scores

- **CodeRabbit (diff quality): 5/5** — no blockers, majors, minors, or nits against the delta itself.
- **Greptile (codebase fit): 5/5** — both `spawn_budget` uses are the exact shapes `CONTRIBUTING.md` prescribes, and the derived constant removes the coupling `test_budget`'s doc warns about.

VERDICT: APPROVE

PR #547, opened 2026-09-09
feat(debug): honour presentation.group and presentation.order

### LINKED ISSUE #318: Compounds: honour preLaunchTask, and decide on presentation/stopAll

vitali87:
Closes #318.

The last honourable part of the issue. `preLaunchTask` landed with the one-member compound launch and `presentation.hidden` with #524; `group` and `order` were refused on the grounds that "croft's picker has no grouping concept yet".

It needs one, and it needs VS Code's specifically, because a `launch.json` is written against VS Code's ordering.

## The comparator is transcribed, not invented

`visible_and_sorted` follows `getVisibleAndSorted` in `debugUtils.ts`. Every rule in it is one croft would plausibly have got wrong on its own:

- an entry that HAS a `presentation` sorts ahead of one that has none, so adding `{"order": 1}` to a single entry lifts it above untouched neighbours;
- within those, a grouped entry sorts ahead of an ungrouped one;
- groups compare by NAME, so grouping is alphabetical rather than first-seen;
- inside a group, a numeric `order` sorts ahead of a missing one.

## Configurations get `presentation` too

Not just compounds. VS Code builds one list from `getConfigurationNames()` covering both kinds and puts it through a single comparator, so a config and a compound naming the same `presentation.group` belong together — sorting the two kinds separately would split that group no matter what the file asked for. croft parsed `presentation` only for compounds.

## Two decisions worth flagging

**The sort is stable**, which is what leaves a workspace declaring no `presentation` in exactly its file order. Without that, shipping this would silently reorder every existing picker. There is a test for precisely that property, and the re-break below confirms it catches an unstable sort.

**Row ids are built before the sort**, from the unsorted indices, so `compound:{i}` and `{i}` keep naming the same entries however the rows move. Indexing the sorted sequence would launch something other than the row says as soon as one entry declares a `presentation`.

One deliberate divergence: VS Code compares group names with locale-aware `localeCompare`, croft compares bytewise. They agree on ASCII group names, and matching it properly would mean taking a collation dependency for the ordering of a picker.

## The refusal path shrinks

What remains in `unsupported_keys` is only what croft cannot READ: a `presentation` that is not an object, a `hidden` that is not a bool. Those keep refusing because `{"hidden": "true"}` is a typo, and launching it unhidden with no signal is worse than the refusal. The message no longer says "does not honour yet" or asks the user to *remove* the key — what is left is a typo to fix, not a feature to drop.

`stopAll` remains out at either value, and #310 still owns the multi-session work.

## Verification

Full suite 4379 passed, clippy clean, fmt clean, doc-ownership and architecture-content checks clean.

Two existing guards used `presentation.order` as their "a shape that still refuses" fixture, and would have gone vacuous the moment this honoured it. Both move to the quoted-bool typo, which still refuses.

Re-break, three mutations against the new tests:

| Mutation | Result |
|---|---|
| an entry with a `presentation` no longer outranks one without | comparator test FAILED |
| groups compare first-seen instead of by name | comparator test FAILED |
| the sort is no longer stable | file-order test FAILED |
| restored | `ok. 7 passed` |

The first attempt at that proof printed nothing for all three mutations and looked like a pass; the harness now fails loudly when a stage produces no `test result:` line, which is what caught it.

## Summary by CodeRabbit

* **New Features**
  * Debug configurations and compounds now honor `launch.json` presentation settings for visibility, grouping, and ordering.
  * Fractional ordering values are supported.
  * Configurations and compounds are sorted together, while entries without presentation settings retain their existing order.
  * Hidden entries are excluded from the debug picker and configuration count.

* **Bug Fixes**
  * Improved validation messaging for malformed presentation settings.

* **Documentation**
  * Updated architecture documentation for presentation and sorting behavior.


vitali87:
The `clippy + tests` failure was `pdf::tests::a_pdftoppm_whose_descendant_lingers_does_not_hold_the_call`, which this branch does not touch — `src/pdf.rs` is not in the diff. It is a load-dependent flake, filed as #548 with the evidence from both the CI run and a separate local failure of a sibling test. Re-triggered with an empty commit.

vitali87:
Both findings fixed in the push above.

**Malformed `group`/`order` now refuse**, like a malformed `hidden` already did. You are right that it contradicted the field's own doc, and `{"order": "1"}" is the same typo class as `{"hidden": "true"}`. One `malformed_presentation_keys` asks that question for all three keys now.

One distinction worth naming: a FRACTION is not a typo. VS Code's schema says `number` and its comparator subtracts one order from the other, so `1.5` is a legal position between two neighbours — the original `as_i64` would have refused it. `order` is a `f64` now, which costs the `Eq` derive on three types; dropped with the rationale this file already records for `ResolvedConfig`, as none of them is ever a map key. A test covers all four shapes.

Configurations stay lenient deliberately: they launch directly rather than through a gate, so there is no refusal path to hang a malformed key on, and VS Code ignores what it cannot read there too. That asymmetry is documented on the function.

**The release note** now states the bytewise-vs-locale caveat instead of claiming exact VS Code placement.

Local: clippy 0, 4380 tests pass. The earlier `clippy + tests` red was unrelated — `apt-get update` hit a transient hash mismatch on the runner image's Chrome repo, which short-circuited the `&&` so poppler-utils was never installed, and the step still went green; filed as #549, with #548 for a separate genuine flake in the same area.

vitali87:

CodeRabbit reviewed an earlier head here and then reported "Review limit reached" on the newest commit, naming the range `ca206a3..b0a8b40` it could not reach — so the `malformed_presentation_keys` work was unreviewed. This is the standing fallback for that gap.

**VERDICT: APPROVE** — CodeRabbit diff-quality 4/5, Greptile codebase-fit 4/5. No blockers, no majors.

### Both earlier CodeRabbit findings verified RESOLVED

- **malformed `group`/`order` dropped silently.** `malformed_presentation_keys` records all three keys, skipping `null` per key. The reviewer checked the honoured and refused sets neither overlap nor leave a hole: `{"order":"1"}`, `{"group":7}`, `{"hidden":1}` are each refused and each yields `None` from the matching accessor, so nothing is both honoured and refused; `{"group":""}` is neither, which is right because VS Code treats a falsy group as no group.
- **release note overclaimed VS Code placement.** Now carries the bytewise-vs-collation caveat in user-facing terms, matching the architecture doc and the comparator's own comment.

### Findings, all addressed in the follow-up commit

- **Five prose sites still enumerated the pre-fix refusal set**, one of them a doc paragraph this PR itself wrote. The code moved and none of its descriptions did — the same class of defect as the original finding, inverted.
- **`"presentation": false` and `""` newly refused**, contradicting the "requesting nothing" carve-out the field's own doc still promised. A one-member compound carrying either used to launch. Carve-out restored, with a test and a positive control (`true` still refuses, or the test would only prove the function never records anything).
- **The `Eq` comment cited a precedent that does not exist** — it said `ResolvedConfig` drops `Eq`; it derives it, ten lines below, because it carries no float. The reason was sound and the citation was invented.
- **An empty `presentation` was read as "no preference"** while the comment above it stated VS Code sorts an entry that HAS a block ahead of one that has none. VS Code tests the object for truthiness, so `{}` is truthy and does sort ahead. croft now matches; the inner checks stay truthiness-based, which is also VS Code's behaviour.
- Nits: `total_cmp` for a genuine total order (an `Equal` answer for NaN would break the transitivity `sort_by` panics on), a comma join now that three keys can appear, a test name asserting the opposite of its body, and a stale "the remedy is stated with its cost" comment.

### Test coverage — two gaps, both closed

The reviewer verified the existing tests are non-vacuous (the nine-item shuffled fixture pins one exact permutation; the `z, a, m` fixture pins stability), and found two pieces of behaviour that stayed green with the code deleted:

1. **A hidden CONFIGURATION** leaving the picker and the badge — every existing test used a compound.
2. **Row ids surviving a reordering sort** — the trap the module doc names, while the only id test used the hidden filter, which does not reorder.

Both now have tests. The id fixture declares its first row last, so a sorted-index bug cannot hide behind entries that were already in order, and it asserts the reorder actually happened before checking the mapping.

### Verified correct, no action needed

Id mapping traced end to end against the dispatch sites; the `i64 → f64` widening safe with NaN unreachable and no float-equality lint exposure; `Eq` removal breaks no call site (nothing puts these types in a set or map, and the only equality uses are `assert_eq!`); the release gate satisfied (bump to 0.1.940 plus a matching notes file); and the config-lenient/compound-strict split judged sound — a configuration has no refusal path and follows VS Code's tolerance, a compound gates a launch and can afford to.

vitali87:
## 🤖 Fallback review — Claude Opus 5 (delta on `b0a8b40`)

CodeRabbit reviewed `b0a8b40` and then reported "Review limit reached" for the range `ca206a3..b0a8b40`, so the follow-up commit was unreviewed. This pass covers it.

**VERDICT: APPROVE** — diff-quality 4/5, codebase-fit 5/5. No blockers, no majors.

### All eight claims verified against the code

Most usefully, the interaction between the two riskiest changes was traced input by input: `false` → neither parsed nor refused; `""` → neither; `null` → neither; `true` → refused, not parsed; `7` → refused, not parsed; `{}` → parsed, not refused. That set matches `asks_for_something` key for key, so the carve-out and the parser cannot disagree. A malformed block still returns `None` from `parse_presentation`, so it can never become a sortable empty presentation.

The empty-block change was checked against VS Code's actual comparator rather than taken on trust: `getVisibleAndSorted`'s `if (!first.presentation) … return 1` tests the object, where `{}` is truthy, while the inner `group` check is truthiness-based — which is exactly the split croft now implements.

### Both new app tests proven non-vacuous by re-breaking

- Removing the configuration half of the `config_count` filter → the hidden-configuration test fails `left: 2, right: 1`.
- Building row ids from the sorted sequence instead of the unsorted list → the id test fails with `row "0" says "Last …" but index 0 is "First"`.

The reviewer also confirmed the id test's comparison corpus is genuinely unsorted (`discover_configs` preserves file order), so it is not comparing a sorted list against itself.

### Two minors, fixed in the commit above this one

Both comment-only, and both the same defect the parent commit existed to remove:

- `parse_compounds` still said `group` and `order` "want a picker ordering that does not exist" — four lines above the correction saying all three are honoured.
- The refusal comment credited the message with naming the cost of running the member directly; the message does not say it, and a test pins that it never mentions `preLaunchTask`. The comment now records why that omission is deliberate rather than describing a message that cannot exist.

Plus the doc nits: the carve-out list is true of the parent key only (inside the block just `null` is carved out, so `{"order": false}` IS recorded), and "croft delivers that by doing nothing" stopped being true of an empty block once the sort began reading it. Also noted where `total_cmp` and VS Code disagree — `-0.0` before `0.0` versus equal — which is expressible in JSON and depended on by nothing.

### Known and left

`{"presentation": {"hidden": "true"}}` on a CONFIGURATION now sorts to the top where VS Code would treat the truthy string as hidden and drop the row. The `hidden` divergence is pre-existing (`as_bool` → `false`); this change only moves where the row lands, and a configuration has no refusal path to report it through. Worth knowing rather than fixing here.

No test pins the three-key join string; the single-key case is covered and the app-level assertion checks substrings.

Issue #548, opened 2026-09-09
Flaky: two pdf spawning tests fail under load and pass on re-run


vitali87:
Two different tests in `src/pdf.rs` failed once each today and passed on re-run, on an unrelated branch (a debug-picker change that touches no PDF code).

**On CI**, PR #547, job `clippy + tests`:

```
test pdf::tests::a_pdftoppm_whose_descendant_lingers_does_not_hold_the_call ... FAILED
thread ... panicked at src/pdf.rs:787:14:
   3: <core::result::Result<alloc::vec::Vec<u8>, std::io::error::Error>>::expect
test result: FAILED. 4378 passed; 1 failed; 11 ignored
```

That is the `.expect("the page was written; a lingering descendant is not a failure")` at line 786, so `run_pdftoppm` returned `Err` rather than the page.

**Locally**, same branch, `cargo test --bin croft` with `RUST_TEST_THREADS=2`:

```
thread pdf::tests::a_failing_pdftoppm_has_both_stderr_bursts_reported panicked at src/pdf.rs:859:13:
both bursts are reported: /tmp/.tmpXXXXXX/pdftoppm exited with exit status: 1: Syntax Error: first
```

Only the FIRST burst arrived. The immediately following full run of the same suite passed both, as did every other run this session, so both are order/load dependent rather than deterministic.

## What they have in common

Both spawn a real `sh` script and race a deadline from `test_budget::spawn_budget`, and both concern what happens to the pipes when a descendant outlives the direct child. The lingering-descendant test deliberately leaves a background `sleep` holding the inherited stdout/stderr write ends:

```rust
let linger = timeout.as_secs_f64() * 6.0;
let script = fake_pdftoppm(tmp.path(), &format!(
    r#"sleep {linger} & for last; do :; done; printf 'x' > "$last.png""#,
));
```

A pipe reaches EOF only when *every* write end closes, so a reader draining to EOF cannot finish while that `sleep` lives; whether the call returns in time then depends on the deadline winning a race against process startup under load. The stderr-bursts failure is the same shape seen from the other side: a second write that had not arrived by the time the drain concluded.

## Why it matters beyond the noise

A flake here is not only a re-run. These two tests are what pin the guarantee that a hung or forking renderer cannot stall croft's frame loop, so a red run is indistinguishable from that guarantee actually regressing, and the honest response to either is the same: re-run and hope.

## Suggested direction

[...]

PR #550, opened 2026-09-09
ci: make the dependency install prove it worked

### LINKED ISSUE #549: CI: a failed apt-get update silently skips the install and the step still passes
The `Install suite runtime dependencies` step in `clippy + tests` can pass while installing nothing, and the failure surfaces minutes later as eight unrelated-looking test failures.

## What happened

On a run of PR #547 (a debug-picker change that touches no PDF code):

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

The step went GREEN. Eight tests then failed with:

```
these tests need a PDF rasteriser (poppler's pdftoppm, or sips on macOS)
```

`a_pdf_opened_from_the_explorer_pages_with_the_arrow_keys`, `arrow_keys_page_through_an_open_pdf`, `the_wheel_does_not_wrap_a_pdf_at_its_ends`, and five more.

## Why the step passed

The step runs:

```yaml
run: |
  sudo apt-get update && sudo apt-get install -y --no-install-recommends zsh poppler-utils
  for d in /usr/share/zsh /usr/local/share/zsh; do
    if [ -d "$d" ]; then sudo chmod -R go-w "$d"; fi
  done
```

`apt-get update` exited non-zero, so the `&&` short-circuited and `apt-get install` never ran. `set -e` did not fire, because a command on the LEFT of an `&&` list is exempt from it by design — that is what `&&` is for. The step then ended on the `for` loop, which succeeded, so the step reported success with neither `zsh` nor `poppler-utils` installed.

A step that installs nothing is indistinguishable from a step that installs everything, which is the failure shape CLAUDE.md catalogues: a check that cannot fail looks exactly like a check that passed.

## The apt error itself is transient and not ours

The hash mismatch is on Google Chrome's repository, which the runner image ships and croft does not need. It will come and go. The bug worth fixing is that croft's CI cannot tell the difference.

## Suggested fix

Stop letting a partial index failure decide whether the install runs, and assert the tools are actually there afterwards — a control in the same run as the claim:

```yaml
run: |
  set -euo pipefail
  # A partial index failure (the runner image ships third-party repos croft
  # does not need) must not skip the install: apt can still resolve these
  # packages from the indexes it did fetch.
  sudo apt-get update || echo "apt-get update reported errors; continuing with the indexes it has"
  sudo apt-get install -y --no-install-recommends zsh poppler-utils
  # Without this the step can only report that it RAN, not that it worked.
  for tool in pdftoppm zsh; do
    command -v "$tool" >/dev/null || { echo "::error::$tool missing after install"; exit 1; }
  done
```

That way a genuinely failed install fails the step it happened in, naming the missing tool, instead of surfacing as eight confusing test failures in a module the PR never touched.

## Related

#548 is a separate, genuine flake in the same area (two `src/pdf.rs` tests racing a deadline). This issue is not that: here the rasteriser was absent, not slow.

vitali87:
Closes #549.

The `Install suite runtime dependencies` step could pass while installing nothing. On PR #547 — a debug-picker change touching no PDF code — `apt-get update` hit a hash mismatch on a Chrome repo the runner image ships and croft does not need, and eight tests then failed with `these tests need a PDF rasteriser`.

## The cause is a bash rule, not apt

A command on the LEFT of an `&&` list is exempt from `set -e` by design. So `apt-get update && apt-get install …` skipped the install when update failed, the script continued to the chmod loop, and the step ended on a success.

Demonstrated rather than asserted, with a control in the same run:

| Command | Exit |
|---|---|
| `bash -e -c 'false && echo ran; for d in /tmp; do :; done'` | **0** — this step's shape: green, nothing installed |
| `bash -e -c 'false; echo unreachable'` | 1 — control: the same failure DOES trip `set -e` outside an `&&` |
| the new script with a missing tool | 1, naming the tool |

## Two changes

A partial index failure no longer decides whether the install runs — apt resolves both packages from the indexes it did fetch, and a genuinely failed *install* still fails the step. And the step asserts `pdftoppm` and `zsh` exist afterwards, because without that it can only report that it RAN, not that it worked, and those are the same colour.

This is self-testing: `pull_request` runs the workflow from the merge commit, so if the new script is wrong, this PR's own job fails.

## Scope

No version bump — the release gate's shipped-file filter covers `src/`, `assets/`, `build.rs` and the manifests, and this touches only the workflow.

#548 is a separate, genuine flake in the same area (two `src/pdf.rs` tests racing a deadline). This one is not that: here the rasteriser was absent, not slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,7 +207,27 @@ jobs:
         # prompt from wedging every spawned zsh before its first prompt
         # (croft never calls compinit; the runner's own zsh startup does).
         run: |
-          sudo apt-get update && sudo apt-get install -y --no-install-recommends zsh poppler-utils
+          set -euo pipefail
+          # `apt-get update && apt-get install` looked equivalent and was not
+          # (#549). A command on the LEFT of an `&&` is exempt from `set -e`
+          # by design, so when `update` failed on a THIRD-PARTY index the
+          # runner image ships and croft does not need (a Chrome repo hash
+          # mismatch), the install was skipped, this step still went green,
+          # and eight PDF tests failed minutes later with "these tests need a
+          # PDF rasteriser" on a PR that touches no PDF code.
+          #
+          # A partial index failure must not decide whether the install runs:
+          # apt resolves these two packages from the indexes it did fetch.
+          sudo apt-get update \
+            || echo "::warning::apt-get update reported errors; installing from the indexes it has"
+          sudo apt-get install -y --no-install-recommends zsh poppler-utils
+          # Without this the step can only report that it RAN, not that it
+          # worked, and those are the same colour. The suite's own skip
+          # message names pdftoppm, so name it here too.
+          for tool in pdftoppm zsh; do
+            command -v "$tool" >/dev/null \
+              || { echo "::error::$tool is missing after the install step"; exit 1; }
+          done
           for d in /usr/share/zsh /usr/local/share/zsh; do
             if [ -d "$d" ]; then sudo chmod -R go-w "$d"; fi
           done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,16 +218,30 @@ jobs:
           #
           # A partial index failure must not decide whether the install runs:
           # apt resolves these two packages from the indexes it did fetch.
-          sudo apt-get update \
-            || echo "::warning::apt-get update reported errors; installing from the indexes it has"
+          # A transient mirror hiccup is what set #549 off in the first
+          # place, and the install below is what actually gates the step, so
+          # retry before giving up on the index refresh.
+          for attempt in 1 2 3; do
+            sudo apt-get update && break
+            echo "::warning::apt-get update failed (attempt $attempt)"
+            sleep 5
+          done
           sudo apt-get install -y --no-install-recommends zsh poppler-utils
           # Without this the step can only report that it RAN, not that it
-          # worked, and those are the same colour. The suite's own skip
-          # message names pdftoppm, so name it here too.
-          for tool in pdftoppm zsh; do
-            command -v "$tool" >/dev/null \
-              || { echo "::error::$tool is missing after the install step"; exit 1; }
-          done
+          # worked, and those are the same colour.
+          #
+          # Each tool is probed the way ITS CONSUMER resolves it, which is not
+          # the same question for the two of them. `pdf::which` walks PATH, so
+          # `command -v` is right (and stricter: it also wants the executable
+          # bit). The terminal tests take `/bin/zsh` by literal path and SKIP
+          # when it is missing, so a zsh present only elsewhere on PATH would
+          # pass a `command -v` probe and leave those tests silently
+          # uncovered - the same false green this step exists to remove, moved
+          # one level up. It holds today only because usrmerge symlinks /bin.
+          command -v pdftoppm >/dev/null \
+            || { echo "::error::pdftoppm is missing after the install step"; exit 1; }
+          [ -x /bin/zsh ] \
+            || { echo "::error::/bin/zsh is missing after the install step"; exit 1; }
           for d in /usr/share/zsh /usr/local/share/zsh; do
             if [ -d "$d" ]; then sudo chmod -R go-w "$d"; fi
           done


vitali87:

CodeRabbit reported "Review limit reached" on this PR with an empty `reviews[]`, so its green check carries no review behind it. This is the standing fallback.

**VERDICT: APPROVE** — CodeRabbit diff-quality 4/5, Greptile codebase-fit 4/5. No blockers, no majors.

### The bash semantics were verified empirically, not assumed

- `set -e` vs the `||` fallback: the fallback runs and the script continues with status 0, as intended.
- The install still fails the step when it genuinely fails — it is a plain command, not part of an AND-OR list, so `set -e` aborts on it. The property the old code kept is kept.
- `command -v` under `set -e` inside a `for` loop was tested in both directions in one run: a missing tool printed the `::error::` and exited 1 without reaching the line after the loop, and the positive control reached it and exited 0.
- The root cause reproduces: `bash -e -c 'false && echo installed; echo continued'` exits 0 — exactly the green step #549 describes.
- Release-gate exemption confirmed by reading the gate's own filter (`^(src/|assets/|build\.rs$|Cargo\.toml$|Cargo\.lock$)`), not by trusting the claim. A workflow-only diff needs no bump.

### Finding, addressed in the follow-up commit

**The `zsh` probe asked the wrong question.** The terminal tests take `/bin/zsh` by literal path and SKIP when it is missing, while the probe checked `PATH`. A zsh present elsewhere on PATH would pass the probe, skip those tests, and leave CI green with zero coverage of the shim they exist to cover — the same false green this step removes, moved one level up. It holds today only because usrmerge symlinks `/bin`.

The asymmetry is worth recording: `pdf::which` walks PATH, so `command -v` is right there and is in fact stricter than the consumer (it also wants the executable bit); and the PDF gate ASSERTS where the zsh gate skips, so a missing rasteriser turns red on its own while a missing zsh does not. Each tool is now probed the way its consumer resolves it.

Also added a retry around the index refresh: a transient mirror hiccup is what set #549 off, and the install is what actually gates the step, so failing the whole job on the first bad fetch would trade one false green for avoidable red.

### Found outside this diff, worth its own issue

`.github/workflows/release.yml` asserts the artifact kind with a `case` that has no `*)` branch. A `case` with no matching branch exits 0, so for any target outside the four in today's matrix the assertion becomes a silent no-op while the step stays green and its comment stays confident. Verified: `bash -e -c 'case "x86_64-pc-windows-msvc" in *-linux-musl) false ;; esac; echo reached'` prints `reached` and exits 0.

There are no remaining `&&` chains in any `run:` block in `ci.yml` after this PR.

### Nitpicks noted, not actioned here

The probe list is hand-maintained against the package list; several other steps rely on the Actions default shell rather than declaring `set -euo pipefail`; and `sonarcloud.yml`'s `|| true` on the clippy report means an empty report imports silently. None blocks this change.

vitali87:
## 🤖 Fallback review — Claude Opus 5 (delta `1eb506c..eeccfe7`)

The earlier fallback approved `1eb506c` and raised the `/bin/zsh` probe; `eeccfe7` fixes that and adds the retry, so this pass reviews code the first review never saw. CodeRabbit remains rate-limited on this PR.

**VERDICT: APPROVE** — diff-quality 5/5, codebase-fit 5/5. No blockers, no majors.

### The two claims, verified by running them

**The probe now asks the consumer's question.** `zsh_or_skip` hardcodes `/bin/zsh` and gates eight tests; `[ -x … ]` was checked against a synthetic usrmerge tree (`bin -> usr/bin`) and the real `/bin`: it resolves through the symlink, and correctly fails for both an absent file and one without the exec bit. The counterpart holds too — `pdf::which` walks `$PATH` and tests `is_file()`, so `command -v` is the matching probe and is, as the comment claims, literally stricter than the consumer.

**The retry loop neither swallows a failure nor aborts early** — the thing that would have reintroduced this PR's own bug. Simulated with `false` in place of both commands under `set -euo pipefail`:

- all three attempts failing → reaches the install (the failing `update` is exempt as the left operand of `&&`, and the body ends in `sleep`, so the loop exits 0);
- `break` fires correctly on a first- or second-attempt success;
- a genuinely failing `apt-get install` → exit 1, with the following line never reached.

### Nits, not actioned

The final iteration's `sleep 5` is dead time (a fully failed refresh burns 15s, not 10s), and the job has no `timeout-minutes`, so three attempts against an unreachable mirror can exceed that in apt's own connection timeouts — a slow job rather than a wrong result, and the shape it already had. `pdfinfo` and `pdftohtml` are also consumed but unprobed; all three ship in `poppler-utils`, so one probe is a valid proxy, and that is unchanged from the previous head.

PR #552, opened 2026-09-09
fix(pdf): scale the stderr settle window with the load the tests run under

### LINKED ISSUE #548: Flaky: two pdf spawning tests fail under load and pass on re-run

vitali87:
Closes #548.

Two tests in `src/pdf.rs` failed once each on branches touching no PDF code, and could not be reproduced on demand. This is the diagnosis, arrived at by measurement, and the fix the measurement supports.

## What is actually wrong

`settle_stderr` returns once the buffer has been quiet for twelve ticks (60 ms). Whether the renderer's SECOND write lands inside that window is a race against **scheduling**, not against the renderer: the bytes can be sitting in the pipe while the reader thread is descheduled.

Measured on this box, five samples idle and five under twelve busy loops on four cores:

| | settle time | bytes held |
|---|---|---|
| idle | 89–116 ms | 40 (both writes) |
| loaded | 69 ms | **19 (the first write alone)** |

Nineteen bytes is `"Syntax Error: first"` by itself. The window closed at 69 ms with the second write still in flight.

## The proof

Controlled, same load, eight attempts each way:

| quiet run | result |
|---|---|
| scaled with load | **0 / 8 failed** |
| production (12 ticks) | **4 / 8 failed** |

Every failure carried the same fingerprint: `settle ended on the QUIET RUN after 69–72 ms holding 19 byte(s)`.

## The change

The window scales in TESTS the way every other spawned-process wait here scales, through `spawn_budget`. **Production keeps its twelve ticks exactly** — `quiet_run_ticks` returns the constant unless a test on this thread has widened it, and the override is `cfg(test)` and thread-local, so parallel tests cannot read each other's.

The instrumentation stays, and it is what made the diagnosis possible. Both failing assertions now report which exit the poll took and what it held, so a future occurrence says whether the window closed early or the bytes never came — two causes needing opposite fixes, which the old message could not tell apart. The lingering-descendant test names its error for the same reason: it failed once on CI saying only that it should not have.

## How this nearly went wrong

The first attempt was this same widening **without** the evidence. It would also have made the flake rarer, suppressing the signal before it had been read — so it was reverted, and only instrumentation shipped, until the numbers arrived.

Before that: six consecutive full-suite runs reproduced nothing, and a twelve-hog load test reproduced nothing. What finally exposed it was measuring the **passing** case, which showed the margin was far thinner than it looked.

## Scope

Bumped to 0.1.941 with notes: the release gate counts this as shipped because `quiet_run_ticks` lives outside the test module, even though production behaviour is byte-for-byte identical. The note says exactly that.

Full suite 4384 passed, clippy clean, fmt clean, doc-ownership clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when collecting PDF renderer error output, helping preserve error details during brief delays or bursts of diagnostic messages.

* **Compatibility**
  * PDF rendering behavior and renderer error reporting remain unchanged in this release.

* **Release**
  * Updated the package version to 0.1.941.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.940"
+version = "0.1.941"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.940"
+version = "0.1.941"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/src/pdf.rs
+++ b/src/pdf.rs
@@ -442,6 +442,89 @@ const MAX_PROBE_STDOUT: u64 = 1 << 20;
 /// pays the whole grace once, on a path that has already failed.
 const STDERR_SETTLE_GRACE: std::time::Duration = std::time::Duration::from_millis(200);
 
+/// How the settle poll ended, recorded under `cfg(test)` only (#548).
+///
+/// Two tests here have failed once each on a loaded machine and could not be
+/// reproduced in six consecutive full-suite runs, so the next occurrence is
+/// the only evidence available. Which of the two exits it took is what
+/// decides the diagnosis, and the failure message could not say:
+///
+/// - ended on the QUIET RUN, missing bytes: the window closed while the
+///   reader thread was descheduled and the bytes were still in the pipe, so
+///   the run is racing scheduling and widening it under load is the fix;
+/// - ended on the DEADLINE: the renderer's second write never arrived at all
+///   within the grace, which is a different bug and a different fix.
+///
+/// Guessing between those two is what this exists to stop. Deliberately NOT
+/// paired with a wider window: a change that makes the flake rarer would
+/// suppress the signal before it has been read.
+#[cfg(test)]
+#[derive(Debug, Clone, Copy)]
+struct SettleOutcome {
+    waited: std::time::Duration,
+    on_quiet: bool,
+    bytes: usize,
+}
+
+#[cfg(test)]
+thread_local! {
+    /// Per THREAD, so tests running in parallel cannot read each other's.
+    static LAST_SETTLE: std::cell::Cell<Option<SettleOutcome>> =
+        const { std::cell::Cell::new(None) };
+}
+
+/// The production quiet run: twelve ticks, 60 ms.
+const QUIET_RUN_TICKS: u32 = 12;
+
+#[cfg(test)]
+thread_local! {
+    /// A test's own quiet run, per THREAD so two tests in parallel cannot
+    /// read each other's (an env var would not have that property).
+    static QUIET_RUN_OVERRIDE: std::cell::Cell<Option<u32>> =
+        const { std::cell::Cell::new(None) };
+}
+
+/// The quiet run this call should use (#548).
+///
+/// A parameter and not a constant because the run is a wall-clock race
+/// against SCHEDULING, and the measurement that settled this issue shows the
+/// margin is thin: on an idle box the poll settles at 89-116 ms with both of
+/// the renderer's writes, while under load it was caught settling at 69 ms
+/// holding only the FIRST - the window closed while the second was still in
+/// flight. Production keeps the twelve ticks it has always had; only a test,
+/// which runs on a machine deliberately saturated with parallel jobs, widens
+/// it.
+fn quiet_run_ticks() -> u32 {
+    #[cfg(test)]
+    if let Some(ticks) = QUIET_RUN_OVERRIDE.with(std::cell::Cell::get) {
+        return ticks;
+    }
+    QUIET_RUN_TICKS
+}
+
+#[cfg(test)]
+fn record_settle(outcome: SettleOutcome) {
+    LAST_SETTLE.with(|c| c.set(Some(outcome)));
+}
+
+/// How the settle poll on this thread last ended, for a failure message.
+#[cfg(test)]
+fn last_settle() -> String {
+    match LAST_SETTLE.with(std::cell::Cell::get) {
+        Some(o) => format!(
+            "settle ended on {} after {:?} holding {} byte(s)",
+            if o.on_quiet {
+                "the QUIET RUN"
+            } else {
+                "the DEADLINE"
+            },
+            o.waited,
+            o.bytes
+        ),
+        None => String::from("settle did not run"),
+    }
+}
+
 /// Test-only: the next `rasterize_page` call for exactly this (path, page)
 /// fails once, simulating a transient rasteriser failure (a spawn refused
 /// under load, an OOM-killed child). Keyed on the full path so parallel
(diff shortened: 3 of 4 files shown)


vitali87:

CodeRabbit reported "Review limit reached" on this PR with an empty `reviews[]`, so its green check carries no review. This is the standing fallback.

**VERDICT: REQUEST_CHANGES** — diff-quality 3/5, codebase-fit 4/5. No blockers, three majors. All are addressed in the push above; the summary is here because the finding is worth recording.

### The central finding: the widening was clamped away, and the tests stopped protecting the constant

`quiet_run` is capped at `ticks_in_grace - 1` — 39 for the 200 ms grace — while `spawn_budget` always produces 48–96 ticks. So `min(…)` yielded a constant **39 on every machine, load or none**: not a scaled window at all, but exactly the fixed constant CONTRIBUTING warns against. Measured by the reviewer with a verbatim copy of the loop:

```
production (12)    exit=QUIET_RUN   84.0 / 82.1 / 82.8 ms
widened x4 (48)    exit=DEADLINE   201.9 / 204.2 / 202.9 ms   effective run 39
widened x8 (96)    exit=DEADLINE   202.4 / 200.5 / 204.0 ms   effective run 39
```

Both tests could then only ever exit on the deadline, and — the real cost — the production constant became untestable. Measured, with a known-positive control in the same run:

```
QUIET_RUN_TICKS cut 12 -> 6, unwidened   both bursts=FALSE   <- control: detectable
QUIET_RUN_TICKS cut 12 -> 6, widened     both bursts=true    <- masked
```

That is the coverage question answered: yes, real coverage was traded away, and the trade was invisible from the diff.

### What replaced it

The rule moved out of the clock. `settle_over` takes its readings from the caller and decides when the buffer has stopped growing; production supplies a sampler that polls and sleeps a tick, the test supplies a scripted one. The flake is gone because the race is gone, not because the window is more patient. The thread-local override, its guard and the scaling arithmetic are all deleted.

The new test pins the constant in both directions — a 10-tick gap must be bridged, a 15-tick gap must not — and the gaps are **absolute on purpose**. Writing them as `QUIET_RUN_TICKS ± 2` reads better and pins nothing, since the fixture then moves with the constant. A re-break caught precisely that: with relative gaps, halving the constant left the test green. With absolute ones:

```
constant halved (12 -> 6)          FAILED
constant cut to 1                  FAILED
quiet exit disabled entirely       FAILED
restored                           passes
```

### The other two majors

- **A doc comment in the diff asserted the opposite of the diff** — "deliberately NOT paired with a wider window", in the commit that paired it with one. A survivor of the instrumentation-only revision. Rewritten.
- **The DEADLINE interpretation guide had been inverted** by the change, and the stale "the cap only binds for a grace under 65 ms" parenthetical no longer held. Both corrected.

### The framing, which was wrong

The reviewer's sharpest point is not about the code. The measurement — production settling at 69 ms holding only the first write — is a statement about a **shipped build**, not about the tests: on a loaded machine a user whose render fails gets half the renderer's complaint, and poppler's useful detail is usually in the later lines. This PR treated that as a test problem. It is now #553, with the numbers and three options, for a decision rather than a guess.

### Verified, no action needed

Thread-local soundness was checked against libtest's source (fresh thread per test regardless of `--test-threads`; `catch_unwind` runs the guard's `Drop`; the `WouldBlock` path shares a thread but still restores) — moot now that the override is gone. The production path is unchanged, confirmed by a non-test `cargo check` with zero warnings. The release gate genuinely requires the bump, since the helpers live outside `mod tests`, and the note is honest about the binary being unchanged.

vitali87:
## 🤖 Fallback review — Claude Opus 5 (rounds 2–4)

CodeRabbit is rate-limited on this PR, so the fallback is the review of record. It took four rounds, and every round found something real.

**Final: VERDICT: APPROVE** — diff-quality 4/5, codebase-fit 5/5, with the three remaining low findings fixed after it and confirmed.

### Round 2 — REQUEST_CHANGES: the fix worked for the wrong reason

The scaled window was clamped away. `quiet_run` is capped at `ticks_in_grace - 1` = 39, while `spawn_budget` always produced 48–96, so the effective run was a constant 39 on every machine — the fixed constant CONTRIBUTING warns against. Both tests could then only exit on the deadline, and the production constant became untestable: halving it left every test green, verified with a control showing the unwidened form does catch it.

That answered the coverage question I had asked: yes, real coverage had been traded away, invisibly from the diff.

### Round 3 — REQUEST_CHANGES: the rework was 5 ms slower than what it replaced

The rule moved out of the clock, which was right. But the sampler slept before the decision, so every quiet exit paid a tick it could not use and returned bytes read 5 ms earlier — measured at 13 polls and 13 sleeps against the original's 13 and 12. The PR's central claim was behavioural identity, so this mattered more than its size.

### Round 4 — APPROVE, with parity verified properly

Tick parity re-checked under a virtual clock across twelve cases, including `grace = 0` and a deadline already passed, with a control confirming the pre-fix form differs. Identical on every path. The constant's pin was re-broken in both directions independently (6 fails one assertion, 24 fails the other), and the measured tolerance — a run of 10 through 14 — matches what the test's comment claims.

Three low findings after it, all fixed:

- the documented grace threshold answered a different question (65 ms was the old clamp's binding point);
- a dead-code justification that this toolchain contradicts;
- `on_quiet` labelling the boundary tick — where the quiet run and the deadline coincide — as a quiet exit, mislabelling precisely the distinction the record exists to make.

### The tail worth recording

Fixing those last two together produced one more error: tightening `on_quiet` moved the very threshold the other fix documents, by one tick, so the corrected numbers were stale in the commit that corrected them. Re-measured under the new formula across every grace from 0 to 300: 60 ms, and 55 ms for an empty buffer. The behavioural reading agrees — a 12-tick run takes 60 ms of wall time, so it cannot fit inside a 56 ms grace whatever the flag says.

Confirmed after that: `on_quiet`'s change is label-only (zero differing cases in returned text or exit tick across every grace and three buffer bodies, with a control proving the flag itself does differ somewhere), no assertion depended on the old labelling, and both nits are done.

PR #554, opened 2026-09-09
ci: fail the artifact-kind check for a target it cannot assert

### LINKED ISSUE #551: release.yml: the artifact-kind assertion is a no-op for any target outside the current matrix
The `case` that verifies a release artifact is the right kind of binary has no `*)` branch, so it asserts nothing for any target it does not name — while the step stays green and its comment stays confident.

```yaml
case "${{ matrix.target }}" in
  *-linux-musl) file "$bin" | grep -q 'ELF 64-bit LSB' ;;
  *-apple-darwin) file "$bin" | grep -q 'Mach-O 64-bit' ;;
esac
```

A `case` with no matching branch exits 0:

```
$ bash -e -c 'case "x86_64-pc-windows-msvc" in *-linux-musl) false ;; *-apple-darwin) false ;; esac; echo reached'; echo "exit=$?"
reached
exit=0
```

For the four targets in today’s matrix the check does what its comment claims — it catches "a cross build that silently produced a host binary". The moment a fifth is added (a gnu triple, a Windows triple), the assertion becomes a silent no-op and the release ships whatever the build produced, with the step reporting success.

This is the same shape as #549: a check that cannot fail is indistinguishable from a check that passed. It is not currently broken, which is exactly why it is worth fixing before it is.

## Fix

Make the unhandled case loud rather than absent:

```yaml
  *) echo "::error::no artifact-kind assertion for ${{ matrix.target }}"; exit 1 ;;
```

Adding a target then requires adding its assertion, which is the intent the comment already states.

Found while reviewing #550.

vitali87:
Closes #551.

The `case` that verifies a release artifact is the right kind of binary has no `*)` arm, and a `case` with no matching branch exits 0. So for any target outside the four in today's matrix, the assertion is a silent no-op — green step, unchecked artifact, and a comment above it still claiming it catches "a cross build that silently produced a host binary".

Not broken today. Guaranteed to break silently the day a fifth target is added, which is exactly when nobody is looking at this line.

## Verified, all three paths

| Case | Result |
|---|---|
| unmatched target, no default arm | **exit 0** — assertion skipped |
| unmatched target, with the new arm | exit 1, `::error::no artifact-kind assertion for <target>` |
| matched target, with the new arm | passes as before |

The four current targets are unaffected.

Also added `set -euo pipefail`: the step relied on the Actions default shell, and `file "$bin" | grep -q` would not have reported a failing `file`.

## Scope

No version bump — the release gate's shipped-file filter covers `src/`, `assets/`, `build.rs` and the manifests, and this touches only a workflow. Same class as #549/#550, found while reviewing that one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Release builds now fail with a clear error when an unsupported target is encountered, preventing invalid binary-type checks from passing silently.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,12 +101,18 @@ jobs:
         # this catches: `file` names the architecture, so a mismatch here is
         # a wrong artifact rather than a missing one.
         run: |
+          set -euo pipefail
           bin="target/${{ matrix.target }}/release/croft"
           test -x "$bin"
           file "$bin"
           case "${{ matrix.target }}" in
             *-linux-musl) file "$bin" | grep -q 'ELF 64-bit LSB' ;;
             *-apple-darwin) file "$bin" | grep -q 'Mach-O 64-bit' ;;
+            # A `case` with no match exits 0, so without this arm the whole
+            # assertion becomes a silent no-op for any target the list does
+            # not name - green step, unchecked artifact, confident comment
+            # above it (#551). Adding a target now has to add its assertion.
+            *) echo "::error::no artifact-kind assertion for ${{ matrix.target }}"; exit 1 ;;
           esac
 
       - name: Package
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,20 +98,27 @@ jobs:
 
       - name: The artifact must exist and be the right kind of binary
         # A cross build that silently produced a host binary is the failure
-        # this catches: `file` names the architecture, so a mismatch here is
-        # a wrong artifact rather than a missing one.
+        # this catches, and catching it means reading the ARCHITECTURE `file`
+        # names, not only the container format. Matching `ELF 64-bit LSB` on
+        # both musl legs let an x86-64 binary pass as aarch64 - the comment
+        # claimed an assertion the check did not make, which is the same
+        # false green as the missing arm below (#551).
         run: |
           set -euo pipefail
           bin="target/${{ matrix.target }}/release/croft"
-          test -x "$bin"
+          test -x "$bin" || { echo "::error::no artifact at $bin"; exit 1; }
           file "$bin"
+          # Exact triples, not suffixes: a typo in a matrix target then reaches
+          # the catch-all rather than silently matching `*-linux-musl`.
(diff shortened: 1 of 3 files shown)


vitali87:

CodeRabbit did review this one (naming the exact head, "no actionable comments"), so this is a second opinion rather than a gap-filler. It found something CodeRabbit did not.

**VERDICT: APPROVE** — diff-quality 5/5, codebase-fit 5/5. No blockers, no majors against the diff itself.

### The finding worth having: the check was vacuous in a second way

The step's comment claimed it catches "a cross build that silently produced a host binary". It did not. Both Linux legs grepped the same string and both Apple legs grepped the same string, so an x86-64 binary passed the aarch64 leg. The missing catch-all arm was one hole; the arch-blind assertion was another, in the same `case`, under a comment asserting otherwise.

Verified: an x86-64 binary matches the x86-64 arm and is rejected by the aarch64 one, which is the discrimination that was absent. Exact triples now, so a typo in a matrix target reaches the catch-all instead of silently matching a suffix.

### What that fix then broke, and what caught it

Naming every triple in a `case` arm made each appear twice in the workflow — and the binstall contract checker asked whether each target appears ANYWHERE in the file. So dropping a target from the build matrix would have left that check green while no release carried the platform. Its own test caught it immediately.

The checker now reads the matrix block rather than the whole file, and a new test pins the exact trap: drop a target from the matrix, leave its name in a `case` arm, and the checker must still object. Re-broken against the old whole-file search to confirm it bites — two failures with the old code, all 145 script tests passing without it.

### Verified, no action needed

Every target the matrix produces still lands on an intended arm (none reaches the catch-all), and the catch-all fires for genuinely unhandled triples. `set -euo pipefail` changes nothing for the existing lines: `-e` was already in force via the Actions default, `file` exits 0 even on a missing path, `test -x` still fails first, and the `grep -q` SIGPIPE hazard is unreachable at this output size (2000 iterations, zero failures). No release leg that previously passed correctly now fails.

### Two more of the same shape, filed separately

The everyday build has the identical arch-blind check, and a release can publish an archive whose checksum line never arrived while the run stays green. Both filed with reproductions rather than folded in here.

PR #555, opened 2026-09-09
feat(debug): run a compound's configurations as several sessions

### LINKED ISSUE #310: Debug compounds: run several configurations in one session set (stop-all semantics)

vitali87:
Closes #310.

croft was single-session by construction — one session or none, with about forty places written against that. A compound needs several, so the model becomes a set with one **focused** session, and every one of those places now says which it means.

## What reads which

| Reads the FOCUSED session | Means ALL sessions |
|---|---|
| call stack, variables, watch | polling the adapters |
| stepping, F6, breakpoint dispatch | stopping |
| the status line | |

## The asymmetry that matters

A configuration launch **replaces** the set; a compound member **joins** it. The ordinary launch path opens by stopping whatever is running — correct for a configuration the user picked, fatal for a compound's second member — so a loop over it would leave only the last member alive. That is why compounds get their own launch path rather than a loop.

## Behaviour

- Members start in **declaration order**, which is what the file asked for and what a client-then-server compound depends on.
- **Stop means stop-all.** Disconnecting only the focused session would leave siblings running with nothing owning them.
- **A failing member does not take its siblings down.** They are real processes; killing them because a later member could not resolve would turn one failure into two. The error names the member, the survivors stay, and the status says how many of how many started.
- **Focus lands on the first member**, since compounds are written server-first.
- **Sessions carry their names**, because with several running the user has to know which are live and which one the stack belongs to.

## Tests

The set is generic over its session type **only** so its rules can be tested without an adapter: a replace clears while a push joins; focus always names a session that exists, and clearing takes the index with it; a stop reaches every session rather than the focused one.

Two existing guards **inverted rather than deleted** — both asserted a multi-member compound is refused and cites this issue. It launches now, so they assert it does not, and the one that mattered still pins that *every* member starts rather than a subset.

## Known limitation, stated not hidden

A member's own `preLaunchTask` is skipped, with a status saying so. Parking is a one-at-a-time mechanism and a set of parked members would interleave unpredictably. The compound's **own** task still runs, and parks the whole member list.

Full suite 4387 passed, clippy clean, fmt clean, doc and architecture checks clean.

## Summary by CodeRabbit

* **New Features**
  * Compound debug configurations now launch multiple sessions in the listed order.
  * Sessions can be focused and managed independently while running concurrently.
  * A member ending or failing to start no longer stops its sibling sessions.
  * Stopping a compound stops all active sessions.

* **Bug Fixes**
  * Compounds with unsupported multi-session `stopAll` settings now clearly report the configuration issue.
  * Members with their own `preLaunchTask` are identified and skipped during compound launches.

* **Documentation**
  * Updated architecture documentation to describe multi-session debugging behavior.


vitali87:
@copilot resolve the merge conflicts in this pull request

vitali87:
Worked through all 11 threads. Four fixed, five are real and filed as follow-ups rather than fixed here, two need no change. Detail below; head is `0e3b6ee`.

## Fixed in this PR

**1. `launch_resolved_into_set` returns `()`, so `Launched` was unconditional** (`mod.rs:20925`) — correct, and it was the most consequential of the eleven. CI proved it in its starkest form: on a runner with no debug adapter the compound reported **"2 session(s), showing ?" over an entirely empty set**. `MemberOutcome::Failed` already existed for exactly this; the launcher now asks the set whether a session appeared:

```rust
let before = self.debug_sessions.len();
self.launch_resolved_into_set(rc);
if self.debug_sessions.len() > before { MemberOutcome::Launched } else { MemberOutcome::Failed }
```

**2. The `started no sessions` disjunction** (`tests.rs:35944`, and the outdated thread on the second guard) — right, and worse than stated: the left arm is satisfied by the zero-session early return, so *neither* assertion constrained the outcome. Compounded by a test comment asserting a false premise ("no adapter is installed, so every member fails to start") — that is true on CI and false on a dev machine, which is why these tests passed locally and failed on CI once I tightened them. They now assert the invariant rather than a host-specific count: a session exists for each one the report claims, and no session is described that cannot be named.

**3. The feedback line was wiped before it was drawn** (found locally, related to your `mod.rs:20992` thread) — `launch_compound_members` built the report and *then* called `reveal_debug_view`, which runs `refresh_run_debug`, which clears `feedback` unconditionally. The only place a skipped member was ever named was assigned and destroyed in the same function. Reveal first, then write.

**4. `stopAll: true` silently ignored** (`mod.rs:24713`) — confirmed: `stopAll` appears nowhere in `src/` outside comments and test fixtures. ARCHITECTURE.md claimed "a compound stating either value gets what it asked for or the default it would have had", which is true for `false`/omitted and false for `true`. The doc now says plainly that `stopAll: true` is not honoured and not refused, and names it as the follow-up rather than part of #310. I did not implement the teardown here — it is a behaviour change to session lifetime, not a fix to this one.

## Real, but follow-ups rather than scope creep

These are correct findings. They need state this PR does not introduce, and bundling them would make an already-large change larger:

- **`Stopped` from a background member is discarded** (`mod.rs:20284`) and **background `Output` never reaches the console** — compounded by there being no way to change focus: `debug_sessions.focus` has exactly one production call site, in `launch_compound_members`. A member that hits a breakpoint is currently unreachable and looks hung. The honest fix is a session switcher, which is its own change.
- **A background member's removal does not refresh the report** (`mod.rs:20292`) — the status and feedback keep naming a session that has ended. Rebuilding that line needs the compound's name retained past launch, which is new state.
- **Status after the focused session is retired no longer names the newly focused one** (`mod.rs:20420`) — same missing state.
- **A stale parked `preLaunchTask` can replace a live compound** (`mod.rs:20984`) — real, and pre-existing for single configs too.
- **The compound is never recorded as the selected debug target** (`mod.rs:24723`), so Restart and the next F5 use the previous configuration.

I will file these against #310 as a follow-up issue rather than leave them in review threads.

## No change

- **`mod.rs:24659`, the error formatter recommending `members[0]`** — it now reads "run its configurations directly" without naming a member, so the specific advice you quote is not in the current text.

Two of your findings overlapped with an independent local review, which is a useful signal on both: the `started no sessions` assertion and the discarded failed-member identity were found twice, separately.

vitali87:
Correction to my previous comment, on the one finding I wrote off.

I said the `mod.rs:24659` thread needed no change because the error text "now reads 'run its configurations directly' without naming a member". That was wrong — I described the message from memory instead of reading it. It hardcodes the first member twice:

```rust
"... fix {} to launch \"{}\" through the compound, or run \"{}\" directly",
plural, members[0].name, members[0].name
```

So for a multi-member compound the remedy did name only the first member, exactly as the review said, and the advice was to launch a subset — the outcome this arm exists to refuse. Fixed in `adec17d`: the message now names every member in order, and the presentation test gained a multi-member case, which had no coverage at all.

That makes it **five** findings fixed, not four. The follow-up list is unchanged.

Also worth recording, since it caused two red CI runs on this PR: three of these tests were asserting the **host** rather than the code. `debug_sessions.names() == ["Server", "Client"]` passes on a machine with a debug adapter and fails on a CI runner without one, and the original test comment asserted the false premise that "no adapter is installed in a test environment, so every member fails to start" — true on CI, false locally. The tests now pin the invariant that was actually broken (a session exists for each one the report claims; no member is described that cannot be named), and I verified it by forcing every member to fail locally: all 20 compound tests pass with zero sessions started, which is the control the earlier versions never had.

vitali87:
Follow-ups filed as #567 — the six deferred findings from this review, with the mechanism and why each is out of scope here. Items 1 (a breakpoint in a background member looks like a hang) and 6 (F5 after the sessions end launches the previous config) are the two a user hits first.

vitali87:
Full accounting of all 12 findings, since two arrived from the re-review after my first pass and my earlier comment predates them. Head is `8f2409e`.

### Fixed and resolved (6)

| Finding | What changed |
|---|---|
| Failed launches reported as successful | `MemberOutcome` now comes from whether the session set actually grew, not from a call that returns `()` |
| `stopAll` ignored for multi-member compounds | Recorded in `unsupported_keys` at `true` with 2+ members, so the compound is refused with the key named |
| Explicit `stopAll` setting incorrectly ignored *(New)* | Same fix — your point that "documenting the mismatch does not prevent the incorrect launch" is the reason it became a refusal rather than a doc note |
| Multi-member error advice recommends one member | The remedy names every member in order; the presentation test gained the multi-member arity, which had no coverage |
| Test accepts compounds that launch no sessions | All three compound tests now assert the invariant, not a host-specific count |
| Test documentation split across unrelated tests *(New)* | Correct — my insertions severed the doc for `a_single_member_compound_launches...` mid-sentence. Reunited. The repo's doc-ownership gate passes either way; it catches a stolen or lost comment, not a severed one, so CI could not have found this |

Also fixed, found locally rather than in a thread: the feedback line was assigned and then wiped by `reveal_debug_view` → `refresh_run_debug`, so the only place a skipped member was ever named never reached the screen.

### Left open, deliberately, and tracked in #567 (6)

Discarded background `Stopped` events · background sessions cannot be inspected or resumed · removed sessions leave stale status and feedback · focus replacement omits the newly focused session · stale pending launch can replace a new compound · compound selection not persisted for restart/F5.

These are real. Each needs either a session switcher (UI) or the compound's identity retained past launch (new state), and neither belongs in a PR this size. I have left the threads open rather than resolving them, so they stay visible against the PR as well as in #567.

### On verification

The `stopAll` fix carries its own control: the one-member case must stay silent, or the rule is one that fires everywhere. Both mutations redden the test — dropping the arity guard fails the one-member arm, and never recording the key reproduces the reported bug exactly as `left: []`.

The test-host problem was the expensive one and caused two red CI runs. Three tests were asserting the runner's capabilities rather than the code: `debug_sessions.names() == ["Server", "Client"]` passes where a debug adapter exists and fails on CI where none does. Verified the fix by forcing every member to fail locally — all 20 compound tests pass with zero sessions started, which is the control the earlier versions never had.

PR #558, opened 2026-09-09
fix(pdf): take the renderer's own output before asking any timing question

### LINKED ISSUE #553: A failing PDF render can report half its error on a loaded machine
Split out of #548, where it surfaced as a test flake and was initially mistaken for one.

## The measurement

`settle_stderr` returns once the renderer's stderr buffer has been quiet for twelve ticks (60 ms), capped by a 200 ms grace. Whether a renderer's SECOND write lands inside that window is a race against scheduling: the bytes can be sitting in the pipe while the reader thread is descheduled.

Measured on a 4-core Linux box, five samples idle and five under twelve busy loops:

| | settle time | bytes held |
|---|---|---|
| idle | 89-116 ms | 40 (both writes) |
| loaded | 69 ms | **19 (the first write alone)** |

Nineteen bytes is `Syntax Error: first` by itself. Under repeated load, 4 of 8 attempts settled early holding only the first write.

## Why this is a production issue and not a test issue

#548 read that measurement as "the test is too impatient" and made the test tolerate it. That was the wrong reading, and it is worth stating plainly: the numbers describe what `settle_stderr` does in a SHIPPED build. On a machine under load, a user whose render fails gets the first line of the renderer's complaint and not the rest — and poppler's useful detail is frequently in the later lines, since the first is often a generic `Syntax Error` while the specific cause follows.

The failure is silent by construction: the error looks complete, and nothing distinguishes "that was all it said" from "the window closed early".

## What is not in question

The 60 ms run is not obviously wrong — it was chosen to bridge the gaps measured between a renderer's own writes, and croft cannot wait on EOF here, because a descendant the renderer forked can hold the pipe open indefinitely (that constraint is documented in `run_pdftoppm` and is not up for revisiting).

## Options

1. **Scale the quiet run by observed load** in production, the way the test helper does for spawn waits. Bounded by the existing 200 ms grace, and paid only on a render that has already failed, so the cost is an error message arriving up to ~140 ms later.
2. **Wait for the direct child's exit before starting the settle**, rather than starting from the exit observation, so the window covers the writes the renderer itself has already made.
3. **Accept it and say so**: document that a loaded machine may truncate a failing renderer's stderr, so the next reader of that code does not have to re-derive it.

Option 1 is cheapest and matches an existing convention. Worth a decision before anything is written, since all three are defensible and only one needs doing.

## Note

The instrumentation added in #548 (`SettleOutcome`, recording which exit the poll took and what it held) is what produced these numbers and is already on `main`, so a future occurrence reports its own cause.

vitali87:
Closes #553.

Taking the decided route — change **when** we start listening, rather than waiting longer.

## What was wrong

A reader thread published the renderer's stderr chunk by chunk. It was right about the constraint: the pipe reaches end of file only when *every* holder of its write end is gone, and a descendant the renderer forked can hold it indefinitely. But it made the answer depend on that thread being **scheduled**.

Measured on a loaded machine: the renderer's second line was still in the pipe when the poll gave up at 69 ms. The user saw a generic `Syntax Error` with the specific cause missing — and nothing to indicate anything had been cut.

## What it is now

| | Before | After |
|---|---|---|
| Who reads the pipe | a background thread | the loop that waits for the exit |
| Mode | blocking | non-blocking |
| After the child exits | poll a shared buffer for quiet | drain until `WouldBlock` |
| Depends on scheduling | yes | **no** |

Everything the renderer itself wrote is in the pipe by the time it exits, so draining until `WouldBlock` takes all of it. `WouldBlock` means *nothing more available*, not *nothing more coming* — the distinction a blocking read cannot make, and the reason end-of-file was never usable here.

The settle poll survives only for a **descendant's** late writes, where it is optional generosity rather than the thing standing between the user and half an error.

## Tests

A new one pins the user's actual case: a renderer that prints a generic line, pauses, prints the cause, then exits. Both lines reported, five times over, under twelve busy loops on four cores.

Two wall-clock tests of the settle poll moved onto the rule itself, in ticks — the rule is what they were about, and they no longer race a clock to assert it.

Full suite 4384 passed, clippy clean; the PDF module passes 30/30 under deliberate load.

## Note

This deletes the reader thread and its lock rather than adding machinery.

## Summary by CodeRabbit

- **Bug Fixes**
  - PDF rendering errors now include the renderer’s complete error output, making failures easier to diagnose.
  - Python Go to Definition remains responsive when whole-project diagnostics are slow or unresponsive, while navigation, hover, and completion continue to work.

- **Documentation**
  - Updated PDF renderer documentation to reflect improved error-output handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,7 +607,7 @@ dependencies = [
 
 [[package]]
 name = "croft-software"
-version = "0.1.941"
+version = "0.1.942"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "croft-software"
-version = "0.1.941"
+version = "0.1.942"
 edition = "2024"
 description = "VSCode-style TUI written in Rust"
 license = "MIT"
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -693,6 +693,10 @@ PDF rasteriser. It prefers `pdftoppm` (poppler) and falls back to macOS `sips`.
 
 **Where links go.** External URLs open with the OS opener behind an http/https/mailto allowlist. A document-supplied `file://` or custom scheme is a one-click app launch and is dropped. Internal GoTo targets flip the page.
 
+**The renderer's stderr is read from one place, without blocking (#553).** A reader thread published it chunk by chunk, which was right about the constraint — the pipe only reaches end of file once every holder of its write end is gone, and a descendant the renderer forked can hold it long after the renderer exited — but it made the answer depend on that thread being SCHEDULED. Measured on a loaded machine, the renderer's second line was still in the pipe when the poll gave up, so a failing render reported its generic first line and dropped the specific cause.
+
+The pipe is non-blocking now and drained by the same loop that waits for the exit. Everything the renderer itself wrote is in the pipe by the time it exits, so a drain-until-`WouldBlock` after the exit takes all of it with no window to lose. `WouldBlock` means "nothing more available", not "nothing more coming" — the distinction a blocking read cannot make, and the reason end-of-file was never usable here. The settle poll survives only for a DESCENDANT's late writes, which is generosity rather than the thing standing between the user and half an error.
+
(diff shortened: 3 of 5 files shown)


vitali87:
@copilot resolve the merge conflicts in this pull request

PR #559, opened 2026-09-13
feat(lua): add tree-sitter syntax highlighting and LSP support

ironwren:
Adds comprehensive Lua language support to Croft:
- **Tree-sitter Grammar**: Adds `tree-sitter-lua` (v0.5) dependency.
- **Syntax Highlighting & Outline**: Configures highlighting queries and symbol extraction for functions, table methods, and colon-notation methods (`function Player:move()`).
- **Language Configuration**: Adds comment tokens (`--`, `--[[ ]]`), bracket indent pairing, and scope identifiers.
- **Built-in LSP Extension**: Adds bundled `lsp-lua` manifest configuring `lua-language-server`, with mappings from VS Code extension IDs (`sumneko.lua`, `luals.lua-language-
  server`).
- **Icons**: Adds Seti Lua icon to the extension widget.
- **Tests**: Adds unit test coverage for Lua outline symbol extraction.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive Lua language support, including syntax highlighting, symbol outlines, comment handling, indentation, and bracket-aware editing.
  - Added built-in integration with `lua-language-server`.
  - Added Lua project root detection and support for common Lua-related VS Code extensions.
  - Added a dedicated Lua icon in the extensions panel.
- **Chores**
  - Updated the application version to 0.1.943.

image


vitali87:
@ironwren 
Built and tested locally at `b67011a`: compiles, `fmt` clean, clippy clean under `-D warnings`, your 4 Lua tests pass. Follows the existing per-language pattern across all 13 files.

Three things before merge:

**1. Version collision.** You claim `0.1.942`, but that shipped on `main` in b34b47f while this was open. PR is `DIRTY` — conflicts in `Cargo.toml`, `Cargo.lock`, and the release note. Rebase onto `main`, bump to `0.1.943`, rename the note. (Not your fault; the release gate can't catch this.)

**2. CI hasn't run.** Only CodeRabbit reported — no `clippy + tests`, no builds, no version gate. Needs a real green run after the rebase.

**3. Table-field functions produce no outline symbol.** `LUA_QUERY` handles `function_declaration` and assignments, but not a `field` inside a `table_constructor`:

```lua
local Player = { move = function() end }   -- no symbol
```

Reproduced on the branch (control: `function greet() end` yields symbols in the same run). Fix:

```scheme
(field
  name: (identifier) @name.function
  value: (function_definition)) @item
```

Plus a test — your outline test covers the three declaration forms only, which is how this slipped through. Worth deciding if `["name"] = function() end` should count too.

Rebased at `0.1.943`, CI green, that pattern in → good from my side.

ironwren:
@vitali87 fixs implemented

vitali87:
Reviewed the three open Copilot threads and the version. Two of the three are worth taking; the third is based on a wrong premise.

**1. `.luarc.jsonc` as a root marker — agreed, worth adding.**

`lua-language-server` reads `.luarc.jsonc` as well as `.luarc.json`, and croft already parses JSONC elsewhere (`crate::tasks::strip_jsonc`, used by `read_jsonc` in `src/import_vscode.rs`), so the format is not foreign to the project. A project whose only config is `.luarc.jsonc` currently gets root detection from `.luacheckrc`/`stylua.toml` or nothing at all.

```toml
root_markers = [".luarc.json", ".luarc.jsonc", ".luacheckrc", ".stylua.toml", "stylua.toml", "selene.toml"]
```

**2. `h.len() == 7` in `lua_highlighting_captures_keywords_and_comments` — agreed, but the reason is worth stating.**

The `7` is not wrong, it is just unexplained, and the two counts in the test genuinely differ: the source ends in a newline, so `compute_line_starts` yields 7 while `src.lines()` yields 6. Asserting the invariant rather than the constant says what is actually meant:

```rust
assert_eq!(h.len(), ls.len());
```

The `lines[0..3]` indexing below it is unaffected — those lines all exist either way.

**3. "lua-language-server ships built in" — no change needed; this is the repo's convention.**

The suggestion is to reword because the manifest says the binary is expected on PATH. But that is true of almost every entry in `TABLE`, and they all use this wording. `assets/extensions/lsp-go/extension.toml` says in its own header comment:

```
# gopls is expected on the user's toolchain PATH (croft does not provision it).
```

and its `TABLE` note is `"gopls ships built in"`. Same for `clangd` and `rust-analyzer`. In this table "ships built in" means croft ships the integration, as distinct from the `""`/no-bundle rows like `"no Vue language server bundled"`. Rewording only the Lua entry would make it the odd one out and imply the others vendor a binary. Recommend resolving this thread unchanged.

---

**Version: please take 0.1.943 — but it currently collides.**

Three PRs were open against a `main` that had already shipped `0.1.942`, and all three claimed a taken or duplicate version. I have re-based the ordering around this PR since it is the only one with a full green CI run:

- **#559 (this one): 0.1.943** — already correct, no change needed.
- #558: moved to 0.1.944.
- #555: moved to 0.1.945.

So nothing to do here for the version, but do not renumber it — the other two are now pinned around it.

vitali87:
@ironwren have a look above

vitali87 (changes requested):
All comments must be cleared

vitali87:
@coderabbitai full review

vitali87 (approved):

Issue #533, opened 2026-09-07
LSP 3.17 workspace/diagnostic: pull whole-project diagnostics from servers that advertise it


vitali87:
Split out of #256, whose design item 2 this was. #256 closed on its acceptance criteria; this part was never built and would otherwise disappear with it.

## What is missing

#256 shipped three of the four paths to whole-project diagnostics:

- **push** (rust-analyzer, basedpyright in workspace mode) — already worked; verified in #296.
- **scope control** — #296.
- **compiler fallback** (`tsc --noEmit` via `Problems: Check Whole Project`) — #530, with the size cap in #531.
- **pull** — not built. This issue.

## Proposed design (from #256, item 2)

Implement LSP 3.17 `workspace/diagnostic` for servers advertising `diagnosticProvider.workspaceDiagnostics`, with `previousResultIds` for incrementality, and refresh on `workspace/diagnostic/refresh`.

vtsls does not advertise it, which is why the compiler fallback exists; taplo and others do.

## Why it is worth doing separately

The compiler fallback spawns a process per sweep and has no incrementality — it re-checks everything each run. A server that supports pull diagnostics can answer the same question over the existing connection with result-id caching, so this is a real improvement for the servers that have it, not a duplicate of #530.

## Notes from building #530/#531

- Diagnostics are keyed by producer in `build_diag_files_by_pane` so a producer's re-run replaces its previous contribution. A pull path should take its own producer id the way the sweep took `u64::MAX` (`PROJECT_CHECK_ID`), not borrow a pane's.
- See #532: a file two producers report cannot be cleared by the second. A pull path adds a third producer and will hit it.
- Whole-workspace work runs off the UI thread here (`src/app/mod.rs`, `jump_to_test_source` and the sweep's worker are the patterns).

## Acceptance criteria

- [ ] A server advertising `workspaceDiagnostics` populates PROBLEMS for unopened files without a manual command.
- [ ] `previousResultIds` are sent, and an unchanged report is not re-processed.
- [ ] `workspace/diagnostic/refresh` triggers a refetch.
- [ ] A server that does not advertise it is untouched — no request, no error.

vitali87:
Groundwork check before anyone starts this — everything needed exists, and there are two specific traps worth knowing up front.

[...]

PR #561, opened 2026-09-18
fix(lsp): a workspace pull no longer wedges every later request (#533)

### LINKED ISSUE #560: Go to Definition silently dead on Python projects since 0.1.939
F12 and Ctrl+click do nothing at all in the editor on Python projects. No error, no status message, nothing in PROBLEMS. Reproduced on three independent setups (two clones on one Mac, plus a second Mac).

**Bisected:** 0.1.938 good, 0.1.939 bad. The only functional commit between them is `902e16c` (#533, PR #544, the whole-project diagnostics pull).

**Mechanism.** `ty` 0.0.73 advertises `workspaceDiagnostics`, accepts the spawn-time `workspace/diagnostic` pull, and then never answers it. `spawn_workspace_pull` held the client `TokioMutex` across that await, so the lock was never released. `open_doc` takes the same lock *on the single-threaded worker loop*, so the loop wedged and every later command queued behind a response that was never coming — definition, hover, completion included.

Instrumented 0.1.941 trace, with the keypress arriving 13s after the pull started:

```
lsp[ty] spawned, ... supports_definition=true
DBG pull: got client lock [ty]
lsp[ty] workspace/diagnostic pull, 0 previous result id(s)
                                   <- "released client lock" never printed
DBG F12 seen mods=KeyModifiers(0x0)
DBG cursor reached request_definition path=.../main.py
                                   <- dies here, no request ever sent
```

The app layer is innocent: the keypress reaches the LSP manager intact and dies waiting on the lock.

Rust projects are unaffected because rust-analyzer does not advertise the capability, so no pull is issued.

vitali87:
Closes #560.

## What broke

F12 and Ctrl+click did nothing at all in the editor on Python projects, from **0.1.939** on. No error, no status message, nothing in PROBLEMS. Reproduced on three independent setups.

Bisected to `902e16c` (#533, PR #544) — the only functional commit between 0.1.938 (good) and 0.1.939 (bad).

## Mechanism

`ty` 0.0.73 advertises `workspaceDiagnostics`, accepts the spawn-time `workspace/diagnostic` pull, and then never answers it. `spawn_workspace_pull` held the client `TokioMutex` across that await, so the lock was never released. `open_doc` takes the same lock **on the single-threaded worker loop**, so the loop wedged and every later command queued behind a response that was never coming — definition, hover and completion included.

Instrumented trace from 0.1.941, keypress arriving 13s after the pull starts:

```
lsp[ty] spawned, ... supports_definition=true
DBG pull: got client lock [ty]
lsp[ty] workspace/diagnostic pull, 0 previous result id(s)
                                   <- "released client lock" never printed
DBG F12 seen mods=KeyModifiers(0x0)
DBG cursor reached request_definition path=.../main.py
                                   <- dies here, no request ever sent
```

Rust projects were unaffected: rust-analyzer does not advertise the capability, so no pull is issued.

## The fix

- `LspClient::detached_server()` clones the `ServerSocket`, so the pull takes the client lock only long enough to detach and then awaits with it **released**. Response routing is per-request (async-lsp assigns the id and stores a fresh `oneshot` per call), so a clone cannot lose or misroute an answer.
- `WORKSPACE_PULL_TIMEOUT` (120s) bounds the request, so an unanswered pull costs one pull rather than parking a task on the result ids forever — the next refresh still runs. Generous by design: a cold whole-project analysis is legitimately slow, and this is a stuck-server backstop, not a latency budget.

The result-id lock is now taken *before* the pull and held to `apply_workspace_report`, so the incrementality guarantee is preserved — two overlapping pulls still cannot take the same `previous` set.

## Verification

Confirmed against real `ty` 0.0.73. Definitions answered in ~1ms **while the pull was still outstanding**, and the pull then gave up cleanly instead of wedging anything:

```
1789741378  lsp[ty] workspace/diagnostic pull, 0 previous result id(s)
1789741384  definition request id=9 ... -> response has_target=true
1789741392  definition request id=15 ... -> response has_target=true
1789741498  lsp[ty] workspace/diagnostic timed out after 120s, giving up
```

`a_hung_workspace_pull_leaves_the_client_lockable` drives the real `spawn_workspace_pull` against a stub server that handshakes, advertises the capability and never answers. It asserts the client stays lockable while the request is outstanding, and **fails** when the deadlock is reintroduced. A precondition asserts the stub really advertises `workspaceDiagnostics`, so it cannot pass by issuing no pull at all; it skips when there is no `python3`.

71 manager tests pass, `cargo fmt --check` clean, clippy clean under `-D warnings`.

## Note for the reviewer

The first commit shipped two tests that could not fail — they modelled the lock discipline against a stand-in mutex and touched no production symbol. The second commit replaces them with the real one above and removes `LspClient::workspace_diagnostics`, which had no callers left and only escaped dead-code analysis by being `pub` in a binary crate.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored Go to Definition for Python projects when the language server does not respond to workspace diagnostics.
  - Prevented stalled diagnostic requests from blocking navigation, hover, completion, and other editor actions.
  - Added a timeout so unresponsive language servers no longer leave requests waiting indefinitely.
  - Improved recovery by excluding servers that repeatedly stall from subsequent diagnostic requests.

- **Release**
  - Updated the application version to 0.1.942.


vitali87:
@greptile review

PR #563, opened 2026-09-21
Add inline color-literal swatches (hex/rgb/hsl) in the editor

vitali87:
## What

Adds inline color-literal swatches: any `#hex`, `rgb()`/`rgba()`, or
`hsl()`/`hsla()` color literal in the buffer gets a background swatch
painted directly behind its text, tinted to the parsed color, with a
legible black/white foreground chosen by luminance.

This works in **any file type/language**, with no LSP dependency - it's a
pure text scan, similar in spirit to VS Code's
[`naumovs.vscode-color-highlight`](https://marketplace.visualstudio.com/items?itemName=naumovs.vscode-color-highlight)
extension, which was a documented gap in `src/vscode_extensions.rs`.

### Not to be confused with #254

croft already has an LSP-driven "document color" feature (`textDocument/documentColor`)
that renders a `■` glyph next to colors, but **only** for languages whose LSP
server reports colors (e.g. `vscode-css-language-server` for CSS). This PR is a
separate, always-on mechanism: it tints the literal text itself, works for
any language (Rust string literals, JS/TS, JSON, comments, plain text, etc.),
and requires no language server at all.

## How

- New `scan_color_swatches()` scans each line for hex/rgb/hsl literals,
  mirroring the existing `scan_bracket_colors` architecture (per-line,
  recomputed on edit, gated by a byte-size cap so large files aren't
  penalized).
- New parser helpers: `parse_hex_color`, `parse_rgb_color`, `parse_hsl_color`,
  `hsl_to_rgb`, plus `swatch_fg_for_bg` for legible foreground selection.
- Rendered in the existing row-paint loop, tinting matched cells' bg/fg while
  preserving other style modifiers (bold/italic/etc. are untouched).
- New `show_color_swatches` field (default `true`) for future toggling.

## Tests

8 new unit tests: hex/rgb/hsl parsing (valid + invalid-length rejection),
foreground-legibility selection, any-language/any-syntax operation
(including inside comments and plain text), rendering correctness, and the
size-gate behavior on large plain-text buffers.

```
cargo test --bin croft swatch
running 8 tests
test result: ok. 8 passed; 0 failed
```

## Demo

Below: a CSS file with hex, `rgba()`, and `hsl()` color literals, each shown
with its swatch background and a readable foreground automatically chosen.

Color swatches demo


vitali87:
@coderabbitai full review

